PR #5, opened 2019-01-26
Add missing re-imports to django.db.models.

K0Te:


mkurnikov:
There is a related django tests folder `prefetch_related`, please uncomment 
https://github.com/mkurnikov/django-stubs/blob/master/scripts/typecheck_tests.py#L220
and fix errors.

K0Te:
I've updated PR, 2 ignored test errors are caused by the fact that Prefetch can add new attributes to model, not sure if there is more clean solution for this.

PR #15, opened 2019-02-04
More fields

aleksanb:
This pr adds \_\_get\_\_ signatures for even more fields which are used in our codebase.


aleksanb:
DecimalField should subclass Field and not IntegerField, so i changed it.

aleksanb:
Unsure how to proceed with the lack of Psychopg stubs, we probably want stubs for the following two files at least: https://github.com/psycopg/psycopg2/blob/master/lib/extras.py and https://github.com/psycopg/psycopg2/blob/master/lib/_range.py, but unsure as to how this would be integrated.

mkurnikov:
Shouldn't `psycopg2` stubs be added in their own `psycopg2-stubs` package in pypi?

aleksanb:
Probably, yes. I didn't really want to be maintaining a separate repo for stubs that I'd only consume through django-stubs, though.

mkurnikov:
Can it be fixed with `# type: ignore` on the `psycopg2` import line? I don't want to add `psycopg2` stubs in this repo, they really do not belong here.

aleksanb:
@mkurnikov updated with type: ignore for now.

mkurnikov:
Thanks!

PR #20, opened 2019-02-06
[WIP] improve model manager support

microblag:
I believe this brings the manger behavior more into line with how django behaves.

Managers instances created as attributes of a Model will have their TypeVar set to the ModelClass

_default_manager will be set to the first manager discovered by default
```
class MyModel(models.Model):
    special_manager = SpecialManager()
    another_manager = AnotherManager()

MyModel.special_manager has type SpecialManager[MyModel]
MyModel.another_manager has type AnotherManager[MyModel]
MyModel._default_manager has type SpecialMangaer[MyModel]
MyModel.objects is undefined
```
If no managers are defined on model then the 'objects' manager will be created as Manager[ModelClass]
```
class MyModel(models.Model):
    pass
MyModel.objects has type Manager[MyModel]
MyModel._default_manager has type Manager[MyModel]
```
If Meta.default_manager_name is set, the attribute named will be used as the Manager
```
class MyModel(models.Model):
    special_manager = SpecialManager()
    another_manager = AnotherManager()
    class Meta:
        default_manager_name = 'anothe_manager'

MyModel.special_manager has type SpecialManager[MyModel]
MyModel.another_manager has type AnotherManager[MyModel]
MyModel._default_manager has type AnotherManager[MyModel]
MyModel.objects is undefined
```


microblag:
After fixing the exception in typecheck_tests.py and running it locally i see i still have a long way to go. Should I close this PR and re-open when I've got things working better, or should i leave it and continue to iterate?

mypy_django_plugin/plugins/models.py
```diff
@@ -200,13 +236,11 @@ def extract_ref_to_fullname(rvalue_expr: CallExpr,
...
+        AddDefaultManager,
         AddIdAttributeIfPrimaryKeyTrueIsNotSet,
         SetIdAttrsForRelatedFields,
         AddRelatedManagers
     ]
     for initializer_cls in initializers:
-        initializer_cls.from_ctx(ctx).run()
-
-    # allow unspecified attributes for now
-    ctx.cls.info.fallback_to_any = True
```
mkurnikov: Leave it be. There was some problems (I'll give you an example later, if you want) with cyclical dependencies in mypy, which prevent from filling model classes with all the attributes now. 
In 0.680, or maybe in 0.690, when mypy guys will finish semanal refactoring, they should be alleviated, and we'll remove this line. For now, it's just a bunch of false positives.

mkurnikov (changes requested):

mkurnikov:
Leave it, I'll be able to review on the go.

django-stubs/contrib/auth/base_user.pyi
```diff
@@ -13,6 +13,8 @@ class AbstractBaseUser(models.Model):
     last_login: Optional[models.DateTimeField] = ...
     is_active: models.BooleanField = ...
     REQUIRED_FIELDS: List[str] = ...
+    class Meta:
+        abstract: bool = True
```
mkurnikov: Replace `True` with `...` / remove `Meta` completely, please. Value will be specified in the actual sources, no need for that here.

mkurnikov (changes requested):

django-stubs/contrib/auth/base_user.pyi
microblag: Yeah, I felt a bit weird putting those in but I haven't figured out how to get the value from the actual code when a stub is present. If you can point me in the right direction I'll gladly update the code and remove those from the stubs.

mkurnikov:
I've added some support for `_default_manager`, could you check latest master and see whether it fixed for your case?

microblag:
@mkurnikov thanks for the update, i'm in a crunch at the moment will hopefully be able to look at it next week

mkurnikov:
I'm closing it for now, if you'd find that current functionality is not enough for your use case, please, create an issue.

PR #23, opened 2019-02-19
Add better typings plus test for transaction.atomic.

syastrov:
- All cases are handled, including bare decorator (@transaction.atomic).
- Decorated function's signature is preserved when type-checking.


mkurnikov:
Thank you very much for the PR! 

Tests for the plugin are extremely slow now, mypy need to load lots of stuff in memory on every test. Could you combine all three your tests into a big one?

syastrov:
You are welcome.

I've merged the tests into one. And I removed some stuff that shouldn't have been in the changes that caused failing tests.

I have force-pushed my changes.

mkurnikov:
Looks really good, run `black` on it so CI would pass (see command in `reformat.xsh`) and I'll merge.

PR #25, opened 2019-02-19
Replaced args:object - kwargs:object  with Tuple and Dict

roderik333:
In django-stubs/views/generic/edit.pyi - replaced every occurence of 'object' in get, post and put with what I belive is the correct type.


mkurnikov:
Look at the last example here
https://mypy.readthedocs.io/en/stable/getting_started.html?highlight=kwargs#more-function-signatures

roderik333:
Ok. Closing.

mkurnikov:
Don't close it, just change them to `Any`, and, if you sure `*args` should be `str`, then change it to `str`. `object` is incorrect type annotation in any case.

PR #29, opened 2019-02-20
QuerySet.in_bulk returns Dict with values of correct model type.

syastrov:
- The keys are still Union[int, str] which doesn't cover all
possibilities, but it's not possible to cover without handling this in
the plugin, because the type of the Dict keys are dynamic depending on
which value for field_name you pass in.


django-stubs/db/models/query.pyi
```diff
@@ -86,7 +86,7 @@ class QuerySet(Iterable[_T], Sized):
     def last(self) -> Optional[_T]: ...
     def in_bulk(
         self, id_list: Any = ..., *, field_name: str = ..., **kwargs: Any
```
mkurnikov: Could you replace `id_list` parameter annotation with the `Sequence[Any]`, and `field_name` with `Optional[str]` while you're here?

syastrov: Hi. I've changed the id_list annotation. I also made a few other improvements and added a test.

mkurnikov:
Hmm, looks like I was wrong, and it's really an `Iterable[Any]`. Makes sense actually, it returns `Dict`, which is inherently unordered, so there's no reason to pass ordered container of values.

mkurnikov:
Thanks!

PR #34, opened 2019-02-28
Fix type errors on other models' managers when using objects = models.Manager() in Model.

syastrov:
I was getting all sorts of strange type errors on other models just because I had `objects` defined on one of my models as models.Manager():
```
class Foo(models.Model):
    objects = models.Manager()

class Bar(models.Model):
    a = models.IntegerField()

Bar.objects.get(id=1).a # Type error saying no attribute 'a' is defined on 'Foo'.
```

I made the following changes:
- Fill in the type argument for that particular instance of the manager, rather than modifying the bases of the Manager type.
- Instantiate a new Instance in `determine_proper_manager_type` so the code doesn't crash under mypy-mypyc.


mypy_django_plugin/main.py
```diff
@@ -55,13 +55,20 @@ def determine_proper_manager_type(ctx: FunctionContext) -> Type:
...
-            ret.type.bases[i] = Instance(base.type, [Instance(outer_model_info, [])])
-            return ret
-    return ret
+            has_manager_base = True
+            break
+
+    if has_manager_base:
+        # Fill in the manager's type argument from the outer model
+        new_type_args = [Instance(outer_model_info, [])]
+        return Instance(typ=ret.type, args=new_type_args, line=ret.line, column=ret.column)
```
mkurnikov: `helpers.reparametrize_instance` could be used here.

mkurnikov (changes requested):

mypy_django_plugin/main.py
syastrov: 👍

syastrov:
I've updated the PR to use `helpers.reparametrize_instance` and fixed a failing typecheck of a Django test.

mkurnikov (approved):

PR #39, opened 2019-03-04
Make CharField(blank=True) not be considered nullable

### LINKED ISSUE #38: CharField with blank=True should not be considered nullable
AFAIK, `CharField(blank=True,...)` just means that the string can be 0-length, and doesn't mean that it's `Optional`/nullable. The documentation on [`blank`](https://docs.djangoproject.com/en/2.1/ref/models/fields/#blank) also seems to suggest this, saying that it "will allow the entry of an empty value", which for a string is just a 0-length string.

I think the offending code is [fields.py:105](
https://github.com/mkurnikov/django-stubs/blob/c962b8ac68856e5ff39748444c20d8fb99ab9f0b/mypy_django_plugin/transformers/fields.py#L105)

I can make a pull request to remove (or comment out) that block of code if needed

Naddiseo:
The documentation on [blank](https://docs.djangoproject.com/en/2.1/ref/models/fields/#blank) says that it "will allow the entry of an empty value", which for a string is just a 0-length string. This patch allows `CharField(blank=True,...)` to no longer be considered `Optional`.

closes #38

### DIFF
--- a/mypy_django_plugin/transformers/fields.py
+++ b/mypy_django_plugin/transformers/fields.py
@@ -101,10 +101,6 @@ def get_private_descriptor_type(type_info: TypeInfo, private_field_name: str, is
 def set_descriptor_types_for_field(ctx: FunctionContext) -> Instance:
     default_return_type = cast(Instance, ctx.default_return_type)
     is_nullable = helpers.parse_bool(helpers.get_argument_by_name(ctx, 'null'))
-    if not is_nullable and default_return_type.type.has_base(helpers.CHAR_FIELD_FULLNAME):
-        # blank=True for CharField can be interpreted as null=True
-        is_nullable = helpers.parse_bool(helpers.get_argument_by_name(ctx, 'blank'))
-
     set_type = get_private_descriptor_type(default_return_type.type, '_pyi_private_set_type',
                                            is_nullable=is_nullable)
     get_type = get_private_descriptor_type(default_return_type.type, '_pyi_private_get_type',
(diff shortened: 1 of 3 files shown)


Naddiseo:
Hmm, from my understanding, the django test that failed /should/ fail type checking...
@mkurnikov, not sure what to do about that one.

mkurnikov:
https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L620
https://github.com/django/django/blob/master/django/core/validators.py#L13

I don't know. 

From my understanding of the code, `blank=False` should disable nullability, if `null=True`. But there's nothing about `blank=True`. 
You can just add this line to `IGNORED_ERRORS` for this specific module.

Naddiseo:
Hmm, interesting. However, allowing a field that doesn't have `null=True` to be `None` can cause the database to throw an IntegrityError, eg:
```python
class MyModel(models.Model):
   field = models.CharField(max_length=1, blank=True)
MyModel(field=None).save()
# raises IntegrityError (in mysql): (1048, "Column 'field' cannot be null")
```

mkurnikov:
I think `blank=True` should give nullability in the `__init__` method, and fail in `create()`. And `reveal_type` should have non-optional type. 
So, your PR is almost right, could you add special casing in `extract_expected_types` similar to https://github.com/mkurnikov/django-stubs/blob/master/mypy_django_plugin/transformers/init_create.py#L169
and add test for that? 

Then, there will be no need for additional line in `IGNORED_ERRORS`, and behaviour you mentioned will be achieved.

Naddiseo:
Let me know if you want me to format it differently, or move the test cases into say model_init.test or nullable_fields.test, but I think code-wise it should do what we're expecting.

mkurnikov:
Looks good! The only thing, could be make two tests out of those three - one for `blank=True`, `null=False`, and one for `blank=True`, `null=True`? 
Tests are very slow, trying to keep their numbers low.

mkurnikov:
Thanks!

PR #41, opened 2019-03-06
Add gzip_page decorator

mattbasta:
Adds the `gzip_page` decorator from `django.views.decorators.gzip`:

https://github.com/django/django/blob/master/django/views/decorators/gzip.py

### DIFF
--- a/django-stubs/views/decorators/gzip.pyi
+++ b/django-stubs/views/decorators/gzip.pyi
@@ -1,3 +1,5 @@
-from typing import Callable
+from typing import Callable, TypeVar
 
-def gzip_page(view_func: Callable) -> Callable: ...
+_C = TypeVar("_C", bound=Callable)
+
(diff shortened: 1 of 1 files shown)


django-stubs/views/decorators/gzip.pyi
```diff
@@ -0,0 +1,3 @@
+from typing import Callable
+
+def gzip_page(view_func: Callable) -> Callable: ...
```
mkurnikov: Please, make it preserve `Callable` like it's done here
https://github.com/mkurnikov/django-stubs/blob/194258ab8e4767cdd8c6593792c3ec6c7697947a/django-stubs/db/transaction.pyi#L34

mattbasta: Updated. Note that the other `views.decorators` methods do not preserve `Callable` like on `atomic`, so that might be something worth cleaning up in the future.

PR #49, opened 2019-03-17
Use less specific types in validators args

kalekseev:
Make it possible to use Tuple[str]


django-stubs/core/validators.pyi
```diff
@@ -38,7 +38,7 @@ class URLValidator(RegexValidator):
     tld_re: Any = ...
     host_re: Any = ...
     schemes: Any = ...
-    def __init__(self, schemes: Optional[List[str]] = ..., **kwargs: Any) -> None: ...
+    def __init__(self, schemes: Optional[Iterable[str]] = ..., **kwargs: Any) -> None: ...
```
mkurnikov: From the implementation, it looks like it should be a `Container`, `Iterable` is not enough. Could you make it a `Collection`, it implements both interfaces?

django-stubs/core/validators.pyi
```diff
@@ -52,7 +52,7 @@ class EmailValidator:
     literal_regex: Any = ...
     domain_whitelist: Any = ...
     def __init__(
-        self, message: Optional[str] = ..., code: Optional[str] = ..., whitelist: Optional[List[str]] = ...
+        self, message: Optional[str] = ..., code: Optional[str] = ..., whitelist: Optional[Iterable[str]] = ...
```
mkurnikov: Here too.

django-stubs/core/validators.pyi
```diff
@@ -119,7 +119,7 @@ class FileExtensionValidator:
     code: str = ...
     allowed_extensions: List[str] = ...
     def __init__(
-        self, allowed_extensions: Optional[List[str]] = ..., message: Optional[str] = ..., code: Optional[str] = ...
+        self, allowed_extensions: Optional[Iterable[str]] = ..., message: Optional[str] = ..., code: Optional[str] = ...
```
mkurnikov: All of them actually.

kalekseev:
@mkurnikov replaced Iterable with Collection.

Also what do you think about adding pyproject.toml file with black configuration (line length = 120) otherwise editors reformats with default line length?
I can also setup pre-commit with black, isort and flake8.

mkurnikov:
Sure, I would gladly accept a PR with those changes. Let's start with the `pyproject.toml`, and add pre-commit hooks later.

PR #53, opened 2019-03-22
Specific return types for values and values list

syastrov:
Resolves https://github.com/mkurnikov/django-stubs/issues/45

Whew... this was tougher to implement than I first imagined :)


mypy_django_plugin/lookups.py
```diff
@@ -0,0 +1,151 @@
+import dataclasses
+from typing import Union, List
+
+from mypy.nodes import TypeInfo
+from mypy.plugin import CheckerPluginInterface
+from mypy.types import Type, Instance
+
+from mypy_django_plugin import helpers
+from mypy_django_plugin.transformers.fields import get_private_descriptor_type
```
mkurnikov: Maybe move `get_private_descriptor_type` to `helpers` as it's used in more than one file now?

mypy_django_plugin/main.py
```diff
@@ -156,7 +157,7 @@ def extract_and_return_primary_key_of_bound_related_field_parameter(ctx: Attribu
         if primary_key_type:
             return primary_key_type
 
-    is_nullable = helpers.get_fields_metadata(ctx.type.type).get(field_name, {}).get('null', False)
+    is_nullable = is_field_nullable(ctx.type.type, field_name)
```
mkurnikov: Access method with `helpers.is_field_nullable` to prevent one import?

mkurnikov (changes requested):
Wow, this one is huge, thanks a lot! 

I'm not going to review it thoroughly, I'm going to trust your expertise and tests here. Just a few remarks. 

I'm going to run it against internal codebases tomorrow, and merge+release if everything is ok.

mkurnikov (approved):

syastrov:
> I'm going to run it against internal codebases tomorrow, and merge+release if everything is ok.

Looking forward to hear if you catch any new errors when running on your project.

syastrov:
I ran on an internal codebase, and there were only 2 differences in the reported errors.

In the first case, compared to master, this error disappeared (was a false positive):

```
def func(arg: Iterable[Tuple[int, datetime]]) -> int:
  pass

func(MyModel.objects.annotate(some_annotation=Max('somefield')).values_list('id', 'some_annotation') # Error occurs here
```

```f.py:63: error: Argument "arg" to "func" has incompatible type "QuerySet[MyModel, Tuple[Any, ...]]"; expected "Iterable[Tuple[int, datetime]]"```

The code was erroring because for all mypy knew, the Tuple could have had any number of elements.
With this PR, mypy knows that the Tuple row type in the QuerySet has exactly 2 elements, even though it considered the type of the second Tuple element as Any (because annotations are not supported).


In the second case, the error message was more specific because the type was inferred.
master:
```f2.py:92: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "str")```

PR:
```f2.py:92: error: Incompatible types in assignment (expression has type "Optional[*str*]", variable has type "str")```

Code looks like this:
```
def f(default_token: Optional[str]):
    if default_token is not None:
        token = default_token # Inferred to be str
    else:
        token = OtherModel.objects.filter( # Error occurs here
            token__isnull=False,
            # ... other conditions here
        ).values_list("token", flat=True).first() # Inferred to be Optional[str]
    if token:
       ...
```

Although the code checks that the `token` var is not None later on (`if token`), I guess mypy doesn't like a variable that can be both Optional and non-Optional depending on the code-path.

**Comment by mkurnikov** [comment]
Internal codebases gave me this
```
[CASE values_with_annotate_inside_the_expressions]
from django.db import models
from django.db.models.functions import Lower

class Publisher(models.Model):
    pass

class Book(models.Model):
    publisher = models.ForeignKey(to=Publisher, on_delete=models.CASCADE, related_name='books')

Publisher().books.values()
reveal_type(Publisher().books.values('name', lower_name=Lower('name')))
```

Revealed `TypedDict` should have `lower_name` as a key, at least as a 
```
lower_name: Any
```

UPD. 
Also this
```
Publisher().books.values('name')[0].pop('name')
```
shows error when it shouldn't.

mkurnikov:
This test also shows a lot of errors with stubs, fixed them in https://github.com/mkurnikov/django-stubs/pull/54

syastrov:
Thanks for the feedback. I'll look at this issue with values and kwargs.
Looks like there also may be some problems with the way the TypedDict is constructed.

mkurnikov:
Tried on internal codebase, no more errors, thanks!

PR #55, opened 2019-03-23
52/model subtypes dont typecheck

syastrov:


syastrov:
The rather annoying errors that I had to add ignores for:
`Definition of ("_default_manager" or "objects") in base class "Parent1" is incompatible with definition in base class "Parent2"`
seem to be a mypy issue. I have made a PR to mypy which fixes them: https://github.com/python/mypy/pull/6585

I'll keep on adding ignores for now...

scripts/typecheck_tests.py
```diff
@@ -213,11 +213,13 @@
...
     'many_to_one': [
-        'Incompatible type for "parent" of "Child" (got "None", expected "Union[Parent, Combinable]")'
+        'Incompatible type for "parent" of "Child" (got "None", expected "Union[Parent, Combinable]")',
+        'Incompatible type for "parent" of "Child" (got "Child", expected "Union[Parent, Combinable]")'
     ],
     'model_meta': [
         '"object" has no attribute "items"',
-        '"Field" has no attribute "many_to_many"'
+        '"Field" has no attribute "many_to_many"',
+        re.compile(r'Definition of "(objects|_default_manager)" in base class "FirstParent" is incompatible with definition in base class "SecondParent"'),
```
mkurnikov: Could you move all those ignores to `__common__`, possibly with the comment with a reference to mypy issue/PR? 

Smth along the
```
re.compile(r'Definition of "(objects|_default_manager)" in base class "[A-Za-z0-9]+" is incompatible with definition in base class "[A-Za-z0-9]+"')
```

syastrov: Done 👍

test-data/typecheck/model_create.test
```diff
@@ -27,7 +27,9 @@ class Parent1(models.Model):
 class Parent2(models.Model):
     id2 = models.AutoField(primary_key=True)
     name2 = models.CharField(max_length=50)
-class Child1(Parent1, Parent2):
+
+# TODO: Remove the 2 expected errors on the next line once mypy issue https://github.com/python/mypy/issues/2619 is resolved:
+class Child1(Parent1, Parent2): # E: Definition of "objects" in base class "Parent1" is incompatible with definition in base class "Parent2" # E: Definition of "_default_manager" in base class "Parent1" is incompatible with definition in base class "Parent2"
```
mkurnikov: Could you change it into multiline error? This is the example how it supported in the test plugin
https://github.com/mkurnikov/pytest-mypy-plugins/blob/master/pytest_mypy/tests/e2e/output.test#L41

syastrov: done

PR #73, opened 2019-05-03
django.contrib.messages.api: use HttpRequest instead of WSGIRequest

scop:
Its error messages explicitly note that an HttpRequest is required.
https://docs.djangoproject.com/en/2.1/_modules/django/contrib/messages/api/#add_message


mkurnikov:
Thanks!

PR #74, opened 2019-05-03
Fix type of prefix arg in staticfiles_urlpatterns

aszlig:
Using a *type* of `None` for the prefix keyword argument is not going to typecheck very well, so it might be a typo here because the default value (rather than the type) is `None`.


aszlig:
@mkurnikov: Hm, is CI currently broken? It can't find `mypy`/`pytest`.

PR #81, opened 2019-05-21
A couple of small fixes

bluetech:


django-stubs/contrib/postgres/fields/jsonb.pyi
```diff
@@ -15,7 +15,11 @@ class JSONField(CheckFieldDefaultMixin, Field):
     default_error_messages: Any = ...
     encoder: Any = ...
     def __init__(
-        self, verbose_name: None = ..., name: None = ..., encoder: Optional[Type[JSONEncoder]] = ..., **kwargs: Any
+        self,
+        verbose_name: Optional[str] = ...,
+        name: None = ...,
```
mkurnikov: `name` should be `Optional[str]` too?

bluetech: Yes, updated.

BTW, thanks for a great project, we already found several bugs using it.

mkurnikov:
Thanks!

PR #84, opened 2019-06-01
fix: remove SuccessMessageMixin.form_valid violation of Liskov substitution principle

### LINKED ISSUE #79: FormMixin is incompatible with SuccessMessageMixin
From what I can tell, the Django sources for these two are fine to be mixed together, however the type annotations seem to disagree:

``` python
from django.views.generic.edit import FormMixin
from django.contrib.messages.views import SuccessMessageMixin

reveal_type(FormMixin.form_valid)
# Revealed type is 'def (self: ..., form: django.forms.forms.BaseForm) -> django.http.response.HttpResponse'
reveal_type(SuccessMessageMixin.form_valid)
# Revealed type is 'def (self: ..., form: django.forms.forms.Form) -> django.http.response.HttpResponseRedirect'

class FooView(FormMixin, SuccessMessageMixin):
    pass

# Definition of "form_valid" in base class "FormMixin" is incompatible with definition in base class "SuccessMessageMixin"
```

I suspect that the signature of `FormMixin` is the one we want to use, so the fix is to update `SuccessMessageMixin` to match. (If you agree that's a good solution I'd be happy to put up a PR)

antonagestam:
fixes #79


mkurnikov:
Thanks!

Issue #83, opened 2019-06-01
"Type[Model]" has no attribute "objects" (again?)


kam1sh:
Hello. I tried to use your package yesterday and I like it, although I've encountered one issue, when the django model I was accessing had no attribute 'objects', according to mypy. So it's like #16:
```python
...
anchor/users/forms.py:26: error: "Type[Model]" has no attribute "objects"
anchor/users/views.py:38: error: "Type[Model]" has no attribute "objects"
...
```
The repo where I found that is https://github.com/kam1sh/anchor.

Also, I looked at django-stubs source code, and the class had no 'objects' field at all. Is it intended?

https://github.com/mkurnikov/django-stubs/blob/0e60b70ae4f521e3a527997d6b79e461d9245dfa/django-stubs/db/models/base.pyi#L9-L16

mkurnikov:
Try to enable plugin in `mypy.ini` via `plugins = mypy_django_plugin.main`. Let me know if it helped.

kam1sh:
I've already enabled that, but did not commited yet =)

Also, without plugin there is no errors at all, but they showed up when I installed `django-stubs` (and `typing` with `typing-extensions` as its dependencies):
```
igor@juno [19:29:48] ~/p/ciconia> mypy anchor
<no output>
igor@juno [19:31:41] ~/p/ciconia> poetry add --dev django-stubs
Using version ^0.12.1 for django-stubs

Updating dependencies
Resolving dependencies... (0.7s)


Package operations: 3 installs, 0 updates, 0 removals

Writing lock file

  - Installing typing (3.6.6)
  - Installing typing-extensions (3.7.2)
  - Installing django-stubs (0.12.1)

igor@juno [19:26:25] ~/p/ciconia> vim setup.cfg # enabled django-stubs plugin
igor@juno [19:32:04] ~/p/anchor> mypy anchor
anchor/users/forms.py:9: error: Name 'forms.UserChangeForm.Meta' is not defined
anchor/users/forms.py:19: error: Name 'forms.UserCreationForm.Meta' is not defined
anchor/users/forms.py:26: error: "Type[Model]" has no attribute "objects"
anchor/users/views.py:38: error: "Type[Model]" has no attribute "objects"
anchor/pypi/models.py:16: error: Module 'django.shortcuts' has no attribute 'reverse'
anchor/pypi/services.py:8: note: unused 'type: ignore' comment
```

mkurnikov:
[...]

PR #85, opened 2019-06-02
Add attributes to the Model and Form classes (#83)

kam1sh:
This commit fixes type error in Model 'objects' (#83) and also adds 'Meta' inner class to the ModelForm, that is used to reference model and its fields ([docs](https://docs.djangoproject.com/en/2.2/topics/forms/modelforms/#modelform))


django-stubs/db/models/base.pyi
```diff
@@ -13,6 +13,7 @@ class Model(metaclass=ModelBase):
     _meta: Any
     _default_manager: Manager[Model]
     pk: Any = ...
+    objects: Manager[Model]
```
mkurnikov, line 16: Please, take a look at the https://github.com/mkurnikov/django-stubs/pull/20 and `_default_manager` in the Django docs. `_default_manager` is the reason why there's no `objects` attribute on the `Model` class.

kam1sh: Um... then I don't understand. I should somehow define `objects` as alias to the `_default_manager`? Because if there is no `objects` then mypy will always fail in cases like mine...

kam1sh: Maybe I can define `objects` and `_default_manager` as a `TypeVar("_T", bound=Manager)`? Will it be appropriate?

pirate: You probably already know this, but just in case not, you can hint objects as `objects: models.Manager` and it will work generically for everything, though it won't be as specific as something like `Manager[Model]`.  We use this in our codebase a lot.

mkurnikov:
I addressed `Form` fix in https://github.com/typeddjango/django-stubs/pull/107, `objects` is not really that easy, closing it for now.

PR #88, opened 2019-06-10
Add `get_storage_class`

cs-cordero:
This function is missing from the storage.pyi file.

Django
master:  https://github.com/django/django/blob/master/django/core/files/storage.py#L351
2.2:  https://github.com/django/django/blob/2.2/django/core/files/storage.py#L357
2.1:  https://github.com/django/django/blob/2.1/django/core/files/storage.py#L355
2.0.13: https://github.com/django/django/blob/2.1/django/core/files/storage.py#L355


sobolevn:
Hi, @cs-cordero! Thanks for bringing this up.
I am a little bit concerned about `Any` type. I can suggest to try using `Storage` as a type here.
https://github.com/typeddjango/django-stubs/blob/master/django-stubs/core/files/storage.pyi#L7

Since I am new here 😄 I would also like to hear @mkurnikov's opinion.

cs-cordero:
HI @sobolevn!  That makes sense.  I've updated it to use `Storage` instead, though of course I defer to what @mkurnikov  thinks.

django-stubs/core/files/storage.pyi
```diff
@@ -43,4 +43,4 @@ class DefaultStorage(LazyObject): ...
 
 default_storage: Any
 
-def get_storage_class(import_path: Optional[str] = None) -> Type[Any]: ...
+def get_storage_class(import_path: Optional[str] = None) -> Type[Storage]: ...
```
sobolevn: I am pretty sure that this notation will disallow subclassing. I guess it is better to write

```
_StorageType = TypeVar('_StorageType', bound=Storage)

def get_storage_class(import_path: Optional[str] = None) -> Type[_StorageType]: 
```

cs-cordero: I could very well be wrong, but I think it does allow subclassing.  This works:

```
class A:
    pass

class B(A):
    pass

class C(B):
    pass

def lol() -> Type[A]:
    return C
```

cs-cordero: Extending the above example:

```
def lol2() -> Type[B]:
    return A # error: Incompatible return value type (got "Type[A]", expected "Type[B]")
```

sobolevn: Ok, looks good for me. Sorry for the extra work. 👍

sobolevn (approved):

django-stubs/core/files/storage.pyi
cs-cordero: no problem!  Happy to help.

mkurnikov (changes requested):
I've added two new folders to the tests typechecking, could you rebase on the latest master. (And close-reopen PR please, if Travis won't recognise it after new commit). 

Otherwise, changes are good, thanks!

PR #100, opened 2019-07-05
Add missing methods to `Field`.

cs-cordero:
I noticed the following fields are missing from the base `Field` class:

* `get_attname` (https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L753)
* `value_from_object` (https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L892)
* `validate` (https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L592)
* `run_validators` (https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L576)


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -69,6 +69,10 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
     def formfield(self, **kwargs) -> FormField: ...
     def contribute_to_class(self, cls: Type[Model], name: str, private_only: bool = ...) -> None: ...
     def to_python(self, value: Any) -> Any: ...
+    def get_attname(self) -> str: ...
+    def value_from_object(self, obj) -> _GT: ...
```
mkurnikov, line 73: Could you extract this line (whuch causes test failures) in a different PR? It's probably not obvious of a change, and requires a test.

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -69,6 +69,10 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
...
+    def validate(self, value, model_instance) -> None: ...
```
mkurnikov, line 74: Could you add annotations to the function arguments?

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -69,6 +69,10 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
...
+    def run_validators(self, value) -> None: ...
```
mkurnikov, line 75: Same here, annotations for arguments

mkurnikov (changes requested):
Thanks for contributing!

mkurnikov:
I addressed those in https://github.com/typeddjango/django-stubs/pull/109

PR #103, opened 2019-07-09
Add correct return annotation to get_storage_class

mkurnikov:
@cs-cordero Sorry, I accidentally messed up your PR https://github.com/typeddjango/django-stubs/pull/88. I've added relevant change here and added you to commit co-authors.


cs-cordero:
@mkurnikov no problem! let me know if there is more that needs doing here.

PR #105, opened 2019-07-10
Makes localtime/localdate accept optional params

### LINKED ISSUE #104: First argument of localdate() and localtime() should be Optional
In timezone.pyi, instead of 

```python
def localtime(value: _AnyTime, timezone: Optional[tzinfo] = None) -> datetime: ...
def localdate(value: _AnyTime, timezone: Optional[tzinfo] = None) -> date: ...
```

these should quite simply be:

```python
def localtime(value: Optional[_AnyTime] = None, timezone: Optional[tzinfo] = None) -> datetime: ...
def localdate(value: Optional[_AnyTime] = None, timezone: Optional[tzinfo] = None) -> date: ...
```

This is true for django >= 1.11, so well within the scope of django-stubs.

Oh, and thank you so much for this great package! <3

sobolevn:
Closes #104


django-stubs/utils/timezone.pyi
```diff
@@ -44,8 +44,8 @@ class override(ContextDecorator):
     def __enter__(self) -> None: ...
     def __exit__(self, exc_type: object, exc_value: object, traceback: object) -> None: ...
 
-def localtime(value: _AnyTime, timezone: Optional[tzinfo] = None) -> datetime: ...
-def localdate(value: _AnyTime, timezone: Optional[tzinfo] = None) -> date: ...
+def localtime(value: Optional[_AnyTime], timezone: Optional[tzinfo] = None) -> datetime: ...
+def localdate(value: Optional[_AnyTime], timezone: Optional[tzinfo] = None) -> date: ...
```
mkurnikov: It's actually optional argument, per https://github.com/django/django/blob/7f19e3713598a37b0809b5434114140170d12e34/django/utils/timezone.py#L211
Could you add `= ...` (and also replace `None` in the second arg with `...` too)? 

UPD. Same for `localtime`

mkurnikov (changes requested):

django-stubs/utils/timezone.pyi
sobolevn: Good catch! Thanks!

mkurnikov (approved):

PR #111, opened 2019-07-19
More accurate `PermissionsMixin.has_perm`

rik:
It only accepts `str` for `perm`


sobolevn:
@rik can please provide a link to the docs?

I have created an issue to add `.github/PULL_REQUEST_TEMPLATE.md` #112

rik:
Sure! https://docs.djangoproject.com/en/2.2/ref/contrib/auth/#django.contrib.auth.models.User.has_perm

To back it up by the code, [`PermissionsMixin.has_perm`]( https://github.com/django/django/blob/7f612eda80db1c1c8e502aced54c2062080eae46/django/contrib/auth/models.py#L256-L269) forwards to [`_user_has_perm`](https://github.com/django/django/blob/7f612eda80db1c1c8e502aced54c2062080eae46/django/contrib/auth/models.py#L171-L183) which forwards to every configured `backend.has_perm`. [`BaseBackend.has_perm`](https://github.com/django/django/blob/7f612eda80db1c1c8e502aced54c2062080eae46/django/contrib/auth/backends.py#L30-L31) checks that the the string is in a set of permissions.

django-stubs/contrib/auth/models.pyi
```diff
@@ -41,7 +41,7 @@ class PermissionsMixin(models.Model):
     user_permissions: models.ManyToManyField = models.ManyToManyField(Permission)
     def get_group_permissions(self, obj: None = ...) -> Set[str]: ...
     def get_all_permissions(self, obj: Optional[str] = ...) -> Set[str]: ...
-    def has_perm(self, perm: Union[Tuple[str, Any], str], obj: Optional[str] = ...) -> bool: ...
+    def has_perm(self, perm: str, obj: Optional[str] = ...) -> bool: ...
     def has_perms(self, perm_list: Union[List[str], Set[str], Tuple[str]], obj: None = ...) -> bool: ...
```
mkurnikov: Could you also make `has_perms` first parameter a `Collection[str]` and cleanup all those `obj: None` that are around?

rik: I've done the `Collection[str]` change but I'm not sure what kind of cleanup you want for the `obj: None`.

PR #113, opened 2019-07-25
values_list for related model id flat=True

mkurnikov:


test-data/typecheck/managers/querysets/test_values_list.yml
```diff
@@ -188,3 +188,20 @@
                 class Entry(models.Model):
                     text = models.CharField(max_length=100)
                     blog = models.ForeignKey(Blog, on_delete=models.CASCADE)
+
+-   case: values_list_flat_true_with_ids
+    main: |
+        from myapp.models import Blog, Publisher
+        reveal_type(Blog.objects.values_list('id', flat=True))  # N: Revealed type is 'django.db.models.query.QuerySet[myapp.models.Blog, builtins.int]'
+        reveal_type(Blog.objects.values_list('publisher_id', flat=True))  # N: Revealed type is 'django.db.models.query.QuerySet[myapp.models.Blog, builtins.int]'
```
sobolevn, line 196: What will happen for `Blog.objects.values_list('id', 'publisher_id', flat=True)`?

mkurnikov: Nothing, it's not valid for Django, and specifically treated in plugin here https://github.com/typeddjango/django-stubs/blob/master/mypy_django_plugin/transformers/querysets.py#L69

sobolevn: > TypeError: 'flat' is not valid when values_list is called with more than one field.

Awesome, thanks!

PR #132, opened 2019-08-12
Refactors bad default values, now using Ellipsis

### LINKED ISSUE #128: Setup flake8-pyi to lint stubs
It will enforce `...` instead of `None`, find useless nodes for annotations, etc.
https://github.com/ambv/flake8-pyi

sobolevn:
Turing on `flake8` for stubs seems like an impossible task. There are so many violations out there.

And I do not want to rewrite them all with `autopep8` or similar tools. I guess that history is more important than some specific style in this case.

So, all in all:
1. `.editorconfig` is introduced
2. all `=None` are refactored to `Optional[ ] = ...`

Closes #128


mkurnikov (approved):

mkurnikov:
Wait, `[*]` with `indent=2` means that all stubs will be indented with 2 spaces?

sobolevn:
Oops, I will fix it in master. That's the correct style: https://github.com/dry-python/returns/blob/master/.editorconfig

mkurnikov:
And `.yaml` will be too... I like `.yaml` with 4 spaces, considering that it has python embedded, and it's always 4 spaces indented.

mkurnikov:
Could you remove `.editorconfig` for now, or at least remove indentation configuration?

PR #135, opened 2019-08-16
Add properties to runserver.Command

Photonios:
See: https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L27


django-stubs/core/management/commands/runserver.pyi
```diff
@@ -1,3 +1,12 @@
+from typing import Tuple
+
 from django.core.management.base import BaseCommand
 
-class Command(BaseCommand): ...
+class Command(BaseCommand):
+    requires_system_checks: bool
+    stealth_options: Tuple[str]
```
sobolevn: Should be `Tuple[str, ...]` to have the same type as other commands. For example: https://github.com/django/django/blob/eac9ab7ebb1ce0cbbc79c4cf65e8f70b0635a240/django/core/management/commands/flush.py#L15

sobolevn (changes requested):
Thanks a lot!

django-stubs/core/management/commands/runserver.pyi
Photonios: Should this definition go into `BaseCommand` then?

sobolevn: Yes, I think so: https://github.com/django/django/blob/eac9ab7ebb1ce0cbbc79c4cf65e8f70b0635a240/django/core/management/base.py#L212-L226

Please, check other props as well.

Photonios: I've updated the type for `stealth_options` and `base_stealth_options` in `BaseCommand` as they were declared as `Any`.

django-stubs/core/management/commands/runserver.pyi
```diff
@@ -1,3 +1,7 @@
 from django.core.management.base import BaseCommand
 
-class Command(BaseCommand): ...
+class Command(BaseCommand):
+    default_addr: str
```
Photonios: Looking at other definitions, I believe this should be `default_addr: str = ...`.. Correct?

(Same applies to the other ones I've added here)

sobolevn: Yes, this is a style thing mostly. Sorry, we don't have a way to check that automatically for now.

Photonios: No problem. Done :+1:

sobolevn (approved):
LGTM! Thanks!

PR #149, opened 2019-08-29
Form.changed_data is not callable fixes #148

### LINKED ISSUE #148: Form changed_data should not be callable
The form stubs have `changed_data` marked as a `Callable[[], List[str]]`, but it's not callable, it's a [property](https://github.com/django/django/blob/master/django/forms/forms.py#L431-L432),  so I think it should just be `List[str]`

jaredlockhart:
When trying to annotate references to `changed_data` I noticed that mypy suggests its type should be `Callable[[BaseForm], List[str]]` but since it's an `@cached_property` it is not callable, so it should be `List[str]`.


django-stubs/forms/forms.pyi
```diff
@@ -29,6 +29,7 @@ class BaseForm:
     fields: Dict[str, Any] = ...
     renderer: BaseRenderer = ...
     cleaned_data: Any = ...
+    changed_data: List[str]: ...
```
sobolevn: Can we make it a `@propery`? I guess it is the closest match.

sobolevn (commented):
Thanks a lot! Good catch.

django-stubs/forms/forms.pyi
jaredlockhart: Oh good idea, I changed it 👍

sobolevn (approved):

sobolevn:
Thanks!

jaredlockhart:
@sobolevn Thanks for merging! When will the next tag be released with this change?

sobolevn:
Sorry, we don't have any release schedule yet. Can you use `master` for now? 

@mkurnikov can you please help us here?

PR #153, opened 2019-09-04
RFC: use DJANGO_SETTINGS_MODULE from env if available

blueyed:
This avoids setting it in setup.cfg explicitly, and follows the usual
pattern of having it defined with Django projects often already.

TODO:

- [ ] doc

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -1,4 +1,5 @@
 import configparser
+import os
 from functools import partial
 from typing import Callable, Dict, List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


sobolevn:
I like this! @mkurnikov what do you think?

sobolevn (approved):

blueyed:
(btw: would help also in case where the env is required already anyway (https://github.com/typeddjango/django-stubs/issues/166#issuecomment-532133894))

mkurnikov (approved):

mkurnikov:
@blueyed Maybe add test for it?

blueyed:
Sure, but likely not in the near future.
Doc needs also to be done.

mkurnikov:
@sobolevn @blueyed I'm going to release 1.2.0 today, this one will get into 1.2.1, when it's done.

sobolevn:
@mkurnikov I don't think that it is a good candidate for `1.2.1` - because it can break things!

My suggestion is to put it into `1.2.0`. Do you need any help from my side?

charettes:
Is this still on some release road map? Maybe only falling back to the environ value if no `mypy.plugins.django-stubs.django_settings_module` option is defined would do a better job at preserving backward compatiblity since this case currently error out.

UnknownPlatypus:
Was done in #2021

PR #157, opened 2019-09-13
Add/fix types for django.utils.safestring.mark_safe

blueyed:
Django code ref: https://github.com/django/django/blob/964dd4f4f208722d8993a35c1ff047d353cea1ea/django/utils/safestring.py#L71-L84


django-stubs/utils/safestring.pyi
```diff
@@ -11,4 +11,4 @@ class SafeText(str, SafeData):
 
 SafeString = SafeText
 
-def mark_safe(s: Any) -> Any: ...
+def mark_safe(s: str) -> SafeText: ...
```
sobolevn, line 14: According to these lines: https://github.com/django/django/blob/964dd4f4f208722d8993a35c1ff047d353cea1ea/django/utils/safestring.py#L82-L83 `mark_safe` can be also used as a decorator. Should we add this into type annotation as well?

blueyed: Likely yes - but I'm not very experienced with type annotations so cannot really tell.

sobolevn: I guess this should be something like this:

```python
from typing import Callable, TypeVar, overload

T = TypeVar('T', bound=Callable)

@overload
def mark_safe(s: T) -> T:
    ...

@overload
def mark_safe(s: str) -> SafeText:
    ...
```

mkurnikov: Just, could you call it `_C` instead of `T` for consistency with other stubs? Otherwise, yes, exactly how @sobolevn suggested, one `@overload` catches all `Callable`s, and other - all strings.

blueyed: @mkurnikov 
Please feel free to adopt this - you can amend or add to my existing commit and push it here then.

mkurnikov:
@blueyed Could you set "allow edits from maintainers"? I don't have permissions to push.

PR #158, opened 2019-09-13
flake8 fixes

blueyed:


django-stubs/db/models/__init__.pyi
```diff
@@ -109,16 +109,6 @@ from .manager import BaseManager as BaseManager, Manager as Manager
 
 from . import lookups as lookups
 
-from .aggregates import (
```
sobolevn, line 112: This is required: https://github.com/django/django/blob/master/django/db/models/__init__.py#L3

sobolevn (changes requested):

sobolevn:
Thanks a lot!

django-stubs/db/models/__init__.pyi
blueyed: It is above already (a bit more complete).
flake8 reported duplicate imports there.

mkurnikov:
Good catch! Is it possible to add flake8 for `.pyi` checking via https://github.com/ambv/flake8-pyi ? (probably pin flake8 dependency too)

blueyed:
Using flake8-pyi additionally sounds good to me.

mkurnikov:
https://github.com/typeddjango/django-stubs/pull/186 for flake8-pyi

blueyed:
Cool, rebased this PR here.

mkurnikov:
How did you find out about those duplicates? I tried to look for this kind of warning in pyflakes/flake8, no luck.

blueyed:
Via `flake8 **/*.pyi`.. ;)

Might make sense to only have a single config by now?
I think per-file-ignores then could be used to handle .py/.pyi differently if needed (and not provided via `flake8-pyi.ini`.

mkurnikov:
> Might make sense to only have a single config by now?
I think per-file-ignores then could be used to handle .py/.pyi differently if needed (and not provided via flake8-pyi.ini.

I tried this approach first, but failed. Problem was that i want to ignore `F401` in `__init__.pyi` files, and check it everywhere else. 
If you'd be able to do it, it would be great.

sobolevn:
@mkurnikov that's how it can be done: https://github.com/dry-python/returns/blob/master/setup.cfg#L41-L49

blueyed:
Closing for now.

PR #159, opened 2019-09-13
flake8: ignore certain errors for .pyi files

blueyed:
They are not checked on CI, but I run flake8 for filetype=python (which is set for `.pyi` also) in Vim/Neovim (via Neomake), so it helps to ignore issues that do not need to be fixed really.


sobolevn (approved):

sobolevn:
Thanks! We have a plan to integrate `flake8` at some point.

PR #162, opened 2019-09-14
WIP: fix scripts/typecheck_tests.py

blueyed:
Exit non-zero if it fails to run.

Currently it fails on CI, but goes unnoticed:
https://travis-ci.com/typeddjango/django-stubs/jobs/235296115

### DIFF
--- a/scripts/typecheck_tests.py
+++ b/scripts/typecheck_tests.py
@@ -47,6 +47,7 @@ if __name__ == '__main__':
     repo_directory = PROJECT_DIRECTORY / 'django-sources'
     mypy_cache_dir = Path(__file__).parent / '.mypy_cache'
     tests_root = repo_directory / 'tests'
+    global_rc = 0
 
     # copy django settings to the tests_root directory
     shutil.copy(PROJECT_DIRECTORY / 'scripts' / 'django_tests_settings.py', tests_root)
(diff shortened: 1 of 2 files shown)


scripts/typecheck_tests.py
```diff
@@ -60,9 +59,18 @@ def replace_with_clickable_location(error: str, abs_test_folder: Path) -> str:
...
         mypy_executable = distutils.spawn.find_executable('mypy')
-        completed = subprocess.run([mypy_executable, *mypy_options], env={'PYTHONPATH': str(tests_root)},
-                                   stdout=subprocess.PIPE, cwd=str(tests_root))
-        sorted_lines = sorted(completed.stdout.decode().splitlines())
+        completed = subprocess.run(
+            [mypy_executable, *mypy_options],
+            env={'PYTHONPATH': str(tests_root)},
+            stdout=subprocess.PIPE,
+            cwd=str(tests_root)
+        )
```
blueyed, line 66: Could be kept as-is maybe, but I think we need to adjust PYTHONPATH there or something similar.

blueyed:
Should be "fixed" now, but it will always exit with 0.
What is the purpose of those tests after all?  Just for information (WIP / current state)?

mkurnikov:
@blueyed Which tests? `typecheck_tests.py`? 

It's basically the only thing that tests stubs for compatibility with current Django codebase and prevents regressions. If it fails on django tests, then it's either an error in stubs, or this specific error should be ignored with error string specified in the file. You can see a lot of them there. It's not perfect, but it was very useful so far in keeping stubs more or less in check. 

I tried many times to use stuff like `retype` to merge stubs into Django codebase and check them against real code, but failed miserably. If you have any suggestions / something changed since that time, I'm very open to new things. But for now, this is the only way in my opinion.
 
As for implementation of `typecheck_tests.py` itself, if code could be improved - I'm all hands for that, I just coded smth really fast to make it work. Feel free to change in any way. 

UPD. Sorry for accidental approve, I should've looked at the changes more thorough. Anyway, it should be in the CI, it can't just always return 0.

mkurnikov (approved):

blueyed:
@mkurnikov 
Thanks for the explanation already.
It is still not clear to me though when / if the script should fail - currently there are many issues being reported, but I also do not know if it is expected to be "silent" currently anyway?

mkurnikov:
I've added error silencing just by looking at every piece of code and seeing whether it's a valid error or false positive. Anyway, I'll look at it now.

mkurnikov (approved):

blueyed:
> I've added error silencing just by looking at every piece of code and seeing whether it's a valid error or false positive.

But while it print things it will always exit with code zero?

I would assume that anything printed is not whitelisted and should cause it to fail?

mkurnikov:
Ok, I've looked at it, it's a mess. It seems I made a mistake way before and made it always return 0.

mkurnikov:
Now I need to fix LOTS of stuff, and on 0.730 everything's horribly broken. Thanks for catch!

UPD. @bluetech Also, sorry for me being arrogant in explaining all this stuff to you, as if you didn't know what you're talking about. I was so sure of myself...

PR #164, opened 2019-09-14
Bump isort, use --diff

blueyed:
TODO:

- [x] squash / squash-merge

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,6 +6,7 @@ skip =
 include_trailing_comma = true
 multi_line_output = 5
 wrap_length = 120
+known_first_party = mypy_django_plugin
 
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -6,6 +6,7 @@ skip =
 include_trailing_comma = true
 multi_line_output = 5
 wrap_length = 120
+known_third_party = mypy_django_plugin
```
sobolevn: This is a `FIRSTPARTY` code I would say. Because it is written right here: https://github.com/typeddjango/django-stubs/tree/master/mypy_django_plugin

blueyed: Fixed.

PR #165, opened 2019-09-14
Fix typecheck tests ci

mkurnikov:
Ok, this turned out to be easier, I think mypy error reporting has changed, so all module-specific ignores didn't apply. 
@blueyed  I've removed your code, if you don't mind. What really should happen, I think, is if any line is not ignored, `global_rc` should change to 1. Please, review it. 

It will hopefully fail now, which means it's working.


blueyed:
Nice.

> @blueyed I've removed your code, if you don't mind.

Ok.
Although I think it would be good to keep it to catch unexpected issues - but I guess that's done now implicitly for non-ignored lines.

It would be better to merge stdout/stderr in order though:
```diff
diff --git i/scripts/typecheck_tests.py w/scripts/typecheck_tests.py
index f2723ca..936444b 100644
--- i/scripts/typecheck_tests.py
+++ w/scripts/typecheck_tests.py
@@ -67,12 +67,11 @@ def replace_with_clickable_location(error: str, abs_test_folder: Path) -> str:
             mypy_argv,
             env={'PYTHONPATH': str(tests_root)},
             stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
         )
-        stdout = completed.stdout.decode()
-        stderr = completed.stderr.decode()
+        output = completed.stdout.decode()

-        sorted_lines = sorted((stdout + stderr).splitlines())
+        sorted_lines = sorted(output.splitlines())
         for line in sorted_lines:
             try:
                 path_to_error = line.split(':')[0]
```

Issue #68, opened 2019-04-18
Requiring field annotations even though type can be inferred from field type


pirate:
Right now mypy is complaining that my fields aren't annotated with types.

![Screen Shot 2019-04-18 at 8 11 05 AM](https://user-images.githubusercontent.com/511499/56360145-abbe6c80-61b1-11e9-9c82-71dba4fed778.png)

Is this expected behavior? If so, why? Cant the type be inferred from the field value itself?

Thanks!

mkurnikov:
Did you set `plugins = mypy_django_plugin.main` in your `mypy.ini`?

pirate:
Yup (and other Django type hints seem to be working).  Here are my two config files:

`mypy.ini`:
```ini
[mypy]
plugins =
    mypy_django_plugin.main
```

`mypy_django.ini`:
```ini
[mypy_django_plugin]
django_settings = penny.settings
ignore_missing_settings = True
ignore_missing_model_attributes = True
```

pirate:
Bump, have any ideas?  Is this expected behavior?  If not, where can we look to solve it? Thanks!

ncvc:
I'm also seeing this issue with
```
django = "=2.2.4"
mypy = "=0.720"
django-stubs = "=1.0.2"
```

Has anyone found a workaround for this?

pirate:
We just added manual type annotation comments to all the fields, never found a solution unfortunately.

sobolevn:
@pirate have you tried `1.0.x`?

pirate:
Yup, still happening on `1.0.2`:

`Pipfile.lock`:
```
"django-stubs": {
            "hashes": [
                "sha256:5018c38542ac1de55ab1782981b1c3eb690e7515a216efc44a9c01862b8cf8b5",
                "sha256:775a3fe5939f2813c6e158b2ea0f3903c91496233e3f2c12384e78915efdf8d8"
            ],
            "index": "pypi",
            "version": "==1.0.2"
        },
```
`mypy.ini`:
```
[mypy]
plugins =
    mypy_django_plugin.main
```
`mypy_django.ini`
```
[mypy_django_plugin]
django_settings = core.settings
ignore_missing_settings = True
ignore_missing_model_attributes = True
```
![image](https://user-images.githubusercontent.com/511499/63127137-7d397e00-bf66-11e9-964f-0e241acda090.png)

aleksanb:
[...]

PR #172, opened 2019-09-18
TestCase for inheritance

benrudolph:
No idea how this thing works, but this is the case that breaks for me @mkurnikov 

https://github.com/typeddjango/django-stubs/issues/68


benrudolph:
yay! i got it to fail with gis models :)

mkurnikov:
Fixed in https://github.com/typeddjango/django-stubs/pull/177. Thanks!

PR #175, opened 2019-09-21
Options.verbose_name_raw is always a string

PeterJCLaw:
At least in the current implementation, `Options.verbose_name_raw` is always a string. This PR changes the type signature to match that in order to reflect that.

My [original code](https://github.com/lamby/django-auto-one-to-one/blob/1e8c51963c3c0ed2d308ad79e14acd75607e8b45/django_auto_one_to_one/models.py) did something like this:
``` python
def foo(model: Type[models.Model], attr: Optional[str] = None) -> None:
    if not attr:
        attr = model._meta.verbose_name_raw

    locals()[attr] = ...  # errors because `attr` is `Optional[str]` here
```

In theory the `not attr` check plus the assignment from a `str` should be enough to let `mypy` know that the type of `attr` is always a `str` later in the function. However because `verbose_name_raw` is currently `Any`, that doesn't happen.


sobolevn:
Thanks!

PR #177, opened 2019-09-23
Add django.contrib.gis.db.models and django.contrib.gis.db.models.fields

### LINKED ISSUE #68: Requiring field annotations even though type can be inferred from field type

mkurnikov:
Fixes https://github.com/typeddjango/django-stubs/issues/68


benrudolph:
🙌

PR #180, opened 2019-09-23
Add support for django-configurations

hishnash:
Add support for django-configurations.

This is enabled if you add a `django_configuration` value to the Plugins settings.

Related to #1709

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,3 +74,18 @@ Otherwise, custom type will be created in mypy, named `MyUser__MyUserManager`, w
 We have Gitter here: <https://gitter.im/mypy-django/Lobby>
 
 If you think you have more generic typing issue, please refer to https://github.com/python/mypy and their Gitter.
+
+
+### Django-configurations support
+
+When using <https://django-configurations.readthedocs.io> you will need to set `django_configuration` in the `[mypy.plugins.django-stubs]` section.
+
+```ini
+[mypy]
+strict_optional = True
+
+; this one is new
+[mypy.plugins.django-stubs]
+django_settings_module = mysettings
+django_configuration = ProductionConfiguration
+```
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -41,9 +41,12 @@ def temp_environ():
         os.environ.update(environ)
 
 
-def initialize_django(settings_module: str) -> Tuple['Apps', 'LazySettings']:
+def initialize_django(settings_module: str, configuration: Optional[str] = None) -> Tuple['Apps', 'LazySettings']:
     with temp_environ():
         os.environ['DJANGO_SETTINGS_MODULE'] = settings_module
+
+        if configuration is not None:
+            os.environ['DJANGO_CONFIGURATION'] = configuration
 
         def noop_class_getitem(cls, key):
             return cls
@@ -60,7 +63,13 @@ def initialize_django(settings_module: str) -> Tuple['Apps', 'LazySettings']:
         apps.get_swappable_settings_name.cache_clear()  # type: ignore
 
         if not settings.configured:
-            settings._setup()
+            # we are not using django configuration
+            if configuration is None:
+                settings._setup()
+            else:
+                from configurations import importer  # type: ignore
+                importer.install(check_options=True)
+                settings._setup()
 
         apps.populate(settings.INSTALLED_APPS)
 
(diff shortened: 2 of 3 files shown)


sobolevn (commented):
Thanks @hishnash!

However, my vote goes against this change for two reasons:
1. This is not `django`-native approach
2. It will bring unneeded complexity to the code and start a pattern of accepting 3rd party code into `django-stubs`

I see this as a separate plugin.

mkurnikov:
I'm with @sobolevn on that. 
I can extract code for settings initialization to make it more pluggable. We can design some kind of hooks to be able to extend library. 
@sobolevn @hishnash Could you think of any other examples of django third party apps for which we'd need hooks?

hishnash:
@mkurnikov @sobolevn more pluggable settings initialisation would be the best option yes.

sobolevn:
@mkurnikov I use https://github.com/sobolevn/django-split-settings, it almost works with `django-stubs` as is, but sometimes requires extra env vars:

```python
from split_settings.tools import include
from os import environ

ENV = environ.get('DJANGO_ENV') or 'development'  # here

base_settings = [
    'components/common.py',
    'components/database.py',

    # Select the right env:
    'environments/{0}.py'.format(ENV),
]

# Include settings:
include(*base_settings)
```

mkurnikov:
@hishnash I tried different refactorings a bit, but I don't really know what would work best. If you have bandwidth, could you make a PR with refactoring you need? Like, extracting initialization into the separate pluggable class or smth.

hishnash:
@mkurnikov I will give it a spin tomorrow, I will need to check how to ensure Mypy inits plugins in the correct order.

or we could have a config param in in the Django-stubs plugin config that points to an alternative package name for init?

brianhelba:
@hishnash @mkurnikov Has there been any progress here? This is an exciting feature.

Is there anything I (or others) could contribute to help?

ghost:
any news with this?

sobolevn:
We have decided to use some sort of a callback, like this:

```cfg
[mypy.plugins.django-stubs]
django_settings_module = mysettings
django_setup_callback = path.to.your.function
```

And the function itself should look like so (probably we will call it from [here](https://github.com/typeddjango/django-stubs/blob/6e5f5f2cdba053f5412247b62e3680221164ebcf/mypy_django_plugin/django/context.py#L70)):

```python
# path/to/your.py
def function(settings_module: str) -> None:
    # By default it will do this:
    settings._setup()
```

This way, you will be able to support any configuration framework of your choice. And we will not have to add the support to the core library.

I will review and merge a PR with the described feature with great pleause. @hishnash are you still interested? 🙂

ticosax:
Can be achieved without additional support and non intrusively
with a custom plugin package
```ini
;setup.cfg
[mypy]
plugins = ./mypy_django_plugin.py
```

```python
# mypy_django_plugin.py
from configurations import importer

from mypy_django_plugin.main import plugin

importer.install()
```

Like this, when the mypy django plugin initialize the settings, the import hook of django-configurations will be installed.

ghost:
the workaround by ticosax is great if you don't need multiple plugins, but in case anyone does, i've created a [gist](https://gist.github.com/3n-k1/b5eef1a771b64f07ca6cc90b320cf3e2) that returns a no-op plugin and sets up the importer

thanks ticosax, i used your snippet as a base :)

Dimitri-WEI-Lingfeng:
> Can be achieved without additional support and non intrusively
> with a custom plugin package
> 
> ```ini
> ;setup.cfg
> [mypy]
> plugins = ./mypy_django_plugin.py
> ```
> 
> ```python
> # mypy_django_plugin.py
> from configurations import importer
> 
> from mypy_django_plugin.main import plugin
> 
> importer.install()
> ```
> 
> Like this, when the mypy django plugin initialize the settings, the import hook of django-configurations will be installed.

Thank you for sharing. It works well!
For more specific:
```python3
# configurations_mypy_django_plugin.py
import os
from configurations.importer import install
from mypy.version import __version__

from mypy_django_plugin import main


def plugin(version):
    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "django_project.config")
    os.environ.setdefault("DJANGO_CONFIGURATION", "Local")
    install()
    return main.plugin(version)
```
```ini
# mypy.ini
[mypy]
plugins = ./configurations_mypy_django_plugin.py
[mypy.plugins.django-stubs]
django_settings_module = "django_project.config"
```

john-sandall:
In case it's useful to anyone, @wlf100220 [solution here](https://github.com/typeddjango/django-stubs/pull/180#issuecomment-820062352) worked for me with a couple tweaks.

Our top-level package is not at the repo root but instead in a `src` folder, and settings is at `./src/project_name/settings.py`. I was getting path issues, resolved by specifying the `mypy_path`:

```
# pyproject.toml
[tool.mypy]
mypy_path = "./src"
plugins = ["./src/configurations_mypy_django_plugin.py"]

[tool.django-stubs]
django_settings_module = "project_name.settings"
```

Brodan:
Will this ever get integrated?

sobolevn:
@Brodan this project is open-source. If you want something, you can provide a design description, get an approval and then send a PR.

Matje1979:
Hello everyone, I am stuck with this problem. I cannot make proposed solutions to work in my case, and I don't unserstand @ticosax solution. 
`
# mypy_django_plugin.py
from configurations import importer

from mypy_django_plugin.main import plugin

importer.install()`

It seems like importing from mypy_django_plugin.main inside mypy_django_plugin.py. Can anyone explain to me what is going on here?

bwo:
@Matje1979 just name your plugin something else.

For me, the proposed solution doesn't work, though.

PR #182, opened 2019-09-23
Get all model bases from runtime context

### LINKED ISSUE #68: Requiring field annotations even though type can be inferred from field type

mkurnikov:
Fixes https://github.com/typeddjango/django-stubs/issues/68
@pirate Try this and let me know if it helps. 

TODO:
- [x] Look for model bases in the `get_base_class_hook` using this method
- [x] Tests

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -6,6 +6,7 @@ from typing import (
 )
 
 from django.core.exceptions import FieldError
+from django.db import models
 from django.db.models.base import Model
 from django.db.models.fields import AutoField, CharField, Field
 from django.db.models.fields.related import ForeignKey, RelatedField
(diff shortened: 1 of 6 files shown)


mkurnikov:
@pirate Does the emoji mean that this PR worked for you?

pirate:
Emoji was from reading the code earlier and seeing that it looked correct superficially, but I just tested it now and it worked! Thanks so much!

mkurnikov:
@msullivan Back then, you've asked about which benefits django-stubs gets from depending on Django runtime. 
This particular issue https://github.com/typeddjango/django-stubs/issues/68 could be solved with three different solutions:
1. Add type annotations to all third-party libs with custom base models.
2. This hack https://github.com/typeddjango/django-stubs/issues/68#issuecomment-534316703
3. And solution presented in this PR, which just reads all model base classes from runtime.

PR #183, opened 2019-09-24
add support for typechecking of filter/get/exclude arguments

mkurnikov:


sobolevn:
Does this cover `F('field_name')`? Should we create a new issue for this?

mkurnikov:
No, it isn't. Create an issue, I'll work on it before release, this sounds like a big problem.

sobolevn:
Done! https://github.com/typeddjango/django-stubs/issues/194

Issue #154, opened 2019-09-05
File.size: Unsupported operand types for "Callable" and "int".


wlisesrivas:
Django File.size is a property type:
`
@cached_property 
def size(self):
...
`
https://github.com/django/django/blob/master/django/core/files/base.py#L32

But in stubs it's treated like callable function (which is wrong):
`
def size(self) -> int: 
...
`
https://github.com/typeddjango/django-stubs/blob/master/django-stubs/core/files/base.pyi#L17

sobolevn:
@wlisesrivas can you please change `def size(self) -> int: ...` to be `@property`?

wlisesrivas:
@sobolevn you mean create a pull-request fixing that? If so, sure.

sobolevn:
Yes 🙂

sobolevn:
Closed by #156

PR #184, opened 2019-09-24
Fix Paginator properties mistyped as Callables

pirate:
Based on this issue https://github.com/typeddjango/django-stubs/issues/154, it looks like every now and then some `cached_properties` get mistyped as `Callable`s instead of `@property`.

I encountered this bug with `Paginator` and figured I'd PR the same fix as the above issue mentions.

**example code triggering the error:**
```python3
print(Paginator(Version.objects.all(), 100).num_pages + 1)
```

**`mypy` output:**
```python3
  47:34  error    mypy         Unsupported operand types for + ("Callable[[], int]" and "int")
```

Here's the relevant django source showing that it's a cached property:
https://github.com/django/django/blob/f97bbad908df128189eff77d98af9a25ed1ecf23/django/core/paginator.py#L94

```python3
class Paginator:
    ...

    @cached_property
    def count(self):
        ...

    @cached_property
    def num_pages(self):
        ...

    @property
    def page_range(self):
        ...

### CODE REVIEW

**Comment by sobolevn** [comment]
@pirate please fix the test here: https://travis-ci.com/typeddjango/django-stubs/jobs/238407074#L556

**Review by sobolevn — APPROVED** [review]

**Comment by mkurnikov** [comment]
@pirate @sobolevn Those two test failures are just false-positives and should be ignored in the `scripts/enabled_test_modules.py` file (add new section for the `pagination` there). It works for the Django test suite, because exception is raised before it can make a call on `int`. The right annotation for `count` is 
```
@property
def count(self) -> int: ...
```

UPD. FWIW https://github.com/django/django/pull/11816

pirate:
Cool, I updated the type signature. I only touched it originally because I noticed it looked strange when I was fixing `num_pages`, I probably should've left it alone and kept my PR focused on the one thing I was actually trying to do haha.

mkurnikov:
For the future, try to avoid using `Union` in the return type. It requires user to make lots of `isinstance` calls before using the value. Either provde `@overload` statements with multiple signatures, or just use `Any`.

mkurnikov:
@pirate, `count` was a function before, and it was incorrect. You did everything right, please return it back to `@property`=) Now `Callable` import is redundant=) 

If you want, I can make it a separate change with ignore statement.

pirate:
I agree about avoiding `Union` because of the `isinstance` painfulness, I haven't tried using `@overload` that much, but we should probably start using it more to fix this type of situation in our own codebase.

I removed the `Callable` import and set it to `@property -> int`.  8 commits for 2 lines... sorry about the slopyness!

mkurnikov:
You still need to ignore those errors in `enabled_test_modules.py` file... Don't worry about number of commits, they'll be squashed anyway. 
Please, also, rebase on master, I've added some new linting.

PR #187, opened 2019-09-25
resolve_template accepts Sequences of str

rik:
- resolve_template is checking for lists or tuples so let's use a generic Sequence
- Union[Template, Template] seemed redundant

### DIFF
--- a/django-stubs/template/response.pyi
+++ b/django-stubs/template/response.pyi
@@ -1,6 +1,6 @@
 import functools
 from http.cookies import SimpleCookie
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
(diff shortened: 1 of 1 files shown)


django-stubs/views/generic/base.pyi
```diff
@@ -24,7 +25,7 @@ class TemplateResponseMixin:
     content_type: Optional[str] = ...
     request: http.HttpRequest = ...
     def render_to_response(self, context: Dict[str, Any], **response_kwargs: Any) -> http.HttpResponse: ...
-    def get_template_names(self) -> List[str]: ...
+    def get_template_names(self) -> Union[Sequence[str], Template, str]: ...
```
sobolevn: This one looks strange to me on the first sight. What the docs say about the return type?

rik: I was to eager to not change my code when I wrote this. Turns out some child classes rely on the fact that it returns a list. I'll remove this change.

PR #189, opened 2019-09-28
fix compatibility with 0.730

### LINKED ISSUE #160: Compatibility with mypy 0.730
Using `mypy 0.730+dev.28423cd78062b81365ec9aaa591b4e823af1606d` I am seeing an error:

> setup.cfg:55: error: Error importing plugin 'mypy_django_plugin.main'

The exception there is:

> No module named 'mypy.newsemanal.semanal'

I've not investigated, but are there plans to make it compatible with the upcoming mypy version?
Pinning it to "<0.730" was done in b11a9a8.

mkurnikov:
Fixes https://github.com/typeddjango/django-stubs/issues/160


sobolevn:
@mkurnikov I guess this should also go to a README in case we strongly require some version of `mypy`.

PR #190, opened 2019-09-28
add mypy compatibility table

mkurnikov:
@sobolevn Like this? Merge, if it's fine.


sobolevn:
@mkurnikov should we also add `django` version in this table?

sobolevn (approved):

sobolevn:
Looks awesome!

Issue #204, opened 2019-10-11
Several stub for `django.db.models.base.Model` are missing


AndreyMZ:
# Bug report

### Steps to reproduce

The following code produces the following warning.

~~~py
from django.db import models

class MyModel(models.Model):
	foo = models.CharField(max_length=255)
	bar = models.CharField(max_length=255)
	
	def clean(self):
		raise self.unique_error_message(MyModel, ('foo', 'bar'))
~~~

> Unresolved attribute reference 'unique_error_message' for class 'MyModel'

### Root cause
No `django.db.models.base.Model.unique_error_message` method in the stub.
https://github.com/typeddjango/django-stubs/blob/8402e7c53ee4d2fa2ea31a506b6a986b67e10c1a/django-stubs/db/models/base.pyi#L11-L44

## How is that should be

`django.db.models.base.Model.unique_error_message` and all other methods of `django.db.models.base.Model` class must be in the stub:

```pyi
from typing import Callable, Type
from django.core.exceptions import ValidationError

# <...>

class Model(metaclass=ModelBase):
	# <...>
	def unique_error_message(self, model_class: Type[_Self], unique_check: Collection[Union[Callable, str]]) -> ValidationError: ... 
	# <...>
```

## System information

- OS: Windows 10
- `python` version: 3.6.8
- `django` version: 2.2.6
- `mypy` version: -
- `PyCharm` version: 2019.2.3 (Community Edition)
- `django-stubs` version: 1.2.0

sobolevn:
@AndreyMZ thanks a lot! Can you please send a PR with the fix?

AndreyMZ:
I have sent a PR with the stub for `django.db.models.base.Model.unique_error_message`.

But I cannot add stubs for other methods (listed below), because I do not know correct type hints.
Maybe for a start it's worth adding stubs with `Any` type hint, isn't it?

Methods of `django.db.models.base.Model` with no stub:
- ~`__init__`~
- `from_db`
- `__repr__`
- `__str__`
- `__eq__`
- `__hash__`
- `__reduce__`
- ~`__getstate__`~
- `__setstate__`
- `_get_pk_val`
- `_set_pk_val`
- ~`get_deferred_fields`~
- ~`refresh_from_db`~
- `serializable_value`
- ~`save`~
- ~`save_base`~
- `_save_parents`
- `_save_table`
- `_do_update`
- `_do_insert`
- ~`delete`~
- `_get_FIELD_display`
[...]

PR #205, opened 2019-10-11
Fix `django.contrib.admin.options.BaseModelAdmin.has_add_permission`

### LINKED ISSUE #203: Signature of "has_add_permission" incompatible with supertype "BaseModelAdmin"
## What's wrong
Getting mypy error:
`project/contact/admin.py:25: error: Signature of "has_add_permission" incompatible with supertype "BaseModelAdmin"`

For example in registering a custom Django Admin such as
```
class ContactAdmin(admin.ModelAdmin):
    list_display = ("__str__",)

    def has_add_permission(self, request: HttpRequest) -> bool:
        return False

admin.site.register(Contact, ContactAdmin)
```
results in the error above for the method `has_add_permission`.

This is possibly due to the fact that https://github.com/typeddjango/django-stubs/blob/8402e7c53ee4d2fa2ea31a506b6a986b67e10c1a/django-stubs/contrib/admin/options.pyi#L89

has     `def has_add_permission(self, request: HttpRequest, obj: Optional[Model] = ...) -> bool: ...`

while https://github.com/django/django/blob/42b9a23267f14be39b9b00958e18d5746783208e/django/contrib/admin/options.py#L467

has   `def has_add_permission(self, request):`

i.e `obj` parameter shouldn't be in stub?

## How is that should be

I don't think this error should appear here. The case of `has_change_permission` for example, works as expected.

## System information

- OS:
- `python` version: 3.7.3
- `django` version: 2.2.5
- `mypy` version: 0.720
- `django-stubs` version: 1.1.0

AndreyMZ:
# I have made things!

## Related issues
Closes #203.
Ref #204.


AndreyMZ:
Is there a contribution guide (I did not find)? If no, I think someone should add some sort of such guide at least to state that contributors must run [black](https://pypi.org/project/black/).

sobolevn:
@AndreyMZ sure thing. Thanks for the suggestion! Issue: https://github.com/typeddjango/django-stubs/issues/206

And thanks a lot for your contribution.

sobolevn:
@AndreyMZ I would appreciate your opinion: https://github.com/typeddjango/django-stubs/pull/319

PR #208, opened 2019-10-15
Cleanup EmailMessage types

kalekseev:
Tried my best to follow https://docs.djangoproject.com/en/2.2/topics/email/#emailmessage-objects and source code


django-stubs/core/mail/message.pyi
```diff
@@ -60,15 +64,17 @@ class EmailMessage:
...
-        body: Optional[str] = ...,
+        body: str = ...,
         from_email: Optional[str] = ...,
-        to: Optional[Union[Sequence[str], str]] = ...,
-        bcc: Optional[Union[Sequence[str], str]] = ...,
+        to: Optional[Sequence[str]] = ...,
+        bcc: Optional[Sequence[str]] = ...,
         connection: Optional[Any] = ...,
-        attachments: Optional[Union[List[Tuple[str, Union[str, bytes], str]], List[MIMEText]]] = ...,
+        attachments: Sequence[
```
sobolevn: Wow! This type looks complex. Let's make an alias.

sobolevn: I am talking about: `Union[MIMEBase, Tuple[Optional[Union[MIMEText, str]], Optional[_AttachmentContent], Optional[str]]]`

django-stubs/core/mail/message.pyi
```diff
@@ -78,26 +84,28 @@ class EmailMessage:
...
         body: str = ...,
         from_email: Optional[str] = ...,
-        to: Optional[List[str]] = ...,
-        bcc: Optional[List[str]] = ...,
+        to: Optional[Sequence[str]] = ...,
+        bcc: Optional[Sequence[str]] = ...,
         connection: Optional[Any] = ...,
-        attachments: None = ...,
+        attachments: Sequence[
+            Union[MIMEBase, Tuple[Optional[Union[MIMEText, str]], Optional[_AttachmentContent], Optional[str]]]
```
sobolevn: The same here.

sobolevn (changes requested):
Thanks a lot!

django-stubs/core/mail/message.pyi
```diff
@@ -60,15 +65,15 @@ class EmailMessage:
...
-        body: Optional[str] = ...,
+        body: str = ...,
         from_email: Optional[str] = ...,
-        to: Optional[Union[Sequence[str], str]] = ...,
-        bcc: Optional[Union[Sequence[str], str]] = ...,
+        to: Optional[Sequence[str]] = ...,
+        bcc: Optional[Sequence[str]] = ...,
         connection: Optional[Any] = ...,
-        attachments: Optional[Union[List[Tuple[str, Union[str, bytes], str]], List[MIMEText]]] = ...,
+        attachments: Sequence[Union[MIMEBase, _AttachmentTuple]] = ...,
```
bluetech: `Optional` should remain here.

django-stubs/core/mail/message.pyi
```diff
@@ -78,26 +83,26 @@ class EmailMessage:
     def attach(
```
bluetech: If you're feeling adventurous, you can have a more precise type here by using an `@overload` -- first case is `filename: MIMEText` and second case is the others.

Same for `_AttachmentTuple` (union of the two options).

bluetech (commented):
The changes LGTM, except one comment.

django-stubs/core/mail/message.pyi
```diff
@@ -60,15 +65,15 @@ class EmailMessage:
     def __init__(
         self,
         subject: str = ...,
-        body: Optional[str] = ...,
+        body: str = ...,
```
mkurnikov: I believe `body` can be `None`, example is here
https://github.com/django/django/blob/master/tests/mail/tests.py#L366

django-stubs/core/mail/message.pyi
```diff
@@ -60,15 +65,15 @@ class EmailMessage:
...
+        to: Optional[Sequence[str]] = ...,
+        bcc: Optional[Sequence[str]] = ...,
         connection: Optional[Any] = ...,
-        attachments: Optional[Union[List[Tuple[str, Union[str, bytes], str]], List[MIMEText]]] = ...,
+        attachments: Optional[Sequence[Union[MIMEBase, _AttachmentTuple]]] = ...,
         headers: Optional[Dict[str, str]] = ...,
-        cc: Optional[Union[Sequence[str], str]] = ...,
-        reply_to: Optional[Union[List[Optional[str]], str]] = ...,
+        cc: Optional[Sequence[str]] = ...,
+        reply_to: Optional[Sequence[str]] = ...,
```
mkurnikov, line 78: `reply_to` might also be a `Sequence[None]`, but from the implementation it looks like an incorrect test
https://github.com/django/django/blob/master/tests/mail/tests.py#L101

django-stubs/core/mail/message.pyi
```diff
@@ -60,15 +65,15 @@ class EmailMessage:
...
-        body: Optional[str] = ...,
+        body: str = ...,
         from_email: Optional[str] = ...,
-        to: Optional[Union[Sequence[str], str]] = ...,
-        bcc: Optional[Union[Sequence[str], str]] = ...,
+        to: Optional[Sequence[str]] = ...,
+        bcc: Optional[Sequence[str]] = ...,
         connection: Optional[Any] = ...,
-        attachments: Optional[Union[List[Tuple[str, Union[str, bytes], str]], List[MIMEText]]] = ...,
+        attachments: Optional[Sequence[Union[MIMEBase, _AttachmentTuple]]] = ...,
```
mkurnikov, line 75: Could be a two-element `Tuple`
https://github.com/django/django/blob/master/tests/mail/tests.py#L434
Also maybe an incorrect test, not sure.

mkurnikov (changes requested):

mkurnikov:
If you have time, it would be better to remove `mail` from `IGNORED_MODULES` here https://github.com/typeddjango/django-stubs/blob/master/scripts/enabled_test_modules.py#L8
and fix all incoming errors for Django test suite typecheck, 
image

This one
```
django-sources/tests/mail/tests.py:1203: error: Argument 1 to "push" of "SMTPChannel" has incompatible type "str"; expected "bytes"
```
should be ignored in the corresponding new section `mail` somewhere here https://github.com/typeddjango/django-stubs/blob/master/scripts/enabled_test_modules.py#L226
see 
https://github.com/python/typeshed/pull/3516

All errors like 
```
django-sources/tests/mail/tests.py:422: error: Item "int" of "Union[Message, str, int, Any]" has no attribute "get_content_type"
```
are addressed in https://github.com/python/typeshed/pull/3504 and should be ignored till new mypy version. 

Others are valid stubs bugs, I believe.

django-stubs/core/mail/message.pyi
kalekseev: I think these two are tests problem

>         'List item 1 has incompatible type "None"; expected "str"',
>        'Incompatible return value type (got "None", expected "bool")',

will try to lend a patch into the django. We also can loose requirements here but i'm not sure, should we loose types if tests are using trash data as input for testing purposes only?

mkurnikov: > 'List item 1 has incompatible type "None"; expected "str"',

It's for `body` and `reply_to` I assume? Ignoring it like you did seems fine and actually the right call.  

What about the second line? Which code does it correspond to? 
> 'Incompatible return value type (got "None", expected "bool")',

kalekseev: https://github.com/django/django/blob/master/tests/mail/tests.py#L1611 this line, the fix should be lambda: False

scripts/enabled_test_modules.py
```diff
@@ -227,6 +227,16 @@
         "Cannot resolve keyword 'porcupine' into field",
         'Argument 1 to "set" of "RelatedManager" has incompatible type "int"',
     ],
+    'mail': [
+        'List item 1 has incompatible type "None"; expected "str"',
+        'Incompatible return value type (got "None", expected "bool")',
+        'Argument 1 to "push" of "SMTPChannel" has incompatible type "str"; expected "bytes"',
+        'Value of type "Union[List[Message], str, bytes, None]" is not indexable',
+        'Incompatible types in assignment (expression has type "bool", variable has type "Union[SMTP_SSL, SMTP, None]")',
```
kalekseev: ignore for https://github.com/django/django/blob/master/tests/mail/tests.py#L1450

django-stubs/core/mail/message.pyi
mkurnikov: Seems reasonable, thanks.

django-stubs/core/mail/message.pyi
```diff
@@ -62,42 +69,44 @@ class EmailMessage:
...
-        mimetype: Optional[str] = ...,
-    ) -> None: ...
+    @overload
+    def attach(self, filename: MIMEText = ...) -> None: ...
+    @overload
+    def attach(self, filename: None = ..., content: _AttachmentContent = ..., mimetype: str = ...) -> None: ...
+    @overload
+    def attach(self, filename: str = ..., content: _AttachmentContent = ..., mimetype: None = ...) -> None: ...
+    @overload
+    def attach(self, filename: str = ..., content: _AttachmentContent = ..., mimetype: str = ...) -> None: ...
```
mkurnikov: Could you explain why there's three overloads? Can't it just be one 
```
@overload
def attach(self, filename: Optional[str] = ..., content: _AttachmentContent = ..., mimetype: Optional[str] = ...)
```
?

mkurnikov: Aside from that, amazing work, thanks=)

kalekseev: I think this will pass args like `(None, attachment, None)` while it's not a valid case (according to my understanding of the source code), we should allow only `(None, attachment, mimetype), (filename, attachment, None), (filename, attachment, mimetype)` combinations.

mkurnikov: Oh, ok. The last two could be merged into one though, with `mimetype: Optional[str] = ...`.

mkurnikov (approved):

mkurnikov:
Thanks for the great work!

PR #209, opened 2019-10-16
More precise annotations of utils.timezone functions that return instances of tzinfo subclasses.

filiplajszczak:
Current annotations for `get_fixed_timezone`, `get_default_timezone` and `get_current_timezone` follow the docstrings of the functions that broadly name returned objects as _tzinfo instance_. It is true, but imprecise, as the returned objects are instances of subclasses of tzinfo. Mypy complains when any of the methods of that subclasses are called on the returned instance eg. `error: "tzinfo" has no attribute "localize"`

Proposed changes should fix it.


django-stubs/utils/timezone.pyi
```diff
@@ -30,10 +32,10 @@ class LocalTimezone(ReferenceLocalTimezone):
 
 utc: UTC = ...
 
-def get_fixed_timezone(offset: Union[timedelta, int]) -> tzinfo: ...
-def get_default_timezone() -> tzinfo: ...
+def get_fixed_timezone(offset: Union[timedelta, int]) -> FixedOffset: ...
```
bluetech: The [Django code](https://github.com/django/django/blob/187a64608d2a207dfc12fdd6816526e9550069b7/django/utils/timezone.py#L28) doesn't return a FixedOffset, it returns a `datetime.timezone`. So I'd either change to `datetime.timezone` or keep it as `tzinfo`.

django-stubs/utils/timezone.pyi
```diff
@@ -30,10 +32,10 @@ class LocalTimezone(ReferenceLocalTimezone):
...
+def get_default_timezone() -> BaseTzInfo: ...
```
bluetech, line 36: This LGTM. Even though it can't capture the exact differences between pytz `_UTCclass`, `_StaticTzInfo` and `_DstTzInfo`, it is an improvement over `tzinfo`.

django-stubs/utils/timezone.pyi
```diff
@@ -30,10 +32,10 @@ class LocalTimezone(ReferenceLocalTimezone):
...
 def get_default_timezone_name() -> str: ...
-def get_current_timezone() -> tzinfo: ...
+def get_current_timezone() -> BaseTzInfo: ...
```
bluetech, line 42: This one is debatable -- it is possible and legal for a user to do `activate(tz)` where `tz` is any `tzinfo`, in which case `get_current_timezone` will not be a pytz timezone. On the other hand, in the vast majority of cases, it will be a pytz, so returning `tzinfo` is annoying (can't call `localize`).

Not sure what is best, but if you decide to keep `BaseTzInfo`, then a comment would be nice.

bluetech (commented):
I also ran into this issue, will be good to fix. Left some comments.

django-stubs/utils/timezone.pyi
filiplajszczak: Ok. It was changed in Django 2.2. I looked into wrong version of the codebase.

django-stubs/utils/timezone.pyi
filiplajszczak: Decision to go this way was typeshed-driven. [Stubs for pytz](https://github.com/python/typeshed/blob/master/third_party/2and3/pytz/__init__.pyi) expose only that class.

django-stubs/utils/timezone.pyi
filiplajszczak: What about `Union[tzinfo, BaseTzInfo]`?

bluetech: Returning a Union is not better than returning just `tzinfo`, in practice. For example, in the typeshed CONTRIBUTING.md file, they are discouraged, with a link to https://github.com/python/mypy/issues/1693.

IMO, practicality beats purity here, so I'd go with `BaseTzInfo`.

sobolevn: Great link! I am with @bluetech here.

filiplajszczak: Great link indeed. Thanks. Leaving it `BaseTzInfo` as it is.

bluetech:
It looks good to me now, though I would add a comment like this above `get_current_timezone()`:

```
Strictly speaking, it is possible to activate() a non-pytz timezone, in which
case BaseTzInfo is incorrect. However, this is unlikely, so we use it anyway,
to keep things ergonomic for most users.
```

filiplajszczak:
OK @bluetech It's included as suggested.

Also, what is breaking the build? I see the output, but what is the difference with the master that causes problems?

filiplajszczak:
@sobolevn Looks like that one has the same build problem as #211

sobolevn:
Yes, I will be able to fix it in the evening (Moscow time). Sorry for the inconvenience.

sobolevn:
Something broke during this time period:

Снимок экрана 2019-10-21 в 23 48 17

My wild guess is that `git submodule` changed somehow. Diff:
- Working: `f452d4232e1c14269bf2f1bd234d6e4ef462dc55` https://travis-ci.com/typeddjango/django-stubs/jobs/246235546#L183
- Failing: `f452d4232e1c14269bf2f1bd234d6e4ef462dc55` https://travis-ci.com/typeddjango/django-stubs/jobs/246498019#L183

At first sight - it does not look like something has changed at all.
I have tried to reset build cache, didn't work.

Local build also fails with the very same reason.
Снимок экрана 2019-10-22 в 0 05 43

Then, I have tried to update `django-sources` to `2.2.6`:

```diff
--- Subproject commit f452d4232e1c14269bf2f1bd234d6e4ef462dc55
+++ Subproject commit aa28213eb51ba26b349c2e51f2defcc7cc87780e
```

Still no luck, the error remains. I am out of ideas. And sadly don't have any more time for now. 
I would appreciate any help 🙂

mkurnikov:
@sobolevn FYI, CI failure happened because of unpinned mypy version. 0.740 broke tests typechecking.

sobolevn:
That's why I use `poetry` everywhere 🙂

PR #215, opened 2019-10-23
AUTH_PASSWORD_VALIDATORS should support OPTIONS

syastrov:
Django allows a Dict as the value for the key 'OPTIONS' inside of the list of validators in AUTH_PASSWORD_VALIDATORS.
See https://docs.djangoproject.com/en/2.2/topics/auth/passwords/#enabling-password-validation

The type annotation is currently overly strict.
Although, it is possible to be more specific with TypedDict, the key 'OPTIONS' is not required to be present, and there is no way to represent that currently, so I have used Union.

# I have made things!

### DIFF
--- a/django-stubs/conf/global_settings.pyi
+++ b/django-stubs/conf/global_settings.pyi
@@ -390,8 +390,8 @@ PASSWORD_RESET_TIMEOUT_DAYS = 3
 # upon login
 PASSWORD_HASHERS: List[str] = ...
 
-AUTH_PASSWORD_VALIDATORS: List[Dict[str, str]] = ...
-
+AUTH_PASSWORD_VALIDATORS: List[Dict[str, Union[str, Dict[str, Any]]]] = ...
+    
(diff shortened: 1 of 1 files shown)


syastrov:
Actually, I found out that it is possible to represent TypedDicts with non-required fields: https://mypy.readthedocs.io/en/latest/more_types.html#mixing-required-and-non-required-items

You have to do something like this:
```
class AuthPasswordValidatorsDictBase(TypedDict):
  NAME: str

class AuthPasswordValidatorsDict(AuthPasswordValidatorsDictBase, total=False):
  OPTIONS: Dict[str, Any]
```

Is it OK to use TypedDict in the stubs?

sobolevn:
@syastrov totally fine by me!

mkurnikov:
I think `Literal` and `TypedDict` should be fine for 1.3.0 version, PyCharm hopefully would catch-up by then. Thank you=)

PR #218, opened 2019-10-27
Properly type JSON responses

niklasb:
# I have made things!

This fixes the types for the `HttpResponse.json()` method, which can return not only JSON objects but also lists, strings or numbers.


sobolevn (approved):

niklasb:
is the CI ok?

sobolevn:
Nope, it has some problems. I am trying to fix it. No luck at the moment 🙁 
Any help is appreciated. See https://github.com/typeddjango/django-stubs/pull/209#issuecomment-544709890

niklasb:
this won't work anyways, apparently mypy does not support recursive type definitions...

PR #220, opened 2019-10-30
Allow returning bool from test_func

### LINKED ISSUE #219: error: Return type "Optional[bool]" of "test_func" incompatible with return type "None" in supertype "UserPassesTestMixin"
# Bug report

Hi,

Thanks for making `django-stubs`. I really appreciate it.

I've encountered an issue that I'm not entirely sure is a bug, but I think it is.

## What's wrong

In my code, I get the following error when running `mypy`:

```
error: Return type "Optional[bool]" of "test_func" incompatible with return type "None" in supertype "UserPassesTestMixin"
```

The relevant code looks like this:

```python
from typing import Optional
from django.contrib.auth.mixins import UserPassesTestMixin
from django.views import View

class NewOrganizationMembership(UserPassesTestMixin, View):
    def test_func(self) -> Optional[bool]:
        organization: Organization = get_object_or_404(
            Organization, pk=self.request.resolver_match.kwargs["pk"]
        )
        user: User = self.request.user
        return user.is_admin_of(organization)
```

In [django-stubs/contrib/auth/mixins.pyi:26](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/contrib/auth/mixins.pyi#L26), `test_func` is set to return `None`:

```python
class UserPassesTestMixin(AccessMixin):
    def test_func(self) -> None: ...
```

## How it should be

From my reading of [the documentation for `UserPassesTestMixin`](https://docs.djangoproject.com/en/2.2/topics/auth/default/#django.contrib.auth.mixins.UserPassesTestMixin), it appears that `test_func` should return `Optional[bool]`: 

```python
from django.contrib.auth.mixins import UserPassesTestMixin

class MyView(UserPassesTestMixin, View):

    def test_func(self):
        return self.request.user.email.endswith('@example.com')
```

However, I'm not entirely sure about this, considering that [Django's `test_func`](https://github.com/django/django/blob/master/django/contrib/auth/mixins.py#L94) doesn't return anything:

```python
class UserPassesTestMixin(AccessMixin):
    """
    Deny a request with a permission error if the test_func() method returns
    False.
    """

    def test_func(self):
        raise NotImplementedError(
            '{0} is missing the implementation of the test_func() method.'.format(self.__class__.__name__)
        )

    def get_test_func(self):
```

I'll make a PR to demonstrate a suggested fix.

## System information

- OS: macOs 10.14.6 (Mojave)
- `python` version: 3.7.4
- `django` version: 2.2.6
- `mypy` version: 0.740
- `django-stubs` version: 1.2.0

yhoiseth:
# Bugfix suggestion

## Related issues

- Closes #219


yhoiseth:
I see that three of eight jobs have failed. It's unclear to me whether the change in this PR has caused any of the failures, considering that [`master` is failing](https://travis-ci.com/typeddjango/django-stubs/builds).

sobolevn:
@yhoiseth no, that's a general problem with the CI. We are working on a fix.

Thanks a lot for your work! We will merge it as soon as it is fixed.

sobolevn (approved):

mkurnikov:
@sobolevn I'm working on fixing CI.

sobolevn:
@mkurnikov thanks a lot!

PR #225, opened 2019-11-11
Pin to 0.740 and fix CI

mkurnikov:


django-stubs/test/client.pyi
```diff
@@ -82,30 +84,22 @@ class Client:
...
-    def post(
-        self, path: str, data: Any = ..., content_type: str = ..., secure: bool = ..., **extra: Any
-    ) -> HttpResponse: ...
-    def head(self, path: str, data: Any = ..., secure: bool = ..., **extra: Any) -> HttpResponse: ...
-    def trace(self, path: str, secure: bool = ..., **extra: Any) -> HttpResponse: ...
+    def request(self, **request: Any) -> Any: ...
+    def get(self, path: str, data: Any = ..., secure: bool = ..., **extra: Any) -> Any: ...
+    def post(self, path: str, data: Any = ..., content_type: str = ..., secure: bool = ..., **extra: Any) -> Any: ...
+    def head(self, path: str, data: Any = ..., secure: bool = ..., **extra: Any) -> Any: ...
+    def trace(self, path: str, secure: bool = ..., **extra: Any) -> Any: ...
```
intgr, line 91: ... But why? Returning `Any` greatly reduces the usefulness of type stubs.

The `HttpResponse` class already contains the attributes that are added by monkey-patching in `Client.request()`.

PR #229, opened 2019-11-13
Better type for force_text using overloads/Literal

syastrov:
- If a str is passed in, it returns a str.
- If strings_only = True and a "protected type" is passed in, returns that type.
- Default if it doesn't match the overloads: return a str

# I have made things!


mkurnikov:
See https://github.com/typeddjango/django-stubs/pull/254

PR #230, opened 2019-11-13
Declare is_relation and related_model on Field

### LINKED ISSUE #228: is_relation and related_model attributes are not declared on the Field stub
# Bug report

## What's wrong

When type checking a function that looks like this with mypy

```python
def my_function(model_cls: Type[models.Model]) -> None:
    try:
        django_field = model_cls._meta.get_field('my_field')
    except FieldDoesNotExist:
        return
    if django_field and django_field.is_relation:
        related_model = django_field.related_model
```

you get the following errors:

```
error: "Field[Any, Any]" has no attribute "is_relation"
error: "Field[Any, Any]" has no attribute "related_model"
```

These are attributes that are always available on Fields and should be declared in the stub. See:

https://github.com/django/django/blob/89368ab6e358ebe29a0417d65209182238daa245/django/db/models/fields/__init__.py#L120

and

https://github.com/django/django/blob/89368ab6e358ebe29a0417d65209182238daa245/django/db/models/fields/__init__.py#L145

I believe the right annotations are `bool` and `Optional[Type[Model]]` respectively.

## How is that should be

There shouldn't be any mypy errors.

## System information

- OS: Mac OS 10.14
- `python` version: 3.6.5
- `django` version: 2.2.4
- `mypy` version: 0.740
- `django-stubs` version: 1.2.0

youssefm:
## Related issues
- Closes #228


youssefm:
@sobolevn I believe this is now ready for review. `related_model` is defined as an attribute on `Field` but as a property on `RelatedField` and `ForeignObjectRel`. Unfortunately, mypy was not happy about this and complaining about incompatible types. I changed the properties to attributes in the hope that they're functionally equivalent. Let me know if it's problematic. Another option might be to declare `related_model` as a property on `Field` even though it's not declared as such in the Django source code.

mkurnikov:
Definiting `related_model` as attribute is fine, but it really should have plugin support to infer the right type of model. Anyway, PR is good to go, thanks!

PR #233, opened 2019-11-20
Add new View.setup method introduced in Django 2.2

rik:
https://docs.djangoproject.com/en/2.2/ref/class-based-views/base/#django.views.generic.base.View.setup
https://github.com/django/django/blob/2.2/django/views/generic/base.py#L83-L87


sobolevn (approved):
Thanks!

PR #235, opened 2019-11-27
Declare base_fields and declared_fields properties on Form

p7g:
# I have made things!

These properties are defined on the `Form` class by `DeclarativeFieldsMetaclass.__new__` [here][1]. The `base_fields` property is mentioned briefly in the documentation [here][2], though I didn't find anything about `declared_fields` when I looked.

I am currently typing some existing code which uses these properties and figured I would make a PR to add them, but if we don't want to declare types for not-so-documented properties like these, I can just declare them in my own code.

[1]: https://github.com/django/django/blob/249a6190aef5e3676c6d9aa6c38cecff895b5993/django/forms/forms.py#L47
[2]: https://docs.djangoproject.com/en/2.2/ref/forms/api/#accessing-the-fields-from-the-form


sobolevn:
Thanks!

PR #237, opened 2019-11-30
Support mypy 0.750

kalekseev:
Release notes https://mypy-lang.blogspot.com/2019/11/mypy-0.html

This pr fixes: 

> *(breaking change)* Symbol node methods .fullname() and .name() were replaced with properties. See this comment for compatibility instructions.

One test failed with:
> /django-stubs/test-data/typecheck/managers/test_managers.yml:312:
> E     main:4: error: Invalid self argument "User_MyManager[User]" to attribute function "get_instance" with type "Callable[[MyManager], int]" (diff)

I made it pass by adding annotation to self but that case may need some extra support inside the plugin until we have runtime generics for managers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ pip install django-stubs
 
 | django-stubs | mypy version | django version | python version
 | ------------ | ---- | ---- | ---- |
+| 1.3.0 | 0.750 | 2.2.x | ^3.6
 | 1.2.0 | 0.730 | 2.2.x | ^3.6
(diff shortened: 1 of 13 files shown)


sobolevn:
@kalekseev oh wow! Thanks a lot! This is really fast PR for a really new version 🚀

mkurnikov:
Thanks for the fast PR! I've fixed CI errors in the https://github.com/typeddjango/django-stubs/pull/238 which it based on yours.

PR #238, opened 2019-11-30
Mypy 0.750 support

mkurnikov:
Supersedes https://github.com/typeddjango/django-stubs/pull/237


jrheard:
this PR refers to django-stubs version 1.3.0, which doesn't seem to exist yet. will that version be published soon? thank you! :)

edit: looks like https://github.com/typeddjango/django-stubs/issues/241 is where i should be looking!

PR #240, opened 2019-12-01
Fix smtp backend open

kalekseev:
Open actually returns None, I overlooked fail_silently condition https://github.com/django/django/blob/master/django/core/mail/backends/smtp.py#L72


sobolevn (approved):
I had to read this django source code three times to actually find `return None` there 😅 
Thanks!

PR #242, opened 2019-12-02
Move BLANK_CHOICE to django.db.models.fields

youssefm:
If you look at https://github.com/django/django/blob/9a17ae50c61a3a0ea6c552ce4e3eab27f796d094/django/db/models/fields/__init__.py#L56

`BLANK_CHOICE` is defined in `django.db.models.fields`, not `django.db.models.fields.files`


sobolevn (commented):
Thanks! Indeed

youssefm:
@sobolevn @mkurnikov tests are now passing with rebase on latest master

PR #244, opened 2019-12-04
Fix return type of `refresh_from_db`

henribru:
`refresh_from_db` is marked as returning the type of the instance, but as can be seen here it actually doesn't return anything: https://github.com/django/django/blob/master/django/db/models/base.py#L585

## Related issues
N/A


sobolevn (approved):
Yes, this is true.

Thanks a lot!

PR #245, opened 2019-12-04
Try to continue ModelInitializer loops despite lookup failures

youssefm:
@mkurnikov this fixed the issue we were talking about on gitter. Not proposing this be merged yet, just want to see how the tests do with this change.

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -118,7 +118,10 @@ class AddRelatedModelsId(ModelClassInitializer):
                     continue
 
                 rel_primary_key_field = self.django_context.get_primary_key_field(related_model_cls)
-                field_info = self.lookup_class_typeinfo_or_incomplete_defn_error(rel_primary_key_field.__class__)
+                try:
+                    field_info = self.lookup_class_typeinfo_or_incomplete_defn_error(rel_primary_key_field.__class__)
+                except helpers.IncompleteDefnException:
+                    continue
                 is_nullable = self.django_context.get_field_nullability(field, None)
                 set_type, get_type = get_field_descriptor_types(field_info, is_nullable)
                 self.add_new_node_to_model_class(field.attname,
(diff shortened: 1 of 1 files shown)


youssefm:
@mkurnikov are tests broken on master? I got the exact same failures on a different PR: https://travis-ci.com/typeddjango/django-stubs/jobs/262363046

mypy_django_plugin/transformers/models.py
```diff
@@ -214,24 +214,27 @@ class AddRelatedManagers(ModelClassInitializer):
...
-            if isinstance(relation, (ManyToOneRel, ManyToManyRel)):
-                manager_info = self.lookup_typeinfo_or_incomplete_defn_error(fullnames.RELATED_MANAGER_CLASS_FULLNAME)
-                self.add_new_node_to_model_class(attname,
-                                                 Instance(manager_info, [Instance(related_model_info, [])]))
+                if isinstance(relation, (ManyToOneRel, ManyToManyRel)):
+                    manager_info = self.lookup_typeinfo_or_incomplete_defn_error(fullnames.RELATED_MANAGER_CLASS_FULLNAME)
+                    self.add_new_node_to_model_class(attname,
+                                                     Instance(manager_info, [Instance(related_model_info, [])]))
+                    continue
+            except:
```
sobolevn: Let's change this to at least `except Exception` please

sobolevn: IMHO, the best way is to use `with suppress(Exception):`

mkurnikov: Actually, try/except wrapping this many lines of code without proper specification which exception to catch is no-go, sorry. 
@youssefm Please, mark it as Draft, so it wouldn't get accidentally merged.

youssefm: yeah I agree 100% - I was just trying to illustrate what a potential fix to the issue might look like. Right now, if one related manager fails to be added for some reason, all other related managers are not added either

youssefm:
@mkurnikov @sobolevn what do you think of this new narrower fix?

mkurnikov:
Aren't all `for` loops in the `ModelClassInitializer` subclasses are suffer from the same problem?

youssefm:
yep, good point @mkurnikov. I actually wrapped the wrong line, but then went through other places where you could potentially call `lookup_typeinfo_or_incomplete_defn_error` or `lookup_class_typeinfo_or_incomplete_defn_error` on user types in for loops and wrapped them in the same way. We could fail to look up a primary key field, a manager, or a related model class and in each of these cases, it seems like the correct behavior should be to move on to the next iteration of the loop instead of giving up entirely

mkurnikov:
Yes, but it also need to `self.api.defer()` if it's not `self.api.final_iteration`, like it's done in 
https://github.com/typeddjango/django-stubs/blob/master/mypy_django_plugin/transformers/models.py#L296.
Semantic analyzer then should return to this loop later, when more `TypeInfo`s are available. 

Seems that the correct solution is to somehow suppress exception only if `self.api.final_iteration` is `True` (execution will never come back to this line of code).

youssefm:
ok I'll close this PR for now. It looks like you have a better fix in mind.

PR #248, opened 2019-12-06
Improve precision of BaseManager.from_queryset type annotation

youssefm:
According to django docs, from_queryset returns a subclass of the manager class it is called on

> For advanced usage you might want both a custom Manager and a custom QuerySet. You can do that by calling Manager.from_queryset() which returns a subclass of your base Manager with a copy of the custom QuerySet method


mkurnikov:
This one should be addressed by the plugin in the `get_dynamic_class_hook`. Till then, it's better to return `Any` in order to avoid false-positives.

youssefm:
not sure I follow - what false positives are you referring to?

youssefm:
Is the concern that this wouldn't handle the methods that were copied over from the queryset class?

mkurnikov:
Yes. Return type should be an `Intersection[_M, queryset_class]`, https://github.com/python/typing/issues/213

youssefm:
I see - thanks for explaining

PR #259, opened 2019-12-11
Move from `IncompleteDefnError` to error flag

mkurnikov:


youssefm:
ahhh this looks like a better approach

mkurnikov:
It's wrong too, I think the one described here https://github.com/typeddjango/django-stubs/pull/245#issuecomment-564356841 is the correct one.

PR #260, opened 2019-12-11
Suppress IncompleteDefnException on final_iteration and continue with the loop

mkurnikov:
Refs https://github.com/typeddjango/django-stubs/pull/245


mkurnikov:
@youssefm Could you try this PR on your codebase?

youssefm:
@mkurnikov thanks for looking into this. This fix appears to be working in our codebase as the related managers are being typed correctly.

PR #268, opened 2019-12-12
add .model field to _BaseQuerySet

si14:
# Added missing `.model` field to `_BaseQuerySet`

Django's QuerySet sets `.model` property here: https://github.com/django/django/blob/d6505273cd889886caca57884fa79941b18c2ea6/django/db/models/query.py#L190

`django-stubs` was missing it, now it (hopefully) won't.

### DIFF
--- a/django-stubs/db/models/query.pyi
+++ b/django-stubs/db/models/query.pyi
@@ -31,6 +31,7 @@ _T = TypeVar("_T", bound=models.Model, covariant=True)
 _QS = TypeVar("_QS", bound="_BaseQuerySet")
 
 class _BaseQuerySet(Generic[_T], Sized):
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks!

si14:
Oh my, that was quick, thank you! Is there a chance of a point release with this thing? Or maybe you plan to cut a release in the near future anyway?

sobolevn:
@si14 1.3.1 was released today, so guess it might take a couple of days at least. Sorry!

si14:
No worries at all, I was just wondering when we can switch back to the upstream, will set myself a reminder to check in a couple weeks, haha. Thanks for a great addition to Django's ecosystem!

PR #273, opened 2019-12-13
update ChoiceField's `choices` kwarg's annotation

jrheard:
per the [docs](https://docs.djangoproject.com/en/3.0/ref/forms/fields/#choicefield), `choices` is:

> Either an iterable of 2-tuples to use as choices for this field, or a callable that returns such an iterable.

---

hi! i encountered this issue when typechecking a django repo at work, and opened this PR to attempt to resolve it. i haven't tested this change locally to ensure that it resolve my issue - is there documentation somewhere on the recommended workflow for doing that? thank you!

### DIFF
--- a/django-stubs/forms/fields.pyi
+++ b/django-stubs/forms/fields.pyi
@@ -207,7 +207,7 @@ class CallableChoiceIterator:
 class ChoiceField(Field):
     def __init__(
         self,
-        choices: _FieldChoices = ...,
+        choices: Union[_FieldChoices, Callable[[], _FieldChoices]] = ...,
         required: bool = ...,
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Looks valid!

mkurnikov:
Hi, thanks for the PR!

I don't use forms myself, so there don't get as much attention as ORM stuff. But, as a first step, you can uncomment `forms_tests` here
https://github.com/typeddjango/django-stubs/blob/master/scripts/enabled_test_modules.py#L7
to check stubs against available Django forms test suite.

mkurnikov (approved):

PR #276, opened 2019-12-16
Expand stubs for django.core.management.color

intgr:
* Many attributes were missing.
* `DEBUG` and `INFO` attributes were never supported by Django.
* The color attributes are dynamically built based on `PALETTES` from termcolors: https://github.com/django/django/blob/master/django/utils/termcolors.py#L75-L94
* The only exception is `ERROR_OUTPUT` which is a legacy alias assigned at: https://github.com/django/django/blob/master/django/core/management/color.py#L54


intgr:
Thank you for the quick merge and release! :)

Issue #264, opened 2019-12-11
Model field validators type is incorrect


sobolevn:
# Bug report

## What's wrong

This code:

```python
@reversion.register()
class AuthorizedDomain(TimedModel):
    """
    This model represents a domain from which users can be authenticated.

    Attributes:
        domain (CharField): a validated domain name, like: `rt.ru`
        comment (CharField): why this domain is included

    Note:
        Only superuser are allowed to create `AuthorizedDomain` instances,
        since it is security related.

    """

    class Meta(object):
        verbose_name = _('AuthorizedDomain')
        verbose_name_plural = _('AuthorizedDomains')

    domain = models.CharField(max_length=50, unique=True, validators=[
        validate_domain,  # type: ignore
    ])
    comment = models.CharField(max_length=140, blank=True)
```

Does not typecheck. Here's my `validate_domain` function:

```python
def validate_domain(domain: str) -> str:
    """
    Validates that domain name is correct.

    >>> validate_domain('ya.ru')
    'ya.ru'

    """
    domain = domain.lower().strip()
    if validators.domain(domain):
        return domain
    raise ValidationError(
        FormatErrorMessages.invalid_domain.format(
            value=domain,
        ),
    )
```

Output:

```
server/bot_app/models.py:42: error: List item 0 has incompatible type "Callable[[str], str]"; expected "Callable[..., None]"
```

## How is that should be

Validator type should match column type. And return value is not important.
Like so: `Callable[[TColumn], ...]`

## System information

- `python` version: 3.7
- `django` version: 2.2
- `mypy` version: 0.750
- `django-stubs` version: 1.3

frnhr:
Should a validator return a value? From [here](https://docs.djangoproject.com/en/3.0/ref/validators/#module-django.core.validators) it seems like it should only raise an exception if needed, otherwise return `None`.

Quote from the link above:

> A validator is a callable that takes a value and raises a ValidationError if it doesn’t meet some criteria. Validators can be useful for re-using validation logic between different types of fields.
> 
> For example, here’s a validator that only allows even numbers:
> 
>     from django.core.exceptions import ValidationError
>     from django.utils.translation import gettext_lazy as _
>     
>     def validate_even(value):
>         if value % 2 != 0:
>             raise ValidationError(
>                 _('%(value)s is not an even number'),
>                 params={'value': value},
>             )

If I'm not mistaken with that, then your validator should look like this and then it should pass the type check:
``` python
[...]

PR #284, opened 2019-12-19
Refactors validators types

sobolevn:
Related to #264


sobolevn:
Ok, this is not as easy as it seems. Why? Because validator can be `BaseValidator` instance. Which is defined here: https://github.com/django/django/blob/d8b6a76bc745b21c6cf2b29c220a91bcae7fd3d7/django/core/validators.py#L298

And then, it will have `code` and `message`. But, in case it is just a function - it won't have these properties. The thing is that currently we don't have the proper typing for `django.core.validators`.

P.S. Btw, there are some unused ignores:

```
UNUSED IGNORES ------------------------------------------------
model_fields: Incompatible type for "size" of "FloatModel" (got "object", expected "Union[float, int, str, Combinable]")
model_fields: Incompatible type for "value" of "IntegerModel" (got "object", expected
model_options: expression has type "Dict[str, Type[Model]]", target has type "OrderedDict
model_enums: 'bool' is not a valid base class
queries: No overload variant of "__getitem__" of "QuerySet" matches argument type "str"
responses: "FileLike" has no attribute "closed"
utils_tests: Argument 1 to "to_path" has incompatible type "int"; expected "Union[Path, str]"
view_tests: Exception must be derived from BaseException
view_tests: No binding for nonlocal 'tb_frames' found
wsgi: "HttpResponse" has no attribute "block_size"
```

sobolevn:
And it can also be `RegexValidator` and `DecimalValidator` and many other types. That even don't have a common parent! WTF is going on! 😵

mkurnikov:
> Btw, there are some unused ignores

Those are for Django 3.0 test check, for some reason output is different for 2.2 and for 3.0. Unused ignores are just noise, until we'd manage to add comparison between outputs of typechecking of Django testsuites. 

About `Callable[..., None]`:
https://www.python.org/dev/peps/pep-0484/#callable
> The arguments of the callback are completely unconstrained in this case (and keyword arguments are acceptable).

If you'd like to have better constraint, maybe use `Callable[[Any], None]`, but it could be prone to false-positives for dubious benefits. I assume validator can actually receive any type as argument, just will always raise `ValidationError` in this case.

sobolevn:
@mkurnikov got it!

This is now a refactoring session. I have improved several types and annotations.

P.S. Tests fail locally because of 

```
E     ../../../../../../private/var/folders/qn/2gssw9hx48g81chw0398hlrr0000gn/T/pytest-mypy-f33ug51p/django.core.exceptions.ImproperlyConfigured: Could not find the GDAL library (tried "gdal", "GDAL", "gdal2.3.0", "gdal2.2.0", "gdal2.1.0", "gdal2.0.0", "gdal1.11.0"). Is GDAL installed? If it is, try setting GDAL_LIBRARY_PATH in your settings.
```

I guess, that this should be skipped if no `gdal` is installed.

mkurnikov (approved):

PR #287, opened 2019-12-24
Reverse dependency on `django.contrib.auth.models.Permission` breaks the `from_queryset` machinery

mkurnikov:
Refs https://github.com/typeddjango/django-stubs/pull/286.

I investigated, and it seems that error occurs because of very involved dependency cycle introduced here https://github.com/typeddjango/django-stubs/blob/master/mypy_django_plugin/main.py#L162. Had no luck fixing it.


UnknownPlatypus:
Closing because the plugin code has been mostly rewritten since then

PR #289, opened 2020-01-01
Add django.contrib.postgres.forms*

asfaltboy:
generated via stubgen and cleaned up a bit


django-stubs/contrib/postgres/forms/array.pyi
```diff
@@ -0,0 +1,41 @@
+from typing import Any, Optional
+
+from django import forms as forms
+
+class SimpleArrayField(forms.CharField):
+    default_error_messages: Any = ...
```
mkurnikov: If attribute is defined on the base class, no need to define it on every field.

mkurnikov (changes requested):
Overall, I'd prefer having as little `Any` throughout the stubs as possible. Better have no method defined than `Any`. So, please, add types instead. 
Thanks!

django-stubs/contrib/postgres/forms/array.pyi
asfaltboy: addressed in 55abc1e77464908da2d53dd30f27b704c16d6d3d

asfaltboy:
Thanks for taking the time to review this. I've cleaned some up methods/attributes that are defined on parent and don't contribute any types. Other methods I derived argument or return types from the code and docs.

Please let me know if this looks better, or is there anything else we can improve further?

mkurnikov:
Thanks, that looks so much better=) 

About CI errors: 
```
django-sources/tests/postgres_tests/test_array.py:887: error: Too few arguments for "get_context" of "Widget"
django-sources/tests/postgres_tests/test_array.py:921: error: Too few arguments for "get_context" of "Widget"
```
those are Liskov violations from Django codebase, you can add method override to the child class with the valid signature (will probably need `# type: ignore` in stubs), or just add ignore line to the proper place in `enabled_test_modules.py`

```
django-sources/tests/postgres_tests/test_json.py:536: error: Incompatible types in assignment (expression has type "Type[Input]", base class "Field" defined the type as "Widget")
```
this is error in `widget:` attribute somewhere in the `forms.Field` I think? Could you fix it in this PR? 

 Last one is just a line in `enabled_test_modules`, one of those is definitely an invalid assignment
```
        class SplitDateTimeRangeField(pg_forms.DateTimeRangeField):
            base_field = forms.SplitDateTimeField

        class SplitForm(forms.Form):
            field = SplitDateTimeRangeField()
```

If any other errors would come up, it's mosly just 
1. If it's error in tests, or wrong type annotation under `assertRaises`, add lines with message to ignores. 
2. If it's inexpressible, or requires some complicated `Union`, especially in the return type - just set to `Any`. 

PR is good, if you work with forms a lot, and want to contribute more - I would be happy to accept even bigger one, it's been mostly overlooked so far.

mkurnikov (approved):

asfaltboy:
Just seen your note after pushing that ^ ... looking again 👀

django-stubs/contrib/postgres/forms/ranges.pyi
```diff
@@ -1,17 +1,24 @@
...
 class BaseRangeField(forms.MultiValueField):
     def __init__(self, **kwargs: Any) -> None: ...
     def compress(self, values: Any) -> Tuple[Optional[Callable], Optional[Callable]]: ...
 
-class IntegerRangeField(BaseRangeField): ...
-class FloatRangeField(BaseRangeField): ...
-class DateTimeRangeField(BaseRangeField): ...
-class DateRangeField(BaseRangeField): ...
+class IntegerRangeField(BaseRangeField):
+    base_field: Union[Type[forms.Field], forms.Field] = ...
```
mkurnikov: `base_field` defined on `forms.Field`, I think. If it may be both `Type[Field]` and `Field`, better fix it there, and remove all this duplication.

django-stubs/contrib/postgres/forms/jsonb.pyi
```diff
@@ -1,8 +1,9 @@
 from django import forms
-from typing import Any
+from typing import Any, Type, Union
 
 class InvalidJSONInput(str): ...
 class JSONString(str): ...
 
 class JSONField(forms.CharField):
+    widget: Union[forms.Widget, Type[forms.Widget]] = ...
```
mkurnikov: Same as in the comment about `base_field`

asfaltboy: Django is weird ... the field classes expect `widget` to define a static attribute specifying the widget class, but then instantiate and set the `self.widget` to the instance in the constuctor 🤦‍♂ ... so if I pull the above line to `forms.Field` I start seeing errors like this:
```
django-sources/tests/postgres_tests/test_ranges.py:723: error: Argument 1 to "render" of "Widget" has incompatible type "str"; expected "Widget"
django-sources/tests/postgres_tests/test_ranges.py:723: error: Too few arguments for "render" of "Widget"
django-sources/tests/postgres_tests/test_ranges.py:727: error: Argument 1 to "render" of "Widget" has incompatible type "str"; expected "Widget"
django-sources/tests/postgres_tests/test_ranges.py:727: error: Argument 2 to "render" of "Widget" has incompatible type "None"; expected "str"
```

I could only resolve this by pulling defining this in child class:

```
class BaseRangeField(forms.MultiValueField):
    widget: Widget
```

In that case, I guess it's better to keep the `Union` on the parent, and add the `Argument 1 to "render" of "Widget" has incompatible type "str"` error under: `enabled_test_modules.py`?

django-stubs/contrib/postgres/forms/ranges.pyi
asfaltboy: addressed in 67d664d

asfaltboy:
@mkurnikov please feel free to take over this if it still has value.

Note: I no longer use Django daily, and don't have the motivation to make this work.

PR #292, opened 2020-01-10
Implement @cached_property attribute inference

intgr:
# I have made things!

Previously, all `@cached_property` attributes were inferred to have `Any` type. Now Mypy can deduce them from the method's return type.

The best demonstration of this is the included test definition:

```python
from typing import List

class Foo:
    @cached_property
    def attr(self) -> List[str]: ...

    reveal_type(attr)      # N: Revealed type is 'django.utils.functional.cached_property[builtins.list*[builtins.str]]'
    reveal_type(attr.name) # N: Revealed type is 'builtins.str'

reveal_type(Foo.attr)      # N: Revealed type is 'django.utils.functional.cached_property[builtins.list*[builtins.str]]'
reveal_type(Foo.attr.func) # N: Revealed type is 'def (*Any, **Any) -> builtins.list*[builtins.str]'

f = Foo()
reveal_type(f.attr)        # N: Revealed type is 'builtins.list*[builtins.str]'
f.attr.name                # E: "List[str]" has no attribute "name"
```


scripts/enabled_test_modules.py
```diff
@@ -457,7 +457,7 @@
         'No overload variant of "join" matches argument types "str", "None"',
         'Argument 1 to "Archive" has incompatible type "None"; expected "str"',
         'Argument 1 to "to_path" has incompatible type "int"; expected "Union[Path, str]"',
-
+        'Cannot infer type argument 1 of "cached_property"',
```
intgr, line 460: Caused by untyped `lambda` used in Django tests (tests/utils_tests/test_functional.py:177):
```python
    def test_cached_property_set_name_not_called(self):
        cp = cached_property(lambda s: None)
```

sobolevn (approved):
Awesome work! Thanks a lot!

mkurnikov:
Thanks!

PR #294, opened 2020-01-10
Fix incorrect type for django.core.files.File.file (#293)

### LINKED ISSUE #293: Incorrect type for django.core.files.File.file
# Bug report

## What's wrong

Trying to assign the result of `open()` to a `File.file` attribute results in the following warning:

```
Incompatible types in assignment (expression has type "IO[Any]", variable has type "StringIO")
``` 

```
# Failing example:

import django.core.files

class MyFile(django.core.files.File):

    def open(self, mode=None):
        # (This is the File.open() method from django)
        if not self.closed:
            self.seek(0)
        elif self.name and os.path.exists(self.name):
            self.file = open(self.name, mode or self.mode)
        else:
            raise ValueError("The file cannot be reopened.")
        return self
```

## How is that should be

No error should be raised.

https://github.com/typeddjango/django-stubs/blob/master/django-stubs/core/files/base.pyi#L11

is `file: StringIO = ...` but should probably be `file: IO[Any] = ...`

## System information

- OS: mac
- `python` version: 3.7.5
- `django` version: 2.2
- `mypy` version: 0.750
- `django-stubs` version: 1.3.0

danifus:
## Related issues
- Closes #293


django-stubs/core/files/base.pyi
```diff
@@ -8,7 +8,7 @@ _T = TypeVar("_T", bound="File")
 
 class File(FileProxyMixin, IO[Any]):
     DEFAULT_CHUNK_SIZE: Any = ...
-    file: StringIO = ...
+    file: IO[Any] = ...
```
sobolevn, line 11: Is it possible to replace `Any` with something more specific here? Or is it the only way?

danifus: I think it might be the only way. The opened file being assigned to the attribute may be either opened as bytes or a string.

Also, from a brief look at how `open()` works in mypy, if the `mode` arg is constant it can infer if it is either `IO[str]` or `IO[bytes]` but if it is a variable it will return `IO[Any]`. Django's `File.open()` method uses a variable for the `mode` arg which means the type in django will be `IO[Any]`.

sobolevn: Thanks!

sobolevn:
Thanks!

PR #296, opened 2020-01-14
Added manager _db property

### LINKED ISSUE #295: Manager has no attribute _db
# Bug report

## What's wrong

The following example from the Django documentation doesn't work:

```python
class PersonQuerySet(models.QuerySet):
    def authors(self):
        return self.filter(role='A')

    def editors(self):
        return self.filter(role='E')

class PersonManager(models.Manager):
    def get_queryset(self):
        return PersonQuerySet(self.model, using=self._db)

    def authors(self):
        return self.get_queryset().authors()

    def editors(self):
        return self.get_queryset().editors()
```

django-stubs is not defining the `_db` property:

```
error: "PersonManager" has no attribute "_db"
```

## System information

- OS: OS X
- `python` version: 3.8.1
- `django` version: 3.0.2
- `mypy` version: 0.761

canassa:
The following example from the Django documentation doesn't work, this pull request adds the `_db` property.

```python
class PersonQuerySet(models.QuerySet):
    def authors(self):
        return self.filter(role='A')

    def editors(self):
        return self.filter(role='E')

class PersonManager(models.Manager):
    def get_queryset(self):
        return PersonQuerySet(self.model, using=self._db)

    def authors(self):
        return self.get_queryset().authors()

    def editors(self):
        return self.get_queryset().editors()
```
## Related issues

- Closes #295


sobolevn (approved):
Thanks!

PR #300, opened 2020-01-17
Add test reproducing foreign key related manager failure

aleksanb:
This is roughly the code i have in my repo that broke between 1.3.2 and 1.3.3, but this test passes :( I'll try to add more code to it hoping it'll fail soon.


UnknownPlatypus:
The test in this pr currently pass, the underlying issue was fixed at some point

PR #301, opened 2020-01-18
Add OutputWrapper write() signature

### LINKED ISSUE #291: Missing OutputWrapper.write() stub
# Bug report

`OutputWrapper.write()` only accepts one argument (as it inherits from `TextIOBase`).

## What's wrong

`OutputWrapper.write()` accepts a style function argument, but this is not in the stub. As a result, this error appears:

```
error: Too many arguments for "write" of "TextIOBase"
```

Example management command:

```python


class NewCommand(BaseCommand):
    def handle(self, *args: Any, **kwargs: Any) -> None:
        self.stdout.write('Doing something', self.style.MIGRATE_LABEL)
```

## How is that should be

There should not be any error.

The stub should be similar to:

```python
class OutputWrapper(TextIOBase):
    ...
    def write(msg: str, style_func: Optional[Callable[[str], str]] = None, ending: Optional[str] = None): ...
```

Maybe it takes `AnyStr` but I'm not sure.

## System information

- OS: Linux
- `python` version: 3.7
- `django` version: 2.2.6
- `mypy` version: 0.761

Tatsh:
I added the `write()` method to the `OutputWrapper` class, which differs from the base class.

I also set the arguments and return values for the `style_func` callable.

## Related issues

- Closes #291


sobolevn (approved):
LGTM 👍

Tatsh:
I was thinking this would come up:

```
django-stubs/core/management/basei:33: error: Signature of "write" incompatible with supertype "TextIOBase" (diff)
```

Is there a project-specified way to deal with this other than `# type: ignore[override]`?

PR #303, opened 2020-01-20
add stub for core/asgi.py from Django 3

gregbaker:
When checking a Django 3 project, I get this error (with sufficient strictness):
```
$ mypy .
foo/asgi.py:12: error: No library stub file for module 'django.core.asgi'
foo/asgi.py:12: note: (Stub files are from https://github.com/python/typeshed)
Found 1 error in 1 file (checked 15 source files)
```

The PR creates a stub for that file, which is new in Django 3.

### DIFF
--- a/django-stubs/core/asgi.pyi
+++ b/django-stubs/core/asgi.pyi
@@ -1,0 +1,3 @@
+from django.core.handlers.asgi import ASGIHandler
+
+def get_asgi_application() -> ASGIHandler: ...


sobolevn:
Thanks! We are currently working on [different version support](https://github.com/typeddjango/django-stubs/pull/288) in django-stubs. It would be an awesome addition!

askoretskiy:
I there any way I could contribute here? I face the same error with new Django 3.0 project.

palfrey:
FYI I also needed a stub ASGIHandler to make this work

palfrey:
This appears to have been fixed in https://github.com/typeddjango/django-stubs/pull/527

PR #305, opened 2020-01-23
Fix apps.get_model() return type

### LINKED ISSUE #304: Too broad return value for apps.get_model()

yhoiseth:
## Related issues

- Closes #304


sobolevn (approved):

yhoiseth:
# Troubleshooting notes

- All jobs pass except _Typecheck Django 2.2 test suite with python 3.7_

## Reproduce locally

Done. (Everything below `UNUSED IGNORES` is present on `master`, as well.)

## Try to fix

`Union[Type[Any], Type[Model]]` doesn't work, either… Weird, considering that `Type[Any]` works.

Two models are failing: `SiteManager` and `Food`. They both have attributes/methods that aren't present on `Model`. Considering that they are both models, I think it makes sense that the type hint `Type[Model]` is more correct than `Type[Any]`. But the test doesn't agree.

It's unclear to me why it works with 3.0 but not with 2.2.

Appending `# type: ignore` to the failing test lines works, but not sure whether that's a good idea. Also unsure what the repo situation with `django-sources/` is.

## Next steps

I don't have any more ideas for how to proceed.

mkurnikov:
I think it requires plugin support, otherwise, it will create a bunch of false-positives everywhere.

UnknownPlatypus:
This was done in commit 89bb6eac756350b1d93e40da973c7c8180d0124d

PR #307, opened 2020-01-27
Allow HTTP error handlers to be strings

JBKahn:
Based on the docs on this page: https://docs.djangoproject.com/en/3.0/ref/urls/

I noticed when adding this to my Django application that the typing information in the django docs don't match those of this repository.

I'm unsure I did this correctly or if any tests are needed/how to add them but I figured I'd contribute this much and if you want to guide me through the process I can help make whatever changes are needed.


sobolevn (approved):
Thanks! Indeed.

PR #310, opened 2020-01-30
Cleanup checks framework

kalekseev:
Improved typings and adjusted exports according to [`__all__`](https://github.com/django/django/blob/efc1c73bf511cbebf03fa5b13fa3a98718daf928/django/core/checks/__init__.py#L19)


sobolevn (approved):
Thanks a lot!

PR #313, opened 2020-01-31
Revert manager behavior for managers without 'objects'

aleksanb:
This fixes reverse many-to-one managers regressions in our project that
were introduced in
https://github.com/typeddjango/django-stubs/commit/665f4d8ea1723d0bcae59904a590ffb03a7c0355.
I've tried to create a reproduce in
https://github.com/typeddjango/django-stubs/pull/300 but was unable to
do so. This patch makes django-stubs work for us again, perhaps you are
able to see why this fix works, because i do not know the reason why my
class doesn't have 'objects'.

We use django 3.0.2 and python 3.8.

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -273,13 +273,16 @@ class AddRelatedManagers(ModelClassInitializer):
             if isinstance(relation, (ManyToOneRel, ManyToManyRel)):
                 try:
                     related_manager_info = self.lookup_typeinfo_or_incomplete_defn_error(fullnames.RELATED_MANAGER_CLASS)  # noqa: E501
-                    if 'objects' not in related_model_info.names:
-                        raise helpers.IncompleteDefnException()
                 except helpers.IncompleteDefnException as exc:
                     if not self.api.final_iteration:
                         raise exc
                     else:
                         continue
+
+                if 'objects' not in related_model_info.names:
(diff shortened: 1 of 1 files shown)


mypy_django_plugin/transformers/models.py
```diff
@@ -273,14 +273,17 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
                     related_manager_info = self.lookup_typeinfo_or_incomplete_defn_error(fullnames.RELATED_MANAGER_CLASS)  # noqa: E501
-                    if 'objects' not in related_model_info.names:
-                        raise helpers.IncompleteDefnException()
                 except helpers.IncompleteDefnException as exc:
                     if not self.api.final_iteration:
                         raise exc
                     else:
                         continue
 
+                if 'objects' not in related_model_info.names:
```
sobolevn, line 282: Thanks! I would be also glad to see a test case for this change.

Can you please provide one that was failing for you?

aleksanb: I’ll try to recreate it again, but i have been unsuccessful so far. I’ll try removing code from our actual project until i have a reproduce instead of creating a test case from the ground up.

sobolevn: thanks!

UnknownPlatypus:
This has been fixed later

PR #316, opened 2020-02-01
Fix cache arg type for django.utils.cache.learn_cache_key()

### LINKED ISSUE #314: django.utils.learn_cache_key incorrect type
# Bug report

## What's wrong
`django.utils.learn_cache_key()` has the incorrect type specified for the `cache` parameter.

https://github.com/typeddjango/django-stubs/blob/f89a0fbbaa53a23b80fd0b9bfe8fdbf71da35de8/django-stubs/utils/cache.pyi#L31

It is currently set as `Optional[LocMemCache]` but in reality this can be any valid type of `BaseCache`. Looking into the function you can see if `cache=None` then it is pulled from `django.core.cache.caches` which returns a `BaseCache` type.

## System information

- OS: Windows 10, 64-bit
- `python` version: 3.8.1, 64-bit
- `django` version: 3.0.2
- `mypy` version: 0.761

vsalvino:
Closes #314 

Some friendly feedback: it would be helpful if this project provided a contributing guide or developer guide with instructions on how to build the project, test the project, what needs to be contributed as part of a pull requests (tests and docs, etc.). I have not performed anything other than manual testing of my own use case for this change.

### DIFF
--- a/django-stubs/utils/cache.pyi
+++ b/django-stubs/utils/cache.pyi
@@ -1,7 +1,6 @@
 from typing import Any, Optional, Tuple
 
 from django.core.cache.backends.base import BaseCache
-from django.core.cache.backends.locmem import LocMemCache
 from django.core.handlers.wsgi import WSGIRequest
 from django.http.response import HttpResponse, HttpResponseBase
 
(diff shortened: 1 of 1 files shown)


sobolevn (approved):

sobolevn:
+100 on `CONTRIBUTING.md`

Thanks a lot!

sobolevn:
@vsalvino I would appreciate your feedback: #319

PR #321, opened 2020-02-06
Fix the type of form.Field.widget

brianhelba:
The `widget` class attribute of `form.Field` expects a `Widget` class, not an instance of `Widget`.

See: https://docs.djangoproject.com/en/3.0/ref/forms/fields/#widget
Also usages at: https://github.com/django/django/blob/3259983f569151232d8e3b0c3d0de3a858c2b265/django/forms/fields.py#L46


sobolevn (approved):
Thanks for the fix!

PR #322, opened 2020-02-12
Improve test client type stubs

intgr:
# I have made things!

1. Added missing `follow: bool` argument to Client request methods.
2. Annotated return types as `HttpResponse` instead of `Any`.
3. Made `Client` class inherit from `RequestFactory`, as it does in Django.

Wrt (2), these return types were reverted from `HttpResponse` to `Any` in PR #225. But I suspect that was simply to silence mypy warnings about "incompatible with supertype RequestFactory", not because there were any issues with the annotation.

Note that `Client.request()` monkey-patches the `HttpResponse` object with some additional attributes. Those attributes were already annotated before, I reordered and added additional comments to make it clear where they come from.

### DIFF
--- a/django-stubs/http/response.pyi
+++ b/django-stubs/http/response.pyi
@@ -62,25 +62,27 @@ class HttpResponseBase(Iterable[Any]):
     def __iter__(self) -> Iterator[Any]: ...
 
 class HttpResponse(HttpResponseBase):
-    client: Client
-    context: Context
     content: Any
     csrf_cookie_set: bool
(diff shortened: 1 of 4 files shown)


sobolevn (commented):
Awesome!

Can please add a typed test case for this?

intgr:
Pushed a test case & several Django suppressions (necessary for Django 2.2 only)

sobolevn (approved):
Awesome, thanks!

PR #323, opened 2020-02-13
MultipleObjectMixin.get_queryset can return Iterables

rik:
https://docs.djangoproject.com/en/3.0/ref/class-based-views/mixins-multiple-object/#django.views.generic.list.MultipleObjectMixin.get_queryset


django-stubs/views/generic/list.pyi
```diff
@@ -17,7 +17,7 @@ class MultipleObjectMixin(ContextMixin):
     paginator_class: Type[Paginator] = ...
     page_kwarg: str = ...
     ordering: Sequence[str] = ...
-    def get_queryset(self) -> QuerySet: ...
+    def get_queryset(self) -> Union[Iterable, QuerySet]: ...
```
sobolevn, line 20: `Iterable` is generic. Can you please add an argument there?
And also a test case for it. Because queryset part is highly complex.

sobolevn (changes requested):

rik:
I believe this was fixed in https://github.com/typeddjango/django-stubs/pull/909

PR #325, opened 2020-02-13
signing.loads allows use timedelta for max_age

artofhuman:
`loads` use `unsign` method  https://github.com/django/django/blob/stable/2.2.x/django/core/signing.py#L134 which allows use timedelta
https://github.com/django/django/blob/stable/2.2.x/django/core/signing.py#L191

`django-stubs` already have right annotation for [unsign](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/core/signing.pyi#L38)


sobolevn (approved):

sobolevn:
Thanks a lot!

PR #290, opened 2020-01-04
Fixes for Manager.from_queryset(), support for QuerySet.as_manager()

mkurnikov:
TODO: 

- [ ] log all new exceptions, instead of crushing the plugin
- [ ] ability to silence errors
- [ ] cleanup the code 

Fixes https://github.com/typeddjango/django-stubs/issues/265
Closes https://github.com/typeddjango/django-stubs/pull/287, closes https://github.com/typeddjango/django-stubs/pull/286, closes https://github.com/typeddjango/django-stubs/pull/272

PR #272, opened 2019-12-13
WIP: QuerySet.as_manager()

mkurnikov:
Fixes https://github.com/typeddjango/django-stubs/issues/265

PR #327, opened 2020-02-13
Make as_manager method use proper generic

### LINKED ISSUE #324: Make as_manager() in queryset generic instead of using Any
# Bug report

## What's wrong
`as_manager()` on querysets currently uses `Any` instead of `_T`.

Like so:

```python
# django-stubs/db/models/query.pyi
    @classmethod
    def as_manager(cls) -> Manager[Any]: ...
```
## How is that should be
Looks like a simple switch to _T fixes it:

```python
# django-stubs/db/models/query.pyi
    @classmethod
    def as_manager(cls) -> Manager[_T]: ...
```

See https://docs.djangoproject.com/en/3.0/topics/db/managers/#creating-a-manager-with-queryset-methods for more info.

silviogutierrez:
- Closes #324


django-stubs/db/models/query.pyi
```diff
@@ -41,7 +41,7 @@ class _BaseQuerySet(Generic[_T], Sized):
         hints: Optional[Dict[str, models.Model]] = ...,
     ) -> None: ...
     @classmethod
-    def as_manager(cls) -> Manager[Any]: ...
+    def as_manager(cls) -> Manager[_T]: ...
```
sobolevn, line 44: Can you please also test it? We have a lot of bugs in QuerySet / Manager topics.
And we want to be sure it is as stable as possible.

sobolevn (changes requested):
Thanks a lot for the fix!

lithammer:
Possible duplicate work in https://github.com/typeddjango/django-stubs/pull/290.

silviogutierrez:
@sobolevn : forgive me if I missed it, but is there a guide to running or developing locally? I'm trying to run tests and doing it through CI passes really takes a while. Got a lot of other fixes too that I'd love to contribute back.

I tried replicating the Travis steps but no luck.

An example here: https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/unit-tests/#running-the-unit-tests

sobolevn:
Yes, we have a really short guide: https://github.com/typeddjango/django-stubs/blob/master/CONTRIBUTING.md

It is not perfect and I am open for any feedback!

RJPercival:
Has this been abandoned?

sciyoshi:
We are waiting for this change as well as it is blocking us upgrading to Python 3.9. Is this still only waiting on tests @sobolevn?

sciyoshi:
It does look like the PR includes a test already - I'm happy to help if there's anything needed to move this forward.

sobolevn:
I would love to merge this after the CI is fixed.

sciyoshi:
After investigating a bit more - looks like this simple approach doesn't quite work and it will probably need some of the more advanced machinery from #272 or #290. I'd recommend closing this PR in favor of one of the other two unless @silviogutierrez has some further input on it.

silviogutierrez:
@sciyoshi : yea, #272 and #290 are much better solutions. Thanks all.

PR #331, opened 2020-02-16
Add db.models.fields.Field method stubs for custom Fields

danifus:
Add stubs for: pre_save, get_db_prep_value, get_db_prep_save

These methods are covered in the django docs and are sometimes used when creating custom fields.


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -108,7 +108,10 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
     def set_attributes_from_name(self, name: str) -> None: ...
     def db_type(self, connection: Any) -> str: ...
     def db_parameters(self, connection: Any) -> Dict[str, str]: ...
+    def pre_save(self, model_instance: Model, add: bool) -> Any: ...
     def get_prep_value(self, value: Any) -> Any: ...
+    def get_db_prep_value(self, value: Any, connection: Any, prepared: bool) -> Any: ...
```
sobolevn, line 113: Are you sure that `connection` is `Any` here?

sobolevn (commented):
Thanks a lot for your help. Just a quick question.

django-stubs/db/models/fields/__init__.pyi
danifus: It's consistent with the other `connection` variables in this file (and across the stubs project from a quick look) but I think it could be compatible with subclasses of `db.backends.base.base.BaseDatabaseWrapper`. Does that look reasonable to you?

sobolevn: Ok, thanks! Let's leave as it is for know. We can create a new issue to work on `connection: Any` latter. Because we try to remove as many `Any` types as we can. @danifus will you please fire a new issue for this?

danifus: I've created #332 to track the `Any` -> `db.backends.base.base.BaseDatabaseWrapper` change.

Will this issue be merged or did you actually want me to create an issue for _this_ merge request :)

sobolevn: Thanks a lot!

sobolevn (approved):

PR #335, opened 2020-02-27
Fix input types of humanize functions

### LINKED ISSUE #334: Stubs for humanize functions should also accept integers
# Bug report

## What's wrong

`ordinal`, `apnumber` and `intword` in [`django.contrib.humanize.templatetags.humanize`](https://docs.djangoproject.com/en/1.11/ref/contrib/humanize/) also accept integers (or anything that can be passed into `int()`)

```
>>> from django.contrib.humanize.templatetags.humanize import apnumber
>>> apnumber(3)
'three'
>>> apnumber(3.0)
'three'
>>> apnumber(Decimal(3))
'three'
>>> apnumber('3')
'three'
```

However running mypy over it produces 
```
error: Argument 1 to "apnumber" has incompatible type "int"; expected "Optional[str]"
```

## How is that should be

`ordinal`,  `apnumber` and `intword` should also accept any input [which have a `__int__` method](https://docs.python.org/3.8/library/typing.html#typing.SupportsInt)

## System information

- OS:
- `python` version: 3.8.1
- `django` version: 2.2.10
- `mypy` version: 0.761
- `django-stubs` version:1.4.0

kevinmarsh:
As mentioned in issue #334 we're being a bit to strict for `ordinal`, `apnumber` and `intword` in [django.contrib.humanize.templatetags.humanize](https://github.com/django/django/blob/stable/3.0.x/django/contrib/humanize/templatetags/humanize.py) only accepting strings as inputs as they can also accept integers.

## Related issues

Closes #334


sobolevn (approved):
Thank you!

sobolevn:
This CI fail is unrelated. We are working on a fix.

PR #338, opened 2020-03-05
Change patch_response_headers argument to Optional[int] (Fixes #337)

### LINKED ISSUE #337: patch_response_headers expecting float for number of seconds
## What's wrong
`patch_response_headers` from `django-stubs.utils.cache.pyi` expects `float` as the second parameter.

Original code:
https://github.com/django/django/blob/master/django/utils/cache.py#L249
Stubs source:
https://github.com/typeddjango/django-stubs/blob/master/django-stubs/utils/cache.pyi#L18

## How is that should be
The source code of django says:
> cache_timeout is in seconds.

Why would it be a float? I checked the history (of the stubs), it used to be an int but was later changed without apparent reason. Also it's nullable, so it could even be `Optional[int]`. Correct me if I'm wrong. I'm willing to do the patching if this is confirmed.

## System information

- OS: ArchLinux
- `python` version: 3.7
- `django` version: 3.0.2
- `mypy` version: 0.761

MarcinWieczorek:
# I have made things!

## Related issues
- Closes #337 

If you have any ideas for tests please suggest them here. Travis fails anyway, but it seems that not because of my fix :)


sobolevn (approved):

Emru1:
any progress with this pull request?

sobolevn:
@MarcinWieczorek can you please rebase it and fix the CI? I would love to get this merge once it is done.

sobolevn (approved):

sobolevn:
Thanks, sorry for the long wait!

MarcinWieczorek:
Cool, thanks

PR #341, opened 2020-03-13
Update to mypy 0.770

intgr:
Mypy was updated to 0.770, see release notes: https://mypy-lang.blogspot.com/2020/03/mypy-0770-released.html

### DIFF
--- a/scripts/enabled_test_modules.py
+++ b/scripts/enabled_test_modules.py
@@ -198,7 +198,7 @@ IGNORED_ERRORS = {
     ],
     'from_db_value': [
         '"Cash" has no attribute',
-        'Argument 1 to "__str__" of "Decimal"',
+        '"__str__" of "Decimal"',
     ],
     'get_object_or_404': [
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks!

intgr:
Beware, I haven't tested the mypy plugin yet, didn't expect it to be merged so fast :D

sobolevn:
The CI passed, so I am confident that it works! 

Is there something I should reconsider? 🙂

intgr:
Cool! I don't have any reason to expect breakage. I will test it with my project soon and let you know if there are any surprises.

scripts/enabled_test_modules.py
```diff
@@ -198,7 +198,7 @@
     ],
     'from_db_value': [
         '"Cash" has no attribute',
-        'Argument 1 to "__str__" of "Decimal"',
+        '"__str__" of "Decimal"',
```
intgr, line 201: Submitted PR to Django to fix their broken test code: https://github.com/django/django/pull/12726

PR #345, opened 2020-03-23
Add RESTRICT and RestrictedError to django.db.models.deletion

sidmitra:
Documentation: https://github.com/django/django/blob/master/django/db/models/deletion.py 

These errors have been in this module for a few years, so are not new.


sobolevn (approved):

sidmitra:
I've updated the failing black formatting issue.

sobolevn (approved):
Thanks!

PR #348, opened 2020-03-29
Add state attribute to Model (#347)

### LINKED ISSUE #347: "Model" has no attribute "_state"
# Bug report

The `mypy` reports that model is missing the `_state` attribute. I've checked the source code and it looks like the `ModelState` is defined, but it isn't included into `Model`.

## What's wrong
Run mypy on:
```python
def is_updated(instance: Model):
    return not instance._state.adding
```

And you will get an error:
```
error: "Model" has no attribute "_state"
```
## How is that should be

No error should be reported since `Model` has the `_state` attribute.

## System information

- OS:
- `python` version: 3.8
- `django` version: 2.2
- `mypy` version: 0.770

I would be happy to provide PR, since the fix is straightforward.

stevanmilic:
# I have made things!

The `_state` attribute is added to `Model` to provide type hints
when its accessed. Altough the `ModelState` was defined it wasn't
included into `Model`.

## Related issues

- Closes #347
- Refs #347


django-stubs/db/models/base.pyi
```diff
@@ -7,6 +7,13 @@ from django.db.models.options import Options
 
 _Self = TypeVar("_Self", bound="Model")
 
+class ModelStateFieldsCacheDescriptor: ...
+
+class ModelState:
+    db: None = ...
```
sobolevn: Is this type correct? Looks like we are missing its other value. And it has to be `Optional[...]`

sobolevn (changes requested):
Thanks a lot! 

When reviewing this PR - I have noticed that we are also missing `from_db` class method that is highly related to the model state: https://github.com/django/django/blame/b9336b78cf2a9a29f4934041c9e221bc68daec80/django/db/models/base.py#L506 

Can you please also add it? Then we can decide in `db: None` issue. :+1:

django-stubs/db/models/base.pyi
stevanmilic: It should be `Optional[str]`

sobolevn (approved):
Thanks a lot!

PR #350, opened 2020-03-31
Improve type of BaseForms.files

quinox:
# I made a thing!

I needed to change the type of `BaseForm.files`, otherwise I can't correctly typecheck my existing, working code. I'm just started with MyPy so I wouldn't surprised if I'm wrong, but this change works for me and doesn't fail the existing testcases.

My existing class:
```python
class MyForm(Form):
  myparam = MultiFileField(...)

  self.clean(self, *args, **kwargs):
      self.files.getlist('myparam', [])
```

The error I got with the old type-hint:
```python
myproject/myapp/forms.py:74:35: error: "Dict[str, Any]" has no attribute "getlist"  [attr-defined]
                combined_files.extend(self.files.getlist('files_0', []))
```

I ran `type(self.files)` in a PDB shell to find the type (`MultiValueDict`), then looked at existing type-hinting and copied the line from `http/request.pyi:HttpRequest.FILES`. 

## Related issues

None that I know of.


sobolevn (approved):
Thanks a lot!

sobolevn:
Just for the reference: it is not `Optional[]` because of this line: https://github.com/django/django/blob/master/django/forms/forms.py#L71

PR #354, opened 2020-04-06
Issue 309

### LINKED ISSUE #309: Can't assign AnonymousUser to request.user
# Bug report

I'm implementing my own logout routine and need to assign AnonymousUser to request.user like:

```
        if hasattr(request, 'user'):
            from django.contrib.auth.models import AnonymousUser

            request.user = AnonymousUser()
```

## What's wrong
The code above has the error:

error: Incompatible types in assignment (expression has type "AnonymousUser", variable has type "AbstractBaseUser")

## How is that should be

request.user should allow both AnonymousUser and AbstractBaseUser

## System information

- OS: Ubuntu 18.04
- `python` version: 3.6
- `django` version: 2.2
- `mypy` version: 0.761

kszmigiel:
# Fix for #309 

- Closes #309


kszmigiel:
@sobolevn according to [this](https://github.com/typeddjango/django-stubs/issues/309#issuecomment-580208162) comment I need to implement logic to discriminate one of the types. I suppose it goes somewhere into mypy django plugin but could you give me a hint where exactly? Thanks in advance

sobolevn (commented):
Thanks a lot!

Can tagged unions be applied here? https://mypy.readthedocs.io/en/stable/literal_types.html#tagged-unions

kszmigiel:
@sobolevn Could you take a look at tests results? I have troubles while importing Literals and unfortunately no ideas how to solve it. What I implemented should work in general, but tests don't pass :/

sobolevn:
@kszmigiel `Literal` is in `typing_extensions` for `python<=3.7` and is in `typing` for `python3.8+`

django-stubs/http/request.pyi
```diff
@@ -51,7 +52,7 @@ class HttpRequest(BytesIO):
     resolver_match: ResolverMatch = ...
     content_type: Optional[str] = ...
     content_params: Optional[Dict[str, str]] = ...
-    user: AbstractBaseUser
+    user: Union[AbstractBaseUser, AnonymousUser]
```
sobolevn, line 55: Let's test it! In the best world `is_authenticated` function should resolve this type.

sobolevn (changes requested):
Thanks a lot!

Next thing we need is a type test (and possibly `is_authenticated` tweak).

django-stubs/http/request.pyi
kszmigiel: I'm not sure how to test it properly yet (unfortunately I'm not so good at testing) so please label it with "help wanted", maybe someone will be able to do this, or I will get back to this later with more experience.

sobolevn: You can have a look at https://github.com/typeddjango/django-stubs/blob/master/test-data/typecheck/models/test_contrib_models.yml and https://github.com/typeddjango/django-stubs/blob/master/test-data/typecheck/contrib/auth/test_decorators.yml

kszmigiel: I've tested the tags, but still have no idea how to make `is_authenticated` resolve `Union[]` type. Any further tips?

test-data/typecheck/test_request.yml
```diff
@@ -14,4 +14,8 @@
             content: |
                 from django.db import models
                 class MyUser(models.Model):
-                    pass
\ No newline at end of file
+                    pass
+-   case: request_object_user_type
+    main: |
+        from django.http.request import HttpRequest
+        reveal_type(HttpRequest.user.tag)  # N: Revealed type is 'Union[Literal['abstract_base_user'], Literal['anonymous_user']]'
```
sobolevn, line 21: Awesome! Now we need to check that `is_authenticated` can potentially check for this tag and discriminate types based on it.

sobolevn: (I am still not sure it is possible, but we are experimenting here)

kszmigiel: Well, I think it can be somehow done in `mypy_django_plugin`, but I'll need more time to investigate it. I also noticed that such such piece of code:

```

def index(request: HttpRequest):
    reveal_type(request.user.tag)
    reveal_type(request.user)
    reveal_type(HttpRequest.user.tag)
(...)
```

returns this in mypy:

```
polls/views.py:6: note: Revealed type is 'Literal['abstract_base_user']'
polls/views.py:7: note: Revealed type is 'django.contrib.auth.models.User'
polls/views.py:8: note: Revealed type is 'Union[Literal['abstract_base_user'], Literal['anonymous_user']]'
```

but I haven't discovered why is so yet. Maybe type discrimination is already done somewhere earlier in plugin's runtime?

sobolevn: We can probably ask this question in https://gitter.im/python/typing
Because I have not tried this or even seen this anywhere.

kszmigiel: I've asked a question a few days ago, but got no response yet :/

django-stubs/contrib/auth/base_user.pyi
```diff
@@ -1,9 +1,15 @@
+import sys
 from typing import Any, Optional, Tuple, List, overload, TypeVar
 
 from django.db.models.base import Model
 
 from django.db import models
 
+if sys.version_info[1] < 8:
```
smithdc1: This could be `if sys.version_info < (3, 8):`
(I find this slightly more readable)

sobolevn: 👍

kszmigiel: done

test-data/typecheck/test_request.yml
kszmigiel: @sobolevn should I still look for an answer or just rebase to squash commits and we're ready to merge?

bullfest: Hey @kszmigiel 
I wrote this based on what I understood from the mypy-documentation, might be something I'm missing, but I believe this is what we're looking for.

This type-checks and accessing `foo`/`bar` on the wrong branch in the if gives the expected errors.

```
from typing import Literal, Union

class Foo:
    is_two: Literal[False]
    bar = "heyy"

class Foo2:
    is_two: Literal[True]
    foo = "Ho"

Foos = Union[Foo, Foo2]

def f(f: Foos):
    if f.is_two:
        print(f.foo)
    else:
        print(f.bar)
```

So I'd think that we'd have to just annotate `is_anonymous` with the `Literal` types in AbstractBaseUser/AnonymousUser, and then it'd work.

Can you fix this or should I open up a new PR based on yours?

bullfest: Hey again @kszmigiel I opened #383 with the change I proposed here. I hope you're okay with me continuing on the work in this PR. :smile:

kszmigiel: @bullfest sorry I had no time to respond to this. Again, great job!

bullfest: No worries, I had no expectation that you would respond in the hour between my comments, it was simply my curiosity and willingness to contribute that couldn't hold me back from opening a PR :sweat_smile:

sobolevn:
Closing, see #383

PR #356, opened 2020-04-13
Fix Modelform `auto_id` to accept Boolean type.

onyekaa:
## I fixed a thing.

Forms have an optional `auto_id` property that is either True, False or a specifically formatted string [per the doc](https://docs.djangoproject.com/en/3.0/ref/forms/api/#django.forms.Form.auto_id). It defaults to a string. While going through existing code I saw that when I set `self.auto_id=True`, it threw:

```
Cannot assign member "auto_id" for type "RoleForm"
  Expression of type "Literal[True]" cannot be assigned to member "auto_id" of class "RoleForm"
    "bool" is incompatible with "str"
```
because one of its two annotations was set to only `str`. This has been fixed.

## Related issues
None


sobolevn (approved):
Thanks a lot!

PR #361, opened 2020-04-23
Add add_to_class method to Model

JorgenPhi:
Add stub for `add_to_class` method in ModelBase. [Link to Django method](https://github.com/django/django/blob/stable/3.0.x/django/db/models/base.py#L323)

## Related issues
Refs #204


django-stubs/db/models/base.pyi
```diff
@@ -27,6 +27,7 @@ class Model(metaclass=ModelBase):
     _state: ModelState
     def __init__(self: _Self, *args, **kwargs) -> None: ...
     @classmethod
```
sobolevn, line 29: You are missing `@classmethod` decorator for `from_db` now 🙂

sobolevn (changes requested):

django-stubs/db/models/base.pyi
JorgenPhi: Thanks

sobolevn (approved):

sobolevn:
Thanks a lot for the fix!

PR #363, opened 2020-04-24
Fix wrong BaseCache typings

### LINKED ISSUE #362: Invalid typings for django.core.cache.backends.base.BaseCache
# Bug report

`django.core.cache.backends.base.Base` typings are invalid. 

According to https://github.com/django/django/blob/447980e72ac01da1594dd3373a03ba40b7ee6f80/django/core/cache/backends/base.py#L108, add operation should be returning `True` if the value was stored, otherwise `False`, but in the typings, it's set to None. 

https://github.com/typeddjango/django-stubs/blob/96914e466bcd7a65fdd92422586f9ebc72fdf8a8/django-stubs/core/cache/backends/base.pyi#L22

Similarly other typings for BaseCache have wrong typings.

## What's wrong

```
foobar.py:688: error: "add" of "BaseCache" does not return a value
```

## How is that should be

Shouldn't raise any error. 

## System information

- `django` version: master branch
- `django-stubs` version: master branch

Ashish-Bansal:
## Description 

Fixes #362

django.core.cache.backends.base.BaseCache were invalid. This PR fixes them in accordance with https://github.com/django/django/blob/447980e72ac01da1594dd3373a03ba40b7ee6f80/django/core/cache/backends/base.py


Ashish-Bansal:
I'm not sure if it's worth adding tests, but I'm not really familiar with the mypy testing framework. 


```
-   case: test_caching_operations
    main: |
        from django.core.cache import cache

        key = "foo"
        value = "bar"

        bool_value = cache.add(key, value)
        reveal_type(bool_value)  # N: Revealed type is 'bool'

        bool_value = cache.touch(key)
        reveal_type(bool_value)  # N: Revealed type is 'bool'

        bool_value = cache.has_key(key)
        reveal_type(bool_value)  # N: Revealed type is 'bool'
```

Will they be around these lines ?

sobolevn (approved):
Thanks a lot!

sobolevn:
@Ashish-Bansal tests are not required for this case. Thanks again! 👍

PR #368, opened 2020-05-08
Update widgets.pyi

jdahlin:
Make ForeignKeyRawIdWidget.rel optional


django-stubs/contrib/admin/widgets.pyi
```diff
@@ -30,10 +30,10 @@ class AdminFileWidget(forms.ClearableFileInput): ...
 def url_params_from_lookup_dict(lookups: Any) -> Dict[str, str]: ...
 
 class ForeignKeyRawIdWidget(forms.TextInput):
-    rel: ManyToOneRel = ...
+    rel: Optional[ManyToOneRel] = ...
     admin_site: AdminSite = ...
     db: None = ...
-    def __init__(self, rel: ForeignObjectRel, admin_site: AdminSite, attrs: None = ..., using: None = ...) -> None: ...
+    def __init__(self, rel: Optional[ForeignObjectRel], admin_site: AdminSite, attrs: None = ..., using: None = ...) -> None: ...
```
sobolevn, line 36: Can you please also fix `using` and `attrs` types? Looks like they should not be `None`.

sobolevn: It also not obvious to me that `rel` can be `None`. This code has zero references of `None`: https://github.com/django/django/blob/3a807a6f597a97b0fa514039a8ea41fb90769981/django/contrib/admin/widgets.py#L135

Can you please share code sample that prove this point?

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -106,8 +106,8 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
     def __get__(self: _T, instance, owner) -> _T: ...
     def deconstruct(self) -> Any: ...
     def set_attributes_from_name(self, name: str) -> None: ...
-    def db_type(self, connection: Any) -> str: ...
-    def db_parameters(self, connection: Any) -> Dict[str, str]: ...
+    def db_type(self, connection: Any) -> Optional[str]: ...
```
sobolevn, line 109: These ones are confirmed:
- https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L694
- https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L717-L722

sobolevn (changes requested):
Thanks a lot for the fix!

Can you please also run `black` to fix the CI?

PR #372, opened 2020-05-23
Allow template render to string helper functions to accept Context

### LINKED ISSUE #371: render_to_string context argument should accept Context
# Bug report

## What's wrong

```
render_test.py:8: error: Argument 2 to "render_to_string" has incompatible type "RequestContext"; expected "Optional[Dict[str, Any]]"
```

```python
link = render_to_string(link_template, RequestContext(request))
```

This is because:

https://github.com/typeddjango/django-stubs/blob/963d50c717282c340fa7d397bdb0402a4e7f9434/django-stubs/template/loader.pyi#L9-L14

## How is that should be

The `render_to_string` function simply passes the `context` argument to the `template.render` function. This function already accepts the correct Union of arguments:

https://github.com/typeddjango/django-stubs/blob/b397ec04a95e009c7f86b6bcb7325d2059384dc1/django-stubs/template/base.pyi#L62-L64

Also, `django.shortcuts.render` and `django.shortcuts.render_to_response` both should also accept the same type for context arg, and currently they do not:

https://github.com/typeddjango/django-stubs/blob/963d50c717282c340fa7d397bdb0402a4e7f9434/django-stubs/shortcuts.pyi#L9-L23

## System information

- OS: MacOS Catalina 10.15.4 (19E287)
- `python` version: 3.8.0
- `django` version: 2.2
- `mypy` version: 0.6.1

I am taking a stab at fixing this in my fork, will submit a PR when ready

asfaltboy:
## Related issues

- Closes #371

### DIFF
--- a/django-stubs/shortcuts.pyi
+++ b/django-stubs/shortcuts.pyi
@@ -9,10 +9,11 @@ from django.http.response import (
 
 from django.db.models import Manager, QuerySet
 from django.http import HttpRequest
+from django.template.context import Context
 
 def render_to_response(
     template_name: Union[str, Sequence[str]],
-    context: Optional[Dict[str, Any]] = ...,
+    context: Optional[Union[Context, Dict[str, Any]]] = ...,
     content_type: Optional[str] = ...,
     status: Optional[int] = ...,
     using: Optional[str] = ...,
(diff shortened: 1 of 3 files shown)


sobolevn (approved):
LGTM 👍

asfaltboy:
Dangit, I think I was wrong to assume all functions should accept Context objects! We should revert this PR, or at least remove the Context types for the following:

* [django.shortcuts.render][1]
* [django.template.loader.render_to_string][2]

[1]: https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#django.shortcuts.render
[2]: https://docs.djangoproject.com/en/2.2/topics/templates/#django.template.loader.render_to_string

To pass Context to these functions, we must call [Context.flatten()][3]

[3]: https://docs.djangoproject.com/en/2.2/ref/templates/api/#django.template.Context.flatten

sobolevn:
@asfaltboy that's ok! Thanks! Waiting for the fix 👍

PR #374, opened 2020-05-29
Allow form.save() in CreateView / UpdateView

quinox:
# I made another thing!

Hello again, after my glorious victory of #350 I've attempted to fix another typing issue I was faced with. 

## Related issues

It looks like #261 is related but I don't think this PR fixes that issue: I'm only fixing `CreateView` and `UpdateView`, not the generic case of `FormView` having a `ModelForm`. I looked at `NewSemanalDjangoPlugin.get_method_hook`, but based on the MyPy documentation about extending MyPy I don't see how to alter the type of the arguments going into `def form_valid`, it only describes how to change the return type.

# Implementation details

This time I managed to even add a testcase! It's not the prettiest testcase in the world, but it was made with love in a test-driven way.

The core solution is simply adding a different `form_valid` signature to `ModelFormMixin`. Just adding it doesn't work though because then it clashes with the parent's `FormMixin` implementation. Therefore I was forced to split `FormMixin` into two classes, `AbstractFormMixin` (everything except `form_valid`) and `FormMixin` (just `form_valid`).

The name `AbstractFormMixin` does not appear in the Django codebase, so I deemed it safe to use, but if you know of a better class name that doesn't risk people confusing this django-stub-specific-class with something that belongs to Django I'll gladly update this PR.

Cheers.


sobolevn:
Thanks a lot, @kszmigiel can you please review this PR? 🙂

test-data/typecheck/test_forms.yml
```diff
@@ -46,6 +46,18 @@
                 reveal_type(self.get_form(form_class))  # N: Revealed type is 'main.MyForm'
                 reveal_type(self.get_form(MyForm2))  # N: Revealed type is 'main.MyForm2'
 
+-   case: updateview_form_valid_has_form_save
```
kszmigiel, line 49: Would be cool to have it tested for CreateView too, the rest LGTM 👍

kszmigiel (approved):

test-data/typecheck/test_forms.yml
kszmigiel: @sobolevn @quinox I've forgotten to submit the review, sorry for that.

sobolevn (approved):

sobolevn:
Thanks a lot!

PR #375, opened 2020-05-29
Increase accuracy of ModelAdmin types

antonagestam:
Hi,

I started working on narrowing some types for `ModelAdmin` and reducing `Any`s. I got a bit ahead of myself and thought I'd make a PR before spending more time on this to verify you're willing to review/work with me once finalized.

I got a [strange error from pytest](https://gist.github.com/antonagestam/54430b1c02081fabd6bbae1eb9046368) on every failure, so I had a hard time working out which errors to expect, so that needs another iteration to fix the test cases. Usually I would try running a test case, see what error is yielded and verify that it makes sense, and then add its `# E: ...` comment. Do you have any ideas about what might be wrong with my setup?

I also ran into a few properties that aren't documented, I've added TODO comments on them and figured maybe we should remove them from the stubs entirely? What do you think?

Edit: I was also unsure about how to introduce "stub-only" types, will prefixing with underscore work, or will those types be exposed as well?


django-stubs/contrib/admin/options.pyi
```diff
@@ -26,8 +30,11 @@ from django.db.models.fields import Field
 
 IS_POPUP_VAR: str
 TO_FIELD_VAR: str
-HORIZONTAL: Any
-VERTICAL: Any
+HORIZONTAL: Final[Literal[1]] = ...
```
sobolevn: Isn't `Literal[1]` just enough?

django-stubs/contrib/admin/options.pyi
```diff
@@ -26,8 +30,11 @@ from django.db.models.fields import Field
...
+VERTICAL: Final[Literal[2]] = ...
+
+# TODO: Is this ok? Is this the way to mark a type as stub-only?
+_Direction = Literal[1, 2]
```
sobolevn: I guess `Union[Literal[1], Literal[2]]` is more readable.

django-stubs/contrib/admin/options.pyi
```diff
@@ -37,24 +44,31 @@ class IncorrectLookupParameters(Exception): ...
...
+    radio_fields: Mapping[str, _Direction] = ...
+    prepopulated_fields: Mapping[str, Sequence[str]] = ...
+    formfield_overrides: Mapping[Type[Field], Mapping[str, Any]] = ...
+    readonly_fields: Sequence[str] = ...
+    ordering: Sequence[str] = ...
+    sortable_by: Sequence[str] = ...
     view_on_site: bool = ...
     show_full_result_count: bool = ...
-    checks_class: Any = ...
+    checks_class: Any = ...  # TODO: This seems to be undocumented, should we remove it?
```
sobolevn: Let it be `Any` for now

django-stubs/contrib/admin/options.pyi
```diff
@@ -93,35 +107,35 @@ class BaseModelAdmin:
     def has_module_permission(self, request: HttpRequest) -> bool: ...
 
 class ModelAdmin(BaseModelAdmin):
-    list_display: Sequence[Union[str, Callable]] = ...
+    list_display: Sequence[Union[str, Callable[[Model], Any]]] = ...
```
sobolevn, line 116: Can we reduce `Any` here? Or is it hard to do?

django-stubs/contrib/admin/options.pyi
```diff
@@ -93,35 +107,35 @@ class BaseModelAdmin:
...
-    action_form: Any = ...
+    add_form_template: str = ...
+    change_form_template: str = ...
+    change_list_template: str = ...
+    delete_confirmation_template: str = ...
+    delete_selected_confirmation_template: str = ...
+    object_history_template: str = ...
+    popup_response_template: str = ...
+    actions: Sequence[Callable[[ModelAdmin, HttpRequest, QuerySet], None]] = ...
+    action_form: Any = ...  # TODO: Undocumented, remove?
```
sobolevn: Let's leave all undocumented feature for now.

sobolevn (commented):
Looks great! Very big thanks! Much awesome!

django-stubs/contrib/admin/options.pyi
antonagestam: Gotcha! 👍

django-stubs/contrib/admin/options.pyi
antonagestam: I actually went for that at first, with a union of callables returning `str` or `bool` which are the two cases mentioned in the docs, but I now think this a case of the docs being vague. I think any type that is `str()`-able is a valid return type here, which is virtually any Python type, right? We could probably change to `object` here, not sure if that adds any value though.

django-stubs/contrib/admin/options.pyi
antonagestam: I'd disagree about the readability, but I'll update accordingly.

antonagestam:
I managed to work around the strange pytest error by just putting the test case in a file and running mypy on it ... incredible how fast one gets used to nice tools!

django-stubs/contrib/admin/options.pyi
```diff
@@ -26,8 +30,11 @@ from django.db.models.fields import Field
 
 IS_POPUP_VAR: str
 TO_FIELD_VAR: str
-HORIZONTAL: Any
-VERTICAL: Any
+HORIZONTAL: Literal[1] = ...
+VERTICAL: Literal[2] = ...
+
+# TODO: Is this ok? Is this the way to mark a type as stub-only?
```
antonagestam: @sobolevn What's your opinion here? Do you use stub-only types in other places? Is this consider OK?

django-stubs/contrib/admin/options.pyi
sobolevn: Ok, let's leave `Any`. Thanks a lot for the explanation!

django-stubs/contrib/admin/options.pyi
sobolevn: Yes, this is fine.

sobolevn (approved):
Thanks 👍

django-stubs/contrib/admin/options.pyi
```diff
@@ -227,9 +247,9 @@ class ModelAdmin(BaseModelAdmin):
     def history_view(self, request: HttpRequest, object_id: str, extra_context: None = ...) -> HttpResponse: ...
 
 class InlineModelAdmin(BaseModelAdmin):
-    model: Any = ...
-    fk_name: Any = ...
-    formset: Any = ...
+    model: Type[Model] = ...
+    fk_name: str = ...
+    formset: BaseFormSet = ...
```
sssbox, line 252: I think this should be: `formset: Type[BaseFormSet] = ...`

sobolevn: @sssbox, please submit a PR if you think that this is correct.

PR #384, opened 2020-06-02
Issue 382

### LINKED ISSUE #382: Incorect type for items method of Sitemap class
# Bug report

## What's wrong
When using the Sitemap class in the following way: 
```py
class ProductSitemap(Sitemap):

    def items(self) -> QuerySet[Product]:
        return Product.objects.all()
```
Will raise the following error:
`Return type "QuerySet[Product]" of "items" incompatible with return type "List[Any]" in supertype "Sitemap"`

## How is that should be

The given snippet should not raise any error. In the Django documentation we can find some example [using a QuerySet](https://docs.djangoproject.com/en/3.0/ref/contrib/sitemaps/#an-example). The type defined in [django-stubs](https://github.com/typeddjango/django-stubs/blob/5c6be7ad121abb2bc8020b0a51c5c0a3981f7216/django-stubs/contrib/sitemaps/__init__.pyi#L19) seems a little bit to restrictive.
It could be replace by: 
- `Iterable[Any]`
- Something similar to [what is done in the paginator](https://github.com/typeddjango/django-stubs/blob/ef8850348b5862d39bfe529a39e71625c5d9c89d/django-stubs/core/paginator.pyi#L27). As the Sitemap class from Django uses Paginator under the hood, this seems the best option to me.

I can do a PR if option number 2 seems good to you.

## System information

- OS:
- `python` version: 3.7.7
- `django` version: 2.2.12
- `mypy` version: 0.770

kszmigiel:
- Closes #382


django-stubs/contrib/sitemaps/__init__.pyi
```diff
@@ -13,10 +13,19 @@ class SitemapNotFound(Exception): ...
 
 def ping_google(sitemap_url: Optional[str] = ..., ping_url: str = ...) -> None: ...
 
+class SupportsLen(Protocol):
+    def __len__(self) -> int: ...
+
+class SupportsCount(Protocol):
+    def count(self) -> int: ...
+
+class SupportsOrdered(Protocol):
```
sobolevn: Thanks a lot! 

The only thing I see here is that we tend to use leading `_` for our internal types. 
Here's where this mistake comes from: https://github.com/typeddjango/django-stubs/blob/ef8850348b5862d39bfe529a39e71625c5d9c89d/django-stubs/core/paginator.pyi#L27

Can you please fix it in both places?

sobolevn (changes requested):

django-stubs/contrib/sitemaps/__init__.pyi
kszmigiel: sure!

sobolevn (approved):

sobolevn:
Thanks a lot!

kszmigiel:
No problem, trying to do my best 😁

PR #385, opened 2020-06-03
Fix parameter types for assertJSONEqual/NotEqual

bullfest:
The previous types were not quite as permissive as one would like (notably the values in the Dict-part don't necessarily have to be strings).

Ideally one would be able to define a recursive type:
```python
JSONType = Union[Dict[str, JSONType], List[JSONType], int, str, float, bool, None]
```
But as MyPy doesn't support this I'd say that either this solution is probably the best we can do.


sobolevn (approved):

sobolevn:
Thanks!

PR #391, opened 2020-06-06
Issue 379

### LINKED ISSUE #379: [Proposal] Tests for Python 3.8
Hi!

During writing tests for my contributions I've noticed that `typing` module is slightly different for Python 3.7 and 3.8, i.e. `TypedDict` is part of `typing_extensions` in 3.7, whereas it's included in `typing` in 3.8. In case we plan to support 3.8 in the future I'd suggest to:

1) refactor our tests so they'd work for both versions
2) go hardcore and drop testing for 3.7 and switch to newer (this would probably be easier)

What do you think guys?

kszmigiel:
- Closes #379


sobolevn (approved):
Great work, @kszmigiel! 

Thanks a lot!

PR #393, opened 2020-06-08
Add more types to HttpResponse and StreamingHttpResponse

danifus:
Type hint for `HttpResponse.getvalue()` was missing.

Also adds more specific types for:

- `HttpResponse.content`
- `StreamingHttpResponse.streaming_content`

Not in this PR but as `StreamingHttpResponse.content` just raises `AttributeError` should we exclude it from the type hint for that property?

```
class StreamingHttpResponse(...):

    @property
    def content(self):
        raise AttributeError(
            "This %s instance has no `content` attribute. Use "
            "`streaming_content` instead." % self.__class__.__name__
        )
```


django-stubs/http/response.pyi
```diff
@@ -62,7 +62,10 @@ class HttpResponseBase(Iterable[Any]):
     def __iter__(self) -> Iterator[Any]: ...
 
 class HttpResponse(HttpResponseBase):
-    content: Any
+    @property
+    def content(self) -> bytes: ...
+    @content.setter
+    def content(self, value: Any): ...
```
sobolevn: Shouldn't `setter` be annotated as `-> None`?

django-stubs/http/response.pyi
```diff
@@ -83,12 +86,16 @@ class HttpResponse(HttpResponseBase):
...
     def json(self) -> Any: ...
+    def getvalue(self) -> bytes: ...
 
 class StreamingHttpResponse(HttpResponseBase):
     content: Any
-    streaming_content: Iterator[Any]
+    @property
+    def streaming_content(self) -> Iterator[bytes]: ...
+    @streaming_content.setter
+    def streaming_content(self, value: Iterable[Any]): ...
```
sobolevn: The same here.

sobolevn (commented):
> Not in this PR but as StreamingHttpResponse.content just raises AttributeError should we exclude it from the type hint for that property?

I think we can ignore it for now.

Thanks a lot for your PR!

django-stubs/http/response.pyi
danifus: Yeah, they should be there. I've added another commit for that.

sobolevn (approved):

danifus:
Hi,

Just checking in to see if this was ok to merge? The build failure was due to the pipeline issues you had in June. Is there a way to clear them?

sobolevn:
@danifus yes, can you please rebase the build?

danifus:
Looks like having different types accepted for a setter and returned from a getter is not supported yet:

https://github.com/python/mypy/issues/3004

That thread initially says that it won't be supported but during discussion they decide it is a common idiom and will support it once an approach is agreed upon.

I'll leave the `content` property alone for the time being and create a new pull request for adding the missing `getvalue()` method. Feel free to close this pull request or leave it open for when the mypy fixes arrive.

sobolevn:
@danifus agreed, let's wait for the `mypy` fix! 👍

terencehonles:
@danifus I believe you should be able to close this now that #712 is merged since this is approximately the same change

danifus:
Thanks @terencehonles! Nice work :)

PR #401, opened 2020-06-16
🐑 Update update_session_auth_hash() stub

### LINKED ISSUE #344: Upgrading to mypy 0.770 and Django-stubs 1.5, breaks mypy_django_plugin
# Bug report
Upgrading mypy and Django-stubs breaks the mypy_django_plugin

## What's wrong

* I did `poetry lock` in my module
* This updated the versions of my dependencies
* This breaks the type checks that I do using `mypy .`. It works without doing these updates

## How is that should be
This should not break my `mypy .` typechecks

## System information

- OS: linux
- `python` version: 3.7.4
- `django` version: 3.0.4
- `mypy` version: 0.770
- `django-stubs` version: 1.5

### LINKED ISSUE #400: update_session_auth_hash() type doesn't match documentation and functionality
# Bug report

## What's wrong

The stub for update_session_auth_hash() is inaccurate, it asks for a WSGIRequest, when
the function itself expects a HttpRequest (with a .session attribute).

In the [documentation](https://docs.djangoproject.com/en/2.2/_modules/django/contrib/auth/#update_session_auth_hash), it is also passed a request.

## How is that should be

The stub should be:

```def update_session_auth_hash(request: HttpRequest, user: AbstractUser) -> None: ...```

## System information

- OS: Ubuntu 18.04
- `python` version: 3.8.3
- `django` version: 2.2.13
- `mypy` version: 0.750
- `django-stubs` version: 1.3.0

### PR DESCRIPTION
**Pull request by coiax**
```
The update_session_auth_hash() stub doesn't match the documented and
actual type that the function expects.

---

Resolves #344 
```

As mentioned in the issue, the [documentation](https://docs.djangoproject.com/en/2.2/topics/auth/default/#django.contrib.auth.update_session_auth_hash) implies this is a regular HttpRequest, and not anything more esoteric.

## Related issues

- Resolves #400


sobolevn (approved):

sobolevn:
Thanks a lot!

PR #406, opened 2020-06-22
explicitly mark value and label types on Choices enum

ehossack:
In addition to the meta properties added
to the Choices enums, value and label are
set for each choice.
This commit explicitly types those instead of the
tuple value and non-extant label types

Relevant django code:
https://github.com/django/django/blob/master/django/db/models/enums.py

Prior to this change, referencing enum labels would produce:
```
error: "MyTextChoicesClass" has no attribute "label"
```


ehossack:
I'm not 100% this is the ideal approach here (using property decorators), but this was the best way I could think of to specify the label/value that django assigns.
Happy for feedback or to make adjustments.

It also wasn't super clear to me if/how tests would be necessary here, but I did try following the CONTRIBUTING.md document and running tests which mostly seemed to fail (looked like I may have missing dependencies? the failures were not clear).
Linting, isort and black ran just fine.

sobolevn (approved):
Thanks a lot! 👍

PR #408, opened 2020-06-22
Script for merging stubs into Django source files for testing

### LINKED ISSUE #405: Testing stubs against Django source code
We need to start using actual Django sources to test stubs. For that, we have to merge our `.pyi` files inside the `.py` files of Django. 

There's a library called LibCST https://github.com/Instagram/LibCST, which allows traversing and modifying Python source code. It's new, Python 3 only, covered with type annotations and well-maintained by Instagram. I think we should use it as a base for our merging tool, instead of 
https://github.com/ambv/retype
https://github.com/google/pytype/tree/master/pytype/tools/merge_pyi
which are both lib2to3 based. 

I used it in the https://github.com/typeddjango/django-stubs/pull/288 to make an example script to add stubs versioning. 

It even has some built-in support for this particular task via https://libcst.readthedocs.io/en/latest/codemods.html#libcst.codemod.visitors.ApplyTypeAnnotationsVisitor

The first step here is to make a simple script that uses this visitor to apply stubs. Later, the `ApplyTypeAnnotationsVisitor` will be subclassed with some custom code, if we'd need any special casing. 

@kszmigiel Could you take a look at it? Maybe make a script which will merge https://github.com/typeddjango/django-stubs/blob/master/django-stubs/shortcuts.pyi into the Django source, and then run mypy on top of it? 
I think you should work on it before you make more changes to the stubs.

kszmigiel:
- Closes #405


kszmigiel:
@mkurnikov @sobolevn I had not much time to sink into libcst docs yet, so here's just initial version on the script, so you can easily check my progress on this in the nearest future :)

mkurnikov:
Looks about right. Now we need to add more steps:
1. Merge all the stubs into Django sources. This will probably create lots of edge cases, at least that how it was with my experiments with `retype` from before.
2. Run mypy on top of annotated Django sources. We can start with whitelisting files to check. 
3. Community call for help, probably. Add it to CI to prevent regressions.

scripts/merge_stubs_into_django.py
```diff
@@ -0,0 +1,59 @@
...
+
+from pathlib import Path
+
+def prRed(skk): print("\033[91m {}\033[00m" .format(skk))
+def prGreen(skk): print("\033[92m {}\033[00m" .format(skk))
+
+context = CodemodContext()
+visitor = ApplyTypeAnnotationsVisitor(context)
+
+stubs_dir = '../django-stubs'
```
mkurnikov, line 13: Please, make paths not `..` relative, but with `Path(__file__).parent` so that script could be run from other directories too.

mkurnikov (changes requested):

scripts/merge_stubs_into_django.py
```diff
@@ -0,0 +1,59 @@
...
+stubs_dict = {}
+sources_dict = {}
+
+stubs_pathlist = Path(stubs_dir).rglob('*.pyi')
+sources_pathlist = Path(sources_dir).rglob('*.py')
+
+for path in stubs_pathlist:
+    str_path = str(path)
+    
+    stubs_dict[str_path.split('/', 2)[-1].split('.')[0]] = str_path
```
mkurnikov, line 25: Same thing here, use `pathlib.Path` everywhere for filesystem paths.

scripts/merge_stubs_into_django.py
```diff
@@ -0,0 +1,59 @@
...
+
+for path in sources_pathlist:
+    str_path = str(path)
+
+    key = str_path.split('/', 3)[-1].split('.')[0]
+
+    if key in stubs_dict:
+        sources_dict[key] = str_path
+
+for key in stubs_dict:
```
mkurnikov, line 35: Please use `enumerate(stubs_dict)`, you're using values of the dict too

UnknownPlatypus:
Closing because stale

PR #413, opened 2020-07-01
Improve partition signature

### LINKED ISSUE #412: Partition type signature can be improved
# Partition type signature can be improved

## What's wrong

It narrows the use case to Model only, even though it works for all kinds of objects.

## How is that should be

It should take a `TypeVar` and return the same type, but divided by the callables return value.

## System information

- OS:
- `python` version:
- `django` version:
- `mypy` version:
- `django-stubs` version:

Swamii:
# Improve partition signature

It narrows the use case to Model only, even though it works for all kinds of objects.
`predicate` could possibly be typed to `Callable[[PartitionMember], Any]`, but maybe that has issues with `lambda`.

## Related issues

- Closes #412


django-stubs/utils/functional.pyi
```diff
@@ -56,4 +56,5 @@ class SimpleLazyObject(LazyObject):
     def __copy__(self) -> List[int]: ...
     def __deepcopy__(self, memo: Dict[Any, Any]) -> List[int]: ...
 
-def partition(predicate: Callable, values: List[Model]) -> Tuple[List[Model], List[Model]]: ...
+PartitionMember = TypeVar("PartitionMember")
```
sobolevn: Let's name it `_PartitionMember`, we don't want it to be public, do we?

sobolevn (changes requested):
Thanks a lot! Just a single thing to fix and we are can merge this 🎉

Swamii:
Hopefully that should do it then. Thanks for the project 👍

sobolevn (approved):

sobolevn:
Thank you for your help!

PR #415, opened 2020-07-06
Don't change type of HttpRequest.user if type has been changed by subclassing

### LINKED ISSUE #407: Unable to subclass HttpRequest to override type of User
I'd like to create a `AuthenticatedHttpRequest` to indicate that a user has been authenticated (for example to annotate the `request` parameter for views with the `login_required` decorator).

I'd expect this to be done like so:
```python
class AuthenticatedHttpRequest(HttpRequest):
    user: User
```

The problem is that the type of `user` is changed by this transformer https://github.com/typeddjango/django-stubs/blob/574a87e68cfb2e6ff66b3d7e1a0bb75f51148a8c/mypy_django_plugin/transformers/request.py#L10-L26
That's called here https://github.com/typeddjango/django-stubs/blob/574a87e68cfb2e6ff66b3d7e1a0bb75f51148a8c/mypy_django_plugin/main.py#L244-L245

rendering it somewhat impossible to declare these kinds of subclasses.

--------------------------------
Not sure what a solution would be here. 

Only changing the type of `user` if the object it's accessed on is an instance of `HttpRequest` (not for subclasses) would possibly be a solution, but would require that all `HttpRequest`-subclasses declare the type of `user` to not fall back to the imprecise type annotations in the `.pyi` file.

Would it be possible to check the type and only change it if it's what's expected from the `.pyi`-file (`Union[AbstractBaseUser, AnonymousUser]`)?

bullfest:
Fix #407

This makes it so that the transformer that sets the type of `HttpRequest.user` to `Union[<user-model>, AnonymousUser]` only changes the type if the type is unchanged from the one declared in the `http/request.pyi` file.

This means that if you subclass `HttpRequest` and change the type of `user` like so:
```python
class AuthenticatedHttpRequest(HttpRequest):
    user: MyUserClass
```
Then the type will remain `MyUserClass`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,19 +90,24 @@ You can use strings instead: `'QuerySet[MyModel]'` and `'Manager[MyModel]'`, thi
 
 Currently we [are working](https://github.com/django/django/pull/12405) on providing `__class_getitem__` to the classes where we need them.
 
-### How can I use HttpRequest with custom user model?
+### How can I create a HttpRequest that's guaranteed to have an authenticated user?
 
-You can subclass standard request like so:
+Django's built in `HttpRequest` has the attribute `user` that resolves to the type
+```python
+Union[User, AnonymousUser]
+```
+where `User` is the user model specified by the `AUTH_USER_MODEL` setting.
 
+If you want a `HttpRequest` that you can type-annotate with where you know that the user is authenticated you can subclass the normal `HttpRequest` class like so:
 ```python
 from django.http import HttpRequest
 from my_user_app.models import MyUser
 
-class MyRequest(HttpRequest):
+class AuthenticatedHttpRequest(HttpRequest):
     user: MyUser
 ```
 
-And then use `MyRequest` instead of standard `HttpRequest` inside your project.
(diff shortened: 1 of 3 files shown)


sobolevn (changes requested):
Thanks a lot for your work! 

Let's add a test for this (because it is a complex feature), and I will be happy to merge it!

README.md
```diff
@@ -90,19 +90,24 @@ You can use strings instead: `'QuerySet[MyModel]'` and `'Manager[MyModel]'`, thi
 
 Currently we [are working](https://github.com/django/django/pull/12405) on providing `__class_getitem__` to the classes where we need them.
 
-### How can I use HttpRequest with custom user model?
+### How can I create a HttpRequest that's guaranteed to have an authenticated user?
```
bullfest, line 93: The previous content here wasn't up to date, the functionality described (using a custom user model) just works out of the box with the vanilla `HttpRequest` class the way django-stubs currently works.

New content describes a similar use case though.

sobolevn: 👍

sobolevn (approved):

bullfest:
Not sure what's up with Travis. Both the checks (queued and successful) refer to the same build.

sobolevn (approved):
Thanks a lot!

PR #418, opened 2020-07-08
Allow more types for args to Paginator.get_page

bullfest:
`Paginator.get_page` delegates handling of the argument to `Paginator.validate_number` and thus can handle a lot more than just `int`. Technically we could let it be `Any`, as `get_page` allows basically any value and has fallbacks. But a sensible middleground is to use any kind of value that can be cast to an integer by `int(value)`.


bullfest:
Relevant section in django source:
https://github.com/django/django/blob/779e615e362108862f1681f965ee9e4f1d0ae6d2/django/core/paginator.py#L41-L69

sobolevn (approved):
Thanks!

Issue #417, opened 2020-07-07
 Incompatible with django-configurations due to the way the settings are imported


snejus:
# Bug report
:bug: 

## What's wrong

Trying to use `django-stubs` together with `django-configurations` is a big pain since the latter interferes and controls how django settings are loaded. By default, the following exception gets thrown when trying to run `mypy`:
```bash
(mypackage-vfeg9Sk1-py3.6) ➜  mypackage IN-3486 ✗ python -m mypy -p src.mypackage                                                       
Error constructing plugin instance of NewSemanalDjangoPlugin

Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/sarunas/.cache/pypoetry/virtualenvs/mypackage-vfeg9Sk1-py3.6/lib/python3.6/site-packages/mypy/__main__.py", line 12, in <module>
    main(None, sys.stdout, sys.stderr)
  File "mypy/main.py", line 89, in main
  File "mypy/build.py", line 180, in build
  File "mypy/build.py", line 227, in _build
  File "mypy/build.py", line 465, in load_plugins
  File "mypy/build.py", line 443, in load_plugins_from_config
  File "/home/sarunas/.cache/pypoetry/virtualenvs/mypackage-vfeg9Sk1-py3.6/lib/python3.6/site-packages/mypy_django_plugin/main.py", line 80, in __init__
    self.django_context = DjangoContext(django_settings_module)
  File "/home/sarunas/.cache/pypoetry/virtualenvs/mypackage-vfeg9Sk1-py3.6/lib/python3.6/site-packages/mypy_django_plugin/django/context.py", line 88, in __init__
    apps, settings = initialize_django(self.django_settings_module)
  File "/home/sarunas/.cache/pypoetry/virtualenvs/mypackage-vfeg9Sk1-py3.6/lib/python3.6/site-packages/mypy_django_plugin/django/context.py", line 70, in initialize_django
    settings._setup()
  File "/home/sarunas/.cache/pypoetry/virtualenvs/mypackage-vfeg9Sk1-py3.6/lib/python3.6/site-packages/django/conf/__init__.py", line 66, in _setup
    self._wrapped = Settings(settings_module)
[...]

Issue #224, opened 2019-11-08
ModuleNotFoundError using a custom apps directory & path


richkirsch:
To make my apps more reusable, I prefer to use an `apps` directory in my project root to house them, with the following structure:
```
--Project/
    -- manage.py
    -- apps/
        -- app1/
            -- app.py
            -- models.py
            -- etc...
    -- etc ...
```

To make life easier, in `manage.py`, I have the following:
```
current_path = os.path.dirname(os.path.abspath(__file__))
sys.path.append(os.path.join(current_path, "apps"))
```

This enables me to use e.g. `from app1 import SomeClass` anywhere in the project, instead of `from apps.app1 import SomeClass`.

However, this is causing `mypy` with `django-stubs` (though not without it) to crash a with `ModuleNotFoundError`:
```
  File "/Users/me/Projects/AwesomeProject/apps/app2/models.py", line 4, in <module>
    from app1.models import SomeModel
ModuleNotFoundError: No module named 'app1'
```

I had a similar issue with PyLint, but was able to solve it with the help of [this StackOverflow page](https://stackoverflow.com/questions/1899436/pylint-unable-to-import-error-how-to-set-pythonpath), taking advatange of an initialisation hook to set the system path in a similar way to above but in `.pylintrc`, like so:
```
[MASTER]
init-hook='
  import os, sys;
  from pylint.config import find_pylintrc;
  sys.path.append(os.path.join(os.path.dirname(find_pylintrc()), 'apps'))'
load-plugins=pylint_django, pylint_celery
...
```

My questions are:
    1) Is there a similar approach I can take?
    2) Otherwise, what are alternative suggestions?

Atala:
I think a workaround is to add the `apps`  folder to your PYTHONPATH :
```
PYTHONPATH=:<path_to_apps_folder> mypy
```

LucidDan:
I do this too, using a dedicated apps directory, and as @Atala says, I adjust my PYTHONPATH, which works fine.

flyte:
This isn't possible when you don't control the calling environment creation such as when mypy is called from the Python extension in VSCode.

Could this module respect the `mypy_path` `mypy.ini` setting and add those paths before trying to import the Django settings file?

https://mypy.readthedocs.io/en/stable/config_file.html#import-discovery

PR #421, opened 2020-07-09
Output a more clear configuration error

### LINKED ISSUE #420: Exceptions should be reported in a more clear way
# Technical improvement suggestion
Not exactly a bug something that should get some more attention: exception handling and reporting them to the user. I've been working on the configuration reading in `main.py` module and found that _handled exceptions_ end up being reported like this:
```bash
$ mypy -p django-stubs --config-file ./test-data/plugins.ini

test-data/plugins.ini:0: error: 'django_settings_module' is not set: no section [mypy.plugins.django-stubs]
Found 1 error in 1 file (checked 477 source files)
```
image

in comparison to

**mypy error with a cmd line argument**
```bash
$ mypy -p django-stubs --config-file ./tst-data/plugins.ini
usage: mypy [-h] [-v] [-V] [more options; see below]
            [-m MODULE] [-p PACKAGE] [-c PROGRAM_TEXT] [files ...]
mypy: error: Cannot find config file './tst/plugins.ini'
```
or the unhandled default
 
```bash
You must install the lxml package before you can run mypy with `--html-report`.
You can do this with `python3 -m pip install lxml`.
  File "/home/sarunas/.pyenv/versions/django-stubs-3.6.11/bin/mypy", line 11, in <module>
    sys.exit(console_entry())
  File "/home/sarunas/.pyenv/versions/3.6.11/envs/django-stubs-3.6.11/lib/python3.6/site-packages/mypy/__main__.py", line 8, in console_entry
  File "mypy/main.py", line 89, in main
  File "mypy/build.py", line 217, in _build
  File "mypy/report.py", line 58, in __init__
  File "mypy/report.py", line 71, in add_report
ImportError
```

While the output is the most visually appealing, I find it very quite opaque and misleading. No source files have been checked by that point, so including the count isn't correct; it's also formatted like an issue, so my first reaction (which is basically muscle memory now) is to switch back to the source code and to have a look at the most recent change that has been made. It takes a lot to find that this is an internal problem rather than something that is related to my source files. 

I'd think that `mypy.Errors` may not be the best solution for this - even though it's a very convenient one. There are a couple of unneeded calls made until it reaches a more related `CompileError` - though it gets thrown plainly, without using its `report_internal_error` method - which seems like a good option that would include some extras for the user for debugging. Though I guess `mypy` may have to change into something more related to this project. Though I haven't seen any of them writing to `stderr` instead of `stdout`. Also, once the issues are raised, they return `None` instead of `NoReturn` - could this indicate that they are not meant to cut off the execution completely?

Though, correct if I'm wrong (I probably am, since I haven't spent days on this :) ), but both of the above seem to be used _during the parsing / linting stage_. Would it be wrong to assume that reading the configuration should be treated as the setup stage? This way, it could report the error in the same format like `mypy` does - could also reuse their stuff. This shouldn't be a significant change I reckon.

snejus:
# I have made things!

This tackles #420; missing configuration is now reported like this:
```
(django-stubs-3.6.11) ➜  django-stubs issue-420 ✗ mypy -p django-stubs --config-file test-data/plugins.ini                                                                                                                                    
usage: (config)
...
[mypy.plugins.django_stubs]
    django_settings_module: str (required)
...
(django-stubs) mypy: error: 'django_settings_module' is not set: no section [mypy.plugins.django-stubs]
```

The function got refactored a little bit and made a little less verbose. I thought this way it will be easier to add new configuration options (a solution to #417 which is on the way for example). 

## Related issues

- Closes #420

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -52,30 +52,37 @@ def add_new_manager_base(ctx: ClassDefContext) -> None:
 
 def extract_django_settings_module(config_file_path: Optional[str]) -> str:
 
-    def raise_error(error_type: int) -> NoReturn:
-        # Considering that the startup time is at least double as long as mypy's, this tries to not make it worse.
-        # The function can possibly run just once and it kills the execution, so it's safe to import and define
-        # the dependencies locally.
+    def exit(error_type: int) -> NoReturn:
+        """Using mypy's argument parser, raise `SystemExit` to fail hard if validation fails.
+
+        Considering that the plugin's startup duration is around double as long as mypy's, this aims to
+        import and construct objects only when that's required - which happens once and terminates the
+        run. Considering that most of the runs are successful, there's no need for this to linger in the
+        global scope.
+        """
         from mypy.main import CapturableArgumentParser
-        usage = "\n".join(["(config)",
-                           "[mypy.plugins.django_stubs]",
-                           "\t(required) django_settings_module: str"])
+
+        usage = '\n'.join(['(config)',
+                           '...',
+                           '[mypy.plugins.django_stubs]',
+                           '\tdjango_settings_module: str (required)',
+                           '...']).expandtabs(4)
         handler = CapturableArgumentParser(prog="(django-stubs) mypy", usage=usage)
-        message = {1: "'django_settings_module' is not set: mypy config file is not specified or found",
-                   2: "'django_settings_module' is not set: no section [mypy.plugins.django-stubs]",
-                   3: "'django_settings_module' is not set: setting is not provided"}[error_type]
-        handler.error(message)
+        messages = {1: "'django_settings_module' is not set: mypy config file is not specified or found",
+                    2: "'django_settings_module' is not set: no section [mypy.plugins.django-stubs]",
+                    3: "'django_settings_module' is not set: the setting is not provided"}
+        handler.error(messages[error_type])
 
     parser = configparser.ConfigParser()
     try:
-        parser.read_file(open(cast(str, config_file_path), "r"), source=config_file_path)
-    except OSError:
-        raise_error(1)
+        parser.read_file(open(cast(str, config_file_path), 'r'), source=config_file_path)
+    except (IsADirectoryError, OSError):
+        exit(1)
 
     section = 'mypy.plugins.django-stubs'
     if not parser.has_section(section):
-        raise_error(2)
-    settings = parser.get(section, 'django_settings_module', fallback=None) or raise_error(3)
+        exit(2)
+    settings = parser.get(section, 'django_settings_module', fallback=None) or exit(3)
     return cast(str, settings).strip('\'"')
 
 
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -3,7 +3,6 @@ from functools import partial
 from typing import Callable, Dict, List, NoReturn, Optional, Tuple, cast
 
 from django.db.models.fields.related import RelatedField
-from mypy.main import CapturableArgumentParser as ErrorHandler
 from mypy.nodes import MypyFile, TypeInfo
 from mypy.options import Options
 from mypy.plugin import (
@@ -52,27 +51,31 @@ def add_new_manager_base(ctx: ClassDefContext) -> None:
 
 
 def extract_django_settings_module(config_file_path: Optional[str]) -> str:
-    usage = "\n".join(["(config)",
-                       "[mypy.plugins.django_stubs]",
-                       "\t(required) django_settings_module: str"])
 
-    def raise_error(handler: ErrorHandler, error_type: int) -> NoReturn:
-        messages = {1: "'django_settings_module' is not set: no mypy config file specified",
-                    2: "'django_settings_module' is not set: no section [mypy.plugins.django-stubs]",
-                    3: "'django_settings_module' is not set: setting is not provided"}
-        handler.error(messages[error_type])
-
-    error = partial(raise_error, ErrorHandler(prog="(django-stubs) mypy", usage=usage))
-    if config_file_path is None:
-        error(1)
+    def raise_error(error_type: int) -> NoReturn:
+        # Considering that the startup time is at least double as long as mypy's, this tries to not make it worse.
+        # The function can possibly run just once and it kills the execution, so it's safe to import and define
+        # the dependencies locally.
+        from mypy.main import CapturableArgumentParser
+        usage = "\n".join(["(config)",
+                           "[mypy.plugins.django_stubs]",
+                           "\t(required) django_settings_module: str"])
+        handler = CapturableArgumentParser(prog="(django-stubs) mypy", usage=usage)
+        message = {1: "'django_settings_module' is not set: mypy config file is not specified or found",
+                   2: "'django_settings_module' is not set: no section [mypy.plugins.django-stubs]",
+                   3: "'django_settings_module' is not set: setting is not provided"}[error_type]
+        handler.error(message)
 
     parser = configparser.ConfigParser()
-    parser.read(cast(str, config_file_path))
+    try:
+        parser.read_file(open(cast(str, config_file_path), "r"), source=config_file_path)
+    except OSError:
+        raise_error(1)
 
     section = 'mypy.plugins.django-stubs'
     if not parser.has_section(section):
-        error(2)
-    settings = parser.get(section, 'django_settings_module', fallback=None) or error(3)
+        raise_error(2)
+    settings = parser.get(section, 'django_settings_module', fallback=None) or raise_error(3)
     return cast(str, settings).strip('\'"')
 
 
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -62,16 +62,17 @@ def extract_django_settings_module(config_file_path: Optional[str]) -> str:
         """
         from mypy.main import CapturableArgumentParser
 
-        usage = '\n'.join(['(config)',
-                           '...',
-                           '[mypy.plugins.django_stubs]',
-                           '\tdjango_settings_module: str (required)',
-                           '...']).expandtabs(4)
-        handler = CapturableArgumentParser(prog="(django-stubs) mypy", usage=usage)
-        messages = {1: "'django_settings_module' is not set: mypy config file is not specified or found",
-                    2: "'django_settings_module' is not set: no section [mypy.plugins.django-stubs]",
-                    3: "'django_settings_module' is not set: the setting is not provided"}
-        handler.error(messages[error_type])
+        usage = """(config)
+        ...
+        [mypy.plugins.django_stubs]
+            django_settings_module: str (required)
+        ...
+        """.replace("\n" + 8 * " ", "\n")
+        handler = CapturableArgumentParser(prog='(django-stubs) mypy', usage=usage)
+        messages = {1: 'mypy config file is not specified or found',
+                    2: 'no section [mypy.plugins.django-stubs]',
+                    3: 'the setting is not provided'}
+        handler.error("'django_settings_module' is not set: " + messages[error_type])
 
     parser = configparser.ConfigParser()
     try:
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -1,9 +1,9 @@
 import configparser
 from functools import partial
-from typing import Callable, Dict, List, Optional, Tuple
+from typing import Callable, Dict, List, NoReturn, Optional, Tuple, cast
 
 from django.db.models.fields.related import RelatedField
-from mypy.errors import Errors
+from mypy.main import CapturableArgumentParser as ErrorHandler
 from mypy.nodes import MypyFile, TypeInfo
 from mypy.options import Options
 from mypy.plugin import (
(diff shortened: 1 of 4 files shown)


mypy_django_plugin/main.py
```diff
@@ -52,25 +52,28 @@ def add_new_manager_base(ctx: ClassDefContext) -> None:
...
-    if not parser.has_option('mypy.plugins.django-stubs', 'django_settings_module'):
-        errors.report(0, None, "'django_settings_module' is not set: setting is not provided",
-                      file=config_file_path)
-        errors.raise_error()
-
-    django_settings_module = parser.get('mypy.plugins.django-stubs', 'django_settings_module').strip('\'"')
-    return django_settings_module
+    parser.read(cast(str, config_file_path))
+
+    section = 'mypy.plugins.django-stubs'
```
sobolevn, line 83: Can we reuse this const from somewhere else? Looks like we should already have it somewhere.

mypy_django_plugin/main.py
```diff
@@ -52,25 +52,28 @@ def add_new_manager_base(ctx: ClassDefContext) -> None:
 
 
 def extract_django_settings_module(config_file_path: Optional[str]) -> str:
-    errors = Errors()
+    usage = "\n".join(["(config)",
+                       "[mypy.plugins.django_stubs]",
+                       "\t(required) django_settings_module: str"])
+
+    def raise_error(handler: ErrorHandler, error_type: int) -> NoReturn:
+        messages = {1: "'django_settings_module' is not set: no mypy config file specified",
```
sobolevn: Let's make an `enum` from keys here. Like `@unique class ErrorType(enum.IntEnum)`?

sobolevn (changes requested):
Thanks a lot! Great work!

Two things:
1. Let's change how we work with constants here
2. Let's test the new output! 

👍

mypy_django_plugin/main.py
snejus: Doesn't seem like it - at least not anywhere in the source code:
image

mypy_django_plugin/main.py
snejus: Very good idea! I think enums access time is faster than dicts. 

Considering that they shouldn't get evaluated until needed for an error, where should they live?

snejus:
Thanks! In terms of testing, would you expect it to be written in the same language as the type checks? I struggled to find any examples. 

If I try removing the configuration from one of the tests now, that's what I get:
```bash
platform linux -- Python 3.6.11, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /home/sarunas/.pyenv/versions/3.6.11/envs/django-stubs-3.6.11/bin/python3.6
cachedir: .pytest_cache
rootdir: /home/sarunas/Documents/misc/django-stubs, inifile: pytest.ini, testpaths: ./test-data/typecheck/test_config.yml
plugins: cov-2.10.0, mypy-plugins-1.3.0
collected 3 items                                                                                                                                                                                                                                                                          

test-data/typecheck/test_config.yml::django_settings_module_missing usage: (config)
[mypy.plugins.django_stubs]
	(required) django_settings_module: str
(django-stubs) mypy: error: 'django_settings_module' is not set: setting is not provided

FAILED
test-data/typecheck/test_config.yml::pyproject_toml_config PASSED
test-data/typecheck/test_config.yml::generate_pyproject_toml_and_settings_file_from_installed_apps_key PASSED

========================================================================================================================================= FAILURES =========================================================================================================================================
______________________________________________________________________________________________________________________________ django_settings_module_missing ______________________________________________________________________________________________________________________________
/home/sarunas/Documents/misc/django-stubs/test-data/typecheck/test_config.yml:2: 
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Critical error occurred
```

Or should this be a standard python test?

mypy_django_plugin/main.py
snejus: Just had a look into this and I have mixed feelings. Enums can take the following form:
```
class ErrorType(enum.Enum):
    SOME_ERROR: "some error because bad bad something"
    BAD_ERROR: "bad error because something"
# or, say
class ErrorType(enum.IntEnum):
    SOME_ERROR: 1
    BAD_ERROR: 2
```

As far as I understand, the `IntEnum` just allows the enum to be referenced like `ErrorType(1)`. The reason I used an int as the key initially was that it's the smallest type that would allow to reference errors without having to add those definitions to the scope before needed.

If an enum is used, this would require a call like `error(ErrorType.MISSING_SECTION`) or similar, i.e. they'd be in the scope before they are needed, which isn't great considering the startup time which is already rather slow and that errors won't get thrown often. 

Also, currently these messages aren't used anywhere else or have any operations done to them. If we, say, had to do any comparison operations, then attaching types to them with enums would make sense; but now they're rather limited, so I'd probably leave it as it is for now, mostly due to performance reasons though. 

Let me know if I'm missing out on something or you have something planned where having them as enums would make more sense. For now I've moved the error handler import to the local scope having done some `python -X importtime` runs - the startup is at least double as sluggish as mypy's :snail:

snejus: Also added a try/except for the file read, and using `ConfigParser.read_file` instead of `read`, since the latter is made for multiple files and does some extras that aren't required here.

snejus:
Also, I didn't find any note in the readme which mentions that some features aren't supported by `Django 2.2` and `python 3.6`. Maybe we should be a bit more clear about it?

sobolevn:
@snejus yes, we must be clear about the minimal requirements.

snejus:
Perfect! Sorry to distract though, what do you reckon about the above? Just realised the second comment ended up being an essay. Skip it - the main idea: at least the way I see it, using enums would add to the already slow startup time, and I struggled to find good reasons to use it - though that's just a shallow look and I can't read your mind so I'd be happy to hear a bit more about this :)

sobolevn:
@snejus seems ok! Let's skip it then!

snejus:
Hey, sorry for disappearing - been a bit too busy with work. 

> Thanks! In terms of testing, would you expect it to be written in the same language as the type checks? I struggled to find any examples.

Do you have a comment about this by any chance? As far as I understand the type checks testing lang (actually, what is it officially called? :) ) cannot handle hard failures/exits - could you confirm this? 

If that's the case I'll add a usual python test for it.

sobolevn:
@snejus you can try https://github.com/typeddjango/django-stubs/tree/master/test-data/typecheck

It probably should work. If not, then we can create `.py` test.

snejus:
Had a deeper look into it - it doesn't seem like it can be done with `pytest-mypy-plugins`, unless there are some undocumented hacks (though none of the current test cases handle `CriticalErrror`s, so I'd assume it's not a thing:
```bash
============================================================ test session starts =============================================================
platform linux -- Python 3.6.11, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /home/sarunas/.pyenv/versions/3.6.11/envs/django-stubs-3.6.11/bin/python3.6
cachedir: .pytest_cache
rootdir: /home/sarunas/Documents/misc/django-stubs, inifile: pytest.ini, testpaths: ./test-data/typecheck/test_config.yml
plugins: cov-2.10.0, mypy-plugins-1.3.0
collected 3 items                                                                                                                            

test-data/typecheck/test_config.yml::config_missing_django_settings_module usage: (config)
[mypy.plugins.django_stubs]
	(required) django_settings_module: str
(django-stubs) mypy: error: 'django_settings_module' is not set: setting is not provided

FAILED
test-data/typecheck/test_config.yml::pyproject_toml_config PASSED
test-data/typecheck/test_config.yml::generate_pyproject_toml_and_settings_file_from_installed_apps_key PASSED

================================================================== FAILURES ==================================================================
___________________________________________________ config_missing_django_settings_module ____________________________________________________
/home/sarunas/Documents/misc/django-stubs/test-data/typecheck/test_config.yml:2: 
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Critical error occurred
```
where the test case looks like this:
```yaml
-   case: config_missing_django_settings_module
    main: |
        pass
    mypy_config: |
        [mypy.plugins.django-stubs]
```

This kinda makes sense - the language is suited to test _type checking_ specifically, whereas this is sort of a test case _setup_ level error.

I guess I'll go ahead with a `.py` test.

snejus:
I've seen you mentioned this PR in #428 - I'll cover django setup validation too.

sobolevn:
@snejus thanks a lot!

sobolevn:
We use `snapshots` for that: https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/test_formatter/test_formatter_output.py

snejus:
Ah, didn't see your message until now and got it done meanwhile. Though would it not be a bit of an overkill to use `snapshots` in this scenario? It's an amazing tool for your use case in WPS - since the output varies a lot. 

For the input validation, the template stays the same and can take 3 error messages (see the tests) - and that's basically it. If new vars get added in the future then it just needs that field name adding there - though that won't happen often I imagine. 

Considering the rest of error handling, the output will look even more boring - the way I see it (do correct me if I'm wrong) is that we're simply gonna catch the exceptions and reraise it with our generic message, like `mypy django stubs have failed initialisation, check a, b, c`. And targeted messages for commonly thrown exceptions.

snejus:
```
(django-stubs-3.6.11) ➜  django-stubs issue-420 ✗ mypy -p django-stubs --config-file test-data/plugins.ini                                                                                                                                    
usage: (config)
...
[mypy.plugins.django_stubs]
    django_settings_module: str (required)
...
(django-stubs) mypy: error: 'django_settings_module' is not set: no section [mypy.plugins.django-stubs]
(django-stubs-3.6.11) ➜  django-stubs issue-420 ✗ echo '\n[mypy.plugins.django-stubs]' >> test-data/plugins.ini                                                                                                                               
(django-stubs-3.6.11) ➜  django-stubs issue-420 ✗ mypy -p django-stubs --config-file test-data/plugins.ini                                                                                                                                    
usage: (config)
...
[mypy.plugins.django_stubs]
    django_settings_module: str (required)
...
(django-stubs) mypy: error: 'django_settings_module' is not set: the setting is not provided
(django-stubs-3.6.11) ➜  django-stubs issue-420 ✗ echo '    django_settings_module = hello' >> test-data/plugins.ini                                                                                                                          
(django-stubs-3.6.11) ➜  django-stubs issue-420 ✗ mypy -p django-stubs --config-file test-data/plugins.ini                                                                                                                                    

  File "/home/sarunas/.pyenv/versions/django-stubs-3.6.11/bin/mypy", line 11, in <module>
    sys.exit(console_entry())
  File "/home/sarunas/.pyenv/versions/3.6.11/envs/django-stubs-3.6.11/lib/python3.6/site-packages/mypy/__main__.py", line 8, in console_entry
  File "mypy/main.py", line 89, in main
  File "mypy/build.py", line 227, in _build
  File "mypy/build.py", line 470, in load_plugins
  File "mypy/build.py", line 448, in load_plugins_from_config
  File "/home/sarunas/Documents/misc/django-stubs/mypy_django_plugin/main.py", line 93, in __init__
    self.django_context = DjangoContext(django_settings_module)
  File "/home/sarunas/Documents/misc/django-stubs/mypy_django_plugin/django/context.py", line 88, in __init__
  File "/home/sarunas/Documents/misc/django-stubs/mypy_django_plugin/django/context.py", line 70, in initialize_django
    settings._setup()
  File "/home/sarunas/.pyenv/versions/3.6.11/envs/django-stubs-3.6.11/lib/python3.6/site-packages/django/conf/__init__.py", line 63, in _setup
  File "/home/sarunas/.pyenv/versions/3.6.11/envs/django-stubs-3.6.11/lib/python3.6/site-packages/django/conf/__init__.py", line 142, in __init__
  File "/home/sarunas/.pyenv/versions/3.6.11/lib/python3.6/importlib/__init__.py", line 126, in import_module
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'hello'
```

Looks like the output is telling me what's next :sweat_smile:

mypy_django_plugin/main.py
```diff
@@ -52,25 +51,39 @@ def add_new_manager_base(ctx: ClassDefContext) -> None:
...
+        Considering that the plugin's startup duration is around double as long as mypy's, this aims to
+        import and construct objects only when that's required - which happens once and terminates the
+        run. Considering that most of the runs are successful, there's no need for this to linger in the
+        global scope.
+        """
+        from mypy.main import CapturableArgumentParser
+
+        usage = '\n'.join(['(config)',
+                           '...',
+                           '[mypy.plugins.django_stubs]',
```
snejus: Just realised that this entire nonsense is a constant, and that the tests already define a string for it:
```
TEMPLATE = """usage: (config)
...
[mypy.plugins.django_stubs]
    django_settings_module: str (required)
...
(django-stubs) mypy: error: 'django_settings_module' is not set: {}
"""
```

Assuming that we follow DRY, and considering that error handling is about to get extended in general, I feel like most of this `exit` function should move somewhere else, say, a module called `errors`? It could live in `mypy_django_plugin/lib/` like `helpers` and `fullnames`. I thought of adding it to the helpers, but it somehow seems wrong to `from helpers import exit` which kills the app.

Additionally, that'd be a good start considering the rest of error handling plans.

sobolevn (approved):
LGTM

@mkurnikov ?
@kszmigiel ?

snejus:
Cool. Let's finish this PR here, (unless there are any further action points that relate to input validation). 

In terms of further **error handling**, #428 is very much related to my #417 which has some work done already, so I believe it may make more sense to continue with it there. I found that #224 also experience similar issues. 

Are you by any chance aware of any other affected tickets? Would be good to source as many as possible in order to build a clear picture and to know which parts people struggle the most with/need to get most attention.

sobolevn:
> Are you by any chance aware of any other affected tickets?

Nope, I guess that you have found everything we had.

kszmigiel:
@sobolevn LGTM 👍 
@snejus great job, love it! ❤️  Does it fix #428 already or should I take care of it after merge?

snejus:
Nope, not yet! Though I'm on it currently, together with #417 to handle `django-configurations`. 

Thanks, by the way! Next step - colours ;)

snejus:
Hey - just checking whether there's anything that is blocking this getting merged in? The further error handling improvements partly depend on this (finally found time to work on it), so it'd be good know what's the status.

sobolevn (approved):

sobolevn:
Thanks a lot!

PR #425, opened 2020-07-14
admin: Allow ModelAdmin.actions to contain str

danifus:
Strings are allowed in ModelAdmin.actions when they refer to a method on the ModelAdmin subclass.

This adds `str` to a `Union` with the existing function signature type


sobolevn (approved):

sobolevn:
Thanks a lot!

PR #427, opened 2020-07-16
Added dynamic class hook for `from_queryset` manager

kszmigiel:


mypy_django_plugin/transformers/managers.py
```diff
@@ -77,8 +167,17 @@ def create_new_dynamic_class(self) -> None:
                 break
             for name, sym in class_mro_info.names.items():
                 if isinstance(sym.node, FuncDef):
+                    # self.copy_method_to_another_class(
```
sobolevn: Do we need these lines? Or can we delete them?

kszmigiel: @sobolevn You're probably right, but at this moment I'm a bit stuck on debugging the existing code (I mean written by @mkurnikov), will fix it later

mypy_django_plugin/transformers/managers.py
```diff
@@ -0,0 +1,183 @@
...
+class ManagerCallback(helpers.DynamicClassPluginCallback):
+
+    def copy_method_to_another_class(
+            self,
+            ctx: ClassDefContext,
+            self_type: Instance,
+            new_method_name: str,
+            method_node: FuncDef) -> None:
+        if method_node.type is None:
+            if not self.semanal_api.final_iteration:
```
mkurnikov: Could `defer_till_next_iteration` method be used here?

mypy_django_plugin/lib/sem_helpers.py
```diff
@@ -0,0 +1,120 @@
...
+from mypy.nodes import Argument, FuncDef, TypeInfo, Var
+from mypy.plugin import ClassDefContext, DynamicClassDefContext
+from mypy.plugins.common import add_method
+from mypy.semanal import SemanticAnalyzer
+from mypy.types import AnyType, CallableType, Instance, PlaceholderType
+from mypy.types import Type as MypyType
+from mypy.types import TypeOfAny, get_proper_type
+
+
+class IncompleteDefnException(Exception):
```
mkurnikov: Why did you create another base exception? There is an `IncompleteDefnError` in `transformers.new_helpers` already

mypy_django_plugin/transformers/managers.py
```diff
@@ -0,0 +1,183 @@
...
+
+        assert isinstance(callee, MemberExpr)
+        assert isinstance(callee.expr, RefExpr)
+
+        # not sure if lookup_typeinfo_or_defer(self.calss_name) could be used here
+
+        base_manager_info = callee.expr.node
+
+        if base_manager_info is None:
+            self.defer_till_next_interation()
```
mkurnikov: It's a typo, should be `iteration`

mypy_django_plugin/transformers/managers.py
```diff
@@ -0,0 +1,183 @@
...
+                args=arguments,
+                return_type=bound_return_type,
+                self_type=self_type)
+
+
+class ManagerFromQuerySetCallback(ManagerCallback):
+    def create_new_dynamic_class(self) -> None:
+        callee = self.call_expr.callee
+
+        assert isinstance(callee, MemberExpr)
```
mkurnikov: I assume for all `get_dynamic_class_hook` invoked on methods, this will be a `MemberExpr`. Could you add a generic subclass of `DynamicClassPluginCallback` called `DynamicClassFromMethodCallback` with `callee` present as a class attribute and set in the `__call__`?

mypy_django_plugin/transformers/managers.py
```diff
@@ -0,0 +1,183 @@
...
+                )
+        new_manager_info.line = self.call_expr.line
+        new_manager_info.defn.line = self.call_expr.line
+        new_manager_info.metaclass_type = new_manager_info.calculate_metaclass_type()
+
+        current_module = self.semanal_api.cur_mod_node
+        current_module.names[self.class_name] = SymbolTableNode(
+                GDEF,
+                new_manager_info,
+                plugin_generated=True)
```
mkurnikov: This could be extracted as a method on base `DynamicClassPluginCallback`

mkurnikov (changes requested):

mypy_django_plugin/lib/helpers.py
```diff
@@ -118,10 +118,37 @@ def __call__(self, ctx: DynamicClassDefContext) -> None:
...
+            new_manager_info,
+            plugin_generated=True)
+        return new_manager_info, current_module
+
     @abstractmethod
     def create_new_dynamic_class(self) -> None:
         raise NotImplementedError
 
+class DynamicClassFromMethodCallback(DynamicClassPluginCallback):
+
```
mkurnikov, line 221: Add class attribute `callee` here, it should be certain that it's a `MemberExpr`. Also, add assert with the type in the `__init__` so that it would fail if this subclass is used incorrectly. It will allow a bit nicer autocompletion and maybe will save you an assert somewhere in subclasses.

mypy_django_plugin/transformers/managers.py
```diff
@@ -0,0 +1,153 @@
...
+                    ctx.cls,
+                    new_method_name,
+                    args=arguments,
+                    return_type=return_type,
+                    self_type=self_type)
+            return
+
+        method_type = method_node.type
+        if not isinstance(method_type, CallableType):
+            if not self.semanal_api.final_iteration:
```
mkurnikov: `defer_till_next_iteration` too?

mypy_django_plugin/transformers/managers.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def copy_method_to_another_class(
+            self,
+            ctx: ClassDefContext,
+            self_type: Instance,
+            new_method_name: str,
+            method_node: FuncDef) -> None:
+        if method_node.type is None:
+            if self.defer_till_next_iteration():
+                return
```
mkurnikov: Should it fail with an exception instead of `return`? Some error description would be useful for later debugging too. 

In the refactoring, I preferred to fail with exception in all places where there's not enough info to resolve types to be able to do a thing. Usually, it's an implementation bug, not the actual missing info. It's also could be a missing piece of info (no class available in the source code at all), then it should be an error to the user before return. Idea was that if those errors won't be silenced, we will catch bugs sooner, and debugging is easier. 
Those exceptions could be silenced later at the top-level. Or they could be raised with a method, like `raise_incomplete_defn_exception` with a type of exception passed, and internally method introspects smth like `DJANGO_STUBS_DEBUG` environment variable. 

So, please, add raising an exception to all the places where there's not enough info to proceed. 

What do you think about this approach?

mypy_django_plugin/transformers/managers.py
```diff
@@ -0,0 +1,153 @@
...
+        if isinstance(bound_return_type, PlaceholderNode):
+            return
+
+        for arg_name, arg_type, original_argument in zip(
+                method_type.arg_names[1:],
+                method_type.arg_types[1:],
+                method_node.arguments[1:]):
+            bound_arg_type = self.semanal_api.anal_type(arg_type, allow_placeholder=True)
+            if bound_arg_type is None and not self.semanal_api.final_iteration:
+                self.semanal_api.defer()
```
mkurnikov: Same thing as above, `defer_till_next_iteration` or/and exception raise, instead of just `return`.

mypy_django_plugin/transformers/managers.py
```diff
@@ -0,0 +1,153 @@
...
+
+        arguments = []
+        bound_return_type = self.semanal_api.anal_type(
+                method_type.ret_type,
+                allow_placeholder=True)
+
+        assert bound_return_type is not None
+
+        if isinstance(bound_return_type, PlaceholderNode):
+            return
```
mkurnikov: Same with an exception instead of `return`.

mkurnikov (changes requested):

mypy_django_plugin/transformers/managers.py
kszmigiel: @mkurnikov will `if not isinstance(method_type, CallableType) and not self.defer_till_next_iteration()` work? I'm not 100% sure if I get it right how deferring works

mypy_django_plugin/transformers/managers.py
kszmigiel: I think that's just right how things should be done, I will follow this approach

mypy_django_plugin/transformers/managers.py
mkurnikov: If it's `not isinstance(CallableType)`, which type it is? Placeholder?

mypy_django_plugin/lib/helpers.py
kszmigiel: Adding an assert in `__init__` makes like no sense, we don't have `DynamicClassDefContext` available up there (only `NewSemanalDjangoPlugin` instance passed in `ManagerFromQuerySetCallback(self)` in `get_dynamic_class_hook`. I will add an assert in `__call__` though :)

mypy_django_plugin/transformers/managers.py
kszmigiel: It might be ` Overloaded`, `CallableType` or `None` actually

mypy_django_plugin/lib/helpers.py
mkurnikov: Yes, in `__call__`, my bad:)

mypy_django_plugin/transformers/managers.py
```diff
@@ -0,0 +1,152 @@
...
+                return_type=bound_return_type,
+                self_type=self_type)
+
+
+class ManagerFromQuerySetCallback(ManagerCallback):
+    def create_new_dynamic_class(self) -> None:
+
+        base_manager_info = self.callee.expr.node
+
+        if base_manager_info is None and not self.defer_till_next_iteration(reason='base_manager_info is None'):
```
mkurnikov, line 26: I don't think this `reason` is useful. `base_manager_info is None` _means_ something, it should be added as the reason. 
How could it happen that `TypeInfo` is `None` here? 

Also, let's see what type `self.callee.expr` is. It's an expression on the left side of the `a.b`. It could be `get_manager_class().from_queryset()` (which will be `CallExpr` as `self.callee.expr`, not a `RefExpr`), so there's no `node` attribute available, and so forth.
If it's not a `RefExpr`, a good idea would be to fallback to `Any` here.

mkurnikov (changes requested):

mypy_django_plugin/lib/helpers.py
```diff
@@ -115,7 +115,7 @@ def copy_method_to_another_class(
                 self_type=self_type)
             return
 
-        method_type: CallableType = method_node.type
+        method_type: CallableType = method_node.type # type: ignore
```
sobolevn: `cast` maybe?

mypy_django_plugin/lib/helpers.py
```diff
@@ -136,7 +136,7 @@ def copy_method_to_another_class(
                 method_node.arguments[1:]):
             bound_arg_type = self.semanal_api.anal_type(arg_type, allow_placeholder=True)
             if bound_arg_type is None and not self.defer_till_next_iteration(reason='bound_arg_type is None'):
-                raise new_helpers.TypeInfoNotFound('of ' + arg_name + ' argument of ' + method_node.fullname)
+                raise new_helpers.TypeInfoNotFound('of ' + arg_name + ' argument of ' + method_node.fullname) # type: ignore
```
sobolevn: Why do we need `type: ignore` here?

kszmigiel: Mypy was resolving `arg_name` as `None`, but now it's changed to:
```
error_msg = 'of {} argument of {}'.format(arg_name, method_node.fullname)
raise new_helpers.TypeInfoNotFound(error_msg)
```
and it works just fine

sobolevn: 👍

mypy_django_plugin/lib/helpers.py
kszmigiel: Applied, works, thanks for the tip 👍

django-stubs/contrib/admin/options.pyi
```diff
@@ -135,7 +135,7 @@ class ModelAdmin(BaseModelAdmin):
     delete_selected_confirmation_template: str = ...
     object_history_template: str = ...
     popup_response_template: str = ...
-    actions: Sequence[Callable[[ModelAdmin, HttpRequest, QuerySet], None]] = ...
+    actions: Sequence[Union[Callable[[ModelAdmin, HttpRequest, QuerySet], None], str]] = ...
```
sobolevn, line 138: I guess that this PR has unrelated changes from other PRs. Is that the case, @kszmigiel?

kszmigiel: Yes, I've updated stub files to the `master` version, I can revert it if needed (this commit: https://github.com/typeddjango/django-stubs/pull/427/commits/b032a3dd494e97a5e8f81bef80596a470a682874)

sobolevn: The proper way to do it is to `rebase` this PR based on current `master`

kszmigiel: 👍 It'll be quite a challenge for me, but I'll try

kszmigiel:
@sobolevn do you have any idea why this test I've added don't pass? I added 
```
from .models import ModelA

reveal_type(ModelA.objects.do_something())
```
in `views.py` of reproduction repository provided by @danifus and it works just fine

My Django version is `3.1b1` and Mypy is `0.782` if that helps

sobolevn (approved):
LGTM! 👍 

Can we merge it? @kszmigiel

sobolevn:
@kszmigiel looks like it is passing now.

kszmigiel:
@sobolevn I think we're ready to merge, but I would like to hear @mkurnikov opinion first. I know there are a few fragments I need to cleanup, but I think I can do this along with writing docs and comments after the rebase to master.

BTW how about getting readthedocs.io page for docs? I can take care of writing the docs as the task for next week - we lack a lot of them, and I could use a short break from coding too.

mypy_django_plugin/lib/helpers.py
```diff
@@ -118,11 +193,46 @@ def __call__(self, ctx: DynamicClassDefContext) -> None:
...
+class DynamicClassFromMethodCallback(DynamicClassPluginCallback):
+    callee: MemberExpr
+
+    def __call__(self, ctx: DynamicClassDefContext) -> None:
+        self.class_name = ctx.name
+        self.call_expr = ctx.call
+
+        assert ctx.call.callee is not None
+        if not isinstance(ctx.call.callee, MemberExpr):
+            # throw error?
```
mkurnikov, line 228: Add `assert` here, instead of isinstance+return.

mypy_django_plugin/transformers/managers.py
mkurnikov: @kszmigiel I still want you to add proper error string here with an explanation of the conditions of why it could be `None`.

test-data/typecheck/managers/querysets/test_from_queryset.yml
```diff
@@ -0,0 +1,236 @@
+-   case: from_queryset_with_base_manager
```
mkurnikov, line 1: According to the test, there has to be some kind of magic somewhere which adds `MyModel` as a generic argument for `NewManager`. Could we get rid of that? 
So, what I'm trying to say is that tests are a bit wrong. `from_queryset` line should be
```
class ModelQuerySet(models.QuerySet[MyModel]):
   ...
NewManager = BaseManager[MyModel].from_queryset(ModelQuerySet)
```
and current line should error out with `Need type annotation for` like it's done here https://github.com/typeddjango/django-stubs/blob/newsemanal-progress/test-data/typecheck/managers/test_managers.yml#L128
and `NewManager` in the current setup should give `Any` as a generic argument, see https://github.com/typeddjango/django-stubs/blob/newsemanal-progress/test-data/typecheck/managers/test_managers.yml#L106

Could be done in a separate PR though.

mkurnikov (changes requested):

kszmigiel:
🎉

mypy_django_plugin/lib/helpers.py
kszmigiel: Done, I will include this in next PR

PR #437, opened 2020-07-30
Fix type of update_fields parameter to Model.save

### LINKED ISSUE #434: Incorrect type annotation for update_fields parameter to Model.save
# Bug report

## What's wrong

Hi there, it looks like the type annotation for the `update_fields` parameter to `Model.save` and `Model.save_base` is incorrect. It's currently typed as `Optional[Union[Sequence[str], str]]` which indicates that a single string can be sent as an argument, but it doesn't look like Django actually supports that. If `update_fields` is not `None`, Django calls `frozenset(update_fields)` which in the case of a single string being submitted will result in a set of one-character strings for each character in the initial string.

The relevant type annotations are here:
https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/base.pyi#L41-L55

The relevant Django code is here:
https://github.com/django/django/blob/stable/3.0.x/django/db/models/base.py#L717

## How is that should be

I think that the correct type for the `update_fields` parameter should be `Optional[Iterable[str]]`. As mentioned above Django passes the value to `frozenset()`, which accepts any `Iterable`, not just `Sequence`.

I realize that this doesn't make much of a difference, as `str` is an `Iterable[str]` so you're allowed to make an invalid call either way, but I think it would be good to have a type annotation that doesn't mislead people into thinking they can pass a single `str` as `update_fields`.

Is there a specific reason why it's typed like this or should I submit a PR to update the type annotation?

## System information

- OS: macOS 10.15.6
- `python` version: `3.7.8`
- `django` version: `3.0.8`
- `mypy` version: `0.770`
- `django-stubs` version: `1.5.0`

ljodal:
Sending in a single string leads to unexpected behavior and the method
accepts any iterable, not just sequences, of strings.

Technically sending in a single string is still allowed because a string
is an iterable of strings, but this way the intention of that argument
is clearer.

## Related issues

Closes #434

### DIFF
--- a/django-stubs/db/models/base.pyi
+++ b/django-stubs/db/models/base.pyi
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Dict, List, Optional, Sequence, Set, Tuple, Type, TypeVar, Union, Collection
+from typing import Any, Callable, Collection, Dict, Iterable, List, Optional, Set, Tuple, Type, TypeVar, Union
 
 from django.core.checks.messages import CheckMessage
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


sobolevn (approved):

sobolevn:
Thanks!

PR #442, opened 2020-08-12
Add route argument to ResolverMatch constructor

federicobond:


sobolevn (approved):
Thanks!

Docs: https://github.com/django/django/blob/2.2.14/django/urls/resolvers.py#L31

PR #443, opened 2020-08-12
Add models.JSONField introduced in Django 3.1

federicobond:
Ref: https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.JSONField


sobolevn:
Currently we don't add django@3 specific things, see #439

mkurnikov:
Fixed in https://github.com/typeddjango/django-stubs/commit/159f8b95ba28a091f9ca4006b106099d5e793949

PR #446, opened 2020-08-18
Return type for translation.get_language should not be Optional[str]

bullfest:
Right now the return type for `django.utils.translation.get_language()` is `Optional[str]`. The only way I can see that it could end up returning `None` regardless of if `trans_real` or `trans_null` is active is if the `LANGUAGE_CODE` settings is `None`.

This would be a clear misconfiguration (and in violation of https://github.com/typeddjango/django-stubs/blob/2829faf1afd949b40248a81846686219d2733bb5/django-stubs/conf/global_settings.pyi#L44). It's still however perfectly possible to set `LANGUAGE_CODE` to `None` in your own settings without mypy complaining. But it's also an indication of your Django setup being clearly misconfigured. I would be willing to look into a way of preventing this if that's what's needed to merge this.


sobolevn (approved):

sobolevn:
Thanks a lot!

PR #448, opened 2020-08-21
Add force_color parameter to django.core.management.color.color_style

zach-waggoner:
`django.core.management.color.color_style` accepts a `force_color` argument. This is useful when overriding `django.utils.log.ServerFormatter` in order to force colors in a non-TTY context.


sobolevn (approved):

sobolevn:
Thanks a lot!

PR #449, opened 2020-08-24
Add "one_to_many", "one_to_one",  "many_to_many," and "many_to_one" attributes to Field base class

### LINKED ISSUE #447: Field class is missing "one_to_many", "one_to_one", "many_to_many", and "many_to_one" attrs
# Bug report

## What's wrong
According to [Django's documentation](https://docs.djangoproject.com/en/3.1/ref/models/fields/#attributes-for-fields-with-relations), these attributes should be present on all Field instances. However, the [Field](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/fields/__init__.pyi#L45) class stub does not include them. Instead, they are declared in the [RelatedField](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/fields/related.pyi#L41) class stub.

## How is that should be
The Field class stub should have the following addition:
```python3
one_to_many: Optional[bool] = ...
one_to_one: Optional[bool] = ...
many_to_many: Optional[bool] = ...
many_to_one: Optional[bool] = ...
```

## System information

- OS:
- `python` version: 3.8
- `django` version: 3.1
- `mypy` version: 0.770

If my understanding of this is correct, I'd be happy to submit a PR. Thank you!

james-perretta:
Closes #447


kszmigiel:
@james-perretta good job! 🎉

sobolevn (approved):

PR #451, opened 2020-08-27
@classproperty decorator has moved in Django 3.1

bullfest:
The `classproperty` decorator was moved in Django 3.1, this is an attempt to reflect that.

From https://docs.djangoproject.com/en/3.1/releases/3.1/#id1
> `django.utils.decorators.classproperty()` decorator is made public and moved to `django.utils.functional.classproperty()`.


bullfest:
Not sure about the status of 3.1 support (mostly due to discussion in #439), but I'll let this stay open in case you think this approach is ok.

mkurnikov:
Fixed in https://github.com/typeddjango/django-stubs/commit/11c3f8cd11a33a4e16ab207bb835285de5d7ed29

PR #453, opened 2020-08-28
Allows FileField storage to be a Callable

### LINKED ISSUE #452: FileField storage parameter should also accept a callable
# Bug report

## What's wrong

When running mypy against a code base with django-stubs installed I get an unexpected typing error:

```
.../models.py:30: error: Argument "storage" to "FileField" has incompatible type "Callable[[], Storage]"; expected "Optional[Storage]"  [arg-type]
```

## How is that should be

The `storage` parameter of the `FileField` class can take either an instance of `Storage` or a no-arg callable that will return an instance of `Storage` as stated here: https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.FileField.storage

## System information

- OS:
- `python` version: 3.8.3
- `django` version: 3.1
- `mypy` version: 0.782

a-musing-moose:
# I have made things!

The `storage` parameter of `FileField` (and by extension `ImageField`)
is not limited to just taking an instance of `Storage`. It can also take
a no-args callable that returns an instance of `Storage`.

## Related issues
- Closes: #452

### DIFF
--- a/django-stubs/db/models/fields/files.pyi
+++ b/django-stubs/db/models/fields/files.pyi
@@ -39,7 +39,7 @@ class FileField(Field):
     def __init__(
         self,
         upload_to: Union[str, Callable, Path] = ...,
-        storage: Optional[Storage] = ...,
+        storage: Optional[Union[Storage, Callable[[], Storage]]] = ...,
         verbose_name: Optional[Union[str, bytes]] = ...,
         name: Optional[str] = ...,
         max_length: Optional[int] = ...,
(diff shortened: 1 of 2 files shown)


sobolevn (approved):

sobolevn:
Can you please fix the formatting?

a-musing-moose:
> Can you please fix the formatting?

I certainly can - it is in an unrelated file i hope that is OK?

sobolevn (approved):

sobolevn:
Thanks!

PR #454, opened 2020-08-28
Allowed to use decimal in combinables

zefciu:
# Decimals can be used alongside floats in F expressions

Django allows to use Decimal values in F expressions just fine. Django-stubs however doesn’t list Decimal as compatible with operations on Combinable. This allows for F expressions with Decimal to check as correct.


kszmigiel:
@zefciu LGTM 👍 
@sobolevn what do you think about this one?

sobolevn (approved):

sobolevn:
Thanks a lot!

PR #457, opened 2020-09-09
Add a unique signature for FileField.contribute_to_class

brianhelba:
Since `FileField.contribute_to_class` uses a `**kwargs`, mypy does not recognize that it is functionally the same as `Field.contribute_to_class`, which does not.


kszmigiel:
LGTM 👍 

@brianhelba thank you!

PR #460, opened 2020-09-11
Fix wrong type for parse_until in template parser

bullfest:
https://docs.djangoproject.com/en/3.1/howto/custom-template-tags/#parsing-until-another-block-tag
> parser.parse() takes a tuple of names of block tags "to parse until"

Should clearly be a tuple of non-specific length (`Tuple[str, ...]`)


sobolevn (approved):

sobolevn:
Thanks!

PR #464, opened 2020-09-17
Check arguments to django.urls.{path,re_path} functions

intgr:
# I have made things!

These functions were added in Django 2.0, but the stubs have been
incomplete since commit 9a6826325716434d0c97ac36c99994f32a9464fb

The implementation is almost identical to `url()` from
`conf/urls/__init__.pyi`


sobolevn (approved):

sobolevn:
Thanks a lot!

PR #470, opened 2020-09-23
Add InlineModelAdmin.has_add_permission definition

### LINKED ISSUE #469: admin.TabularInline: Signature of "has_add_permission" incompatible with supertype "BaseModelAdmin"
## What's wrong

The signature for the `admin.TabularInline.has_add_permission` method seems to be wrongly typed, resulting in the following error when a subclass declares a type according to the Django documentation:
```
admin.TabularInline: Signature of "has_add_permission" incompatible with supertype "BaseModelAdmin"
```

Example signature resulting in this error:
```python
    def has_add_permission(self, request: HttpRequest, obj: Model) -> bool:
        return False
```

The problem seems to be caused by the TabluarInline directly inheriting the types from here
https://github.com/typeddjango/django-stubs/blob/19c73a106dc918606f8a9c3db018ccd3f18c1e06/django-stubs/contrib/admin/options.pyi#L109

When in reality the InlineModelAdmin has a different signature for the `has_add_permission` function:
https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#django.contrib.admin.InlineModelAdmin.has_add_permission

## How is that should be

The `has_add_permission` should be typed specifically for the `InlineModelAdmin` class rather than inheriting from `BaseModelAdmin`.

It would be a good idea to verify other `has_x_permission` methods at the same time, as the inline admin functions might differ from the base model admin

## System information

- OS: Linux
- `python` version: 3.8.5
- `django` version: 3.1.1
- `mypy` version: 0.770

JoaRiski:
Add a proper definition for the InlineModelAdmin's has_add_permission method.

Fixes #469

**NOTE: This change is entirely untested, so take extra care when reviewing.** I'm unsure if the type check works in stub inheritance the same way it does in actual code, but if yes it could be this change will hit the same "incompatible with supertype" error as mentioned in #469


blueyed (approved):
LGTM.

The difference with InlineModelAdmin is that the parent object gets passed in there.

https://github.com/django/django/blob/742d6bc8dbc3868201895933e1c40c36f46462c7/django/contrib/admin/options.py#L2087
https://github.com/django/django/blob/742d6bc8dbc3868201895933e1c40c36f46462c7/django/contrib/admin/options.py#L2170-L2177

Can you rebase it to re-trigger CI?

JoaRiski:
Looks like this needs to entirely remove the function from `BaseModelAdmin` and implement the `ModelAdmin` specific variant directly to `ModelAdmin`, as signature changes in subclasses don't seem to be allowed.

Would that sound reasonable?

JoaRiski:
@blueyed made the change proposed above, still untested (just used GitHub's editor), but I don't think there should be an issue unless `BaseModelAdmin` is used in unexpected places.

Care to re-trigger the workflow?

sobolevn (commented):
Or maybe we can use `# type: ignore` on `Inline`'s version of this method.

I think, that it will be more accurate. In this case, subclasses of `BaseModelAdmin` would still have `has_add_permission` method.

JoaRiski:
I haven't checked how inheritance works with `# type: ignore` in the base class. Does that approach allow us to still provide proper type hints for inheritance, or would it make essentially just untyped?

sobolevn:
Mypy copies the method definition from parent. It should be typed as we define it. But, we just silence the error there.

PR #472, opened 2020-09-29
Add deferrable argument to unique constraint (Django 3.1)

ticosax:
# Add deferrable argument to unique constraint

https://docs.djangoproject.com/en/3.1/ref/models/constraints/#uniqueconstraint


mkurnikov:
@sobolevn Same thing as in https://github.com/typeddjango/django-stubs/pull/489#issuecomment-763205001 ?

sobolevn (approved):

sobolevn:
Thanks, @ticosax and @mkurnikov! 👍

PR #475, opened 2020-09-29
Use AbstractBaseUser instead of AbstractUser where possible

### LINKED ISSUE #431: update_session_auth_hash second param should be `AbstractBaseUser`
https://github.com/typeddjango/django-stubs/blob/master/django-stubs/contrib/auth/__init__.pyi#L32

Currently needlessly limiting it to `AbstractUser` when in fact `update_session_auth_hash` looks like so:

```python
def update_session_auth_hash(request, user):
    """
    Updating a user's password logs out all sessions for the user.

    Take the current request and the updated user object from which the new
    session hash will be derived and update the session hash appropriately to
    prevent a password change from logging out the session from which the
    password was changed.
    """
    request.session.cycle_key()
    if hasattr(user, 'get_session_auth_hash') and request.user == user:
        request.session[HASH_SESSION_KEY] = user.get_session_auth_hash()
```

`AbstractBaseUser`, meanwhile, has `get_session_auth_hash` even on this set of typestubs

```python
class AbstractBaseUser(models.Model):

    ...

    def get_session_auth_hash(self) -> str: ...

   ...
```

Should prefer the superclass here.

Thanks for a great set of stubs otherwise!

henribru:
There's a few different places where `AbstractUser` is used even though they work with `AbstractBaseUser`, this fixes that. Note that I've left the use of `AbstractUser` in `user_passes_test`, as `staff_member_required` uses it with the assumption that the user has a `is_staff` attribute.

## Related issues
- Closes #431


sobolevn (approved):

sobolevn:
Thanks a lot!

PR #476, opened 2020-09-29
Template Loader.get_template_sources should yield Origins

wkschwartz:
`django.template.loaders.base.Loader.get_template_sources` [should](https://docs.djangoproject.com/en/3.1/ref/templates/api/#django.template.loaders.base.Loader.get_template_sources) yield `Origins`. Currently it's returning `None`. Objects that yield [should](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#generators) be marked as iterables.

I'm marking this PR as "allow edits by maintainers." Please don't wait for me if you like this patch and it needs changes before being merged!


sobolevn (approved):

sobolevn:
Thanks!

PR #478, opened 2020-10-05
Add stubs for django.core.asgi

rafis:
```
$ mypy app
app\asgi.py:12: error: Skipping analyzing 'django.core.asgi': found module but no type hints or library stubs
app\asgi.py:12: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```


Goldziher:
Thanks, but this is already included in latest version of the code.

PR #481, opened 2020-10-07
renderer = BaseRenderer instead of EngineMixin in Widget.render()

### LINKED ISSUE #480: Widget.render() renderer argument should accept BaseRenderer
# Bug report

## What's wrong

I've got a class that is a subclass of [`Widget`](https://docs.djangoproject.com/en/3.1/ref/forms/widgets/#widget) and overrides the [`render()`](https://docs.djangoproject.com/en/3.1/ref/forms/widgets/#django.forms.Widget.render) method.

The class + overridden method look something like this:

```python
import typing as t

from django import forms
from django.forms.renderers import BaseRenderer
from django.utils.html import format_html
from django.utils.safestring import SafeText

class SomeWidget(forms.TextInput):
    def render(
        self,
        name: str,
        value: str,
        attrs: t.Any = None,
        renderer: t.Optional[BaseRenderer] = None,
    ) -> SafeText:
        html = super().render(name, value, attrs=attrs, renderer=renderer)
        extra_html = 'extra-html-here'
        html += extra_html
        return format_html(html, value=value)
```
This produces the following error:
```
error: Argument 4 of "render" is incompatible with supertype "Widget"; supertype defines the argument type as "Optional[EngineMixin]"  [override]
note: This violates the Liskov substitution principle
note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

This is because of this part:
https://github.com/typeddjango/django-stubs/blob/be3fc259d5c13ad22cc22ab52596fd802ebe2205/django-stubs/forms/widgets.pyi#L44-L46

## How is that should be
In django, the [`Widget.render()`](https://github.com/django/django/blob/599f2f79e22aaa475e45b72768babf70ac71a6b2/django/forms/widgets.py#L244-L247) method uses the `renderer` argument to trigger the [`_render()`](https://github.com/django/django/blob/599f2f79e22aaa475e45b72768babf70ac71a6b2/django/forms/widgets.py#L249-L252) method, which in turn triggers `renderer.render()`.

The [`EngineMixin` class](https://github.com/django/django/blob/2d2fbc3a70e4c3c38f0f78869dacd5588aa6a3bd/django/forms/renderers.py#L28-L39) does not have a `render()` method, but [`BaseRenderer`](https://github.com/django/django/blob/2d2fbc3a70e4c3c38f0f78869dacd5588aa6a3bd/django/forms/renderers.py#L19-L25) does. Therefore, `renderer` should accept type `Optional[BaseRenderer]` instead of `Optional[EngineMixin]`.

## System information

- OS: macOS Catalina v10.15.6
- `python` version: 3.7.7
- `django` version: 2.2.16
- `mypy` version: 0.761

mariamrf:
## Related issues
- Closes #480


sobolevn (approved):
Thanks a lot!

PR #485, opened 2020-10-11
updated package setup

Goldziher:
Rebased version of the previous PR:

add pre-commit
improve the travis setup
cleanup and update the md files
improve the project folder structure
add stubgen scripts
update setup.py

PR #487, opened 2020-10-12
fix mypy version < to <=

### LINKED ISSUE #486: No compatibility with the new version of mypy
# Bug report

## What's wrong

After updating mypy to the newest version I got this Error message:
```console
django-stubs 1.6.0 requires mypy<0.790,>=0.782, but you'll have mypy 0.790 which is incompatible.
```

## How is that should be

could you please update django-stubs so it can be compatible with the new version of mypy, thank you

## System information

- OS: ubuntu 20.04
- `python` version: 3.8
- `django` version: 3.1.2
- `mypy` version: 0.790

kszmigiel:
closes #486


kszmigiel:
@sobolevn So 1.6.1 it is 😂

setup.py
```diff
@@ -21,9 +21,9 @@ def find_stub_files(name: str) -> List[str]:
     readme = f.read()
 
 dependencies = [
-    "mypy>=0.782,<0.790",
-    "typing-extensions",
-    "django",
+    'mypy>=0.782,<=0.790',
```
sobolevn: I guess it should be `'mypy>=0.782` 🤔

setup.py
```diff
@@ -21,9 +21,9 @@ def find_stub_files(name: str) -> List[str]:
...
+    'typing-extensions',
+    'django',
```
sobolevn, line 26: Btw, let's use `black` on `setup.py` as well.

django-stubs/test/runner.pyi
```diff
@@ -70,7 +70,7 @@ class ParallelTestSuite(TestSuite):
     processes: Any = ...
     failfast: Any = ...
     def __init__(self, suite: Any, processes: Any, failfast: bool = ...) -> None: ...
-    def run(self, result: Any): ...
+    def run(self, result: Any, debug: Optional[bool]) -> Any: ...
```
sobolevn: Does `debug` have a default? If so, we need to specify that.

sobolevn (changes requested):
Thanks a lot!

django-stubs/test/runner.pyi
```diff
@@ -70,7 +70,7 @@ class ParallelTestSuite(TestSuite):
     processes: Any = ...
     failfast: Any = ...
     def __init__(self, suite: Any, processes: Any, failfast: bool = ...) -> None: ...
-    def run(self, result: Any, debug: Optional[bool]) -> Any: ...
+    def run(self, result: Any, debug: Optional[bool] = False) -> Any: ...
```
sobolevn, line 73: `= ...`

sobolevn (changes requested):
Thanks! The last nitpick!

Goldziher:
There is already a PR that is doing this, https://github.com/typeddjango/django-stubs/pull/485

also the changes to the test runner are incorrect here - its an override.

.travis.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
     - name: Lint with black
       python: 3.7
-      script: 'black --check django-stubs/'
+      script: 'black --check django-stubs/ setupy.py'
```
sobolevn, line 41: `setup.py`

sobolevn (changes requested):

sobolevn:
Fixed in #501

setup.py
oladhari: @kszmigiel and @sobolevn sorry for commenting on merged PR, does this means that it supports any new version of mypy, now mypy has updated version to `0.812` in the table in Readme file:https://github.com/typeddjango/django-stubs#version-compatibility
django-stubs supports till mypy version 0.790
thank you

PR #489, opened 2020-10-14
Add algorithm parameter (Django 3.1)

okapies:
Add `algorithm` parameter or field, which is introduced in Django 3.1, to the following functions/classes:

- `django.core.signing.Signer.__init__`
- `django.core.signing.base64_hmac`
- `django.utils.crypto.salted_hmac`
- `django.contrib.auth.tokens.PasswordResetTokenGenerator`

## Related issues
- django/django#12374
- django/django#12454
- https://code.djangoproject.com/ticket/27468


sobolevn:
Thanks a lot! I am going to leave this hanging for a while 😞 

Please, see #262

sobolevn (approved):

okapies:
@sobolevn I see. I also think there might be a problem around the versioning. Please leave it until the issue is resolved.

mkurnikov:
@sobolevn I think it should be backwards compatible? The parameter is optional, the worst case will be a false-positive, which is fine. Right? Please merge it, if you agree with this.

PR #490, opened 2020-10-14
Use more specific base classes in model exceptions

noamkush:
This changes the base classes for `Model.DoesNotExist` and `Model.MultipleObjectsReturned` to the correct base classes.


sobolevn (approved):

sobolevn:
Thanks!

PR #496, opened 2020-10-19
user parameter for login accepts None

### LINKED ISSUE #495: django.contrib.auth.login should accept Optional[AbstractBaseUser] as second argument
# Bug report

## What's wrong

Django project's `login` function handles None as second argument- https://github.com/django/django/blob/master/django/contrib/auth/__init__.py#L87

However, the stubs only accept `AbstractBaseUser`

## How is that should be

https://github.com/typeddjango/django-stubs/blob/8729a74f8141e24ca630507e4271f984a91a53d2/django-stubs/contrib/auth/__init__.pyi#L26 Should accept `Optional[AbstractBaseUser]` as second argument

## System information

- OS:
- `django` version: 3.1.2

giladsheffer:
## Related issues

Closes #495


django-stubs/contrib/auth/__init__.pyi
```diff
@@ -23,7 +23,7 @@ def load_backend(path: str) -> ModelBackend: ...
 def get_backends() -> List[ModelBackend]: ...
 def authenticate(request: Any = ..., **credentials: Any) -> Optional[AbstractBaseUser]: ...
 def login(
-    request: HttpRequest, user: AbstractBaseUser, backend: Optional[Union[Type[ModelBackend], str]] = ...
+    request: HttpRequest, user: Optional[AbstractBaseUser], backend: Optional[Union[Type[ModelBackend], str]] = ...
```
sobolevn, line 26: Can you please also specify that `backend` has a default value `= ...`?
https://github.com/django/django/blob/master/django/contrib/auth/__init__.py#L87

sobolevn (changes requested):
Thanks! 👍

django-stubs/contrib/auth/__init__.pyi
sobolevn: Oh, it is already there! Sorry!

sobolevn (approved):

PR #497, opened 2020-10-22
Allow overridable checks to be CheckMessage

brianhelba:
The checks framework expects return types to be `List[CheckMessage]`. This allows all overridable checks-returning methods to return the more general `CheckMessage`, instead of the `Error` subclass.


sobolevn:
@brianhelba thanks!

We have a failing test:

> django-sources/tests/admin_checks/tests.py:248: error: Argument 1 to "append" of "list" has incompatible type "str"; expected "CheckMessage"

Which is now probably this line: https://github.com/django/django/blob/509d9da26fb92a8d566ec105ff40bb024803ceaa/tests/admin_checks/tests.py#L264

We have to ignore it.

sobolevn (approved):

sobolevn:
Good catch! 👍

PR #498, opened 2020-10-22
Fix #224 add paths in mypy's mypy_path option for import discovery

### LINKED ISSUE #224: ModuleNotFoundError using a custom apps directory & path

danifus:
Search paths in mypy_path config option for module resolution

This allows packages to be located outside of PYTHONPATH and still be imported.

Adds a test for this feature

## Related issues
Closes #224

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -1,8 +1,10 @@
 import configparser
+import sys
 from functools import partial
 from typing import Callable, Dict, List, NoReturn, Optional, Tuple, cast
 
 from django.db.models.fields.related import RelatedField
+from mypy.modulefinder import mypy_path
 from mypy.nodes import MypyFile, TypeInfo
 from mypy.options import Options
 from mypy.plugin import (
@@ -91,6 +93,10 @@ class NewSemanalDjangoPlugin(Plugin):
     def __init__(self, options: Options) -> None:
         super().__init__(options)
         django_settings_module = extract_django_settings_module(options.config_file)
+        # Add paths from MYPYPATH env var
+        sys.path.extend(mypy_path())
+        # Add paths from mypy_path config option
+        sys.path.extend(options.mypy_path)
         self.django_context = DjangoContext(django_settings_module)
 
     def _get_current_queryset_bases(self) -> Dict[str, int]:
--- a/test-data/typecheck/test_config.yml
+++ b/test-data/typecheck/test_config.yml
@@ -38,3 +38,30 @@
               class MyModel(models.Model):
                   user = models.ForeignKey('auth.User', on_delete=models.CASCADE)
 
+-   case: add_mypy_path_to_package_search
+    main: |
+        import extra_module
+    mypy_config: |
+        [mypy]
+        mypy_path = ./extras
+        [mypy.plugins.django-stubs]
+        django_settings_module = mysettings
+    files:
+        -   path: extras/extra_module.py
+            content: |
+                def extra_fn():
+                  pass
+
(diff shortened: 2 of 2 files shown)


mypy_django_plugin/main.py
```diff
@@ -91,6 +92,7 @@ class NewSemanalDjangoPlugin(Plugin):
     def __init__(self, options: Options) -> None:
         super().__init__(options)
         django_settings_module = extract_django_settings_module(options.config_file)
+        sys.path.extend(options.mypy_path)
```
sobolevn, line 99: What about `MYPYPATH` env var?

sobolevn (commented):
Thanks a lot! Quick questions.

mypy_django_plugin/main.py
danifus: Good point. It wasn't looking there but I've updated it so that it will. I haven't included a test for the env var though. Is there a good way of setting env vars in the test suite?

I was looking at https://github.com/python/mypy/blob/master/mypy/modulefinder.py#L541 for inspiration. Is there anything else we should incorporate from there?

test-data/typecheck/test_config.yml
```diff
@@ -38,3 +38,16 @@
               class MyModel(models.Model):
                   user = models.ForeignKey('auth.User', on_delete=models.CASCADE)
 
+-   case: add_mypy_path_to_package_search
```
sobolevn, line 41: We can add a test for `MYPYPATH` as well.
`pytest-mypy-plugins` supports `env:` section.
see https://github.com/typeddjango/pytest-mypy-plugins/blob/master/pytest_mypy_plugins/tests/test-paths-from-env.yml

sobolevn (changes requested):
Thanks a lot!

Let's add one more test and we are fine! 👍

test-data/typecheck/test_config.yml
danifus: I've added the test for MYPYPATH env var. Let me know if there is anything else I should add. Thanks for the help!

sobolevn (approved):
Thanks a lot! Great work.

PR #499, opened 2020-10-22
Ensure that all registered checks take the same parameters

brianhelba:
The Django API requires that registered checks take an `app_configs` argument, which is a `List` of `AppConfig`. While checks could only accept an `Iterable`, in principle they can rely on having a more specific type of `List`, and internally use more powerful operations (like random access) on it. This is practically also passed by keyword, so all parameters should be specified with a `= ...` to indicate that they are keywords.

Django always passed `app_configs` as a list, but checks can accept and use a more general more `Sequence`.

The Django API also requires that registered checks take `**kwargs`.

This results in the canonical parameters of:
`app_configs: Optional[Sequence[AppConfig]] = ..., **kwargs: Any`


brianhelba:
@sobolevn I'm a little unsure about the second commit in this PR:
> Make all registered checks have a more specific parameter type
> 
> Django always passes registered checks a `List` or `None` as the `app_configs` argument. While checks could only accept an `Iterable`, in principle they can rely on having a more specific type of `List`, and internally use more powerful operations (like random access) on it.
>
> This results in the canonical parameters of:
> `app_configs: Optional[List[AppConfig]] = ..., **kwargs: Any`

What's the policy on specification of parameters? If we *know* that a function will be passed `List` when called by Django, should we require that it accept a `List`, or should we only require that it accept the more general, weaker `Iterable`?

Depending on the decision, I can drop the second commit.

sobolevn:
Well, it depends 🙂 

Do we have any specific constraints that force us to use `List` and not just `Iterable`?
Btw, there are also protocols like `Sequence` in between `Iterable` and `List`.

brianhelba:
In the case of these functions, I *think* they don't do random access, and would probably still work if they were called with a more general `Iterable`, but we know that Django itself will always call them with a `List`.

I think `Sequence` is often useful for user-facing APIs (those are called directly by users), since people (and even upstream Django :disappointed_relieved:) often treat `Tuple` and `List` interchangeably. For a Django-facing API like this one (meaning only Django should ever call these functions, outside of test cases), I think it's actually simpler to just guarantee that a `List` will be passed.

For the code here, this doesn't matter as much. However, what we decide here will also affect how all user-defined checks are expected to be declared. [Django's docs on this](https://docs.djangoproject.com/en/3.1/topics/checks/#writing-your-own-checks) only talk about using `List`, so I'd be fine if we keep that strong bound.

sobolevn:
Yes, `Sequence` is good enough! 👍

brianhelba:
@sobolevn I've updated this to have checks accept a `Sequence[AppConfig]`.

sobolevn:
@brianhelba please, run `black` on your changes.

brianhelba:
Done.

sobolevn (approved):
Thanks a lot for your hard work 👍

PR #504, opened 2020-10-27
make BaseModelAdmin generic to properly type methods dealing with models

### LINKED ISSUE #482: BaseModelAdmin should be a generic class
# Bug report

## What's wrong

I have a class inheriting from `BaseModelAdmin` and a method `delete_model` defined as such:
```python
def delete_model(self, request: HttpRequest, obj: AgendaItem) -> None:
```
`AgendaItem` is a Django model

However, mypy raises an error:

```
app/chronology/admin.py: note: In member "delete_model" of class "AgendaItemAdmin":
app/chronology/admin.py:376: error: Argument 2 of "delete_model" is
incompatible with supertype "ModelAdmin"; supertype defines the argument type as
"Model"  [override]
        def delete_model(self, request: HttpRequest, obj: AgendaItem) -> N...
        ^
```

because in stubs this function is defined as:

```python
def delete_model(self, request: HttpRequest, obj: Model) -> None: ...
```

## How is that should be

The solution is to make `BaseModelAdmin` a generic class with model type variable.

## System information

- OS: ArchLinux
- `python` version: 3.8.6
- `django` version: 3.1.2
- `mypy` version: 0.782

ghost:
# i have made things!

## related issues

closes #482

this is my first time really working with mypy like this, apologies if i made any mistakes that i didn't catch


ghost:
black failed in ci, but when i ran it locally it wanted to make a *lot* of completely unrelated changes, which i figured wouldn't really be good for this pr

django-stubs/contrib/admin/options.pyi
```diff
@@ -199,7 +199,7 @@ class ModelAdmin(BaseModelAdmin):
     ) -> None: ...
     def save_form(self, request: Any, form: Any, change: Any): ...
     def save_model(self, request: Any, obj: Any, form: Any, change: Any) -> None: ...
```
sobolevn: Do we need to type `obj` as `_T` here?

sobolevn (changes requested):
Thanks! 👍 

Please, reformat this file only, I think that it is ok to trust black to do its job.
Maybe it has a new version / rule to apply.

django-stubs/contrib/admin/options.pyi
ghost: sure do! i was really tired when i worked on this, sorry :p

sobolevn: No worries! Thanks a lot for your help!

ghost:
i added a new typevar `_ModelT` that's bound to `Model`, and if an argument to a method looked like it should be generic, i typed it as `_ModelT`. i wasn't super sure on some of them, but i figured it would be better to be safe, since i don't think it'll really create any issues

sobolevn (changes requested):
Awesome work! 👍 

Let's add a single test case for `ModelAdmin` class! Currently, we are terrible at test-coverage 😞 
You can take some inspiration from https://github.com/typeddjango/django-stubs/blob/master/test-data/typecheck/contrib/admin/test_options.yml

ghost:
i'd love to add a test case, but i'm not quite clear on what i would test. just write subclass of `ModelAdmin` to make sure it typechecks?

sobolevn:
I was thinking about revealing type of https://github.com/typeddjango/django-stubs/pull/504/files#diff-8143d231653e2933a801943d02bd50968776e6f1805a191aad7012d4881b129cR162

ghost:
would something like this work? reveal_type worked, but the output looked really unintuitive, so i figured it might be best to just have the field typecheck

if that's good with you, i'll add something that doesn't typecheck to make sure the constraint works properly

sobolevn (approved):

sobolevn:
Thanks a lot!

sobolevn:
> if that's good with you, i'll add something that doesn't typecheck to make sure the constraint works properly

Feel free to send a new PR with an extra test 👍

PR #506, opened 2020-10-28
Add django.forms.widgets.FILE_INPUT_CONTRADICTION symbol

brianhelba:
https://github.com/django/django/blob/302caa40e4caab7d95ef7d0a88a90f935039ab09/django/forms/widgets.py#L399


django-stubs/forms/widgets.pyi
```diff
@@ -71,6 +71,8 @@ class HiddenInput(Input):
 class MultipleHiddenInput(HiddenInput): ...
 class FileInput(Input): ...
 
+FILE_INPUT_CONTRADICTION: Any
```
sobolevn, line 74: Shouldn't it be `object`?

brianhelba: I was following the example of other some other sentinel objects. For example:
https://github.com/django/django/blob/302caa40e4caab7d95ef7d0a88a90f935039ab09/django/db/models/fields/mixins.py#L3
https://github.com/typeddjango/django-stubs/blob/cb1fa08a8298717c3284ea1b895290de150626c8/django-stubs/db/models/fields/mixins.pyi#L5

I could not find any examples in `django-stubs` of these sorts of sentinel objects being annotated as `: object`.

I can change this, if you think it should be `object`.

sobolevn (approved):

sobolevn:
Thanks a lot!

PR #508, opened 2020-10-29
Ensure that all registered checks return a list of CheckMessage

brianhelba:
This changes missing and `Any` return types on registered checks to always return `List[CheckMessage]`.

In many cases, check functions were already annotated to return in a more specific type than `CheckMessage` (e.g. `Error`). In these cases, it's assumed that the existing annotation is correct, and the more specific type is kept.


django-stubs/core/checks/model_checks.pyi
```diff
@@ -1,8 +1,8 @@
 from typing import Any, List, Optional, Sequence
 
-from django.core.checks.messages import Warning
+from django.core.checks.messages import CheckMessage, Warning
 
 from django.apps.config import AppConfig
 
 def check_all_models(app_configs: Optional[Sequence[AppConfig]] = ..., **kwargs: Any) -> List[Warning]: ...
```
brianhelba: I'm not sure if leaving the more specific functions alone is the best policy. Assuming that the current annotation is correct, in practice the current Django implementation does only return `Warning`. However, this is just accidental, as the API allows it to return the more general `CheckMessage`.

Should we annotate around what the implementation actually does or what the API can guarantee?

sobolevn: > what the API can guarantee?

Let's stick with this 👍

PR #509, opened 2020-10-29
Provide a more specific signature for django.core.checks.register

brianhelba:
This also corrects the fact that `run_checks` only returns a concatenated `list` of all registered check results, which may only contain `CheckMessage`:
https://github.com/django/django/blob/302caa40e4caab7d95ef7d0a88a90f935039ab09/django/core/checks/registry.py#L60


sobolevn (approved):

sobolevn:
Thanks!

Issue #511, opened 2020-10-30
Proposal: Allow narrowing ORM fields


antonagestam:
I am currently working on implementing [phantom types for Python](https://github.com/antonagestam/phantom-types), which, when implemented throughout an application, would allow statically proving that a value saved on model is validated.

To be able to use this with django-stubs there would need to be some way of narrowing the types that ORM fields hold.

One example would be a field for country codes:

```python
from phantom.ext.iso3166 import CountryCode


class Member(models.Model):
    name = CharField()
    country: CountryCode = CharField(max_length=2)
```

This makes mypy error if I try to assign a string value to the field:

```python
m = Member()
m.country = "SE"  # E: Incompatible types in assignment (expression has type "str", variable has type "CountryCode")
```

Unless I first prove that the string is a valid country code:

```python
country_code = "SE"
assert isinstance(country_code, CountryCode)
m.country = country_code
# ... or
m.country = CountryCode.parse(country_code)
```

Unfortunately mypy raises an error when annotating ORM fields:

```
Incompatible types in assignment (expression has type "CharField[Union[str, int, Combinable, None], Optional[str]]", variable has type "CountryCode") [assignment]
```

Although my usecase is phantom types, I believe a feature that enables this would also work equally well with `typing.NewType` types.

This can be worked around with using the usual dirty tricks of ignore comments or using e.g. `typing.cast(CountryCode, CharField())`. But it would be very nice to see builtin support for this. I think it would be optimal if this builtin support also verifies that the annotated type is a subtype of the type that the field holds, so that the type annotated to a `CharField` must be a subtype of `str` and so on.

What do you think?

_PS.: This is a proposal to get a discussion started, I'm not asking anyone to implement this for me._

sobolevn:
Hi, @antonagestam! 

You can use this notation:

```python
score: 'models.Field[int, int]' = models.IntegerField(default=0)
```

Will it work for you?

The problem is that you currently can specify wrong generic params:

```python
score: 'models.Field[str, str]' = models.IntegerField(default=0)  # wrong, but typechecks!
```

antonagestam:
@sobolevn Cool, I did not know about that! I think I'm running into a variance issue though. 

I setup a minimal Django project to test with, and testing with this definition just to have less moving parts to think about (also tested with `NewType` with same results):

```python
from __future__ import annotations

from django.db import models


class CountryCode(str):
    pass


class Member(models.Model):
    name = models.CharField(max_length=200)
[...]

PR #513, opened 2020-10-31
Make Field typevars contravariant

antonagestam:
#### Related issues

- Refs #511


sobolevn (changes requested):
Thanks! 

Let's fix the tests and add a new one with your sample from the original issue! 
I would love to get this merged.

antonagestam:
@sobolevn mypy doesn't allow the return type to be contravariant, which is probably sane, but I think this means it will be very hard to support a notation like this. Using a cast works as expected but is a bit disappointing.

```python
country = cast("models.Field[CountryCode, CountryCode]", models.CharField(max_length=2))
```

PR #515, opened 2020-10-31
make FormMixin generic to allow proper typing for LoginView

### LINKED ISSUE #514: LoginView has a typing issue
# bug report

## what's wrong

django's LoginView (django.contrib.auth.views.LoginView) violates the Liskov substitution principle. we see [here](https://github.com/django/django/blob/4f61810751751b8c5070ce038ea57e949650e9e3/django/contrib/auth/views.py#L90) that the view's `form_valid` method calls `form.get_user()`, meaning that `form` must be of type `AuthenticationForm`. however, `FormMixin` defines `form` as `BaseForm`, and attempting to type `form` properly results in a violation of the liskov substitution principle

the following minimal failing example overrides LoginView and copies the `form_valid` method exactly (with type hints added):

```py
from django.contrib.auth.views import LoginView
from django.forms import BaseForm
from django.contrib.auth import login as auth_login
from django.http import HttpResponseRedirect

class MyLoginView(LoginView):
    def form_valid(self, form: AuthenticationForm) -> HttpResponseRedirect:
        """Security check complete. Log the user in."""
        auth_login(self.request, form.get_user())
        return HttpResponseRedirect(self.get_success_url())
```

it fails with the following error:

```
$ mypy testproj/views.py 
testproj/settings.py:28: error: Need type annotation for 'ALLOWED_HOSTS' (hint: "ALLOWED_HOSTS: List[<type>] = ...")
testproj/views.py:7: error: Argument 1 of "form_valid" is incompatible with supertype "FormMixin"; supertype defines the argument type as "BaseForm"
testproj/views.py:7: note: This violates the Liskov substitution principle
testproj/views.py:7: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
Found 2 errors in 2 files (checked 1 source file)
```

## how is that should be

this shouldn't fail. i'm not exactly sure if this can be fixed in django-stubs or if this is something that needs to be fixed in django

clearly django's LoginView class expects form to be an `AuthenticationForm` due to the `get_user` method call, but typing it as such would violate the liskov principle

## system information

- os: fedora 32
- `python` version: 3.8.6
- `django` version: 3.1.2
- `mypy` version: 0.790

ghost:
# i have made things!

## related issues

closes #514

tests look good, i added a new one, let me know if you want any additional tests


sobolevn (approved):
Awesome work! Thanks!

Let's ask someone else for a quick look, maybe we are missing something!
And then merge it 🛷

Goldziher (approved):

sobolevn:
🎉

ghost:
awesome, thanks!

PR #516, opened 2020-10-31
Fix inaccurate ModeAdmin type.

sssbox:
Hello

formset of `InlineModelAdmin` should be a subclass of `BaseFormSet` not an instance of it.

## Related issues

-Refs https://github.com/typeddjango/django-stubs/pull/375/files#r514897398

All tests that passed before this change pass afterwards.


adamchainz:
Thanks for the PR, this ended up being fixed in duplicate PR #521

Issue #512, opened 2020-10-30
user.is_staff complex type


amikrop:
Hello, as I am not sure if this is an issue related to django-stubs, or djangorestframework-stubs, or just my code, I am posting here as well. I created this issue https://github.com/typeddjango/djangorestframework-stubs/issues/109
Seems `user.is_staff` does not agree with the `bool` return type annotation I gave to my method. Is this an issue for fixing in django-stubs or djangorestframework-stubs? If it is an issue of my code, please suggest what I could do.

sobolevn:
Can you please post what `reveal_type(user)` and `reveal_type(user.is_stuff)` produces?

amikrop:
user: Revealed type is 'Union[django.contrib.auth.models.User, django.contrib.auth.models.AnonymousUser]'
user.is_staff: Revealed type is 'Union[django.db.models.fields.BooleanField[Union[builtins.bool, django.db.models.expressions.Combinable], builtins.bool], builtins.bool]'

ghost:
could always just use `bool(user.is_active)`. calling `user.clean_fields()` will ensure the instance has a bool there

amikrop:
> could always just use `bool(user.is_active)`. calling `user.clean_fields()` will ensure the instance has a bool there

Yep, I was hoping to solve it without adding more visual clutter to the code.

MarcinWieczorek:
I've read your comments and referred issues/PRs and I'm still not sure what is the solution other than wrapping it in `bool()`. I think that assuming bool should be totally okay, but I don't understand how it works in django/drf in general. I can use a model's let's say `FloatField` as a `float`. Here it breaks because of the union, and this (failing) testcase shows it.

```patch
diff --git a/tests/typecheck/contrib/auth/test_misc.yml b/tests/typecheck/contrib/auth/test_misc.yml
index b9207ce..169f9fb 100644
--- a/tests/typecheck/contrib/auth/test_misc.yml
+++ b/tests/typecheck/contrib/auth/test_misc.yml
@@ -13,3 +13,21 @@
         from django.test import Client
 
         get_user(Client())
+-   case: test_user_fields
+    main: |
+        from typing import Union
+        from django.contrib.auth.models import User
+        from django.contrib.auth.models import AnonymousUser
+
+        anonymous: AnonymousUser
+        is_staff1: bool = anonymous.is_staff
+        is_active1: bool = anonymous.is_active
+
+        user: User
+        is_staff2: bool = user.is_staff
[...]

PR #517, opened 2020-11-01
Add `is_active` field to `django.contrib.auth.AbstractBaseUser` model

### LINKED ISSUE #212: Django model field
Hi, I am using django and I here is a bug I encountered, the method `django.contrib.auth.authenticate` returns and instance of `django.contrib.auth.models.AbstarctBaseUser`.

If I check for `user.is_active` pylance claims **Cannot access member "is_active" for type "AbstractBaseUser"** even though it is a defined field on the `AbstractBaseUser` class as shown in the screenshot.

Screenshot from 2020-08-07 14-34-08

-   Language Server version: pylance (v2020.8.0)
-   OS and version: Ubuntu 20.04 64bit
-   Python version: python 3.8.2 (pyenv, venv)

yankyhoffman:
Add the `is_active` property on the `django.contrib.auth.AbstractBaseUser` model as defined in the [Django source code](https://github.com/django/django/blob/master/django/contrib/auth/base_user.py#L52).

## Related issues

- Closes [#212 (Pylance)](https://github.com/microsoft/pylance-release/issues/212)


sobolevn:
Maybe `is_active: Literal[True] = ...`?

yankyhoffman:
The Django user model defines the `is_active` property and it [defaults](https://docs.djangoproject.com/en/3.1/topics/auth/customizing/#django.contrib.auth.models.CustomUser.is_active) to `True`, I wouldn't say it should be defined as `Literal` i.e. a sub class might want to override it.

(Or perhaps I don't yet fully understand the definition of `Literal[True]` (I didn't really use too much types yet, still trying to break into the habit)).

sobolevn:
Yes, `Literal` is too much. It should be `is_active: bool = ...` then! 👍

yankyhoffman:
> Yes, `Literal` is too much. It should be `is_active: bool = ...` then!

Fair point, as this is indeed the point of a property in an abstract class.

I followed your suggestion in my newest commit.

sobolevn (approved):

sobolevn:
Ok, we have a problem

> ../../home/travis/build/typeddjango/django-stubs/django-stubs/contrib/auth/models.pyi:81: error: Incompatible types in assignment (expression has type "BooleanField[Union[bool, Combinable], bool]", base class "AbstractBaseUser" defined the type as "bool")

yankyhoffman:
Wow! I see I still have plenty to learn regarding using a typing system on python.

What I don't understand is why doesn't `bool` match the requirements of `db.models.BooleanField`? As I see defined in the [`db.models.fields`](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/fields/__init__.pyi#L265:L268) its `_pyi_private_set_type` and `_pyi_private_get_type` types should match `bool` (at least as far as I understand).

sobolevn:
I recently saw an issue about that as well: https://github.com/typeddjango/django-stubs/issues/512

Probably, the easiest way for now is to annotate `is_active` as it asks.

yankyhoffman:
I updated the type to reflect the implemented type in the sub class in my latest commit.

I ran `pytest` locally with `dev-requirements.txt` installed and the tests which failed previously have passed, however [1 test](https://github.com/typeddjango/django-stubs/blob/master/tests/typecheck/models/test_inheritance.yml#L94) (out of the 187) has failed, with
```
django.core.exceptions.ImproperlyConfigured: Could not find the GDAL library (tried "gdal", "GDAL", "gdal3.1.0", "gdal3.0.0", "gdal2.4.0", "gdal2.3.0", "gdal2.2.0", "gdal2.1.0", "gdal2.0.0"). Is GDAL installed? If it is, try setting GDAL_LIBRARY_PATH in your settings. (diff)
```
at the bottom of the stack trace, might be due to my environment setup..,

**Edit**: It indeed appears that it was due to my environment being different from the testing environment as the test have passed. (I just realized that it might be due to me running the tests on a different version of python.)

sobolevn:
Thanks a lot! I invite you to investigate this field issue further! 👍

LanDinh:
So jumping in on this discussion because this produced a type error on my project \*laugh\*

In my project, I subclass from `AbstractBaseUser` instead of subclassing from `AbstractUser`, because `AbstractUser` defines a bunch of fields that I don't like - and I override the `is_active` attribute as a `@property`.

So at least in my opinion, I think that annotating `AbstractBaseUser.is_active: bool` was the correct thing to do, and that `AbstractUser` is the one who breaks inheritance principles, and should thus receive a `# type: ignore[override]` - so that people who decide to go the `AbstractBaseUser` way can do so without having to add a that type ignore. Should those people end up turning that property into a `BooleanField` instead, that will have been their conscious choice, which means that they should be aware of the type hinting consequences.

BUT before I create a PR for this, I was hoping to get some other opinions on that :)

sobolevn:
@LanDinh your solution implies that all other subclasses of `AbstractBaseUser` would have to use `# type: ignore` when defining `is_active` as a `BooleanField`. Which is not good in my opinion.

For me, this solution seems better.

LanDinh:
Good thing I decided to ask for opinions first. I totally forgot that any mismatch with supertypes will be checked, and not just the most immediate parent...^^ Thanks for reminding me of that!

PR #522, opened 2020-11-06
change get_user to use a protocol requiring a session 

ghost:
# i have made things!

this pr defines a "_HasSession" protocol, and update `contrib.auth.get_user` to use it
`get_user` only requires a session field, and idiomatic django testing frequently calls get_user with a TestClient object

example use case (based off tests i have written and found in the wild):

```py
class ViewTests(TestCase):
    def test_login_view(self):
        data = ...
        self.client.post('/login', data)
        user = auth.get_user(self.client)
        self.assertFalse(user.is_anonymous)
```


django-stubs/contrib/auth/__init__.pyi
```diff
@@ -27,7 +32,7 @@ def login(
 ) -> None: ...
 def logout(request: HttpRequest) -> None: ...
 def get_user_model() -> Type[Model]: ...
-def get_user(request: HttpRequest) -> Union[AbstractBaseUser, AnonymousUser]: ...
+def get_user(request: _HasSession) -> Union[AbstractBaseUser, AnonymousUser]: ...
```
sobolevn: Thanks! 

I am just wondering if `Union[HttpRequest, TestClient]` is better.

ghost: no problem!

i had that thought too. i thought iot might be better to do a protocol in case anyone wanted to implement something that acted like a client/request, but thinking on it now, i don't think that would be that common. using a union would definitely be easier to read

django-stubs/contrib/auth/__init__.pyi
```diff
@@ -27,7 +27,7 @@ def login(
 ) -> None: ...
 def logout(request: HttpRequest) -> None: ...
 def get_user_model() -> Type[Model]: ...
-def get_user(request: HttpRequest) -> Union[AbstractBaseUser, AnonymousUser]: ...
+def get_user(request: Union[HttpRequest, Client]) -> Union[AbstractBaseUser, AnonymousUser]: ...
```
sobolevn, line 29: Let's also add two tests cases for it: with `request` and with `TestClient`!

ghost:
sorry for taking so long, tests are written!

sobolevn:
Awesome! Thanks!

Can you please check why the tests are failing?

ghost:
it's an issue with django 2.2 test suite. i have no idea what the issue is, but it looks like it isn't anything to do with this patch. i can't reproduce it in upstream master, and i can't reproduce it in a clean virtualenv with django 2.2

the issue is in `django-source/tests/serializers/tests.py` line 357:

```py
class Serializer(serializers.json.Serializer):
# django-source/tests/serializers/tests.py:357: error: Name 'serializers.json.Serializer' is not defined
```

but like i said, can't reproduce :/

edit: gimme a bit and i'll bisect

ghost:
huh, bisect found that da0f2c838a14640b72c9933f5e86603fd683e6f0 is the first commit to show the issue, but i have no idea why that would affect that

ghost:
found it. the test client is `django.test.client.Client`, and that was causing the error for some reason

ghost:
lmk if this looks alright and i'll rebase it so that it's not completely incoherent lol

sobolevn (approved):

sobolevn:
Thanks a lot for the help!

PR #523, opened 2020-11-06
Return a `User` model from `get_user_model` (as opposed to `Model`)

yankyhoffman:
The current return type defined for `django.contrib.auth.get_user_model` is `models.Model`.

While it is true that when extending the user model the only real requirement might be that the model storing the user data should be a database model, Django [recommends](https://docs.djangoproject.com/en/3.1/topics/auth/customizing/#specifying-a-custom-user-model) that one should should inherit from `AbstractBaseUser` and extend the custom model further
> The easiest way to construct a compliant custom user model is to inherit from **AbstractBaseUser**.
(which in turn inherits from `models.Model`).

Updating the return type should fix issues like the following screenshot
image
currently using `get_user_model` will display an error when trying to call the [`set_password`](https://github.com/django/django/blob/master/django/contrib/auth/base_user.py#L98:L100)  method set on `AbstractBaseUser`.


django-stubs/contrib/auth/__init__.pyi
```diff
@@ -26,7 +25,7 @@ def login(
     request: HttpRequest, user: Optional[AbstractBaseUser], backend: Optional[Union[Type[ModelBackend], str]] = ...
 ) -> None: ...
 def logout(request: HttpRequest) -> None: ...
-def get_user_model() -> Type[Model]: ...
+def get_user_model() -> Type[AbstractBaseUser]: ...
```
sobolevn, line 28: It would be ideal to add a `mypy` plugin for this function using `get_function_hook`.

@yankyhoffman do you want to give this a try?

sobolevn (commented):
Thanks!

django-stubs/contrib/auth/__init__.pyi
sobolevn: And we also need tests for it! 👍

ghost: @sobolevn looks like that plguin code has been there for a year and a half: https://github.com/typeddjango/django-stubs/blame/a3624dec36814bf8118d2c5f1ee3fca2e4f3b48c/mypy_django_plugin/main.py#L189

ghost: and i can also confirm that it works properly. the following code (in a project with an installed app `my_users` containing the model `MyUser`) typechecks just fine

```py
from typing import Type

from django.contrib.auth import get_user_model
from my_users.models import MyUser

x: Type[MyUser] = get_user_model()
```

ghost: i think a test and a comment on `get_user_model` explaining that the plugin takes care of it would be good though

sobolevn: @3n-k1 thanks! 👍

mkurnikov:
I'm going to close it for now, as it has already been handled by the plugin. @yankyhoffman, if you still interested in this issue, please provide an example where it doesn't work as expected.

PR #524, opened 2020-11-06
Add Point and Polygon to contrib.gis.geos

clarityflowers:
The typings for these already existed, but they weren't included in `__init__.py`. This PR adds them!


sobolevn (approved):
Thanks!

Ref https://github.com/django/django/blob/stable/2.2.x/django/contrib/gis/geos/__init__.py

PR #526, opened 2020-11-08
create monkeypatching function for adding get_item dunder 

### LINKED ISSUE #507: Generic Type Error with ModelAdmin when running in strict mode
# Bug report

## What's wrong

I might be crazy here, but when I subclass from `admin.ModelAdmin` and run mypy with `--strict` enabled I get an error, `error: Missing type parameters for generic type "ModelAdmin"`.  It's completely non-obvious to me how to exactly resolve this. The following code will fail on my machine:

```python
from my_app import models as db

class FooAdmin(admin.ModelAdmin):
    pass

admin.site.register(db.Foo, FooAdmin)
```

This code: 

```python
class FooAdmin(admin.ModelAdmin[db.Foo]):
    pass
```

Generates a type error: `TypeError: 'MediaDefiningClass' object is not subscriptable`

The _only_ way I've been able to resolve this is to do the following:
```python

if TYPE_CHECKING:
    class FooAdmin(admin.ModelAdmin[db.Foo]): ...
else:
    class FooAdmin(admin.ModelAdmin):
```

But that feels like a pretty annoying workaround for all of our admin pages

## How is that should be

I _believe_ I should be able to pass in `db.Foo` when annotating the type for `FooAdmin` like so: `class FooAdmin(admin.ModelAdmin[db.Foo])` although I'm not sure on exactly how to go about this. Any guidance would be much appreciated here

## System information

- OS: macOS
- `python` version: 3.8.6
- `django` version: 2.2.16
- `mypy` version: 0.70

ghost:
# i have made things!

## related issues

closes #507

refs #504, #515 

wasn't too sure about the best way to implement this, so i figured i'd make a draft and ask your thoughts


ghost:
if this is good, i'll add in the rest of the stuiff from #504 and #515

sobolevn (commented):
Awesome! Thank you!

The most interesting question is: how should we distribute this code?
By that I mean:
1. We can ship it with `django-stubs` package, but this way we would have to add `django-stubs` to direct dependencies (not dev-dependencies). And this will also add our dependencies as transitive ones: https://github.com/typeddjango/django-stubs/blob/aab8acf2ea5075c349ec4e05b4ee95c29bfb5e81/setup.py#L23-L27 I guess many people might not want `mypy` to be their direct dependency 🤔 

2. We can create a new package a keep it in sync with `django-stubs`. It won't have any dependencies at runtime, but this will complicate our tooling and setup. I am not sure if Python has a concept of monorepos.

I would love to hear a feedback on this 👍

ghost:
oh look, my favorite part of programming, real world complications!

i think a monorepo is best. if someone makes a class generic, they'd have to submit pull requests to two different repositories, and that just seems way too much for everyone involved. i think a directory structure like this might be best:

```
.
├── dev-requirements.txt
├── django-source
├── django-stubs
├── django_stubs_ext
│   ├── django_stubs_ext
│   │   └── [code goes here]
│   ├── setup.py
│   └── etc., etc.
├── setup.cfg
├── setup.py
└── etc., etc.
```

and then when releases happen, just push both to pypi

sobolevn:
One more thing to consider: #262 

At some point in time we plan to support different versions of django, well, differently.
So, this "patcher" thing should be dependent to stubs in different versions.

The good thing is that we are free to patch more classes to be `Generic` that we actually need.

ghost:
i don't think multiple versions would be too difficult with this. make the `__need_generic` list contain tuples with a version restriction:

```py
__need_generic: List[Tuple[str, Type[Any]]] = [
    ('3', FormMixin),
    ('2', SomeClass),
    ('', VersionIndependentClass),
]

filter(lambda x: str(django.VERSION[0]) == x[0] or x[0] == '', __need_generic)
```

although i'm sure a more sophisticated solution would also work

ghost:
first time writing an actual package, let me know if it looks alright?

i basically copied everything from the root package and deleted what wasn't necessary, then added django_stubs_ext as a requirement to the root package

django_stubs_ext/README.md
```diff
@@ -0,0 +1,43 @@
...
+import django_stubs_ext
+
+django_stubs_ext.monkeypath()
+```
+
+This only needs to be called once, so the call to `monkeypatch` should be placed in your top-level urlconf.
+
+## Version compatibility
+
+Since django-stubs supports multiple Django versions, this package takes care to only monkey-patch the features needed by your django version, and decides which features to patch at runtime.
```
sobolevn: Let's also add a note that 
> Anyway, it is safe, because we only add `__class_getitem__` magic method that does nothing in runtime

django_stubs_ext/dev-requirements.txt
```diff
@@ -0,0 +1,7 @@
+wheel
+mypy==0.790
+pre-commit==2.7.1
+pytest==6.1.1
+-e .
+
+#gitpython==3.1.9
```
sobolevn: Can we remove this line?

sobolevn: 🤔 

Do we even need this file?

django_stubs_ext/django_stubs_ext/monkeypatch.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(self, cls: Type[_T], version: Optional[int] = None) -> None:
+        """Set the data fields, basic constructor."""
+        self.version = version
+        self.cls = cls
+
+
+# certain django classes need to be generic, but lack the __class_getitem__ dunder needed to
+# annotate them: https://github.com/typeddjango/django-stubs/issues/507
+# this list stores them so `monkeypatch` can fix them when called
+__need_generic: List[MPGeneric] = [
```
sobolevn: niptick: let's make it protected, not private.

django_stubs_ext/django_stubs_ext/monkeypatch.py
```diff
@@ -0,0 +1,43 @@
...
+    version-independent.
+
+    This is slightly overkill for our purposes, but useful for future-proofing against any
+    possible issues we may run into with this method.
+    """
+
+    version: Optional[int]
+    cls: Type[_T]
+
+    def __init__(self, cls: Type[_T], version: Optional[int] = None) -> None:
```
sobolevn: Let's follow the same format `sys.version` uses. A tuple of `int, int, int, str?`

django_stubs_ext/requirements.txt
```diff
@@ -0,0 +1,2 @@
+django
```
sobolevn: Do we need this file?

django_stubs_ext/setup.py
```diff
@@ -0,0 +1,43 @@
...
+
+setup(
+    name="django-stubs-ext",
+    version="0.1.0",
+    description="Monkey-patching and extensions for django-stubs",
+    long_description=readme,
+    long_description_content_type="text/markdown",
+    license="MIT",
+    url="https://github.com/typeddjango/django-stubs",
+    author="Maksim Kurnikov",
```
sobolevn: You are the author! 🙂

sobolevn (commented):
Thanks a lot!

django_stubs_ext/dev-requirements.txt
ghost: you're right. monorepo, all the development is gonna be done from within the main repo :p

django_stubs_ext/requirements.txt
ghost: it's technically true from an end-user perspective, the package does rely on django, but if someone installs it, they'll probably have django installed anyway

django_stubs_ext/django_stubs_ext/monkeypatch.py
ghost: we can, but i think that would really complicate the logic with no added benefit. django uses semver, so only the major releases matter. this way we don't have to parse and check a version constraint

django_stubs_ext/django_stubs_ext/monkeypatch.py
ghost: can do!

django_stubs_ext/setup.py
ghost: thanks, i appreciate that! figured i'd leave it since it's your project :p

django_stubs_ext/requirements.txt
ghost: ahhhh missed the point of requirements.txt. nevermind, we can remove it

django_stubs_ext/django_stubs_ext/monkeypatch.py
sobolevn: Ok, fair enough! 👍

django_stubs_ext/setup.py
sobolevn: ok, let's stick to our roots and make @mkurnikov the one in charge 😆

ghost: i'm too tired to read usernames right now haha

.travis.yml
```diff
@@ -54,4 +54,4 @@ before_install: |
   sudo apt-get install -y binutils libproj-dev gdal-bin
   pip install -U pip setuptools wheel
 install: |
-  pip install -r ./dev-requirements.txt
+  pip install -e ./django_stubs_ext -r ./dev-requirements.txt
```
sobolevn: Maybe we should add it to `./dev-requirements`? Things like this are hard to remember during the initial project setup. And this might complicate it for new devs!

ghost:
don't merge just yet, i still need to add the rest of the generic patches :p

dev-requirements.txt
```diff
@@ -9,3 +9,4 @@ pytest==6.1.1
 pytest-mypy-plugins==1.6.1
 psycopg2-binary
 -e .
+-e ./django_stubs_ext
```
sobolevn, line 11: Oh, it is already here. So, we should probably clean up the `.travis.yml`?

django_stubs_ext/LICENSE.txt
```diff
@@ -0,0 +1,8 @@
+Copyright (c) Maxim Kurnikov.
```
sobolevn: I think that one license is enough.

sobolevn: Let's remove this one.

django_stubs_ext/setup.py
mkurnikov: I'd appreciate it if you leave me as the main author in all those places. But if it matters, we could add the `CONTRIBUTORS.md` file with proper acknowledgments, what do you think?

.travis.yml
ghost: for sure. i had it in the wrong place in dev-requirements which was causing issues, so i tried to fix it this way

django_stubs_ext/setup.py
sobolevn: @mkurnikov sure thing! 👍

ghost:
alright, i think i'm all set on my end. i can't think of anything else i would want to add or update. does this look good to everyone else?

ghost:
writing a release script for it

sobolevn (approved):

sobolevn:
Thanks a lot! Let's merge it! 👍

ghost:
ayyyy fantastic! thanks for the help through all of this :)

setup.py
```diff
@@ -24,6 +24,7 @@ def find_stub_files(name: str) -> List[str]:
     "mypy>=0.790",
     "typing-extensions",
     "django",
+    "django-stubs-ext",
```
NyanKiyoshi, line 27: Couldn't this rather be put as an extra dependency? For example, like so:
```py
setup(
    ...,
    # or another name/way of doing
    extras_require={"full": ["django-stubs-ext"]},
)
```

Then users could run `pip install django-stubs[full]` (or the other way around, have full by default and minimal for people that want the minimal, i.e. no monkey patching)

This would prevent adding additional files and dependencies that are not required.

sobolevn: It is a direct dependency, but you can just ignore it completely. It does not get easier that that.

ghost: i'm not sure the savings of a relatively small package are worth the possibility of library consumers accidentally missing the `[full]` part of the install step and wondering what's wrong

i do have a bug report in the mypy repo about this, but it's seen no activity in almost half a year, so i'm not sure we should hold our breath haha

PR #532, opened 2020-11-19
Fix signature of RemoteUserBackend.configure_user

lambdalisue:
# I have made things!

The function signature of `django.contrib.auth.backends.RemoteUserBackend.configure_user` has changed from 2.1 on 2.2 but django-stubs did not follow that.

##### stable/2.1.x
https://github.com/django/django/blob/stable/2.1.x/django/contrib/auth/backends.py#L163

```python
    def configure_user(self, user):
        """
        Configure a user after creation and return the updated user.
        By default, return the user unmodified.
        """
        return user
```

##### stable/2.2.x ~ stable/3.1.x
https://github.com/django/django/blob/stable/2.2.x/django/contrib/auth/backends.py#L177
https://github.com/django/django/blob/stable/3.0.x/django/contrib/auth/backends.py#L236
https://github.com/django/django/blob/stable/3.1.x/django/contrib/auth/backends.py#L222

```python
    def configure_user(self, request, user):
        """
        Configure a user after creation and return the updated user.
        By default, return the user unmodified.
        """
        return user
```

## Related issues

I couldn't found any.


django-stubs/contrib/auth/backends.pyi
```diff
@@ -35,6 +35,6 @@ class AllowAllUsersModelBackend(ModelBackend): ...
 class RemoteUserBackend(ModelBackend):
     create_unknown_user: bool = ...
     def clean_username(self, username: str) -> str: ...
-    def configure_user(self, user: User) -> User: ...
+    def configure_user(self, request: Any, user: User) -> User: ...
```
sobolevn: I guess it should be `request: HttpRequest`

sobolevn (changes requested):
Thanks for the fix!

django-stubs/contrib/auth/backends.pyi
lambdalisue: The function is used in `authenticate()` method and the `request` type, which is directly passed, in that function is `Any`

https://github.com/django/django/blob/stable/2.2.x/django/contrib/auth/backends.py#L150-L160
https://github.com/typeddjango/django-stubs/blob/master/django-stubs/contrib/auth/backends.pyi#L14

Shall I

1. Let it go?
2. Fix type definition of `authenticate()` as well?

I'm not sure if `request` in `authenticate()` is `HttpRequest` or can be others.

sobolevn: Any ideas what else can it be?

lambdalisue: No idea. I just said I'm not sure while I didn't wrote the definition of `authenticate()`.

sobolevn: let's change both then!

lambdalisue: @sobolevn Done

sobolevn (approved):
Thanks a lot!

PR #536, opened 2020-12-01
Fix incorrect and add missing annotations for loaddata command

wkschwartz:
- Added some of the instance attributes from command arguments: `ignore`, `using`, `app_label`, `verbosity`, `format`.
- Added class attribute: `help`.
- Fixed return type of `find_fixtures`.


wkschwartz:
CI appears to have failed booting without actually running the tests. If the maintainers are interested in this patch, would they please re-start CI? I don't think I can do it myself.

sobolevn:
Done!

django-stubs/core/management/commands/loaddata.pyi
```diff
@@ -6,10 +6,16 @@ from django.core.management.base import BaseCommand
 READ_STDIN: str = ...
 
 class Command(BaseCommand):
+    ignore: bool = ...
+    using: str = ...
+    app_label: str = ...
+    verbosity: int = ...
+    format: str = ...
     missing_args_message: str = ...
+    help: str = ...
```
sobolevn: We already have `help` here: https://github.com/typeddjango/django-stubs/blob/master/django-stubs/core/management/base.pyi#L38

sobolevn (changes requested):
Thanks!

Please, double check that all new attrs are not duplicating ones from `BaseCommand` 👍

django-stubs/core/management/commands/loaddata.pyi
wkschwartz: Thanks. I believe d3ff5415db9306b2d5a766b2167a9d2e1a548ce2 resolves this.

wkschwartz:
CI seems ill. `sudo apt-get install binutils libproj-dev gdal-bin` is 404ing. eg: https://github.com/typeddjango/django-stubs/pull/536/checks?check_run_id=1481582406#step:3:63

```
Err:26 http://security.ubuntu.com/ubuntu bionic-updates/main amd64 libpoppler73 amd64 0.62.0-2ubuntu2.10
  404  Not Found [IP: 52.147.219.192 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/p/poppler/libpoppler73_0.62.0-2ubuntu2.10_amd64.deb  404  Not Found [IP: 52.147.219.192 80]
```

All tests and lints pass locally for me with Python 3.7 set up per https://github.com/typeddjango/django-stubs/blob/master/CONTRIBUTING.md

PR #541, opened 2020-12-15
Stricter return type annotations for template.Library

### LINKED ISSUE #483: Function is untyped after decorator transformation
# Bug report

## What's wrong

I have an inclusion tag defined using a decorator:

```python
@register.inclusion_tag("admin/widgets/related_widget_wrapper.html")
def agenda_cell(
    room: Room, hour: datetime, agenda_item: Optional[AgendaItem] = None
) -> AgendaCellContext:
    ...
```

Mypy gives me here this error:

```
app/chronology/templatetags/agenda.py:34: error: Function is untyped after
decorator transformation  [misc]
    def agenda_cell(
    ^
```

suggesting that `inclusion_tag` returns an untyped function.

## How is that should be

`inclusion_tag` and other templatetag decorators should return the function of the same types.
e.g. using typevars

## System information

- OS: ArchLinux
- `python` version: 3.8.6
- `django` version: 3.1.2
- `mypy` version: 0.782

timmartin:
This adds more constrained return types to the template.Library methods. These are overloaded so that they can be used as function decorators with and without arguments.

## Related issues

Closes #483


timmartin:
I've written stubs that I think are correct, but it's pretty hard to figure out all the overloads. What's the best way to test this?

sobolevn (commented):
Awesome! Can you please write a couple of tests for it?

sobolevn (approved):
Awesome! 👍

PR #542, opened 2020-12-16
convince mypy that user.is_staff (and friends) are booleans

### LINKED ISSUE #512: user.is_staff complex type

ghost:
# i have made things!

## related issues

closes #512

this should be enough to convince mypy that the `is_staff`, `is_active`, and `is_superuser` are (for all intents and purposes) bools

alternatively we could set things up so that `BooleanField`s are *always* seen by mypy as a bool, but i think that a conservative approach is best

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -22,7 +22,7 @@ from mypy_django_plugin.django.context import DjangoContext
 from mypy_django_plugin.lib import fullnames, helpers
 from mypy_django_plugin.transformers import fields, forms, init_create, meta, querysets, request, settings
 from mypy_django_plugin.transformers.managers import create_new_manager_class_from_from_queryset_method
-from mypy_django_plugin.transformers.models import process_model_class
+from mypy_django_plugin.transformers.models import process_model_class, set_auth_user_model_boolean_fields
 
 
 def transform_model_class(ctx: ClassDefContext, django_context: DjangoContext) -> None:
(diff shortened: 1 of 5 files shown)


mypy_django_plugin/main.py
```diff
@@ -270,8 +270,12 @@ def get_attribute_hook(self, fullname: str) -> Optional[Callable[[AttributeConte
             return partial(settings.get_type_of_settings_attribute, django_context=self.django_context)
 
         info = self._get_typeinfo_or_none(class_name)
+        if info and info.has_base(fullnames.PERMISSION_MIXIN_CLASS_FULLNAME) and attr_name == "is_superuser":
+            return partial(set_auth_user_model_boolean_fields, django_context=self.django_context)
         if info and info.has_base(fullnames.HTTPREQUEST_CLASS_FULLNAME) and attr_name == "user":
             return partial(request.set_auth_user_model_as_type_for_request_user, django_context=self.django_context)
+        if info and info.has_base(fullnames.ABSTRACT_USER_MODEL_FULLNAME) and attr_name in ("is_staff", "is_active"):
```
sobolevn, line 277: I am a big fan of class-level immutable constants for cases like this. Do you think that this is worth refactoring?

sobolevn (commented):
Awesome! Thanks!

mypy_django_plugin/main.py
sobolevn: `frozenset` basically 🙂

ghost: do you mean moving `ABSTRACT_USER_MODEL_FULLNAME` to the `NewSemanalDjangoPlugin` class?

sobolevn: No, I was thinking about `('is_staff', 'is_active')`

ghost: ahhh
yeah, tbh i was trying to think of a way to turn it into a dict, so you could just pass in the fullname and attr_name and get the correct partial as a result, but i'm not sure that's doable in a clean way :/

i don't think that one bit is worth refactoring, but i absolutely want to refactor those three if statements

sobolevn: Ok 👍

sobolevn (approved):

PR #543, opened 2020-12-16
Django admin additions

intgr:
# I have made things!

Added more precise typing for Django admin symbols.


django-stubs/contrib/admin/sites.pyi
```diff
@@ -39,28 +54,28 @@ class AdminSite:
...
     @empty_value_display.setter
-    def empty_value_display(self, empty_value_display: Any) -> None: ...
+    def empty_value_display(self, empty_value_display: str) -> None: ...
     def has_permission(self, request: WSGIRequest) -> bool: ...
     def admin_view(self, view: Callable, cacheable: bool = ...) -> Callable: ...
     def get_urls(self) -> List[URLResolver]: ...
     @property
     def urls(self) -> Tuple[List[URLResolver], str, str]: ...
-    def each_context(self, request: Any): ...
+    def each_context(self, request: WSGIRequest) -> Dict[str, Any]: ...
```
sobolevn, line 71: Are you sure it is not `HTTPRequest`?

sobolevn (commented):
Thanks! Only `WSGIRequest` catches my eye.

django-stubs/contrib/admin/sites.pyi
intgr: Other request arguments in this file are typed as `WSGIRequest`, so I picked that for consistency. I don't know when one should be preferred over the other.

Do you think I can change all of them to `HttpRequest`?

sobolevn: I am really not sure. Let's keep it this way for now. Thanks!

sobolevn (approved):

intgr:
Thanks :)

PR #547, opened 2020-12-18
Add Field to django_stubs_ext.django_stubs_ext.patch._need_generic

### LINKED ISSUE #546: Field missing from django_stubs_ext.patch._need_generic
# Bug report

A custom `Field` subclass does not type-check properly, even when using django_stubs_ext.

## What's wrong

The `Field` class is not included in `django_stubs_ext.patch._need_generic`.

## How is that should be

If the `Field` class were included in `django_stubs_ext.patch._need_generic`, the following would type check properly instead of causing a crash:
```
class Foo(Field[int, int]):
    pass

class Bar(TextField[str, str]):
    pass
```

## System information

- OS: Debian
- `python` version: 3.8.6
- `django` version: 3.1.4
- `mypy` version: 0.790
- `django-stubs-ext` version: 0.1.0

markedwards:
## Related issues

- Closes #546

### DIFF
--- a/django_stubs_ext/django_stubs_ext/patch.py
+++ b/django_stubs_ext/django_stubs_ext/patch.py
@@ -3,6 +3,7 @@ from typing import Any, Generic, List, Optional, Tuple, Type, TypeVar
 from django import VERSION as VERSION
 from django.contrib.admin import ModelAdmin
 from django.contrib.admin.options import BaseModelAdmin
+from django.db.models.fields import Field
 from django.db.models.manager import BaseManager
 from django.db.models.query import QuerySet
 from django.views.generic.edit import FormMixin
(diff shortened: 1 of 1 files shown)


django_stubs_ext/django_stubs_ext/patch.py
```diff
@@ -38,6 +39,7 @@ def __repr__(self) -> str:
     MPGeneric(ModelAdmin),
     MPGeneric(FormMixin),
     MPGeneric(BaseModelAdmin),
+    MPGeneric(Field),
```
sobolevn, line 42: Maybe we can also provide an `extra=` argument to `.monkeypatch(extra=)` call?

So user will have a chance to also patch custom classes in case this is required.

sobolevn (commented):
Thanks! What do you think about my suggestion?

django_stubs_ext/django_stubs_ext/patch.py
markedwards: I wouldn't object, but I'm not sure I understand the use case.  I would say it should be a goal of django-stubs to patch everything covered by its typing implementation, and I don't see a use case for patching objects outside of the scope of django-stubs.

Do you have a specific use case in mind?

sobolevn: No 🙂

sobolevn (approved):

sobolevn:
Thanks!

PR #548, opened 2020-12-18
Update django_stubs_ext documentation to reference settings

### LINKED ISSUE #544: TypeError: object is not subscriptable, with django_stubs_ext
# Bug report

I'm attempting to implement django_stubs_ext, but I still get a `TypeError` when adding a type parameter to a class that ought to be supported.

## What's wrong

I have installed django_stubs_ext, and added `django_stubs_ext.monkeypatch()` to manage.py, asgi.py, and wsgi.py.  However, I still get `TypeError: ... object is not subscriptable` when trying to add type parameters.

Example:

```
class Test(ModelAdmin[Project]):
    pass
```

This results in: `TypeError: 'MediaDefiningClass' object is not subscriptable`.

The error only manifests when I execute `mypy` from the command line.  My Django application runs fine, and I see that I can do `ModelAdmin[Project]` from inside a django shell.  So it just appears that the monkeypatch doesn't get applied when mypy runs from the command line.

## How is that should be

Maybe I misunderstand something, but shouldn't the above work?

## System information

- OS: Debian
- `python` version: 3.8.6
- `django` version: 3.1.4
- `mypy` version: 0.790
- `django-stubs-ext` version: 0.1.0

markedwards:
This change fixes two issues in the django_stubs_ext documentation:

1) The call to `django_stubs_ext.monkeypatch()` works best in settings
2) The example code had a typo ("monkeypath" -> "monkeypatch")

## Related issues

- Closes #544

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,12 +90,12 @@ Install it:
 pip install django-stubs-ext  # as a production dependency
 ```
 
-And then place in your `manage.py`, `wsgi.py`, and `asgi.py` files:
+And then place in your top-level settings:
 
 ```python
 import django_stubs_ext
 
-django_stubs_ext.monkeypath()
+django_stubs_ext.monkeypatch()
 ```
 
 2. You can use strings instead: `'QuerySet[MyModel]'` and `'Manager[MyModel]'`, this way it will work as a type for `mypy` and as a regular `str` in runtime.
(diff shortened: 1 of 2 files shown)


sobolevn (approved):

sobolevn:
Thank you!

PR #558, opened 2021-01-19
Missing Django 3.1 items

mkurnikov:
Closes https://github.com/typeddjango/django-stubs/pull/556

PR #559, opened 2021-01-19
add value_to_string method to Field

mkurnikov:
Refs https://github.com/typeddjango/django-stubs/pull/558#discussion_r560497169


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -142,6 +142,7 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
     def cached_col(self) -> Col: ...
     def value_from_object(self, obj: Model) -> _GT: ...
     def get_attname(self) -> str: ...
+    def value_to_string(self, obj: Any) -> str: ...
```
sobolevn: Is it `Any` or `Model`? Like it was here: https://github.com/typeddjango/django-stubs/pull/559/files#diff-1f20b551c410ef7758e9395bf08460ce59b9d22096dc6c973e6f461816d146dbL86

PR #560, opened 2021-01-31
Allow ModelAdmin.actions to be set to None.

charettes:
`None` is an allowed value to disable all actions for a particular `ModelAdmin`.

https://docs.djangoproject.com/en/3.1/ref/contrib/admin/actions/#disabling-all-actions-for-a-particular-modeladmin


sobolevn (approved):
Thanks!

PR #564, opened 2021-02-08
fix: add headers to http.HttpResponseBase constructor

### LINKED ISSUE #563: "headers" missing from http.HttpResponseBase stub?
# Bug report

## What's wrong

https://github.com/typeddjango/django-stubs/blob/8f9e77ce39cb581ca220fb6f91197885e462a7c4/django-stubs/http/response.pyi#L21 defines the types for the http.HttpResponseBase class; however, it seems to be missing the 'headers' parameters, as defined in: https://github.com/django/django/blob/50a5f8840fa564dcefdb1fa5c58f06fcd472ee70/django/http/response.py#L99

This is causing me problems in typing my own project but I'm new to mypy so happy to have any misconceptions on my side corrected.

## How is that should be

I believe headers should be `Optional[Dict(str, str)]`?

## System information

- OS: Arch Linux
- `python` version: 3.8
- `django` version: 2.2.13
- `mypy` version: 0.780

ChasNelson1990:
Closes #563


sobolevn (changes requested):
Thanks! 👍 

Let's also type this attribute: https://github.com/django/django/blob/master/django/http/response.py#L100

ChasNelson1990:
@sobolevn - am I right in saying I will need to type the whole ResponseHeaders class for this?

django-stubs/http/response.pyi
```diff
@@ -11,19 +11,30 @@ from django.template import Context, Template
...
+    def pop(self, key: str, default: Optional[str]=...) -> str: ...
+    def setdefault(self, key: str, value: str) -> None: ...
 
 class HttpResponseBase(Iterable[Any]):
     status_code: int = ...
     cookies: SimpleCookie = ...
     reason_phrase: str = ...
     charset: str = ...
     closed: bool = ...
+    headers: Union[ResponseHeaders, Dict] = ...
```
sobolevn: `.headers` are always `ResponseHeaders`

sobolevn (commented):
Thanks! 👍

ChasNelson1990:
Hmmm... pre-commit works fine locally (so why the GH action lint fails is confusing) but pytest fails with:
```
____________________________________ ERROR collecting django_stubs_ext/tests/test_monkeypatching.py _____________________________________
ImportError while importing test module '/home/chas/gliff/django-stubs/django_stubs_ext/tests/test_monkeypatching.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.9/importlib/__init__.py:127: in import_module
django_stubs_ext/tests/test_monkeypatching.py:10: in <module>
    from django_stubs_ext import patch
E   ImportError: cannot import name 'patch' from 'django_stubs_ext' (unknown location)
```

ChasNelson1990:
Do I need to install the stubs modules with `pip install -e .`?

ChasNelson1990:
> Hmmm... pre-commit works fine locally (so why the GH action lint fails is confusing) but pytest fails with:

D'oh - local was on master

django-stubs/http/response.pyi
```diff
@@ -6,24 +6,35 @@ from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple, Type, U
...
 from django.test.client import Client
-
+from django.utils.datastructures import CaseInsensitiveMapping
 from django.template import Context, Template
 from django.urls import ResolverMatch
 
 class BadHeaderError(ValueError): ...
 
+class ResponseHeaders(CaseInsensitiveMapping):
+    store: Dict = ...
```
sobolevn: `Dict[str, str]` isn't it?

django-stubs/http/response.pyi
```diff
@@ -6,24 +6,35 @@ from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple, Type, U
...
+    def __init__(self, data: Dict) -> None: ...
+    def _convert_to_charset(self, value: Union[bytes, str], charset: str, mime_encode: str=...) -> str: ...
+    def __delitem__(self, key: str) -> None: ...
+    def __setitem__(self, key: str, value: str) -> None: ...
+    def pop(self, key: str, default: Optional[str]=...) -> str: ...
```
sobolevn: `default: Optional[str] = ...`

django-stubs/http/response.pyi
```diff
@@ -6,24 +6,35 @@ from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple, Type, U
...
-
+from django.utils.datastructures import CaseInsensitiveMapping
 from django.template import Context, Template
 from django.urls import ResolverMatch
 
 class BadHeaderError(ValueError): ...
 
+class ResponseHeaders(CaseInsensitiveMapping):
+    store: Dict = ...
+    def __init__(self, data: Dict) -> None: ...
```
sobolevn: The same here 🤔

sobolevn (commented):
Thanks again! I guess that this is the last iteration 👍

sobolevn (approved):
Great work! 🚀

PR #566, opened 2021-02-11
Update static.pyi

tkrapp:
# Correct function signature of serve function
# document_root and show_indexes are now marked as Optional since they have a default value.

## Related issues
None

### DIFF
--- a/django-stubs/views/static.pyi
+++ b/django-stubs/views/static.pyi
@@ -3,7 +3,7 @@ from typing import Any, Optional
 from django.http.request import HttpRequest
 from django.http.response import FileResponse
 
-def serve(request: HttpRequest, path: str, document_root: str = ..., show_indexes: bool = ...) -> FileResponse: ...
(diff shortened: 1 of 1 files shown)


django-stubs/views/static.pyi
```diff
@@ -3,7 +3,7 @@ from typing import Any, Optional
 from django.http.request import HttpRequest
 from django.http.response import FileResponse
 
-def serve(request: HttpRequest, path: str, document_root: str = ..., show_indexes: bool = ...) -> FileResponse: ...
+def serve(request: HttpRequest, path: str, document_root: Optional[str] = ..., show_indexes: Optional[bool] = ...) -> FileResponse: ...
```
sobolevn: Thanks! But, looks like `show_indexes` is `bool`: https://github.com/django/django/blob/master/django/views/static.py#L19

sobolevn (changes requested):

django-stubs/views/static.pyi
tkrapp: Sorry, you're right. I'm fairly new to type hints.

sobolevn (approved):
Thanks!

PR #568, opened 2021-02-15
Correct order for FileField __init__ args

### LINKED ISSUE #567: db.models.fields.files.FileField __init__ args incorrect order
## What's wrong

The order for args in `db.models.fields.files.FileField.__init__` is incorrect.

In Django it is:
```python
    def __init__(self, verbose_name=None, name=None, upload_to='', storage=None, **kwargs):
```
see ( https://github.com/django/django/blob/3fa1ed53be370f4b1a94d78b56ff30d23b131623/django/db/models/fields/files.py#L227 ) 

but currently in django-stubs we have:

```python
   def __init__(
        self,
        upload_to: Union[str, Callable, Path] = ...,
        storage: Optional[Union[Storage, Callable[[], Storage]]] = ...,
        verbose_name: Optional[Union[str, bytes]] = ...,
        name: Optional[str] = ...,
        max_length: Optional[int] = ...,
        unique: bool = ...,
        blank: bool = ...,
        null: bool = ...,
        db_index: bool = ...,
        default: Any = ...,
# etc.
```
( see https://github.com/typeddjango/django-stubs/blob/993dd816a4ddf0048f69c9a114de6183d23ffd31/django-stubs/db/models/fields/files.pyi#L39 )

so you cannot use `verbose_name` and `name` as positional arguments.

e.g.:
```python
class Thing(models.Model):
    my_upload = models.FileField("My great upload", upload_to="/uploads")
```
results in:
```
error: "FileField" gets multiple values for keyword argument "upload_to"  [misc]
```

## How is that should be

I think it should instead be:
```python
    def __init__(
        self,
        verbose_name: Optional[Union[str, bytes]] = ...,
        name: Optional[str] = ...,
        upload_to: Union[str, Callable, Path] = ...,
        storage: Optional[Union[Storage, Callable[[], Storage]]] = ...,
        max_length: Optional[int] = ...,
        unique: bool = ...,
# etc...
```

## System information

- OS:
- `python` version: 3.6
- `django` version: 2.2.15 -> present
- `mypy` version: mypy 0.800
- `django-stubs` version: latest

danthedeckie:
## Related issues

Closes: https://github.com/typeddjango/django-stubs/issues/567

### DIFF
--- a/django-stubs/db/models/fields/files.pyi
+++ b/django-stubs/db/models/fields/files.pyi
@@ -38,10 +38,10 @@ class FileField(Field):
     upload_to: Union[str, Callable] = ...
     def __init__(
         self,
-        upload_to: Union[str, Callable, Path] = ...,
-        storage: Optional[Union[Storage, Callable[[], Storage]]] = ...,
         verbose_name: Optional[Union[str, bytes]] = ...,
         name: Optional[str] = ...,
+        upload_to: Union[str, Callable, Path] = ...,
+        storage: Optional[Union[Storage, Callable[[], Storage]]] = ...,
         max_length: Optional[int] = ...,
         unique: bool = ...,
         blank: bool = ...,


sobolevn (approved):
Thanks!

PR #569, opened 2021-02-18
Mention issue about as_manager along with a workaround

syastrov:
# I have made things!

Document common issue and workaround.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,9 +130,10 @@ Example:
 
 ```python
 from django.db import models
+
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks!

sobolevn (approved):

kalekseev:
Great! So the trick was to assign `as_manager()` to intermediate variable, I was testing with `objects = QuerySet.as_manager()()` and that form doesn't work

PR #570, opened 2021-02-21
feat(add-get-lines): Add get_inlines method

melvyn-apryl:
Document ModelAdmin.get_inlines(). Available since [Django 3.0](https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_inlines):

>The get_inlines method is given the HttpRequest and the obj being edited (or None on an add form) and is expected to return an iterable of inlines. You can override this method to dynamically add inlines based on the request or model instance instead of specifying them in ModelAdmin.inlines.


sobolevn (approved):

sobolevn:
Thanks! 👍

PR #571, opened 2021-03-04
Tweak the constructor signature for ManyToManyField

leamingrad:
Prior to this change, ManyToManyField was declared as being generic in
`_ST` and `_GT`, but also used the `_T` Typevar in its `__init__` signature.
This caused mypy to add `_T` to the variables it was generic in when used
as an alias.

The symptom of this problem was that mypy would show an error with the
message "Type application has too few types (3 expected)" where a
ManyToManyField alias was declared, but adding an extra argument would
fail because the type only takes two arguments.

This change brings the signature of ManyToManyField in line with
ForeignKey and OneToOneField.

## Related issues

None - I can raise one if required.

### DIFF
--- a/django-stubs/db/models/fields/related.pyi
+++ b/django-stubs/db/models/fields/related.pyi
@@ -189,7 +189,7 @@ class ManyToManyField(RelatedField[_ST, _GT]):
     swappable: bool = ...
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks!

PR #573, opened 2021-03-06
Fix Field arguments variance

### LINKED ISSUE #511: Proposal: Allow narrowing ORM fields

antonagestam:
## Related issues

- Closes #511
- Refs #513

### DIFF
--- a/tests/typecheck/fields/test_base.yml
+++ b/tests/typecheck/fields/test_base.yml
@@ -163,3 +163,5 @@
         book.published = 2006  # E: Incompatible types in assignment (expression has type "int", variable has type "Year")
         book.published = Year(2006)
         reveal_type(book.published)  # N: Revealed type is 'main.Year*'
+        def accepts_int(arg: int) -> None: ...
+        accepts_int(book.published)
--- a/tests/typecheck/fields/test_base.yml
+++ b/tests/typecheck/fields/test_base.yml
@@ -151,3 +151,15 @@
 
                 class MyModel(AuthMixin, models.Model):
                     pass
+-   case: can_narrow_field_type
+    main: |
+        from typing import cast, NewType
+        from django.db import models
+        Year = NewType("Year", int)
+        class Book(models.Model):
+            published = cast(models.Field[Year, Year], models.IntegerField())
+        book = Book()
+        reveal_type(book.published)  # N: Revealed type is 'main.Year*'
+        book.published = 2006  # E: Incompatible types in assignment (expression has type "int", variable has type "Year")
+        book.published = Year(2006)
+        reveal_type(book.published)  # N: Revealed type is 'main.Year*'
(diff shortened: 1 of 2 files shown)


tests/typecheck/fields/test_base.yml
```diff
@@ -150,4 +150,16 @@
...
+-   case: can_narrow_field_type
+    main: |
+        from typing import cast, NewType
+        from django.db import models
+        Year = NewType("Year", int)
+        class Book(models.Model):
+            published = cast(models.Field[Year, Year], models.IntegerField())
+        book = Book()
+        reveal_type(book.published)  # N: Revealed type is 'main.Year*'
+        book.published = 2006  # E: Incompatible types in assignment (expression has type "int", variable has type "Year")
```
sobolevn, line 163: Let's also test that it can be used as an `int`, just in case.

tests/typecheck/fields/test_base.yml
```diff
@@ -150,4 +150,16 @@
...
-                    pass
\ No newline at end of file
+                    pass
+-   case: can_narrow_field_type
+    main: |
+        from typing import cast, NewType
+        from django.db import models
+        Year = NewType("Year", int)
+        class Book(models.Model):
+            published = cast(models.Field[Year, Year], models.IntegerField())
```
sobolevn, line 160: What will happen without `cast`?

Let's also test it as well.

sobolevn (commented):
Thanks a lot!

tests/typecheck/fields/test_base.yml
sobolevn: I mean that `book.published` can be passed to, for example, `def accepts_int(arg: int): ...`

sobolevn (approved):
Thanks a lot! I am going to merge this as-is. Plus we can continue with modifing our plugin later on.

Great work! 👍

PR #576, opened 2021-03-19
Update decorators.pyi to match Django 3.1

tkrapp:
# Make decorators.py compatible with Django 3.1
This commit resembles the state of the file in Django 3.1. It is not 100% compatible with Django 3.0 which misses the sync* and async* decorators.

## Related issues

None


django-stubs/utils/decorators.pyi
```diff
@@ -12,5 +12,7 @@ class classonlymethod(classmethod): ...
 def method_decorator(decorator: Union[Callable, Iterable[Callable]], name: str = ...) -> Callable[[_T], _T]: ...
 def decorator_from_middleware_with_args(middleware_class: type) -> Callable: ...
 def decorator_from_middleware(middleware_class: type) -> Callable: ...
-def available_attrs(fn: Callable): ...
 def make_middleware_decorator(middleware_class: Type[MiddlewareMixin]) -> Callable: ...
+def sync_and_async_middleware(func: Callable) -> Callable: ...
```
sobolevn: Can we type `Callable` better?

1. Does it change the `Callable` signature? If no, use `_CallableType = TypeVar('_CallableType', bound=Callable)` and `def sync_only_middleware(func: _CallableType) -> _CallableType: ...`
2. If it does change the signature, we can use our `plugin` to fix this

sobolevn: It does not change the type at all: https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/utils/decorators.py#L155-L162

So, we are going to use `_CallableType = TypeVar('_CallableType', bound=Callable)`

sobolevn (approved):

PR #577, opened 2021-03-19
MIMEBase parameter for EmailMessage.attach()

egrubbs:
# Fixed EmailMessage.attach()

Ever since Django 1.2, when the django/core/mail/message.py file was added, the first parameter EmailMessage.attach() could be a MIMEBase instance, and not just a MIMEText instance.

https://github.com/django/django/blob/stable/1.2.x/django/core/mail/message.py

## Related issues

I couldn't find any open issues for this.

## Validation

You can verify that mypy no longer reports an error for the MIMEImage (subclass of MIMEBase) parameter with the text snippet.
I didn't see any tests for the EmailMessage methods, but I could add one if you want.

```
from email.mime.text import MIMEText
from email.mime.image import MIMEImage
from django.core.mail.message import EmailMessage


message = EmailMessage()
message.attach("myfilename", "mycontent", "text/plain")
message.attach(MIMEText("mytext"))
message.attach(MIMEImage("mydata", "image/png"))
```


sobolevn (commented):
Thanks a lot!

> I didn't see any tests for the EmailMessage methods, but I could add one if you want.

Yes, please!

egrubbs:
@sobolevn, I added a test. It's very rudimentary. I wasn't sure how to check that it gets an error if an invalid type is used as a parameter. I didn't see any examples of something like pytest.raises() inside the yaml tests.

tests/typecheck/test_mail.yml
```diff
@@ -0,0 +1,19 @@
...
+        message.attach("myfilename", "mycontent", "text/plain")
+
+        mime_text = MIMEText("mytext")
+        reveal_type(mime_text)  # N: Revealed type is 'email.mime.text.MIMEText'
+        message.attach(mime_text)
+
+        mime_image = MIMEImage("mydata", "image/png")
+        reveal_type(mime_image)  # N: Revealed type is 'email.mime.image.MIMEImage'
+        message.attach(mime_image)
+        reveal_type(message.attachments)  # N: Revealed type is 'builtins.list[Any]'
```
sobolevn, line 19: Here's an example of how we can track errors: https://github.com/dry-python/returns/blob/master/typesafety/test_curry/test_partial/test_partial_generic.yml#L1-L21

sobolevn (approved):
Thanks!

PR #580, opened 2021-04-02
Create related managers from generated managers

kalekseev:
Right now related managers don't detect generated managers, this pr fixes that by generating related manager for each related field. There's a problem left, types may conflict because of the same name, eg:

```
a.py:
AddressManager = Manager.from_queryset(AddressQuerySet)
class Address:
     ...

b.py:

class Profile:
     ...

c.py:
UserManager = Manager.from_queryset(UserQuerySet)
class User:
    address = ForeignKey(Address)
    profile = ForeignKey(Profile)
    objects = UserManager()

d.py:
from a import * # that will import generated User_RelatedManager
from b import * # that will import generated User_RelatedManager too
# as a result mypy will complain that two entities with the same name have different types
```

Possible solutions:
1) append model name to generated name, for example above names would be: `Address_User_RelatedManager` and `Profile_User_RelatedManager`
2) generate related manager at the same time when the manager generated, place it at the same module and reuse with all relations

The second approach is probably better because it returns the same type for all related managers, as a result code like below won't complain that two different types assigned to `qs` variable:

```
if cond:
   qs = address.users  
else:
   qs = profile.users
```

Thoughts? @sobolevn

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -275,6 +275,8 @@ class AddRelatedManagers(ModelClassInitializer):
                 # create new RelatedManager subclass
                 parametrized_related_manager_type = Instance(related_manager_info, [Instance(related_model_info, [])])
                 default_manager_type = related_model_info.names["objects"].type
+                if default_manager_type is None:
+                    default_manager_type = self.try_generate_related_manager(related_model_cls, related_model_info)
                 if (
                     default_manager_type is None
                     or not isinstance(default_manager_type, Instance)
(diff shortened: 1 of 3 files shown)


sobolevn:
Hi! Thanks a lot!

I have requested a review from @mkurnikov, because he is the codeowner of this feature.
I know that there were some problems with it.

Considering `from some import *` - I don't think that this is a big problem. `*` should not really be widely used and that is just one more reason why.

kalekseev:
> Considering from some import * - I don't think that this is a big problem. * should not really be widely used and that is just one more reason why.

I think the UX would be much better if we won't mess with user's codebase, imagine someone runs into that error message about types that he can't even find in his code.
Adding model name to the type shouldn't be the problem for django-stubs, it's one line change + rename in tests.

mkurnikov:
Generating related manager in the module of the model that has it as an attribute was just a quick fix to the problem. Both proposed solutions are valid. Second one is better though, I'm just a bit worried that you'll get into issues with import cycles. Anyway, the first one works too. 

Upd.
Is there any way maybe to make those generated related managers not participate in the import-star at all? Is there's not, maybe there's should be? 

I'm going to review the actual code later this week, @sobolevn, if you think the code is fine, you can go ahead and merge it. 

Upd 2. 
Just reviewed it a bit from the phone, @kalekseev, you should add your "double import star" test too, in case there'll be different behaviour later.

kalekseev:
Finished pr, used first option to prevent import conflicts (added model class name to generated related manager name), added test for star import.

sobolevn (approved):
Thanks a lot!

PR #581, opened 2021-04-02
compatibility matrix doc change for new django-stubs version

### LINKED ISSUE #575: Docs: about compatibility with mypy version
# Bug report

## What's wrong

the documentation is mentioning that django-stubs is compatible only with mypy version 0.790 , now we have new version 0.812 in mypy, updating versions does not show any error but wondering why readme file is not mentioning the support of higher mypy versions

## How is that should be

django-stubs | mypy version | django version | python version
-- | -- | -- | --
1.7.0 | >0.790 | 2.2.x \|\| 3.x | ^3.6

## System information

- OS:
- `python` version:3.8
- `django` version:3.1.7
- `mypy` version:0.812
- `django-stubs` version:1.7.0

highpost:
- Closes #575

I saw this and thought it would be easy to close. Please note that I set the release number to 1.8 because in your comment you said that you would be making a new release soon. I can easily change it back to 1.7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ We rely on different `django` and `mypy` versions:
 
 | django-stubs | mypy version | django version | python version
 | ------------ | ---- | ---- | ---- |
+| 1.8.0 | 0.812 | 2.2.x \|\| 3.x | ^3.6
 | 1.7.0 | 0.790 | 2.2.x \|\| 3.x | ^3.6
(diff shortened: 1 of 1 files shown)


sobolevn (approved):

sobolevn:
Sorry, it took me longer to release `1.8.0` than I thought: https://pypi.org/project/django-stubs/1.8.0/

PR #583, opened 2021-04-05
Add URLPattern options to django.url.conf since they were missing.

### LINKED ISSUE #582: Type hints for django.urls.conf.* are incorrect
# Bug report

## What's wrong

It all starts with the documentation of including a list of `URLPattern` into `django.urls.conf.include`, as documented in https://docs.djangoproject.com/en/3.1/topics/http/urls/#including-other-urlconfs - look at the example for `extra_patterns`.

When then looking through the Django source itself for the implementation of `django.urls.conf.include`, the code path leads to an output of `Tuple[List[URLPattern], Optional[str], Optional[str]]: ...`, as opposed to only the current type hint of `Tuple[List[URLResolver], Optional[str], Optional[str]]: ...`.

When then thinking about the use case of inputting that output into `path` or `re_path`, it's notable that this isn't supported; only the `URLResolver` case is supported as output of `include` - here's the currently accepted case for `path` as an example:
```
@overload
def path(
    route: str, view: List[Union[URLResolver, str]], kwargs: Dict[str, Any] = ..., name: str = ...
) -> URLResolver: ...
```

You can even fall further down the rabbit hole by looking at the implementation of `django.urls._path`, which ends up calling the constructor for `URLResolver` in the case of included `List[URLPattern]` - and the type hints for that constructor are also missing for this case - they are even missing for the `List[URLResolver]` case!

How did I find this? I'm augmenting Django in my repo to make it easier to use for me, since I know that "I'll always provide this information, so let me write a helper function so that I don't have to repeat myself all the time". Which led me to add my own implementation of `include` which I use instead of Django's `django.url.include`, an abbreviated version looks a bit like this:
```
def include(module_name: str, namespace: str) -> Tuple[List[URLPattern], str, str] :
  # This is how Django imports the urlpatterns from a different module.
  # I thus expect there to be the property urlpatterns and silence any type checker complaints with # type: ignore comments.
  # If I mess this up, the server won't start up anyways, which will be noticable right away.
  module = import_module(module_name) 

  # In my own urlpatterns list, I provide 3-tuples - the route to be used, a class-based view, as well as an additional string argument.
  # path here is the regular django.urls.path, in the instance which creates URLPatterns.
  patterns = [
    path(route, view.as_view(custom_view_arg = arg), name = arg)
    for route, view, arg in module.urlpatterns
  ]

  return patterns, namespace, namespace
```

If I then try to use this snippet like so:
```
urlpatterns = [
    path('test/', include('test.app.urls', namespace = 'test')),
]
```
I get type warnings, since there is no overridden version of `path` which accepts `List[URLPattern]`. This code passes through just fine, server is up and well, and serving my endpoints - which is why I ended up literally stepping through the Django implementation to make sure that I missed nothing, and compared my findings with the provided type hints in django-stub.

## How is that should be

Proposed fix for `django-stubs.urls.conf.pyi`:

The hints for `include` should be adapted like so:
```
def include(arg: Any, namespace: Optional[str] = ...) -> Tuple[List[Union[URLResolver, URLPattern]], Optional[str], Optional[str]]: ...
```
Note that `List[URLResolver]` -> `List[Union[URLResolver, URLPattern]]`.

The hints for `path` should be adapted like so:
```
@overload
def path(
    route: str, view: List[Union[URLResolver, URLPattern, str]], kwargs: Dict[str, Any] = ..., name: str = ...
) -> URLResolver: ...
```
Note that `Union[URLResolver, str]` -> `Union[URLResolver, URLPattern, str]`

I stopped myself from getting lost in the rabbit hole when I saw that `URLResolver` doesn't accept any kind of list as `urlconf_name`, even though these code paths get hit even with vanilla Django, so I didn't think too much about how to fix that - especially since I'm not affected by it.

The only reason why I didn't directly create a PR for this is that even though I've looked at the Django source and the django-stubs type hints extensively to customize Django to fit my needs, I've never really looked at how testing is done in either of these projects, and I wasn't really in the mood to figure that out at 1am in the morning :D If anyone wants to give me some pointers on that, I can create a PR to adjust the type hints as I mentioned here.

Sorry for the lengthy post!

## System information

- `python` version: 3.9.2
- `django` version: 3.1.2
- `mypy` version: 0.812

LanDinh:
# I have made things!

It's possible to pass in `List[URLPattern]` as argument to `django.urls.include`, which then results in `List[URLPattern]` as output of that function. The current type hints for all of `django.urls.include`, `django.urls.path` and `django.urls.re_path` don't support this use case - this PR adjust their type hints to support it.

## Related issues

Fixes #582


LanDinh:
Some follow-up work might be necessary on `URLResolver`, specifically the `urlconf_name` field - despite its name, it can also receive `List[URLResolver]` and `List[URLPattern]` as input (just look at the comments of the constructor in the Django implementation!), but I'm not sure what the exact full type / options are and don't want to mess up stuff, so I've left it out of this PR.

I took a quick look at the tests; it's not quite straightforward (I've never written any type checking test before). Is there some documentation somewhere that I could refer to on how to specify the expected output of cases? E. g. tests to make sure that `List[URLPattern]` will be expected and not raise a type error.

sobolevn:
Yes,
- https://github.com/typeddjango/pytest-mypy-plugins
- error: https://github.com/typeddjango/pytest-mypy-plugins/blob/master/pytest_mypy_plugins/tests/test-simple-cases.yml#L52-L55

sobolevn (commented):
Awesome, thank you!

Can we please test that now it works?

LanDinh:
I've tested it with my local snippet, it works :) Am now looking into writing a test for that.

LanDinh:
Test added.

sobolevn (approved):
Thanks!

PR #585, opened 2021-04-11
Add type hints for the postgres CursorDebugWrapper and add missing method to the BaseCursorDebugWrapper

LanDinh:
# Add type hints for the postgres CursorDebugWrapper and add missing method to the BaseCursorDebugWrapper

`django.db.backends.utils.CursorDebugWrapper` was missing annotations for the `debug_sql` method, and `django.db.backends.postgres.base` did not expose `CursorDebugWrapper`.

## Related issues

None


django-stubs/db/backends/postgresql/base.pyi
```diff
@@ -15,3 +17,7 @@ class DatabaseWrapper(BaseDatabaseWrapper):
     _named_cursor_idx: int = ...
     @property
     def pg_version(self) -> str: ...
+
+class CursorDebugWrapper(BaseCursorDebugWrapper):
+    def copy_expert(self, sql: str, file: IOBase, *args: Any): ...
```
sobolevn, line 22: I see that these methods return something, but it is not clear what 🤔 
https://github.com/django/django/blob/main/django/db/backends/postgresql/base.py#L343-L350

django-stubs/db/backends/utils.pyi
```diff
@@ -24,16 +25,19 @@ class CursorWrapper:
         tb: Optional[types.TracebackType],
     ) -> None: ...
     def callproc(self, procname: str, params: List[Any] = ..., kparams: Dict[str, int] = ...) -> Any: ...
-    def execute(
-        self, sql: str, params: Optional[Union[Sequence[_SQLType], Mapping[str, _SQLType]]] = ...
-    ) -> Optional[Any]: ...
-    def executemany(
-        self, sql: str, param_list: Sequence[Optional[Union[Sequence[_SQLType], Mapping[str, _SQLType]]]]
-    ) -> Optional[Any]: ...
+    def execute(self, sql: str, params: _ExecuteParameters = ...) -> Optional[Any]: ...
```
sobolevn: `Optional[Any]` is still just `Any`.

django-stubs/db/backends/postgresql/base.pyi
LanDinh: Yeah, so this is from psycopg2, not from Django itself - and I could not figure out the type hints of that one.... Partly also because it's a wheel and thus written in C and I'm not familiar with the translation between the c original code and the generated python.... And the documentation doesn't talk about the return type at all...

If you can make sense out of that, that would be amazing^^ I've tried and failed...

psycopg2 documentation: https://www.psycopg.org/docs/cursor.html
implementation of copy_expert: https://github.com/psycopg/psycopg2/blob/master/psycopg/cursor_type.c#L1577

sobolevn (approved):
Thanks!

PR #586, opened 2021-04-13
[3.2] Add durable argument to atomic

hannseman:
https://docs.djangoproject.com/en/3.2/topics/db/transactions/#django.db.transaction.atomic


sobolevn:
So, basically since the `1.8` relase, we going to add `django@3.2` features in. 
@hannseman can you please edit our table here? https://github.com/typeddjango/django-stubs#version-compatibility

hannseman:
> So, basically since the 1.8 relase, we going to add django@3.2 features in.

That's great news!

> @hannseman can you please edit our table here? https://github.com/typeddjango/django-stubs#version-compatibility

Not really sure what we want here, do you want me to add a new entry for 1.9 with only `3.2.x`?

sobolevn:
> do you want me to add a new entry for 1.9 with only 3.2.x?

Yes, please!

sobolevn (approved):
Thanks a lot!

PR #588, opened 2021-04-13
[3.2] Add missing exceptions

hannseman:
https://docs.djangoproject.com/en/3.2/ref/exceptions/#module-django.core.exceptions


sobolevn (approved):
💪

PR #589, opened 2021-04-13
[3.2] Adjust contrib.admin

hannseman:
* [`contrib.admin.action`](https://github.com/django/django/blob/stable/3.2.x/django/contrib/admin/decorators.py#L1)
* [`contrib.admin.display`](https://github.com/django/django/blob/stable/3.2.x/django/contrib/admin/decorators.py#L32)
* [`contrib.admin.EmptyFieldListFilter`](https://github.com/django/django/blob/stable/3.2.x/django/contrib/admin/filters.py#L434)
* [`contrib.admin.AdminSite.enable_nav_sidebar`](https://github.com/django/django/blob/stable/3.2.x/django/contrib/admin/sites.py#L58) (Actually added in Django 3.1)
* [`contrib.admin.AdminSite.final_catch_all_view`](https://github.com/django/django/blob/stable/3.2.x/django/contrib/admin/sites.py#L58)


django-stubs/contrib/admin/decorators.pyi
```diff
@@ -1,6 +1,12 @@
-from typing import Any, Callable, Optional, Type
+from typing import Callable, Optional, Type, Sequence
 
 from django.contrib.admin.sites import AdminSite
 from django.db.models.base import Model
 
+def action(
+    function: Optional[Callable] = None,
```
sobolevn: Can we describe the internals of `Callable` here? Maybe partially: just params or return type? Or it can be literally any callable?

sobolevn: Btw, we use `...` instead of plain values.

hannseman: > Can we describe the internals of Callable here? Maybe partially: just params or return type? Or it can be literally any callable?

Sure!

I found some other stuff from `contrib.admin` that was missing as well.

sobolevn (approved):

sobolevn:
Thanks a lot!

hannseman:
@sobolevn thanks for the review!

PR #595, opened 2021-04-16
Add `no_key` parameter to `select_for_update` queryset method (#594)

### LINKED ISSUE #594: select_for_update should accept no_key parameter
# Bug report

The `select_for_update` queryset method now accepts the `no_key` parameter: https://docs.djangoproject.com/en/3.2/ref/models/querysets/#select-for-update (django 3.2). The parameter isn't included in the stub function.

## What's wrong

```python
queryset.select_for_update(no_key=True)
# Unexpected keyword argument "no_key" for "select_for_update" of "_BaseQuerySet"  [call-arg]
```
## How is that should be

No error should be reported.

## System information

- OS: Linux
- `python` version: 3.9
- `django` version: 3.2
- `mypy` version: 0.812

I'll try to provide a PR soon.

stevanmilic:
## Related issues
- Closes #594
- Refs #594


sobolevn (approved):
Thanks!

PR #597, opened 2021-04-20
Handle GenericForeignKey class typeinfo lookup failure.

charettes:
This addresses an obscure crash we're getting when defining a GenericForeignKey subclass on a model.

Not sure how this slipped through type checking since `helpers.lookup_class_typeinfo -> Optional[TypeInfo]` while `.get_private_descriptor_type(type_info: TypeInfo, ...)` so this should be a clear type violation.

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -198,7 +198,12 @@ class DjangoContext:
                 # it's generic, so cannot set specific model
                 field_name = field.name
                 gfk_info = helpers.lookup_class_typeinfo(api, field.__class__)
-                gfk_set_type = helpers.get_private_descriptor_type(gfk_info, "_pyi_private_set_type", is_nullable=True)
+                if gfk_info is None:
+                    gfk_set_type = AnyType(TypeOfAny.unannotated)
+                else:
+                    gfk_set_type = helpers.get_private_descriptor_type(
+                        gfk_info, "_pyi_private_set_type", is_nullable=True
+                    )
                 expected_types[field_name] = gfk_set_type
 
         return expected_types
--- a/tests/typecheck/fields/test_generic_foreign_key.yml
+++ b/tests/typecheck/fields/test_generic_foreign_key.yml
@@ -19,3 +19,26 @@
                     pass
                 class Tag(models.Model):
                     content_object = fields.GenericForeignKey()
+-   case: generic_foreign_key_subclass_could_point_to_any_model_and_is_always_optional
+    main: |
+        from myapp.models import Tag, User
+        myuser = User()
+        Tag(content_object=None)
+        Tag(content_object=myuser)
+        Tag.objects.create(content_object=None)
+        Tag.objects.create(content_object=myuser)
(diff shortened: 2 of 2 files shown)


mypy_django_plugin/django/context.py
```diff
@@ -198,7 +198,12 @@ def get_expected_types(self, api: TypeChecker, model_cls: Type[Model], *, method
                 # it's generic, so cannot set specific model
                 field_name = field.name
                 gfk_info = helpers.lookup_class_typeinfo(api, field.__class__)
-                gfk_set_type = helpers.get_private_descriptor_type(gfk_info, "_pyi_private_set_type", is_nullable=True)
+                if gfk_info is None:
```
sobolevn, line 201: Thanks! Can you please provide a simple test for this? 
Otherwise, we can face regressions in the future! 😞

sobolevn (changes requested):
Great catch! 👍

charettes:
@sobolevn I came up with a regression test that alters the type of a field to one without any type info associated to it to reproduce the failure.

```
tests/typecheck/fields/test_generic_foreign_key.yml::generic_foreign_key_subclass_could_point_to_any_model_and_is_always_optional Traceback (most recent call last):
  File "mypy/checkexpr.py", line 3892, in accept
  File "mypy/checkexpr.py", line 270, in visit_call_expr
  File "mypy/checkexpr.py", line 347, in visit_call_expr_inner
  File "mypy/checkexpr.py", line 852, in check_call_expr_with_callee_type
  File "mypy/checkexpr.py", line 911, in check_call
  File "mypy/checkexpr.py", line 1023, in check_callable_call
  File "mypy/checkexpr.py", line 723, in apply_function_plugin
  File "/Users/charettes/Workspace/django-stubs/mypy_django_plugin/transformers/init_create.py", line 63, in redefine_and_typecheck_model_init
    return typecheck_model_method(ctx, django_context, model_cls, "__init__")
  File "/Users/charettes/Workspace/django-stubs/mypy_django_plugin/transformers/init_create.py", line 38, in typecheck_model_method
    expected_types = django_context.get_expected_types(typechecker_api, model_cls, method=method)
  File "/Users/charettes/Workspace/django-stubs/mypy_django_plugin/django/context.py", line 201, in get_expected_types
    gfk_set_type = helpers.get_private_descriptor_type(gfk_info, "_pyi_private_set_type", is_nullable=True)
  File "/Users/charettes/Workspace/django-stubs/mypy_django_plugin/lib/helpers.py", line 157, in get_private_descriptor_type
    sym = type_info.get(private_field_name)
AttributeError: 'NoneType' object has no attribute 'get'
```

I also tried to investigate why this wasn't caught by mypy through placing `reveal_type` in the `elif contenttypes_in_apps and isinstance(field, GenericForeignKey)` branch but it seems it never ventures there for a reason I couldn't figure out 🤷 .

mypy_django_plugin/django/context.py
charettes: Makes sense, added a regression test.

sobolevn (approved):
Thanks!

PR #600, opened 2021-04-22
Update BaseCommand.requires_system_checks for django >=3.2

abhiabhi94:
# I have made things!
Updated the type definition of the `required_system_checks` attribute on the `BaseCommand`.

- the documentation for this specific change in django can be seen at https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/#django.core.management.BaseCommand.requires_system_checks

## Related issues
None


abhiabhi94:
As far I could see(I am shortsighted), i didn't see a test for the modules defined inside the `managment` package. If acceptable, I could write a test for the above change or other too maybe.

Also, I don't think the contributing instructions as of now mention the installation of `GDAL` as requirement for the running of tests. The test below actually seems to require it
https://github.com/typeddjango/django-stubs/blob/8f97bf880d3022772581ea4cf8b5bf5297a27bad/tests/typecheck/models/test_inheritance.yml#L94-L115

I can maybe add the link to [django's geospatial installation instructions](https://docs.djangoproject.com/en/dev/ref/contrib/gis/install/geolibs/) here, if acceptable.

django-stubs/core/management/base.pyi
```diff
@@ -38,7 +39,7 @@ class BaseCommand:
     help: str = ...
     output_transaction: bool = ...
     requires_migrations_checks: bool = ...
-    requires_system_checks: bool = ...
+    requires_system_checks: Union[bool, List[Tags], Tuple[Tags], str]  = ...
```
sobolevn: Maybe `Tuple[Tags, ...]`?

sobolevn (commented):
Thanks! Test would be awesome!

django-stubs/core/management/base.pyi
abhiabhi94: yes, thanks for the feedback.

sobolevn (approved):

PR #598, opened 2021-04-20
Updated the BaseQuerySet to Django 3.2

maximmasiutin:
There is no _BaseQuerySet in Django 3.2 LTS.

The .pyi files didn't match the .py files. 

Besides that, the static code analyzer Pyre gave an error: 
`The modelled function "django.db.models.query.QuerySet.get" is an imported function, please model "django.db.models.query._BaseQuerySet.get" directly.`

With the above patch, the query.pyi  matches the query.py from the Django 3.2 LTS. 

I have also updated some other files just to use the QuerySet instead of the non-existing _BaseQuerySet.

PR #601, opened 2021-04-23
Added a few definitions to resolve incompatibility with Pyre/Pysa

maximmasiutin:
If we take https://github.com/facebook/pyre-check and use the typeddjango/django-stubs rather than the modules provided by the pyre-check, there will be errors:

```
../taint/django_sources_sinks.pysa:13:0 `django.http.request.HttpRequest.accepted_types` is not part of the environment!
../taint/django_sources_sinks.pysa:23:0 The modelled function `django.http.request.HttpRequest.__repr__` is an imported function, please model `object.__repr__` directly.
../taint/django_sources_sinks.pysa:64:0 `django.http.response.HttpResponseBase.__bytes__` is not part of the environment!
../taint/django_sources_sinks.pysa:71:0 `django.http.response.HttpResponse.__bytes__` is not part of the environment!
../taint/django_sources_sinks.pysa:74:0 The modelled function `django.http.response.HttpResponse.__iter__` is an imported function, please model `django.http.response.HttpResponseBase.__iter__` directly.
../taint/django_sources_sinks.pysa:131:0 Class `django.contrib.sessions.backends.base.SessionBase` has no attribute `_session_key`.
```

This pull request fixes the errors above mentioned, by defining missing member functions and properties.


sobolevn (approved):
Thanks!

Question: do we need to somehow test `pyre` compat with our project?

maximmasiutin:
@sobolevn Thank you very much! However, for the full compatibility with Pyre, the https://github.com/typeddjango/django-stubs/pull/598 should also be resolved. I don't know why it fails the checks. The commit that I was proposed is in full compatibility with Django 3.2 LTS. There is no "_BaseQuerySet" in Django 3.2 LTS.

PR #603, opened 2021-04-27
Replace models.Model annotations with type variables

### LINKED ISSUE #263: Incorrect type for `get_object()` in view mixins
# Bug report

## What's wrong

This code:

```python
class SubmissionRetryView(SubmissionCreateView, UpdateView):
    """This view is used to send the application submission once again."""

    pk_url_kwarg = ''  # do not allow to query by pk
    slug_url_kwarg = 'uid'
    slug_field = 'uid'
    context_object_name = 'instance'

    def get_queryset(self):
        """Returns custom queryset for each user."""
        return filter_submissions_to_retry(self.request.user.link.profile)

    def get_object(
        self, queryset: Optional[QuerySet[ApplicationSubmission]] = None,
     ) -> ApplicationSubmission:
        """Returns the post-processed copy of an application submission."""
        instance = super().get_object(queryset=queryset)
        reveal_type(instance)
        return prepare_submission_for_retry(instance)
```

Where `prepare_submission_for_retry` is `def prepare_submission_for_retry(instance: ApplicationSubmission) -> ApplicationSubmission:`

Raises this error:

```python
server/web_app/views/submission.py:124: note: Revealed type is 'django.db.models.base.Model'
server/web_app/views/submission.py:125: error: Argument 1 to "prepare_submission_for_retry" has incompatible type "Model"; expected "ApplicationSubmission"
```

## How is that should be

`get_object` should return instance type from `QuerySet`.

## System information

- `python` version: 3.7
- `django` version: 2.2
- `mypy` version: 0.750
- `django-stubs` version: 1.3

sondrelg:
Tries to implement @sobolevn's suggested fix for `get_object` and related types. 

Found one more application in `list.pyi` which seemed to apply.

Refs https://github.com/typeddjango/django-stubs/issues/263 (and possibly closes it)

Closes #263
Closes #598


django-stubs/views/generic/list.pyi
```diff
@@ -7,17 +7,19 @@ from django.views.generic.base import ContextMixin, TemplateResponseMixin, View
...
     paginate_by: int = ...
     paginate_orphans: int = ...
     context_object_name: Optional[str] = ...
     paginator_class: Type[Paginator] = ...
     page_kwarg: str = ...
     ordering: Sequence[str] = ...
-    def get_queryset(self) -> QuerySet: ...
+    def get_queryset(self) -> QuerySet[T]: ...
     def get_ordering(self) -> Sequence[str]: ...
     def paginate_queryset(self, queryset: _BaseQuerySet, page_size: int) -> Tuple[Paginator, int, QuerySet, bool]: ...
```
sobolevn: `QuerySet[T]` as well

django-stubs/views/generic/list.pyi
```diff
@@ -7,17 +7,19 @@ from django.views.generic.base import ContextMixin, TemplateResponseMixin, View
 from django.db.models import Model
 from django.http import HttpRequest, HttpResponse
 
+T = TypeVar('T', bound=Model)
+
 class MultipleObjectMixin(ContextMixin):
     allow_empty: bool = ...
     queryset: Optional[QuerySet] = ...
```
sobolevn: `Optional[QuerySet[T]]`

sobolevn (commented):
Thanks! Let's try it out!

sobolevn:
I would also appreciate a test or two for these new types!

sondrelg:
I'll add some 👍

django-stubs/views/generic/detail.pyi
```diff
@@ -1,20 +1,22 @@
...
+from typing import Any, Optional, Type, TypeVar
 
 from django.views.generic.base import ContextMixin, TemplateResponseMixin, View
 
 from django.db import models
 from django.http import HttpRequest, HttpResponse
 
+T = TypeVar('T', bound=models.Model)
+
 class SingleObjectMixin(ContextMixin):
```
sobolevn: Don't forget to use `Generic[T]` superclass!

django-stubs/views/generic/detail.pyi
```diff
@@ -1,20 +1,22 @@
...
+    model: Type[T] = ...
+    queryset: models.query.QuerySet[T] = ...
     slug_field: str = ...
     context_object_name: str = ...
     slug_url_kwarg: str = ...
     pk_url_kwarg: str = ...
     query_pk_and_slug: bool = ...
-    def get_object(self, queryset: Optional[models.query.QuerySet] = ...) -> models.Model: ...
-    def get_queryset(self) -> models.query.QuerySet: ...
+    def get_object(self, queryset: Optional[models.query.QuerySet] = ...) -> T: ...
```
sobolevn, line 18: `queryset: Optional[models.query.QuerySet[T]]`

django-stubs/views/generic/list.pyi
```diff
@@ -7,17 +7,19 @@ from django.views.generic.base import ContextMixin, TemplateResponseMixin, View
...
     paginate_by: int = ...
     paginate_orphans: int = ...
     context_object_name: Optional[str] = ...
     paginator_class: Type[Paginator] = ...
     page_kwarg: str = ...
     ordering: Sequence[str] = ...
-    def get_queryset(self) -> QuerySet: ...
+    def get_queryset(self) -> QuerySet[T]: ...
     def get_ordering(self) -> Sequence[str]: ...
     def paginate_queryset(self, queryset: _BaseQuerySet, page_size: int) -> Tuple[Paginator, int, QuerySet, bool]: ...
```
sobolevn: `Tuple[Paginator, int, QuerySet[T], bool]:`

django-stubs/views/generic/detail.pyi
sondrelg: I did this on purpose actually, and was hoping to discuss it.

Here's the Django code: https://github.com/django/django/blob/main/django/views/generic/detail.py#L30

Seeing that the **default** behavior of `get_object` is to call `get_queryset`, would it not make sense that the `queryset` optional kwarg should be `QuerySet[Any]`? Passing `QuerySet[T]` would result in the same behavior as passing nothing.

Let me know if that was unclear 🙂

django-stubs/views/generic/detail.pyi
sondrelg: I might be out of my depth here - what do you mean? 

Btw, if you think it would be easier, feel free to jump in and take over at any point to round off the PR 👏

sobolevn:
Closes #598

sobolevn:
Closes #263

PR #604, opened 2021-04-30
Correct type of core.management.commands.runserver.Command.default_port

henzef:
Strangely this variable is actually a string:
https://github.com/django/django/blob/8bcb00858e0ddec79cc96669c238d29c30d7effb/django/core/management/commands/runserver.py#L33


sobolevn (approved):
Thanks!

PR #610, opened 2021-05-13
Use BaseDatabaseWrapper class instead of sqlite3's wrapper

intgr:
# I have made things!

Code dealing with DatabaseWrapper instances should be capable of dealing
with any database, not just sqlite3.


sobolevn (approved):
Thanks!

PR #612, opened 2021-05-13
Improve hints for database connections (DatabaseWrapper)

intgr:
* `django.db.{connection, connections, router}` are now hinted -- including `ConnectionHandler` and `ConnectionRouter` classes.
* Several improvements to `BaseDatabaseWrapper` attribute hints.
* In many places, database connections were hinted as `Any`, which I changed to `BaseDatabaseWrapper`.
* In a few places I added additional `SQLCompiler` hints.
* Minor tweaks to nearby code.


django-stubs/db/backends/base/operations.pyi
```diff
@@ -8,12 +8,8 @@ from django.db.backends.utils import CursorWrapper
 from django.db.models.base import Model
 from django.db.models.expressions import Case, Expression
 from django.db.models.sql.compiler import SQLCompiler
-
-from django.db import DefaultConnectionProxy
 from django.db.models.fields import Field
 
-_Connection = Union[DefaultConnectionProxy, BaseDatabaseWrapper]
```
intgr, line 15: Here dropped the distinction between `DefaultConnectionProxy` and `BaseDatabaseWrapper`.

django-stubs/core/management/commands/inspectdb.pyi
```diff
@@ -10,7 +11,9 @@ class Command(BaseCommand):
     def normalize_col_name(
         self, col_name: str, used_column_names: List[str], is_relation: bool
     ) -> Tuple[str, Dict[str, str], List[str]]: ...
-    def get_field_type(self, connection: Any, table_name: Any, row: Any) -> Tuple[str, Dict[str, str], List[str]]: ...
+    def get_field_type(
+        self, connection: BaseDatabaseWrapper, table_name: Any, row: Any
```
sobolevn: Is `table_name` really `Any`?

sobolevn: Looks like `str` to me

django-stubs/db/backends/base/base.pyi
```diff
@@ -39,24 +43,23 @@ class BaseDatabaseWrapper:
...
-    close_at: Optional[Any] = ...
+    close_at: Optional[float] = ...
     closed_in_transaction: bool = ...
     errors_occurred: bool = ...
     allow_thread_sharing: bool = ...
-    run_on_commit: List[Any] = ...
+    run_on_commit: List[Callable[[], None]] = ...
     run_commit_hooks_on_set_autocommit_on: bool = ...
-    execute_wrappers: List[Any] = ...
+    execute_wrappers: List[Callable[[Any, Any, Any, Any, Any], None]] = ...
```
sobolevn: Why is it `Callable[[Any, Any, Any, Any, Any], None]`? Can we be more specific?

django-stubs/db/backends/base/client.pyi
```diff
@@ -4,6 +4,6 @@ from django.db.backends.base.base import BaseDatabaseWrapper
 
 class BaseDatabaseClient:
     executable_name: Any = ...
-    connection: Any = ...
+    connection: BaseDatabaseWrapper
```
sobolevn, line 7: Why not `connection: BaseDatabaseWrapper = ...`?

tests/typecheck/db/test_connection.yml
```diff
@@ -0,0 +1,13 @@
+-   case: raw_default_connection
```
sobolevn, line 1: Should we test something else, what do you think? SInce there are a lot of changes 🤔

sobolevn (changes requested):
Thanks a lot! Great work! 👍

django-stubs/core/management/commands/inspectdb.pyi
intgr: Yeah, I wanted to keep the scope down to `connection` arguments rather than chase every rabbit hole. This PR already took longer than I was hoping. But I'll add the hint for `table_name`.

django-stubs/db/backends/base/base.pyi
intgr: This is one of those rabbit holes... The documentation is hand-wavy https://docs.djangoproject.com/en/3.2/topics/db/instrumentation/ and I still don't understand what the return type is supposed to be, but something like...
```python
_SQLType = Union[None, bool, int, float, Decimal, str, bytes, datetime, UUID]
_ExecuteParameters = Optional[Union[Sequence[_SQLType], Mapping[str, _SQLType]]]
_ExecuteWrapper = Callable[
    [
        Callable[[str, Union[_ExecuteParameters, Sequence[_ExecuteParameters]], bool, Dict[str, Any]], Any],
        str,
        Union[_ExecuteParameters, Sequence[_ExecuteParameters]],
        bool,
        Dict[str, Any],
    ],
    Any,
]
```

And this is probably still wrong because I believe `_SQLType` could also contain DB-backend-specific types like `cx_Oracle.Binary`.

intgr: I guess hinting the type of the `params` argument is counter-productive. If users want to use this, unless they add an `if TYPE_CHECKING` guard and import the internal `_ExecuteParameters` type, they would need to annotate that argument as the type soup:
`Union[Optional[Union[Sequence[Union[None, bool, int, float, Decimal, str, bytes, datetime, UUID]], Mapping[str, Union[None, bool, int, float, Decimal, str, bytes, datetime, UUID]]]], Sequence[Optional[Union[Sequence[Union[None, bool, int, float, Decimal, str, bytes, datetime, UUID]], Mapping[str, Union[None, bool, int, float, Decimal, str, bytes, datetime, UUID]]]]]]`

And it would break every time something gets added to `_SQLType`.

I'll just use `Any` for the `params` argument for now.

django-stubs/db/backends/base/client.pyi
intgr: To me the ` = ...` seems to imply that this attribute has a value at class-definition-time. But the `connection` attribute only exists on instances, not the class itself.

But I don't feel strongly about this, I can add the `...` back, what do you think?

sobolevn (approved):
Thanks!

I will merge this as-is, feel free to add more tests later of you find them useful.

PR #615, opened 2021-05-17
[WIP/RFC] Revisit patching of mypy builtins (reveal_{type,locals})

### LINKED ISSUE #609: Improve monkey-patching for `reveal_{type,locals}
After doing #591 I've thought that it might be better to only do this patching during the `"django.setup()"` - like initially suggested here.

I think it a) should not require to use the `patch()` method, and b) should not be in place during runtime in general (after the (apparently necessary/unavoidable) "setup").

_Originally posted by @blueyed in https://github.com/typeddjango/django-stubs/issues/590#issuecomment-819423678_

blueyed:
Fixes https://github.com/typeddjango/django-stubs/issues/609
Reverts https://github.com/typeddjango/django-stubs/commit/ee58b18f1577eb31fb5faaae94725a18625904fc

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -1,3 +1,4 @@
+import builtins
 import os
 import sys
 from collections import defaultdict
@@ -59,6 +60,13 @@ def initialize_django(settings_module: str) -> Tuple["Apps", "LazySettings"]:
 
         models.QuerySet.__class_getitem__ = classmethod(noop_class_getitem)  # type: ignore
         models.Manager.__class_getitem__ = classmethod(noop_class_getitem)  # type: ignore
+
+        # Define mypy builtins, to not cause NameError during setting up Django.
+        # TODO: temporary/unpatch
+        assert not hasattr(builtins, "reveal_type")
+        builtins.reveal_type = lambda _: None
+        assert not hasattr(builtins, "reveal_locals")
+        builtins.reveal_locals = lambda: None
 
         from django.apps import apps
         from django.conf import settings
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ But, it does not make any sense to use this project without `mypy`.
 
 ### mypy crashes when I run it with this plugin installed
 
-Current implementation uses Django runtime to extract models information, so it will crash, if your installed apps or `models.py` is not correct. For this same reason, you cannot use `reveal_type` inside global scope of any Python file that will be executed for `django.setup()`.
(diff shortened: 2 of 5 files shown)


mypy_django_plugin/django/context.py
```diff
@@ -60,6 +61,13 @@ def noop_class_getitem(cls, key):
         models.QuerySet.__class_getitem__ = classmethod(noop_class_getitem)  # type: ignore
         models.Manager.__class_getitem__ = classmethod(noop_class_getitem)  # type: ignore
 
+        # Define mypy builtins, to not cause NameError during setting up Django.
+        # TODO: temporary/unpatch
```
sobolevn, line 65: 👀

sobolevn (commented):
Looks great, thanks!

mypy_django_plugin/django/context.py
blueyed: The idea here is to unset it afterwards again / only have it during the `"django.setup()"` - but likely/apparently mypy sets/overrides it anyway.
(I've only quickly tested it, and it worked as expected.)

Any suggestion for a test for this?

sobolevn: We can create a new test with something like this:

```yaml
-   case: patch_reveal_type_during_django_setup
    main: |
        # main is empty on purpose, because we test the `django.setup()` step:
    installed_apps:
        - myapp
    files:
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                # Will be executed during `django.setup()`:
                a = 1
                reveal_type(a)
                reveal_locals()
```

🤔

sobolevn (approved):
Thanks a lot, @blueyed !

mypy_django_plugin/django/context.py
aleksanb: I tried upgrading to current master of django-stubs, and the assert on this line is failing for me, while our current version of django-stubs runs fine.
```
    assert not hasattr(builtins, "reveal_type")
AssertionError
```
Could this assert be too conservative?

sobolevn: What do you mean by "too conservative"? I am open to any other alternatives that work 🙂

aleksanb: I'm not quite sure myself hehe.
The precondtion fails when i'm trying to type check locally, so therefore i thought that perhaps this assert might not be needed / be rejecting valid django projects.

sobolevn: Do you patch your builtins somehow? Maybe some other tools do it for you?
Or is it a problem of `django-stubs` alone?

Anyways, I think you should open a new issue.

PR #587, opened 2021-04-13
[3.2] Adjust model indexes

hannseman:
https://docs.djangoproject.com/en/3.2/ref/models/indexes/#django.db.models.Index

PR #616, opened 2021-05-19
Adjust Postgres indexes for Django 3.2

antonagestam:
Adds support to the Postgres-specific indexes for instantiating with [`*expressions`](https://docs.djangoproject.com/en/3.2/ref/models/indexes/#expressions) and the new [`include` argument](https://docs.djangoproject.com/en/3.2/ref/models/indexes/#include).

## Related issues

- Refs #587 (builds on top of, briefly mentioned this change in discussion).

### DIFF
--- a/tests/typecheck/models/test_indexes.yml
+++ b/tests/typecheck/models/test_indexes.yml
@@ -3,6 +3,7 @@
         from django.db import models
         from django.db.models.functions import Lower
         models.Index(fields=["foo"])
+        models.Index(models.F("bar"), name="test")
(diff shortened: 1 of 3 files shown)


tests/typecheck/models/test_indexes.yml
```diff
@@ -3,6 +3,7 @@
         from django.db import models
         from django.db.models.functions import Lower
         models.Index(fields=["foo"])
+        models.Index(models.F("bar"), name="test")
```
antonagestam, line 6: Thought it was worth adding a case with only an expression here.

sobolevn (approved):
Great work! Thanks!

PR #617, opened 2021-05-19
Add stub for OpClass

antonagestam:
Adds type hints for the new Postgres-specific [`OpClass`](https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/indexes/#opclass-expressions). I missed this one in the previous PR :)

## Related issues

- Refs #616

### DIFF
--- a/django-stubs/contrib/postgres/indexes.pyi
+++ b/django-stubs/contrib/postgres/indexes.pyi
@@ -2,7 +2,7 @@ from typing import Optional, Sequence, Union
 
 from django.db.models.query_utils import Q
 
-from django.db.models import Index
+from django.db.models import Index, Func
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks again!

sobolevn:
@antonagestam are you already using it with `django@3.2`? Do you have any other issues? I am thinking about making a new release for `django@3.2`

antonagestam:
@sobolevn We have upgraded to Django 3.2 without issues (🎉), but we have a few post-upgrade tasks that I'm dealing with now, hence this and the previous PR. I think `QuerySet.alias()` is missing too, but that's the only other outstanding one I've bumped into so far.

Thanks again for reviewing and merging so quickly 🙌

PR #618, opened 2021-05-19
Add stub for QuerySet.alias()

antonagestam:
I reasoned that this method can use a self type directly since it doesn't add attributes to model instances. Though, that's assuming the WIP for typing `.annotate()` doesn't do smart work with lookups, e.g. to give type errors for `.annotate(full_name=Concat(...)).filter(flul_nmae="")`?

Does this need tests or is it fine as-is?

## Related issues

- Refs #617


django-stubs/db/models/query.pyi
```diff
@@ -103,6 +103,7 @@ class _BaseQuerySet(Generic[_T], Sized):
     def prefetch_related(self: _QS, *lookups: Any) -> _QS: ...
     # TODO: return type
     def annotate(self, *args: Any, **kwargs: Any) -> QuerySet[Any]: ...
+    def alias(self: _QS, *args: Any, **kwargs: Any) -> _QS: ...
```
sobolevn, line 106: Do we need to edit our plugin to support this new method?

antonagestam: @sobolevn I'm not sure. Would the plugin currently give a type error for `.alias(intermediary_field=...).filter(intermediary_field__gt=...)`?

antonagestam: I experimented a little bit, and this doesn't give a type error (even though `other_thing` isn't a field or an annotation, e.g. simulating misspelling the annotation):

```python
ModelA.objects.annotate(things=Count("field")).filter(other_thing__gt=1)
```

Unless there's some other aspect that the plugin does that I'm missing, I think this change should be fine.

sobolevn: Will this work? `blogs = Blog.objects.alias(entries=Count('entry')).filter(entries__gt=5)`

antonagestam: Yes, that will pass, what I'm saying is just that kwargs to `filter()` already aren't inspected at all.

All of these pass:

```python
Blog.objects.alias(entries=Count('entry')).filter(entries__gt=5)
Blog.objects.annotate(anything=Count("woot")).filter(not_a_valid_lookup="woot")
Blog.objects.annotate(entries=Count('entry')).filter(entries__gt="test")
Blog.objects.annotate(anything=Count("woot")).filter(this_does_not_yield_error="woot")
```

PR #619, opened 2021-05-20
Document limitation with `django_stubs_ext.monkeypatch()`

bobwhitelock:
This monkey patching will only work when using Python 3.7+, which introduced the `__class_getitem__` magic method. This tripped me up for a while as I was on Python 3.6 and wondered why the monkey patching was not working, therefore documenting this to point this out to future readers.


sobolevn (approved):
Thanks! Good addition 👍

PR #621, opened 2021-05-23
Add `db_collation` parameter to `CharField` and `TextField`

ykiu:
Add `db_collation` parameter to `CharField` and `TextField`. The option has been added in django 3.2😃

https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.CharField.db_collation
https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.TextField.db_collation

### DIFF
--- a/django-stubs/db/models/fields/__init__.pyi
+++ b/django-stubs/db/models/fields/__init__.pyi
@@ -264,6 +264,31 @@ class TextField(Field[_ST, _GT]):
     _pyi_private_get_type: str
     # objects are converted to string before comparison
     _pyi_lookup_exact_type: Any
+    def __init__(
+        self,
+        verbose_name: Optional[Union[str, bytes]] = ...,
+        name: Optional[str] = ...,
+        primary_key: bool = ...,
+        max_length: Optional[int] = ...,
+        unique: bool = ...,
+        blank: bool = ...,
+        null: bool = ...,
+        db_index: bool = ...,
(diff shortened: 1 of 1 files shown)


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -263,6 +264,31 @@ class TextField(Field[_ST, _GT]):
     _pyi_private_get_type: str
     # objects are converted to string before comparison
     _pyi_lookup_exact_type: Any
+    def __init__(
```
sobolevn, line 267: Isn't it the same as in `CharField`?

sobolevn (commented):
Thanks! 👍

django-stubs/db/models/fields/__init__.pyi
ykiu: Thanks for the quick response!

Yes, the definition of `TextField.__init__()` is the same as that of `CharField`.

I'm not sure if I got your concern correctly, but if you are concerned about introducing unnecessary redundancy, I think we need to repeat the definition because `CharField` is not a subclass of `TextField` nor vice versa.

sobolevn: 👍

sobolevn (approved):
Thanks!

ykiu:
Thank you🙂

Issue #622, opened 2021-05-24
`concrete` attribute not seen in fields returned by `_meta.get_fields()`


pawmar:
# Bug report

## What's wrong

When iterating over fields returned by `_meta.get_fields()` and checking if they're concrete fields (have db column) I get:

`error: Item "Field[Any, Any]" of "Union[Field[Any, Any], ForeignObjectRel]" has no attribute "concrete"  [union-attr]`

example code (`obj` is a django db model instance):

```python
for field in obj._meta.get_fields():
    if field.concrete:
       continue
    # ...
```

## How is that should be

There should be no error as `concrete` is valid attribute here.

## System information

- OS:
- `python` version: 3.9.3
- `django` version: 3.2.2
- `mypy` version: 0.812
- `django-stubs` version: 1.8.0

sobolevn:
Yes, I can confirm: `concrete` is there on `Field` instances https://github.com/django/django/blob/d3d95d645fc9f36355d27598475612de43587db1/django/db/models/fields/__init__.py#L770

@pawmar can you please submit a PR?

pawmar:
@sobolevn added PR

PR #623, opened 2021-05-25
add concrete attr to db.models.Field #622

pawmar:
Adding missing `concrete` attribute to `db.models.Field`.

related issue: https://github.com/typeddjango/django-stubs/issues/622


sobolevn (approved):
Thanks!

PR #628, opened 2021-05-31
fix CursorWrapper __iter__ method

horpto:
I had a bunch of errors like `"None" has no attribute "__next__"`.


django-stubs/db/backends/utils.pyi
```diff
@@ -7,7 +7,7 @@ from uuid import UUID
 logger: Any
 
 # Python types that can be adapted to SQL.
-_SQLType = Union[None, bool, int, float, Decimal, str, bytes, datetime, UUID]
+_SQLType = Union[None, bool, int, float, Decimal, str, bytes, date, datetime, UUID, Tuple[Any, ...], List[Any]]
```
sobolevn, line 10: @LanDinh can you please take a look at this change? 🙂

sobolevn (approved):

sobolevn:
Thank you!

PR #631, opened 2021-06-03
fix(ResolverMatch): Added _func_path

### LINKED ISSUE #630: "ResolverMatch" has no attribute "_func_path"
# Bug report

## What's wrong

`error: "ResolverMatch" has no attribute "_func_path"  [attr-defined]`
Not finding _func_path as an attribute, but in Django ResolverMatch class _func_path initialized under conditional segments and whatever the condition it will get initialized. So it should be added in stubs as well.

## How is that should be

It should recognize _func_path as an attribute to ResolverMatch 

## System information

- OS:
- `python` version: 3.7.8
- `django` version: 3.2.3
- `mypy` version: 0.812

abhishek0220:
ResolverMatch class in Django contains _func_path variable and it will get initialized in init method whatever the condition, so it should be added in type stubs as well. https://github.com/django/django/blob/main/django/urls/resolvers.py#L48

# I have made things!

## Related issues

- Closes #630 
- Refs #630


django-stubs/urls/resolvers.pyi
```diff
@@ -14,6 +14,7 @@ class ResolverMatch:
     namespace: str = ...
     view_name: str = ...
     route: str = ...
+    _func_path: str = ...
```
sobolevn: We tend to use `_func_path: str` notation for things that don't have a class-level default value. Is this the case?

sobolevn (commented):
Thanks!

django-stubs/urls/resolvers.pyi
abhishek0220: Yes

sobolevn (approved):
Thanks!

PR #633, opened 2021-06-04
get_language can sometimes return None

syastrov:
From https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.translation.get_language
> Returns the currently selected language code. Returns None if translations are temporarily deactivated (by deactivate_all() or when None is passed to override()).

# I have made things!


sobolevn (approved):
👍

PR #639, opened 2021-06-10
Adds support for pyproject.toml files

### LINKED ISSUE #638: Configure django-stubs with pyproject.toml
Hello,

The django-stubs README includes an example on how to configure it using the `mypy.ini` or `setup.cfg` files:

```
[mypy]
plugins =
    mypy_django_plugin.main

[mypy.plugins.django-stubs]
django_settings_module = "myproject.settings"
```

But since mypy recently added support for pyproject.toml files I think that it would be nice to include a pyproject.toml example too.

canassa:
Since mypy 0.900 the pyproject.toml files are supported.

This PR adds a support for it. It searchs for a `tool.django-stubs` section. This is an example configuration:

```
[tool.django-stubs]
django_settings_module = "config.settings.local"
```

Fixes #638

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -3,6 +3,7 @@ import sys
 from functools import partial
 from typing import Callable, Dict, List, NoReturn, Optional, Tuple, cast
 
+import toml
 from django.db.models.fields.related import RelatedField
 from mypy.modulefinder import mypy_path
 from mypy.nodes import MypyFile, TypeInfo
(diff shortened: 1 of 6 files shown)


canassa:
The tests are breaking but I don't think that it's related to my changes 🤔

sobolevn:
Yes, fails are related. `mypy@0.900` changed error messages format, so our tests now fail 😢

canassa:
@sobolevn 

Could you provide some pointers on how to fix those tests? I might give it go

sobolevn:
@canassa I would be super grateful! But, it will require a lot of manual labor 😞 

So, let's take a look at what has changed:

```
__________________________ method_decorator_function ___________________________
/home/runner/work/django-stubs/django-stubs/tests/typecheck/utils/test_decorators.yml:20: 
E   Expected:
E     main:10: note: Revealed type is 'def (self: main.TestView, request: django.http.request.HttpRequest, *args: Any, **kwargs: Any) -> django.http.response.HttpResponseBase' (diff)
E   Actual:
E     main:10: note: Revealed type is "def (self: main.TestView, request: django.http.request.HttpRequest, *args: Any, **kwargs: Any) -> django.http.response.HttpResponseBase" (diff)
E   
E     E: ...ote: Revealed type is 'def (self: main.TestView, request: django.http...
E     A: ...ote: Revealed type is "def (self: main.TestView, request: django.http...
```

Notice, that `'` changed to be `"` in the new message from `mypy`. We have around ~140 failing tests because of this. Full list: https://github.com/typeddjango/django-stubs/pull/639/checks?check_run_id=2796956336

canassa:
@sobolevn I can do that I think.

But updating those tests would also break them if they are executed against mypy <0.900 so enforcing mypy >= 0.900 would be necessary. What do you think about that?

sobolevn:
> so enforcing mypy >= 0.900 would be necessary

Yes, sure! 
Let's rebase this PR.

canassa:
It's rebased!

Please note that on the setup.py file the mypy version is still set to `mypy>=0.790`.

sobolevn:
@canassa I have fixed it in https://github.com/typeddjango/django-stubs/commit/8c387e85fe5d3c3759b4ecbba643b0f2491fd063

mypy_django_plugin/main.py
```diff
@@ -79,18 +80,54 @@ def exit(error_type: int) -> NoReturn:
...
-    return cast(str, settings).strip("'\"")
+    def exit_toml(error_type: int) -> NoReturn:
+        from mypy.main import CapturableArgumentParser
+
+        usage = """(config)
+        ...
+        [tool.django-stubs]
+        django_settings_module = str (required)
+        ...
+        """.replace(
```
sobolevn: There's https://docs.python.org/3/library/textwrap.html#textwrap.dedent 
Maybe it is better?

sobolevn (changes requested):
Looks good. Let's add several tests cases:
- Just `.toml`
- Both `.toml` and `.cfg` files are there
- Wrong `.toml` header
- Missing required `.toml` setting
- Wrong type for the setting in `.toml`

I will be happy to merge and release it!

canassa:
@sobolevn 

I added the tests! The only scenario that I couldn't test is "Both .toml and .cfg files are there". If I understood correctly, the plugin is loaded by mypy and mypy decides which configuration will be used, so we don't have much say in there.

mypy_django_plugin/main.py
canassa: I added that change in an isolated commit just in case you don't like so it's easier to revert 😆 

https://github.com/typeddjango/django-stubs/pull/639/commits/a503ea1de5f41bd5bd6f57449551700fe6947513

The only problem that I found with `textwrap.dedent` is that the first line must also be indented otherwise it doesn't work. So this won't work anymore:

```
    usage = """(config)
    ....
    """
```

and I had to change it to be like this:

```
    usage = """
    (config)
    ....
    """
```

And that actually changes how the django-stubs displays those errors to the user - hence the updated tests.

canassa: Hmm I just realized that I could `.lstrip()` the string no changes in the tests are necessary:

```
handler = CapturableArgumentParser(prog="(django-stubs) mypy", usage=textwrap.dedent(usage).lstrip())
```

What do you think?

sobolevn (approved):
Looks great, thank you!

graingert:
Thanks a lot!

PR #640, opened 2021-06-11
Allow Collection for 'fields' and 'exclude' of form model helpers (#637)

### LINKED ISSUE #637: "field" and "exclude" arguments of model_to_dict() do not accept sets
# Bug report

## What's wrong

Our test suite contains code that simplifies to this fragment:
```py
from typing import Mapping

from django.forms import model_to_dict


def check(instance: Model, data: Mapping[str, object]) -> None:
    assert data == model_to_dict(instance, fields=data.keys())
```

When checking that with mypy, it reports:
```
testcase.py:8: error: Argument "fields" to "model_to_dict" has incompatible type "AbstractSet[str]";
expected "Union[List[Union[Callable[..., Any], str]], Sequence[str], Literal['__all__'], None]"
[arg-type]
        assert data == model_to_dict(instance, fields=data.keys())
```

## How is that should be

The implementation of `model_to_dict()` only needs `__bool__()` and `__contains__()` to be provided by the `fields` and `exclude` arguments, so passing a keys set should not be flagged as an error.

I think a solution could be to replace `Sequence` in the stubs annotation with `Collection`.

## System information

- OS: Ubuntu Linux 18.04
- `python` version: 3.6.9
- `django` version: 2.2.1
- `mypy` version: 0.812

mthuurne:
## Related issues

Closes #637

There are several functions and classes in `django.forms.models` that take a `fields` or `exclude` argument. Previously, `Sequence` was used to annotate these, but the code of Django (I checked version 3.2.4) doesn't require `__getitem__()` to be implemented, so requiring `Collection` instead is sufficient.

The practical advantage of requiring `Collection` is that a set, such as the key set of a dictionary, can be passed without first having to convert it to a list or tuple.

---

By the way, the full definition of `_Fields` is:
```py
_Fields = Union[List[Union[Callable, str]], Collection[str], Literal["__all__"]]
```
The `Literal["__all__"]` is for accepting `ALL_FIELDS`, but I don't know what `List[Union[Callable, str]]` is for. I couldn't find any code or documentation that treats elements of `fields` and `exclude` as callable.

### DIFF
--- a/django-stubs/forms/models.pyi
+++ b/django-stubs/forms/models.pyi
@@ -37,7 +37,7 @@ from django.db.models import ForeignKey
 
 ALL_FIELDS: str
 
-_Fields = Union[List[Union[Callable, str]], Collection[str], Literal["__all__"]]
+_Fields = Union[Collection[str], Literal["__all__"]]
 _Labels = Dict[str, str]
 _ErrorMessages = Dict[str, Dict[str, str]]
 
--- a/django-stubs/forms/models.pyi
+++ b/django-stubs/forms/models.pyi
@@ -2,6 +2,7 @@ from datetime import datetime
 from typing import (
     Any,
     Callable,
+    Collection,
     Dict,
     Iterator,
     List,
@@ -36,7 +37,7 @@ from django.db.models import ForeignKey
 
 ALL_FIELDS: str
 
-_Fields = Union[List[Union[Callable, str]], Sequence[str], Literal["__all__"]]
+_Fields = Union[List[Union[Callable, str]], Collection[str], Literal["__all__"]]
 _Labels = Dict[str, str]
 _ErrorMessages = Dict[str, Dict[str, str]]
 
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ with open("README.md") as f:
     readme = f.read()
 
 dependencies = [
-    "mypy>=0.790",
+    "mypy>=0.790,<0.900",
     "typing-extensions",
     "django",
     "django-stubs-ext",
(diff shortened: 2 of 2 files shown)


django-stubs/forms/models.pyi
```diff
@@ -36,7 +37,7 @@ from django.db.models import ForeignKey
 
 ALL_FIELDS: str
 
-_Fields = Union[List[Union[Callable, str]], Sequence[str], Literal["__all__"]]
+_Fields = Union[List[Union[Callable, str]], Collection[str], Literal["__all__"]]
```
sobolevn: Btw, can we improve `Callable` type here? Right now it is just `Callable[..., Any]`. Maybe it can be more specific?

sobolevn (approved):
Thanks!

sobolevn:
@mthuurne can you please pin `mypy<0.900`?

django-stubs/forms/models.pyi
mthuurne: As I wrote in the initial PR message, I don't see the code using `fields` or `exclude` as a callable anywhere, at least not in Django 3.2.4. Are you certain that's a feature that (still) exists?

mthuurne:
> @mthuurne can you please pin `mypy<0.900`?

I modified the version range in `setup.py`; I hope I got the syntax right.

django-stubs/forms/models.pyi
sobolevn: Ok, let's remove it then! 👍

mthuurne: I pushed a commit containing its removal.

PR #641, opened 2021-06-11
fix tests for mypy 0.900

canassa:
Fixes the tests for mypy >= 0.900

## WIP:

```basic_new_typeinfo``` requires a new `line` parameter now. I think that I need some advice on where to get that info from.

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -20,7 +20,7 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
 
     assert isinstance(base_manager_info, TypeInfo)
     new_manager_info = semanal_api.basic_new_typeinfo(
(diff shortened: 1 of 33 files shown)

### CODE REVIEW

File: mypy_django_plugin/transformers/managers.py [hunk 1]
```diff
@@ -20,7 +20,7 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
 
     assert isinstance(base_manager_info, TypeInfo)
     new_manager_info = semanal_api.basic_new_typeinfo(
-        ctx.name, basetype_or_fallback=Instance(base_manager_info, [AnyType(TypeOfAny.unannotated)])
+        ctx.name, basetype_or_fallback=Instance(base_manager_info, [AnyType(TypeOfAny.unannotated)]), line=ctx.call.line
```
canassa, line 23: I found a "line" in the ctx object and I used it in here, but it would be a good idea to make sure that's correct since I have no idea what I am doing 😄

sobolevn (approved):
Great work! Thank you! 🤩

PR #642, opened 2021-06-14
Patch detail and list generic view base classes

SmileyChris:
Similar to the already patched `FormMixin`, the detail and list generic view classes are also missing __class_getitem__ and choke when you try to use their generics.


sobolevn (approved):
Thanks!

PR #646, opened 2021-06-15
Copy decorated queryset methods to manager too

p7g:
# I have made things!

This pull request adds support for copying decorated method types from the QuerySet to the Manager when using `Manager.from_queryset`. It assumes that the decorator returns a function with the same signature as the decorated function, which handles common cases like `lru_cache` and `transaction.atomic`.

## Related issues

The issue is that only `FuncDef` nodes are copied to the manager class type generated by a call to `Manager.from_queryset`. Most notably, this leaves out decorated methods. I don't think there's an existing issue for this.

Here's an example:

```python
from django.db import models, transaction

class TestQuerySet(models.QuerySet["Test"]):
    @transaction.atomic
    def test_queryset_method(self):
        return self

TestManager = models.Manager.from_queryset(TestQuerySet)

class Test(models.Model):
    objects = TestManager()

if TYPE_CHECKING:
    reveal_type(Test.objects.test_queryset_method)
```

At the `reveal_type` there's a type error about TestManager having no attribute `test_queryset_method`. With this fix it works as expected.

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -1,4 +1,4 @@
-from mypy.nodes import GDEF, FuncDef, MemberExpr, NameExpr, RefExpr, StrExpr, SymbolTableNode, TypeInfo
+from mypy.nodes import GDEF, Decorator, FuncDef, MemberExpr, NameExpr, RefExpr, StrExpr, SymbolTableNode, TypeInfo
 from mypy.plugin import ClassDefContext, DynamicClassDefContext
 from mypy.types import AnyType, Instance, TypeOfAny
 
@@ -67,6 +67,11 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
             break
         for name, sym in class_mro_info.names.items():
             if isinstance(sym.node, FuncDef):
-                helpers.copy_method_to_another_class(
-                    class_def_context, self_type, new_method_name=name, method_node=sym.node
-                )
+                func_node = sym.node
+            elif isinstance(sym.node, Decorator):
+                func_node = sym.node.func
+            else:
+                continue
+            helpers.copy_method_to_another_class(
+                class_def_context, self_type, new_method_name=name, method_node=func_node
+            )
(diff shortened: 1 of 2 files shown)


mypy_django_plugin/transformers/managers.py
```diff
@@ -67,6 +67,11 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
             break
         for name, sym in class_mro_info.names.items():
             if isinstance(sym.node, FuncDef):
-                helpers.copy_method_to_another_class(
-                    class_def_context, self_type, new_method_name=name, method_node=sym.node
-                )
+                func_node = sym.node
+            elif isinstance(sym.node, Decorator):
```
sobolevn, line 71: We totally need a test for this!

sobolevn (changes requested):
Thanks a lot, this is good enough for now.
Please, add a simple test and I will merge it.

mypy_django_plugin/transformers/managers.py
p7g: Good call, I'll work on adding one tonight 👍

p7g: The test I added fails in this way on master:
```
________________________ from_queryset_with_decorated_queryset_methods ________________________
/Users/patrick/code/django-stubs/tests/typecheck/managers/querysets/test_from_queryset.yml:187: 
E   Expected:
E     main:2: note: Revealed type is "myapp.models.MyModel_NewManager[myapp.models.MyModel]"
E     main:3: note: Revealed type is "builtins.str" (diff)
E   Actual:
E     main:2: note: Revealed type is "myapp.models.MyModel_NewManager[myapp.models.MyModel]"
E     main:3: error: "MyModel_NewManager[MyModel]" has no attribute "queryset_method" (diff)
E     main:3: note: Revealed type is "Any"          (diff)
E   
E     E: main:3: note: Revealed type is "builtins.str"...
E     A: main:3: error: "MyModel_NewManager[MyModel]" has no attribute "queryset_...
E                ^
```

sobolevn (approved):
Thanks!

Issue #578, opened 2021-03-19
Wrong type annotation for create method of a Manager


tcleonard:
# Bug report

## What's wrong

Queryset manager `create` method has the wrong type definition, it seems to expect a queryset type back I think.
Mypy raises an error when trying to override the original `create` method.

For example:
```python
class MyModelManager(models.Manager):

    def create(self, *args, **kwargs) -> MyModel:
          return super().create(*args, **kwargs)


class MyModel(models.Model):

   objects = MyModelManager()
```

raises the following error:
```
Signature of "create" incompatible with supertype "_BaseQuerySet"
```

## System information

- OS: Linux
- `python` version: 3.8
- `django` version: 3.1.7
- `mypy` version: 0.790
- `django-stubs` version: 1.7.0

sobolevn:
Can you please try to replace `*args, **kwargs` with the precise signature?

tcleonard:
which should be? You mean listing all model attributes?

sobolevn:
Oh, I am sorry 🙂 
I have read your original post incorrectly. 

Can you please try to use `class MyModelManager(models.Manager[MyModel]):` instead?

tcleonard:
I still get the same error

tcleonard:
my error was not due to the return type of the create (I was changing the types of some of the arguments of the method)

tcleonard:
actually it was both...

So even simply:
```python
class MyModelManager(models.Manager['MyModel']):

    def create(self, *args, **kwargs) -> MyModel:
          return super().create(*args, **kwargs)


class MyModel(models.Model):

   objects = MyModelManager()
```

still raises `Signature of "create" incompatible with supertype "_BaseQuerySet"`.

syastrov:
I can't reproduce this. Both examples pass for me locally.
CI will probably agree:
https://github.com/typeddjango/django-stubs/pull/653

sobolevn:
I will close this, because #653 does not reproduce this issue.

Feel free to reopen / submit a reproducable example, so we can fix it!

PR #653, opened 2021-06-27
Check whether reported issue #578 actually fails currently

syastrov:
See #578


sobolevn:
Thanks a lot, do you think that this test has a value on its own?
Or can I close it?

syastrov:
I suppose it doesn't hurt to leave it.

sobolevn:
👍

PR #656, opened 2021-06-30
Fix type of MigrationLoader.applied_migrations

darius-blink:
# Fix type of MigrationLoader.applied_migrations

The type of `MigrationLoader.applied_migrations` was set as `Set[Tuple[str, str]]`. This is no longer correct. Looking through the Django code, the property is set in `db/migrations/loader.py`, lines `216-220`:

> ```python
> if self.connection is None:
>     self.applied_migrations = {}
> else:
>     recorder = MigrationRecorder(self.connection)
>     self.applied_migrations = recorder.applied_migrations()
> ```

Where the `MigrationRecorder.applied_migrations` is defined in `db/migrations/recorder.py`, line `72` (with a very helpful doc string as well):
> ```python
> def applied_migrations(self):
>     """
>     Return a dict mapping (app_name, migration_name) to Migration instances
>     for all applied migrations.
>     """
>     if self.has_table():
>         return {(migration.app, migration.name): migration for migration in self.migration_qs}
>     else:
>         # If the django_migrations table doesn't exist, then no migrations
>         # are applied.
>         return {}
> ```

This PR simply updates the type stub to reflect the Django code.


sobolevn (approved):
Thanks!

PR #657, opened 2021-07-02
Add distinct, order_by and all method signatures for _ValuesQuerySet

MichaelAquilina:
# I have made things!

Solves incorrect results reported by `django-stubs` when using `values_list` and `values`:

```
core/models/foo.py:100: error: "_ValuesQuerySet[Foo, Any]" has no attribute "distinct"
core/models/foo.py:119: error: "_ValuesQuerySet[Foo, str]" has no attribute "distinct"
```

```
core/tasks.py:32: error: "_ValuesQuerySet[Foo, int]" has no attribute "order_by"
```

```
core/tasks.py:154: error: "_ValuesQuerySet[Foo, int]" has no attribute "order_by"
```

### DIFF
--- a/django-stubs/db/models/query.pyi
+++ b/django-stubs/db/models/query.pyi
@@ -169,6 +169,9 @@ class _ValuesQuerySet(Generic[_T, _Row], Collection[_Row], Reversible[_Row], Que
     def latest(self, *fields: Any, field_name: Optional[Any] = ...) -> _Row: ...  # type: ignore
     def first(self) -> Optional[_Row]: ...  # type: ignore
     def last(self) -> Optional[_Row]: ...  # type: ignore
+    def distinct(self, *field_names: Any) ->  _ValuesQuerySet[_T, _Row]: ... # type: ignore
+    def order_by(self, *field_names: Any) -> _ValuesQuerySet[_T, _Row]: ... # type: ignore
+    def all(self) -> _ValuesQuerySet[_T, _Row]: ... # type: ignore
 
 class RawQuerySet(Iterable[_T], Sized):
     query: RawQuery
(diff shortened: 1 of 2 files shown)


django-stubs/db/models/query.pyi
```diff
@@ -169,6 +169,9 @@ class _ValuesQuerySet(Generic[_T, _Row], Collection[_Row], Reversible[_Row], Que
     def latest(self, *fields: Any, field_name: Optional[Any] = ...) -> _Row: ...  # type: ignore
     def first(self) -> Optional[_Row]: ...  # type: ignore
     def last(self) -> Optional[_Row]: ...  # type: ignore
+    def distinct(self) ->  _ValuesQuerySet[_T, _Row]: ... # type: ignore
+    def order_by(self, *field_names: Any) -> _ValuesQuerySet[_T, Row]: ... # type: ignore
```
sobolevn: Should it be `_Row`?

sobolevn (changes requested):
Thanks! Can you please add a single test for these three methods?

You can add them into any existing test as well.

django-stubs/db/models/query.pyi
MichaelAquilina: Yep nice catch :)

It seems like typos like this just cause the return type to transform to `Any` based on the tests I wrote? I wasnt aware of this

MichaelAquilina:
@sobolevn thanks for the review! I've fixed the `_Row` name and added a test case which covers all the addition :)

MichaelAquilina:
Somewhat related to this change.

I think the following classes appear to be unused in the project? It was somewhat confusing understanding their purpose while working on this change:

```
152:class ValuesListIterable(BaseIterable[Tuple]): ...
153:class NamedValuesListIterable(ValuesListIterable): ...
155:class FlatValuesListIterable(BaseIterable):
```

sobolevn (approved):
Looks great, thanks!

sobolevn:
@MichaelAquilina looks like they are unused indeed. Let's remove them in the next PR.

PR #659, opened 2021-07-03
Use `Sequence` instead of `List` for `path` param

davidszotten:
Unlike `List`, which is invariant, `Sequence` is covariant, which lets
`path` accept lists of subsets of the `Union` as well.

I believe this is safe, as django doesn't mutate this input. I found
[this comment](https://github.com/python/mypy/issues/3351#issuecomment-300447832) helpful


sobolevn (approved):
LGTM, thanks!

Issue #608, opened 2021-05-01
Get rid of `ValuesQuerySet`


sobolevn:
Problems with `QuerySet` / `ValuesQuerySet` separation:
- There are a lot of places, where we use `QuerySet` annotation, so `ValuesQuerySet` is not allowed there. 
- We cannot simply change it, because `_BaseQuerySet` is not enough for small portion of cases
- `T` type var has `models.Model` bound, which stops us from annotating thing with, for example, `QuerySet[SomeTypedDict]`, which is valid when we use `.values` and `SomeTypedDict` represents the final result of this call

sobolevn:
Related: https://github.com/typeddjango/django-stubs/pull/603/commits/6522f30cdb9027483f46d77167394c84eb7b7f4b

syastrov:
But should a `QuerySet` where `.values` was called still have the original `Model` in its type?

It might be relevant if I were to call `.filter` after calling `.values` to know which type of `Model` is being queried so that the type-checker can generate an error if I filtered on an invalid field.

If Django hadn't allowed users to "flaunt their individualism", this would be a lot simpler :)

https://docs.djangoproject.com/en/3.2/ref/models/querysets/#django.db.models.query.QuerySet.values
> The people who made Django prefer to put all the SQL-affecting methods first, followed (optionally) by any output-affecting methods (such as values()), but it doesn’t really matter. This is your chance to really flaunt your individualism.

Also, another weird edge-case is that you can call `.bulk_create` after calling `.values`. There, knowledge of the `Model` is also useful for type-checking.

However, we can just decide to go after simplicity rather than absolute correctness, and accept that these uncommon patterns of use are likely to result in less type-checking or false positives.

sobolevn:
I thing that we can try to add the second type argument dynamically to `QuerySet` instance when we need to (basically after `.values` / similar calls). This way, we would still have pretty user-facing API like `QuerySet[MyModel]` and all the features of `ValuesQuerySet`.

Later, we can use `TypeArgTuple` to represent this natively using variadic generics.

syastrov:
That would be nice. Would you check the behavior in other type-checkers if you haven't provided all the type parameters?
It would be pretty nasty if you always got an error there.

I'm not sure I see how using variadic generics will help? It seems like we have some fixed number of type parameters, but we should allow some defaults for them if not provided (except that is not currently possible).

syastrov:
[...]

PR #660, opened 2021-07-03
chore: remove unused class stubs from query.pyi

MichaelAquilina:
# I have made things!

Cleanup what appear to be unused class stubs.

Follow up from #657


sobolevn (approved):
Related #608 

Thanks!

Issue #144, opened 2019-08-25
ValuesQuerySet is removed and not available


sobolevn:
I have tried to use `ValuesQuerySet` that is returned in:

```python
def post_dates_by_author(author_id: int):
    """Returns published blog posts."""
    x = BlogPost.objects.filter(
        author__id=author_id,
    ).values_list('created_at', flat=True)
    reveal_type(x)
    return x
```

Reproduction link: https://github.com/sobolevn/django_stubs_example/blob/master/server/apps/main/logic/repo.py

Output:

```
» PYTHONPATH="$PYTHONPATH:$PWD" mypy server
server/apps/main/logic/repo.py:20: note: Revealed type is 'django.db.models.query.ValuesQuerySet[server.apps.main.models.BlogPost, datetime.datetime]'
```

But, it looks like `ValuesQuerySet` is removed in `1.9`, and there's not such class: https://github.com/django/django/blob/aed89adad54a977829c4f180b036033e031ebcc7/docs/releases/1.9.txt#L1069

So, it is not clear how to annotate your function to return `ValuesQuerySet` when it is missing from django.

mkurnikov:
I meant it to be stubs-only class, maybe I should've start it with `_` or smth. Anyway, I haven't thought about the annotation of the return value... 
Is it possible to annotate it with the string based annotation, like `'ValuesQuerySet'`?  Will mypy be ok with that? 

If it's not, then I guess the only way is to go back to the multi-parameter generics then... Support of the `annotate()` will require third generic param, and it's very unobvious to the users.

sobolevn:
That's my attempt:

```python
import typing

if typing.TYPE_CHECKING:
    from django.db.models.query import ValuesQuerySet

def test() -> 'ValuesQuerySet[BlogPost]':
    return BlogPost.objects.all().values_list('id', flat=True)
```

Output:

```
» PYTHONPATH="$PYTHONPATH:$PWD" mypy server
server/apps/main/logic/repo.py:19: error: "ValuesQuerySet" expects 2 type arguments, but 1 given
```

I was not sure about the second argument. And I have tried `int`:

```python
def test() -> 'ValuesQuerySet[BlogPost, int]':
    return BlogPost.objects.all().values_list('id', flat=True)
```

And it worked.

mkurnikov:
[...]

PR #661, opened 2021-07-03
Add more method signatures for _ValuesQuerySet

MichaelAquilina:
# I have made things!

Follow up from #657 - I found more cases in my codebase where method signatures where incorrectly missing.

Looking at https://github.com/typeddjango/django-stubs/issues/608 it seems like this issue has been discussed at length already about why this is an issue.

My initial thoughts was that maybe there was a way of having:

- an internal `_QuerySet[_T, _Row]` type
- an external facing alias `QuerySet[_T]` which is equivalent to `_QuerySet[_T, _T]`
- values list would then return `_QuerySet[_T, _Row]` as appropriate.

But based on the conversation you've had it seems like this might be harder than I assume it is?

The downside of keeping it in the current way is that we essentially need to redefine all `QuerySet` method signatures twice: Once for `QuerySet` and once for `_ValuesQuerySet`

### DIFF
--- a/django-stubs/db/models/query.pyi
+++ b/django-stubs/db/models/query.pyi
@@ -154,6 +154,9 @@ class _ValuesQuerySet(Generic[_T, _Row], Collection[_Row], Reversible[_Row], Que
     def distinct(self, *field_names: Any) ->  _ValuesQuerySet[_T, _Row]: ... # type: ignore
     def order_by(self, *field_names: Any) -> _ValuesQuerySet[_T, _Row]: ... # type: ignore
     def all(self) -> _ValuesQuerySet[_T, _Row]: ... # type: ignore
+    def annotate(self, *args: Any, **kwargs: Any) -> _ValuesQuerySet[_T, Any]: ...
+    def filter(self, *args: Any, **kwargs: Any) -> _ValuesQuerySet[_T, _Row]: ...
+    def exclude(self, *args: Any, **kwargs: Any) -> _ValuesQuerySet[_T, _Row]: ...
 
 class RawQuerySet(Iterable[_T], Sized):
     query: RawQuery
--- a/tests/typecheck/managers/querysets/test_values_list.yml
+++ b/tests/typecheck/managers/querysets/test_values_list.yml
@@ -29,11 +29,15 @@
 -   case: values_list_supports_queryset_methods
     main: |
         from myapp.models import MyUser
+        from django.db.models.functions import Length
         query = MyUser.objects.values_list('name')
         reveal_type(query.order_by("name").get())  # N: Revealed type is "Tuple[builtins.str]"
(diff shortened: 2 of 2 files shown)


tests/typecheck/managers/querysets/test_values_list.yml
```diff
@@ -29,11 +29,15 @@
...
         from myapp.models import MyUser
+        from django.db.models.functions import Length
         query = MyUser.objects.values_list('name')
         reveal_type(query.order_by("name").get())  # N: Revealed type is "Tuple[builtins.str]"
         reveal_type(query.distinct("name").get()) # N: Revealed type is "Tuple[builtins.str]"
         reveal_type(query.distinct().get()) # N: Revealed type is "Tuple[builtins.str]"
         reveal_type(query.all().get()) # N: Revealed type is "Tuple[builtins.str]"
+        reveal_type(query.filter(age__gt=16).get()) # N: Revealed type is "Tuple[builtins.str]"
+        reveal_type(query.exclude(age__lte=16).get()) # N: Revealed type is "Tuple[builtins.str]"
+        reveal_type(query.annotate(name_length=Length("name")).get()) # N: Revealed type is "Any"
```
MichaelAquilina, line 40: might be worth checking this one is correct - I wasnt 100% sure about the approach for `annotate`

django-stubs/db/models/query.pyi
```diff
@@ -154,6 +154,9 @@ class _ValuesQuerySet(Generic[_T, _Row], Collection[_Row], Reversible[_Row], Que
     def distinct(self, *field_names: Any) ->  _ValuesQuerySet[_T, _Row]: ... # type: ignore
     def order_by(self, *field_names: Any) -> _ValuesQuerySet[_T, _Row]: ... # type: ignore
     def all(self) -> _ValuesQuerySet[_T, _Row]: ... # type: ignore
+    def annotate(self, *args: Any, **kwargs: Any) -> _ValuesQuerySet[_T, Any]: ...
```
sobolevn, line 157: Please, see https://github.com/typeddjango/django-stubs/pull/398

We are working on fixing `Any` there

sobolevn (approved):
Ok, let's keep `Any` for now, but after  https://github.com/typeddjango/django-stubs/pull/398 we would need to change that!

sobolevn:
> My initial thoughts was that maybe there was a way of having

Do you have some free time to try this out? 🙂 
I would love to have this `ValuesQuerySet` removed for good!

MichaelAquilina:
> Do you have some free time to try this out? slightly_smiling_face
I would love to have this ValuesQuerySet removed for good!

I'd definitely want to give it a try at the very least :) is there anything in particular I should be aware of before diving in?

sobolevn:
See #608 and #144 and #33

PR #662, opened 2021-07-04
chore: remove excess # type: ignore comments in _ValuesQuerySet

MichaelAquilina:
# I have made things!

I noticed that a lot of `# type: ignore` comments (including mine) might have been added over time without really understand why they were needed.

I've left the ones that seem to cause errors. From what I can tell, this is because `_ValuesQuerySet` inherits both `Collection[_Row]` and `QuerySet[_T]`. Because `QuerySet[_T]` also inherits `Collection[_T]`, this seems to confuse mypy:

```
Cannot determine consistent method resolution order (MRO) for "_ValuesQuerySet"
```


django-stubs/db/models/query.pyi
```diff
@@ -140,20 +140,20 @@ _Row = TypeVar("_Row", covariant=True)
 class _ValuesQuerySet(Generic[_T, _Row], Collection[_Row], Reversible[_Row], QuerySet[_T], Sized): # type: ignore
     def __len__(self) -> int: ...
     def __contains__(self, x: object) -> bool: ...
-    def __iter__(self) -> Iterator[_Row]: ...  # type: ignore
-    @overload  # type: ignore
+    def __iter__(self) -> Iterator[_Row]: ...
+    @overload
     def __getitem__(self, i: int) -> _Row: ...
     @overload
     def __getitem__(self: _QS, s: slice) -> _QS: ... # type: ignore
```
MichaelAquilina, line 147: probably makes sense to leave a comment explaining why this is needed for the next person - but wanted to check with what the best way to phrase it is

sobolevn (approved):
Great work! 👍

sobolevn:
I have added https://github.com/typeddjango/django-stubs/pull/663 to check for this. Good catch!

PR #666, opened 2021-07-04
fix(BaseCache): corrected return type of get_many func

### LINKED ISSUE #665: Wrong return type for get_many function BaseCache
# Bug report

## What's wrong

As per BaseCache definition, get_many function can return `Dict[str, Any]` type of object, but in stubs, it is written as `Dict[str, Union[str, int]]`

## How is that should be

It should return an object of type `Dict[str, Any]`

## System information

- OS:
- `python` version: 3.7.8
- `django` version: 3.1
- `mypy` version: 0.902

abhishek0220:
# I have made things!

get_many function can return Dict[atr, Any] type of object, but in stubs, it is written as Dict[str, Union[str, int]]

## Related issues

- Closes #665


django-stubs/core/cache/backends/base.pyi
```diff
@@ -24,7 +24,7 @@ class BaseCache:
     def set(self, key: Any, value: Any, timeout: Any = ..., version: Optional[Any] = ...) -> None: ...
     def touch(self, key: Any, timeout: Any = ..., version: Optional[Any] = ...) -> bool: ...
     def delete(self, key: Any, version: Optional[Any] = ...) -> None: ...
-    def get_many(self, keys: List[str], version: Optional[int] = ...) -> Dict[str, Union[int, str]]: ...
+    def get_many(self, keys: List[str], version: Optional[int] = ...) -> Dict[str, Any]: ...
```
sobolevn: Why we have `keys` as `List[str]` and not `List[Any]`?

sobolevn: And is it a `List`? Or `Sequence`?

sobolevn (commented):
Thanks! One question though.

django-stubs/core/cache/backends/base.pyi
abhishek0220: Seems it depend on the SubClass implementation of BaseCache. As in Django implementation, I think it can be Sequence, and keys child can be of type Any.

sobolevn: Let's change it then!

abhishek0220: Sure, adding new changes

abhishek0220: should I change for other functions as well?

sobolevn: Please change them, if they have the same problem.

abhishek0220: Done

sobolevn (commented):
Looks like `key` is not `Any` after all. Facts:
- https://github.com/django/django/blob/main/django/core/cache/backends/base.py#L280-L292
- https://docs.djangoproject.com/en/3.2/topics/cache/#cache-key-warnings

I don't have much experience with caching internals, sorry.
And this seems tricky to me.

What's your vision on this?

abhishek0220:
Yep, you're right I too generally used this with str, but in one of the implementations Django-Redis https://github.com/jazzband/django-redis/blob/master/django_redis/client/default.py#L240-#L245 they defined key as Any so I thought it may possible to take the key as Any. <b>But seeing this we can say the key should be of type `str`. </b>

I too don`t have much experience with caching, I just used this a few days back and got this.

abhishek0220:
Seems purely dependent on subclass implementation, I think it's safe to use it as `Any`. As in Redis implementation, I`m able to create a key of any type.

sobolevn:
Ok, let's stick to `Any`, I don't mind. Thanks a lot!

sobolevn (approved):

PR #668, opened 2021-07-11
Add end-of-file-fixer hook to pre-commit

flaeppe:
Got pointed with setting up this hook here via https://github.com/typeddjango/djangorestframework-stubs/pull/156


sobolevn (approved):
Thanks 👍

PR #670, opened 2021-07-17
widen type of make_password and get_hasher parameters

timnyborg:
make_password() will accept a subclass of BasePasswordHasher as a hasher, but currently this is flagged by mypy.
Likewise, [get_hasher() will accept a BasePasswordHasher](https://github.com/django/django/blob/56f9579105c324ff15250423bf9f8bdf1634cfb4/django/contrib/auth/hashers.py#L118) (and immediately return it): 

Current false negative:
```python
from django.contrib.auth import hashers

class LegacyPasswordHasher(hashers.PBKDF2PasswordHasher):
    algorithm = "pbkdf2"
    digest = hashlib.sha512
    ...

def make_legacy_password(password: str, salt: str = None) -> str:
    return hashers.make_password(password, salt, hasher=LegacyPasswordHasher())


error: Argument "hasher" to "make_password" has incompatible type "LegacyPasswordHasher"; expected "str"

```


sobolevn (approved):
Thanks!

timnyborg:
I've updated the request with the code blackened by pre-commit

PR #671, opened 2021-07-17
Replace accepts expression as string

### LINKED ISSUE #503: 🧵 False positive for strings used in models.functions.Replace
# Bug report

## What's wrong

The stub for `django.db.models.functions.Replace`  is too restrictive, and doesn't permit strings to be used as
an argument, while in reality and in the [Django docs](https://docs.djangoproject.com/en/3.1/ref/models/database-functions/#replace), strings are allowed.

```python

from django.db.models import Value
from django.db.models.functions import Replace

expression = Replace("sketch", Value("spam"), Value("ham"))
# error: Argument 1 to "Replace" has incompatible type "str"; expected "Combinable"  [arg-type]
```

## How is that should be

It should not cause any problems. The expression above should cause no type errors, as it is a perfectly acceptable way of calling the function.

## System information

- OS:
- `python` version: Python 3.8.5
- `django` version: 3.0.10
- `mypy` version: 0.782

timnyborg:
Fixes #503 

Currently, the typical use of Replace (see for example the [django docs](https://docs.djangoproject.com/en/3.2/ref/models/database-functions/#replace)) where the expression is a string, will throw an error: 
```
error: Argument 1 to "Replace" has incompatible type "str"; expected "Combinable"
```
I'm uncertain whether Combinable should be replaced with its subclass Expression.


sobolevn (approved):
Thanks!

PR #675, opened 2021-07-25
Add datetime to DateTimeField

snmishra:
DateTimeField was missing datetime as a valid set type. But Django clearly accepts datetime.


sobolevn (approved):
Good catch!

PR #677, opened 2021-07-27
Add @display to django.contrib.admin

mwilliamson-foodsteps:
The `display` decorator is defined in `django.contrib.admin.decorators`, but
isn't included in `django.contrib.admin`, which is how the Django docs describe
its usage.


sobolevn (approved):
Thanks!

PR #679, opened 2021-07-28
Change PasswordResetForm.save() to accept HttpRequest, not WSGIRequest.

### LINKED ISSUE #552: False positive type failure in PasswordResetForm
# Bug report

## What's wrong

False positive type failure in PasswordResetForm:

> error: Argument "request" to "save" of "PasswordResetForm" has incompatible type "HttpRequest"; expected "Optional[WSGIRequest]"

```python
from django.contrib.auth.forms import PasswordResetForm
request = HttpRequest()

form = PasswordResetForm()
form.save(request=request)  # complains about incompatible type.
```

## How is that should be

No error should be thrown in mypy, as the code as written works fine.

## System information

- OS: Ubuntu 20.04
- `python` version: 3.8.5
- `django` version: 3.0.11
- `mypy` version: 0.782

richardxia:
This changes the `request` argument of `PasswordResetForm.save()` to be an `Optional[HttpRequest]` rather than `Optional[WSGIRequest]`. In general, Django's APIs mostly only care about getting an `HttpRequest`, not a `WSGIRequest`, since the latter is just a lightweight wrapper that bridges a raw request from WSGI to Django's `HttpRequest` type.

## Related issues
Closes #552.

### DIFF
--- a/django-stubs/contrib/auth/forms.pyi
+++ b/django-stubs/contrib/auth/forms.pyi
@@ -6,6 +6,7 @@ from django.contrib.auth.models import User
 from django.contrib.auth.tokens import PasswordResetTokenGenerator
 from django.core.exceptions import ValidationError
 from django.core.handlers.wsgi import WSGIRequest
+from django.http.request import HttpRequest
 
 UserModel: Any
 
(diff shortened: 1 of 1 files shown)


sobolevn (approved):

sobolevn:
Thanks!

richardxia:
Thank you!

PR #680, opened 2021-07-29
Fix/673/from queryset then custom qs method

### LINKED ISSUE #673: Error calling custom queryset method on from_queryset generated manager after calling other QuerySet method (e.g. `filter`)
`MyModel.objects.filter(...).my_method()` no longer works on a custom queryset after https://github.com/typeddjango/django-stubs/pull/398 was merged.

There are no existing tests for this.

syastrov:
# I have made things!

## Related issues

Closes #673

### DIFF
--- a/tests/typecheck/managers/querysets/test_from_queryset.yml
+++ b/tests/typecheck/managers/querysets/test_from_queryset.yml
@@ -1,10 +1,11 @@
 -   case: from_queryset_with_base_manager
     main: |
         from myapp.models import MyModel
-        reveal_type(MyModel().objects)  # N: Revealed type is "myapp.models.MyModel_NewManager[myapp.models.MyModel]"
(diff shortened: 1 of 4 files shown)


tests/typecheck/managers/querysets/test_from_queryset.yml
```diff
@@ -1,9 +1,11 @@
 -   case: from_queryset_with_base_manager
     main: |
         from myapp.models import MyModel
-        reveal_type(MyModel().objects)  # N: Revealed type is "myapp.models.MyModel_NewManager[myapp.models.MyModel]"
+        reveal_type(MyModel().objects)  # N: Revealed type is "myapp.models.NewManager[myapp.models.MyModel]"
```
syastrov, line 4: This is since the code no longer has to generate a new version of the manager with a type var, since it now already has one.

sobolevn (approved):
Great work, as always! 👍

PR #682, opened 2021-07-30
extend ProtectedError and RestrictedError stubs

timnyborg:
types for their init methods and *_objects properties

Solves `"ProtectedError" has no attribute "protected_objects"` and similar false positives


django-stubs/db/models/deletion.pyi
```diff
@@ -14,8 +14,13 @@ def RESTRICT(collector, field, sub_objs, using): ...
 def SET(value: Any) -> Callable: ...
 def get_candidate_relations_to_delete(opts: Options) -> Iterable[Field]: ...
 
-class ProtectedError(IntegrityError): ...
-class RestrictedError(IntegrityError): ...
+class ProtectedError(IntegrityError):
+    protected_objects: set[Model]
```
sobolevn: We use `typing.Set`

sobolevn (changes requested):
Thanks!

timnyborg:
Thanks, I've replaced with set with typing.Set

sobolevn (approved):

sobolevn:
Thanks!

PR #683, opened 2021-07-30
Change return type of queryset.bulk_update to int

abhiabhi94:
- the change was made in the commit https://github.com/django/django/commit/cd124295d882e13cff556fdeb78e6278d10ac6d5.

I'm not sure if this can be merged as of now, since this applies only to the `main` branch of `django` as of now, and will be released in `4.0`.


sobolevn:
> I'm not sure if this can be merged as of now

Yes, let's keep it `None` for now. We only support `django@3` for now.

flaeppe:
Is it possible to merge this now? Can't see that it was part of #786

sobolevn:
@flaeppe sure! Can you please rebase it?

abhiabhi94:
if it's okay, i will rebase it in some time.

abhiabhi94:
updated.

For what it is worth, [`django 4.0` has dropped support for `python 3.6` and `3.7`](https://docs.djangoproject.com/en/4.0/faq/install/#what-python-version-can-i-use-with-django), maybe we don't need to test against those versions in the CI?

sobolevn (approved):
Thanks!

PR #686, opened 2021-07-31
Update type of BinaryField to include memoryview as well

### LINKED ISSUE #667: BinaryField should have type `memoryview`
# Bug report

## What's wrong

`BinaryField` is annotated as type `bytes`.

## How is that should be

`BinaryField` can be of type `memoryview` (see [BinaryField.to_python()](https://github.com/django/django/blob/2a04e24d2dfc8e60a66e4369d970913cb2112d91/django/db/models/fields/__init__.py#L2272)).

## System information

- OS: macOS Big Sur (11.4)
- `python` version: 3.8.3
- `django` version: 3.1.13
- `mypy` version: 0.812

abhiabhi94:
## Related issues
- Closes #667 

I'm not sure the test is placed in the correct file. will update if required.

### DIFF
--- a/django-stubs/db/models/fields/__init__.pyi
+++ b/django-stubs/db/models/fields/__init__.pyi
@@ -493,7 +493,7 @@ class FilePathField(Field[_ST, _GT]):
     ): ...
 
 class BinaryField(Field[_ST, _GT]):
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks!

PR #689, opened 2021-08-11
Return Promise for lazy functions.

### LINKED ISSUE #688: `gettext_lazy` should return a Promise object
# Bug report

The type annotation added in #126 is incorrect. 

## What's wrong

```
from django.utils.translation import gettext_lazy, ngettext_lazy, pgettext_lazy, ugettext_lazy
from django.utils.functional import Promise

a = gettext_lazy("Help")
b = ugettext_lazy("Help")
c = pgettext_lazy("Help")
d = ngettext_lazy("help", "helps")
e = ugettext_lazy("Help")

assert isinstance(a, Promise)
assert isinstance(b, Promise)
assert isinstance(c, Promise)
assert isinstance(d, Promise)
assert isinstance(e, Promise)
```

## How is that should be

For all lazy functions above, django-stubs consider the return type to be `str`, which is actually incorrect. All of them are in fact instances of `django.utils.functional.Promise`

PIG208:
Django uses a proxy method for lazy translation functions that actually returns instances of `Promise` instead of `str`. To be accurate, we should change the signatures for these functions.

## Related issues

- Closes #688
- Refs #688


django-stubs/utils/translation/__init__.pyi
```diff
@@ -3,6 +3,7 @@ from contextlib import ContextDecorator
 from typing import Any, Callable, Optional, Union
 
 from django.core.handlers.wsgi import WSGIRequest
+from django.utils.functional import Promise
```
sobolevn: I think that `Promise` is a new thing, isn't it? Looks like that latest `3.2.x` does not have this primitive: https://github.com/django/django/tree/3.2.6/django/utils

sobolevn: Moreover, even in latest `master` it is just a plain class with no methods or anything: https://github.com/django/django/blob/main/django/utils/functional.py#L68

PIG208: Hmm.. `Promise` has been there for quite a while: https://github.com/django/django/commit/e4e28d907aee9f0e3977e2c6237406d7746c5719. But you are right that `Promise` is just a plain class. Since the proxy method knows about the `resultclasses`, I guess creating generics like `Promise[str]` would be better?

sobolevn: Maybe we can add these methods to `Promise` as well? https://github.com/django/django/blob/3.2.6/django/utils/functional.py#L84-L191

PIG208:
Hi, I have updated the PR. I think it is probably not necessary to make `Promise` generic at this point, having these methods should be sufficient.

django-stubs/utils/functional.pyi
```diff
@@ -20,7 +20,25 @@ class cached_property(Generic[_T]):
...
+# Promise is only subclassed by a proxy class defined in the lazy function
+# so it makes sense for it to have all the methods available in that proxy class
+class Promise:
+    def __init__(self, args: Any, kw: Any) -> None: ...
+    def __reduce__(self) -> Tuple[Any, Tuple[Any]]: ...
+    @classmethod
+    def __prepare_class__(cls) -> None: ...
+    @classmethod
+    def __promise__(cls, method_name: str) -> Callable: ...
+    def __text_cast(self) -> str: ...
```
sobolevn: We 100% do not need private methods.

django-stubs/utils/functional.pyi
```diff
@@ -20,7 +20,25 @@ class cached_property(Generic[_T]):
...
+    def __bytes_cast(self) -> bytes: ...
+    def __bytes_cast_encoded(self) -> bytes: ...
+    def __cast(self) -> Union[bytes, str]: ...
+    def __lt__(self, other: Any) -> bool: ...
+    def __hash__(self) -> int: ...
+    def __mod__(self, rhs: Any) -> str: ...
+    def __add__(self, other: Any) -> str: ...
+    def __radd__(self, other: Any) -> str: ...
+    def __deepcopy__(self, memo: Any): ...
```
sobolevn, line 27: Should return `-> Promise`. Ideally it should return self-type or bound-self-typevar, but `Promise` will be good enough.

django-stubs/utils/functional.pyi
```diff
@@ -20,7 +20,25 @@ class cached_property(Generic[_T]):
...
+    @classmethod
+    def __prepare_class__(cls) -> None: ...
+    @classmethod
+    def __promise__(cls, method_name: str) -> Callable: ...
+    def __text_cast(self) -> str: ...
+    def __bytes_cast(self) -> bytes: ...
+    def __bytes_cast_encoded(self) -> bytes: ...
+    def __cast(self) -> Union[bytes, str]: ...
+    def __lt__(self, other: Any) -> bool: ...
+    def __hash__(self) -> int: ...
```
sobolevn: I think `__hash__` can be removed.

django-stubs/utils/functional.pyi
PIG208: Looks like the implementation forces evaluation of the lazy value when any of these methods are called. [`__add__` for example](https://github.com/django/django/blob/main/django/utils/functional.py#L195), returns whatever the casted value's `__add__` returns.

django-stubs/utils/functional.pyi
PIG208: You are right! These are removed.

django-stubs/utils/functional.pyi
PIG208: Removed.

PIG208:
There is still a problem with this. `Promise` [proxied other methods of resultclasses](https://github.com/django/django/blob/3.2.6/django/utils/functional.py#L111-L119), so we probably want those be available on the `Promise` object as well. Otherwise, these methods won't be usable. I will try to update the plugin for this.

PIG208:
This is my first time working on a plugin. I hope this implementation does not go horribly wrong. Another feature that I'm planning to add is to being able to determine what attributes are available on `Promise[_ResultClassT]` by calculating the mro of the proxied object.

Which hook should I use, `get_customize_class_mro_hook`, `get_attribute_hook` (combined with adding `__getattr__` to `Promise`) or `get_method_hook`? Any help will be appreciated!

PIG208:
I believe that this implementation should be good enough for another review. @sobolevn could you take a look at this?

I ditched the generic approach and reimplement this with a simpler plugin handling the use case for `lazystr` (and `*_lazy` as well) [suggested by](https://chat.zulip.org/#narrow/stream/3-backend/topic/The.20proper.20way.20to.20deal.20with.20.60Promise.60.20objects/near/1415829) @andersk . This is tested with some additional test cases and the entire codebase of Zulip (<strike>edit, there still seems to be some issues to be resolved. Will update soon </strike>). The earlier commits are left unchanged.

Looks like the CI is still failing for another unrelated issue. Will look into this in a separate PR probably. Otherwise, I think we are good to go.

PIG208:
@sobolevn Hi! The PR has been rebased and the CI is passing. I think it is ready for review.

mypy_django_plugin/transformers/functional.py
```diff
@@ -0,0 +1,40 @@
...
+        method_name = ctx.context.name
+    elif isinstance(ctx.context, CallExpr) and isinstance(ctx.context.callee, MemberExpr):
+        method_name = ctx.context.callee.name
+    else:
+        ctx.api.fail(f'Cannot resolve the attribute of "{ctx.type}"', ctx.context, code=ATTR_DEFINED)
+        return AnyType(TypeOfAny.from_error)
+
+    str_info = helpers.lookup_fully_qualified_typeinfo(helpers.get_typechecker_api(ctx), f"builtins.str")
+    assert str_info is not None
+    method = str_info.get(method_name)
```
sobolevn: I think that you should use `analyze_member_access` here from `mypy/checkmember.py`

PIG208: Updated.

sobolevn (approved):
Let's try it out! Thank you!

hannseman:
This PR causes errors when passing `gettext_lazy` strings to parameters expecting strings like `verbose_name` and `help_text` on model fields:

```python
error: Argument "verbose_name" to "PositiveBigIntegerField" has incompatible type
"_StrPromise"; expected "Optional[str]"  [arg-type]
            null=True, blank=True, verbose_name=_("number")
```

(where `_` is defined as `from django.utils.translation import gettext_lazy as _`)

sobolevn:
Please, open a new issue. I am not releasing new version yet, we have time to fix it.

terencehonles:
@PIG208 was there a reason `_StrPromise`

https://github.com/typeddjango/django-stubs/blob/fa972f1632fb69861c913f2bf359fcf4d2284d7a/django-stubs/utils/functional.pyi#L31-L46

Is a subclass of `Sequence[str]` and not `str` itself? A string is slice-able so I understand the thought process at a high level, but `lazystr` expects/returns a string not a sequence of strings and I'm not sure why this was made this generic / explicit. This has caused a lot of downstream code (even in this codebase) to need to declare everything as `StrOrPromise`. While this doesn't seem like that big of a lift, I'm not sure it should be needed at all.

I'm raising this question here instead of creating a separate issue since I figured I'd ask about the intent before completely suggesting this should change (maybe there was a reason I am not aware of).

It seems like `Promise` should have been made generic and returns and possibly subclasses the type it is so that callers can treat a promise and the object itself interchangeably.

@sobolevn not sure if you have any thoughts or would prefer this to be a separate issue.

PIG208:
@terencehonles  Thanks for bringing this up!

`_StrPromise` is a subclass of `Sequence[str]` because this is how [`builtins.str`](https://github.com/python/typeshed/blob/main/stdlib/builtins.pyi#L411-L586) is typed in typeshed. `_StrPromise` is defined only because operators don't work well with mypy plugins, while we want methods on a regular `str` to be all available. These are just the implementation details.

The *point* of having `_StrPromise` is to avoid using them with regular `str` interchangeably. This can cause real bugs when dealing with third-party libraries, because `Promise` is not `str`.

We did not make `Promise` generic because that would require more work to implement a full-blown mypy plugin to support arbitrary type arguments, while the most prevalent use case of `Promise` is with lazily evaluated strings.

andersk:
See also https://github.com/typeddjango/django-stubs/pull/1139#issuecomment-1232167698 for concrete examples of what would go wrong if we incorrectly told mypy that `_StrPromise` is a subclass of `str`.

terencehonles:
ok, thanks for the explanation.

> The _point_ of having `_StrPromise` is to avoid using them with regular `str` interchangeably. This can cause real bugs when dealing with third-party libraries, because `Promise` is not `str`.

A `Promise[str]` seems like it should be mostly considered a string and I'm not sure I agree or have run into an issue that this comment refers to, but I understand it would fail an `isinstance(...)` check and for mypy and potential future compilation would be concerned that would be an issue.

I had believed that Django's Promise implementation should be a more or less drop in replacement for the object it's promising since I thought it's blockingly coerced into the object when it's used, but it sounds like there's been some discussion of this and I will check out the comment @andersk pointed to :+1:

PR #695, opened 2021-08-15
Minor fixes to improve django-stubs

PIG208:
## Related issues
None. This batch of fixes are found the project of fixing 2k+ errors in the django-stubs migration of Zulip (zulip/zulip#18777). As they are mostly minor fixes, I integrate them in this PR.


django-stubs/contrib/staticfiles/views.pyi
```diff
@@ -1,6 +1,6 @@
 from typing import Any
 
-from django.core.handlers.wsgi import WSGIRequest
+from django.http.request import HttpRequest
 from django.http.response import FileResponse
 
-def serve(request: WSGIRequest, path: str, insecure: bool = ..., **kwargs: Any) -> FileResponse: ...
+def serve(request: HttpRequest, path: str, insecure: bool = ..., **kwargs: Any) -> FileResponse: ...
```
sobolevn, line 6: 👍

django-stubs/contrib/auth/backends.pyi
```diff
@@ -10,9 +10,7 @@ _AnyUser = Union[Model, AnonymousUser]
 UserModel: Any
 
 class BaseBackend:
-    def authenticate(
-        self, request: HttpRequest, username: Optional[str] = ..., password: Optional[str] = ..., **kwargs: Any
-    ) -> Optional[AbstractBaseUser]: ...
+    def authenticate(self, request: Optional[HttpRequest], **kwargs: Any) -> Optional[AbstractBaseUser]: ...
```
sobolevn, line 13: Yeap, current `3.2.x` django agrees with this signature: https://github.com/django/django/blob/3.2.6/django/contrib/auth/backends.py#L9

django-stubs/core/mail/backends/base.pyi
```diff
@@ -13,4 +13,4 @@ class BaseEmailBackend:
     def __exit__(
         self, exc_type: Type[BaseException], exc_value: BaseException, traceback: types.TracebackType
     ) -> None: ...
-    def send_messages(self, email_messages: Iterable[EmailMessage]) -> int: ...
+    def send_messages(self, email_messages: List[EmailMessage]) -> int: ...
```
sobolevn: Can you please explain this change? Why `Iterable` is not good enough?

sobolevn: Moreover, `List` is invariant. It might cause problems.

sobolevn (commented):
Thanks a lot! Great to hear that `zulip` is going to use `django-stubs` 🎉

django-stubs/core/mail/backends/base.pyi
PIG208: Given the way `send_messages` is invoked (https://github.com/django/django/blob/3.2.6/django/core/mail/__init__.py#L83-L87), it is more appropriate to type it as list-like object. This allows the implementation of subclasses to use methods like `__len__` and `__getitem__`, and overwritting the method with signature of `def send_messages(self, email_messages: Iterable[EmailMessage]) -> int` will still be acceptable.

PIG208: Since `List` is invariant, perhaps `Sequence` will be better here. The PR has been updated.

sobolevn (approved):
Thanks a lot! 👍

django-stubs/contrib/staticfiles/views.pyi
sobolevn: Btw, if you have time/will you can refactor almost all `WSGIRequest` references into `HttpRequest`

PR #705, opened 2021-09-06
Fix `BaseModelForm` not working with generic type parameter

### LINKED ISSUE #703: `BaseModelForm` should be patched to be generic
# Bug report

## What's wrong

https://github.com/typeddjango/django-stubs/commit/eb702384a8c77dcdaa57d5e56a1bd21272e04eb8 (released in 1.9.0 made `BaseModelForm` generic, but `django_stubs_ext.monkeypatch()` has not been updated to add patch in the `__class_getitem__` method for the class.

Running `mypy --disallow-any-generics test.py` on:

```
from django import forms
import django_stubs_ext

django_stubs_ext.monkeypatch()


class FooForm(forms.ModelForm):
    pass
```

results in:

```
test.py:8: error: Missing type parameters for generic type "ModelForm"
```

but if one tries to add the generic param:

```
from django import forms
import django_stubs_ext

django_stubs_ext.monkeypatch()


class FooForm(forms.ModelForm[Model]):
    pass
```

now `mypy` will not complain, but actually running the code (with `python test.py`) will result in:

```
TypeError: 'ModelFormMetaclass' object is not subscriptable
```

## How is that should be

`BaseModelForm` should be added to: https://github.com/typeddjango/django-stubs/blob/f98330356c663b9dddfe029d28c9968008098a4d/django_stubs_ext/django_stubs_ext/patch.py#L40

## System information

- OS: Docker image `python:3.9.6-slim-buster`
- `python` version: Docker image `python:3.9.6-slim-buster`
- `django` version: 3.2.7
- `mypy` version: 0.910
- `django-stubs-ext` version: 0.3.0

ruohola:
Resolves: https://github.com/typeddjango/django-stubs/issues/703

### DIFF
--- a/django_stubs_ext/tests/test_monkeypatching.py
+++ b/django_stubs_ext/tests/test_monkeypatching.py
@@ -5,6 +5,8 @@ from typing import Optional
 import pytest
 from _pytest.fixtures import FixtureRequest
 from _pytest.monkeypatch import MonkeyPatch
+from django.db.models import Model
+from django.forms.models import ModelForm
 from typing_extensions import Protocol
 
 import django_stubs_ext
@@ -45,6 +47,9 @@ def test_patched_generics(make_generic_classes: _MakeGenericClasses) -> None:
     for el in _need_generic:
         if el.version is None:
             assert el.cls[type] is el.cls  # `type` is arbitrary
+
+    class TestForm(ModelForm[Model]):
+        pass
 
 
 @pytest.mark.parametrize(
--- a/django_stubs_ext/django_stubs_ext/patch.py
+++ b/django_stubs_ext/django_stubs_ext/patch.py
@@ -6,6 +6,7 @@ from django.contrib.admin.options import BaseModelAdmin
 from django.db.models.fields import Field
 from django.db.models.manager import BaseManager
 from django.db.models.query import QuerySet
+from django.forms.models import BaseModelForm
 from django.views.generic.detail import SingleObjectMixin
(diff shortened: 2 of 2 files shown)


sobolevn:
Thanks a lot, @ruohola 
Please, also add a test for it, so it won't happen again 🙂

ruohola:
> Thanks a lot, @ruohola
> Please, also add a test for it, so it won't happen again 🙂

Done I guess. Not sure if you had something else in mind.

sobolevn (approved):
Thanks!

ruohola:
Fixed CI.

django_stubs_ext/tests/test_monkeypatching.py
```diff
@@ -46,6 +47,9 @@ def test_patched_generics(make_generic_classes: _MakeGenericClasses) -> None:
         if el.version is None:
             assert el.cls[type] is el.cls  # `type` is arbitrary
 
+    class TestForm(forms.ModelForm[type]):
```
sobolevn: You need to: 
1. use an actual model type here (if it is already created)
2. just add a `# type: ignore`, because we test a different thing

ruohola: 👍 All fixed, no need for the `# type: ignore`.

sobolevn (approved):
Thanks a lot!

sobolevn:
I am releasing `django-stubs-ext@0.3.1` right now

PR #708, opened 2021-09-08
Add through_defaults for RelatedManager methods

snchildress:
This PR:
- adds the `create` method stub of the RelatedManager class
- adds types for the `through_defaults` kwarg for the `create`, `add`, and `set` methods of the RelatedManager class

Documentation: https://docs.djangoproject.com/en/3.2/ref/models/relations/

### DIFF
--- a/django-stubs/db/models/manager.pyi
+++ b/django-stubs/db/models/manager.pyi
@@ -1,6 +1,7 @@
 import datetime
 from typing import (
     Any,
+    Callable,
     Dict,
     Generic,
     Iterable,
@@ -121,10 +122,23 @@ class Manager(BaseManager[_T]): ...
 
 class RelatedManager(Manager[_T]):
     related_val: Tuple[int, ...]
-    def add(self, *objs: Union[_T, int], bulk: bool = ...) -> None: ...
+    def add(
+        self,
+        *objs: Union[_T, int],
+        bulk: bool = ...,
+        through_defaults: Optional[Union[Dict[str, Any], Callable]] = None...,
+    ) -> None: ...
+    def create(
+        self, **kwargs: Any, through_defaults: Optional[Union[Dict[str, Any], Callable]] = None...,
+    ) -> None: ...
     def remove(self, *objs: Union[_T, int], bulk: bool = ...) -> None: ...
     def set(
-        self, objs: Union[QuerySet[_T], Iterable[Union[_T, int]]], *, bulk: bool = ..., clear: bool = ...
+        self,
+        objs: Union[QuerySet[_T], Iterable[Union[_T, int]]],
+        *,
+        bulk: bool = ...,
+        clear: bool = ..., 
+        through_defaults: Optional[Union[Dict[str, Any], Callable]] = None...,
     ) -> None: ...
     def clear(self) -> None: ...
 


django-stubs/db/models/manager.pyi
```diff
@@ -121,10 +122,23 @@ class Manager(BaseManager[_T]): ...
 
 class RelatedManager(Manager[_T]):
     related_val: Tuple[int, ...]
-    def add(self, *objs: Union[_T, int], bulk: bool = ...) -> None: ...
+    def add(
+        self,
+        *objs: Union[_T, int],
+        bulk: bool = ...,
+        through_defaults: Optional[Union[Dict[str, Any], Callable]] = None...,
```
sobolevn, line 129: Is it just `Callable`? Or can we be more specific?

django-stubs/db/models/manager.pyi
```diff
@@ -121,10 +122,23 @@ class Manager(BaseManager[_T]): ...
...
+    ) -> None: ...
     def remove(self, *objs: Union[_T, int], bulk: bool = ...) -> None: ...
     def set(
-        self, objs: Union[QuerySet[_T], Iterable[Union[_T, int]]], *, bulk: bool = ..., clear: bool = ...
+        self,
+        objs: Union[QuerySet[_T], Iterable[Union[_T, int]]],
+        *,
+        bulk: bool = ...,
+        clear: bool = ..., 
+        through_defaults: Optional[Union[Dict[str, Any], Callable]] = None...,
```
sobolevn, line 141: The same here.

django-stubs/db/models/manager.pyi
```diff
@@ -121,10 +122,23 @@ class Manager(BaseManager[_T]): ...
...
     related_val: Tuple[int, ...]
-    def add(self, *objs: Union[_T, int], bulk: bool = ...) -> None: ...
+    def add(
+        self,
+        *objs: Union[_T, int],
+        bulk: bool = ...,
+        through_defaults: Optional[Union[Dict[str, Any], Callable]] = None...,
+    ) -> None: ...
+    def create(
+        self, **kwargs: Any, through_defaults: Optional[Union[Dict[str, Any], Callable]] = None...,
```
sobolevn, line 132: And here

sobolevn (changes requested):
Thanks a lot! We also need tests for this 🙂

django-stubs/db/models/manager.pyi
snchildress: The Django docs generically state that as of 3.1, `through_defaults` can be a callable: https://docs.djangoproject.com/en/3.2/ref/models/relations/#django.db.models.fields.related.RelatedManager.add

snchildress (commented):
Thanks for the feedback! I'll add tests before resubmitting for review

sobolevn:
What kind of callable? `Callable[[], None]`? `Callable[[Model], bool]`? Any callable? 🙂

rockstarr-programmerr:
Hi guys, can I resurrect this PR :) Regarding the discussion around which kind of callable, I had a look at the source code of Django, and they actually mean a callable (`Callable[[], Any]`) can be used as a value in the `through_defaults` dictionary. So my suggestion for the typing of `through_defaults` is just:
```
through_defaults: Optional[Dict[str, Any]] = ...
```

rockstarr-programmerr:
Actually, since `ManyToManyRelatedManager` is different from `RelatedManager`, I suggest a new stub class

```
class ManyToManyRelatedManager(Manager[_T]):
    def add(
        ...
        through_defaults: Optional[Dict[str, Any]] = ...,
    ) -> None: ...
```

intgr:
> Hi guys, can I resurrect this PR :)

Yes please! This PR has been sitting idle for 1.5 years. Create a new PR and reference this one. Though I'll try to remember to credit @snchildress as a co-author too.

> Actually, since ManyToManyRelatedManager is different from RelatedManager, I suggest a new stub class

👍 Seems like Django calls this `ManyRelatedManager`?

I think this will require some plugin changes as well, shouldn't be too hard. Because right now we assume everything is `RelatedManager`.

flaeppe:
Superseded by #1943

PR #710, opened 2021-09-08
Add SmallAutoField

c-py:
# Add support for SmallAutoField

`SmallAutoField` was [introduced in Django 3.0](https://docs.djangoproject.com/en/3.2/releases/3.0/). This PR adds a type definition for this field. I have copied the implementation for `BigAutoField`.

## Related issues

None, as far as I can see.


sobolevn (changes requested):
Thanks! We need a test for this feature 🙂

c-py:
@sobolevn I wasn't confident while adding this test. Let me know if the test should do something different.

sobolevn (approved):

sobolevn:
Thanks, exactly what we needed.

PR #422, opened 2020-07-11
Improve *HttpResponse.getvalue() types

danifus:
- Makes `StreamingHttpResponse.getvalue()` return type more specific
- Adds missing `HttpResponse.getvalue()` definition

PR #712, opened 2021-09-09
update typing on HttpResponse and StreamingHttpResponse

terencehonles:
This fixes a regression in https://github.com/typeddjango/django-stubs/pull/695

# Description

While the documentation for `HttpResponse` and `StreamingHttpResponse` *says* `content` and `streaming_content` [should be bytestrings](https://github.com/django/django/blob/ecf87ad513fd8af6e4a6093ed918723a7d88d5ca/django/http/response.py#L324-L327) or [an iterable of bytestrings](https://github.com/django/django/blob/0a28b42b1510b8093a90718bafd7627ed67fa13b/django/http/response.py#L395-L399) respectively, this is not what the API supports ([here](https://github.com/django/django/blob/ecf87ad513fd8af6e4a6093ed918723a7d88d5ca/django/http/response.py#L342-L362) and [here](https://github.com/django/django/blob/0a28b42b1510b8093a90718bafd7627ed67fa13b/django/http/response.py#L415-L427)) and there are tests which make sure the API supports more than bytestrings ([here](https://github.com/django/django/blob/0a28b42b1510b8093a90718bafd7627ed67fa13b/tests/cache/tests.py#L2250), [here](https://github.com/django/django/blob/0a28b42b1510b8093a90718bafd7627ed67fa13b/tests/i18n/urls.py#L8), etc). Before assigning `content` or `streaming_content` the code paths will call  `self.make_bytes` to coerce the value to bytes.


terencehonles:
It looks like the property change is also captured by https://github.com/typeddjango/django-stubs/pull/393, but the regression is still an issue that this PR only addresses AFAIK

django-stubs/http/response.pyi
```diff
@@ -101,8 +101,11 @@ class HttpResponse(HttpResponseBase):
 
 class StreamingHttpResponse(HttpResponseBase):
     content: Any
-    streaming_content: Iterator[bytes]
-    def __init__(self, streaming_content: Iterable[bytes] = ..., *args: Any, **kwargs: Any) -> None: ...
+    @property
```
sobolevn: Can you please add a test for this? This seems to be a common issue.

sobolevn (changes requested):
👍

django-stubs/http/response.pyi
terencehonles: yeah I'll get to that

django-stubs/http/response.pyi
```diff
@@ -10,6 +10,31 @@ from django.test.client import Client
 from django.urls import ResolverMatch
 from django.utils.datastructures import CaseInsensitiveMapping
 
+_T = TypeVar('_T')
+_U = TypeVar('_U')
+
+class _PropertyDescriptor(Generic[_T, _U]):
```
terencehonles, line 16: I used https://github.com/python/mypy/issues/3004#issuecomment-493741241 as a reason to test to see if I could implement the property with a descriptor. I'm not sure if there are any negative side effects, but properties *are* descriptors and it looks like the types resolve correctly now (my original code using properties did not work when adding the tests)

django-stubs/http/response.pyi
```diff
@@ -21,7 +46,7 @@ class ResponseHeaders(CaseInsensitiveMapping):
     def pop(self, key: str, default: Optional[str] = ...) -> str: ...
     def setdefault(self, key: str, value: str) -> None: ...
 
-class HttpResponseBase(Iterable[Any]):
+class HttpResponseBase:
```
terencehonles, line 49: This is not iterable directly, but its subclasses are

terencehonles:
This looks like this PR actually completely superceeds #393 since `getvalue()` is already typed from https://github.com/typeddjango/django-stubs/pull/422

terencehonles:
@sobolevn looks like the CI is now passing again, so a final review would be appreciated (just letting you know I'm done, but not a sense of urgency)

sobolevn (approved):
Awesome work! Thank you! 👍

PR #713, opened 2021-09-09
update all path related operations to have more accurate types

terencehonles:
# Description

Updates typing to better match https://github.com/django/django/blob/master/django/contrib/staticfiles/finders.py

_PathType was borrowed from https://github.com/typeddjango/django-stubs/blob/master/django-stubs/core/management/commands/compilemessages.pyi


terencehonles:
I accidentally created a PR against my own fork :facepalm: (I wish GitHub would stop making that so easy to not notice). This PR doesn't look like it has any merge conflicts, but I haven't looked at it since I wrote it in https://github.com/terencehonles/django-stubs/pull/1 and I am submitting this to hopefully merge it in, but also to run the CI against these changes.

django-stubs/contrib/staticfiles/finders.pyi
```diff
@@ -1,37 +1,45 @@
...
     storage: Storage = ...
     def __init__(self, storage: Optional[Storage] = ..., *args: Any, **kwargs: Any) -> None: ...
 
 class DefaultStorageFinder(BaseStorageFinder): ...
 
-def find(path: str, all: bool = ...) -> Optional[Union[List[str], str]]: ...
+@overload
+def find(path: str, all: Literal[True]) -> List[str]: ...
+@overload
+def find(path: str, all: Literal[False] = False) -> Optional[str]: ...
```
sobolevn: We use `= ...` to define default values.

sobolevn (changes requested):
Thanks! A simple test would be very helpful!

terencehonles:
> Thanks! A simple test would be very helpful!

@sobolevn done, I used https://github.com/django/django/blob/main/tests/staticfiles_tests/test_finders.py for inspiration :sweat_smile:

sobolevn (approved):
Looks great! Thanks!

PR #714, opened 2021-09-09
add `HashedFilesMixin.url` argument `force`

terencehonles:
# Description

This informs the type checker that `force` is valid keyword by widening https://github.com/typeddjango/django-stubs/blob/a1f3712c4334efa2cb461b727b673857c674d0e4/django-stubs/core/files/storage.pyi#L18 to [match the implementation](https://github.com/django/django/blob/0a28b42b1510b8093a90718bafd7627ed67fa13b/django/contrib/staticfiles/storage.py#L161)


sobolevn (approved):

sobolevn:
Thanks!

PR #717, opened 2021-09-22
Fix nullability for fields with non-None default. Fixes #444

### LINKED ISSUE #444: values_list returns Optional[Type] for field when field has a default value and is not nullable
# Bug report
## What's wrong

```
from django.db import models

class MyModel(models.Model):
    my_field = models.DecimalField(decimal_places=2, max_digits=10, default=0)

result = MyModel.objects.values_list('my_field', flat=True).get(id=1)
reveal_type(result) # Revealed type is: Optional[Decimal]
```

## How is that should be

The type of `result` is expected to be `Decimal` as the field is not nullable.

It seems to be due to the use of the `default` value for the field.
The culprit is probably this line: https://github.com/typeddjango/django-stubs/blob/31e795016f154309e675c85616f4f8af033c0860/mypy_django_plugin/django/context.py#L241
It should return `False` if there is a default instead.

There doesn't seem to be a test-case for this in: https://github.com/typeddjango/django-stubs/blob/31e795016f154309e675c85616f4f8af033c0860/test-data/typecheck/managers/querysets/test_values_list.yml

===

Edit:
It seems that the function should still return True if there is a default if `method` is `create` or `__init__`, but not in other cases.

dfn-certling:
# I have made things!

## Related issues

- Closes #444 

The original suggestion in the issue breaks `tests/typecheck/models/test_create.yml::when_default_for_primary_key_is_specified_allow_none_to_be_set` so there needs to be a separate handling when creating primary key fields with a default.

Running the unit tests fails one test that is already broken for me in the fork:

```
FAILED tests/typecheck/models/test_inheritance.yml::django_contrib_gis_base_model_mixin_inheritance
```

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -277,7 +277,9 @@ class DjangoContext:
         if method == "create":
             if isinstance(field, AutoField):
                 return True
-        if isinstance(field, Field) and field.has_default():
+            if isinstance(field, Field) and field.primary_key and field.has_default() and field.default is not None:
+                return True
+        if isinstance(field, Field) and field.has_default() and field.default is None:
             return True
         return nullable
 
--- a/tests/typecheck/managers/querysets/test_values_list.yml
+++ b/tests/typecheck/managers/querysets/test_values_list.yml
@@ -281,3 +281,22 @@
                 class Blog(models.Model):
                     num_posts = models.IntegerField()
                     text = models.CharField(max_length=100, blank=True)
+
+-   case: values_list_field_defaults
+    main: |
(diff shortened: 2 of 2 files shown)


sobolevn (approved):
Looks awesome! Thanks!

daniel-orca:
@sobolevn, @dfn-certling Can you please push those changes? it would be great

intgr (approved):
I rebased this and it looks good to me. Not sure why this hasn't been merged already.

intgr:
I see, #1020 has been merged which already solves this use case.

The fix there seems more correct (using field's `null=True` argument instead of `default=`). Closing this PR.

PR #724, opened 2021-10-04
fix `django.contrib.auth.models.Group.natural_key` no return type

### LINKED ISSUE #723: `django.contrib.auth.models.Group.natural_key` no return type
https://github.com/typeddjango/django-stubs/blob/fb125609818057dff8aad807637b3ca9cca5e194/django-stubs/contrib/auth/models.pyi#L37-L42

should be
```pyi
def natural_key(self) -> Tuple[str]: ...
```

star2000:
## Related issues

Closes #723


sobolevn (approved):

sobolevn:
Thanks a lot! 👍

Issue #700, opened 2021-09-04
mypy crashes with `TypeError: Object of type AnyType is not JSON serializable`


kalekseev:
# Bug report

mypy crashes at the end of successful check trying to write cache. Not sure if that's `django-stubs` problem and can't test if mypy works without `django-stubs` because of many type checks errors mypy doesn't write cache at the end.

## What's wrong

```
  File "/opt/hostedtoolcache/Python/3.9.6/x64/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/mypy/__main__.py", line 11, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "mypy/main.py", line 87, in main
  File "mypy/main.py", line 165, in run_build
  File "mypy/build.py", line 179, in build
  File "mypy/build.py", line 254, in _build
  File "mypy/build.py", line 2697, in dispatch
  File "mypy/build.py", line 3021, in process_graph
  File "mypy/build.py", line 3138, in process_stale_scc
  File "mypy/build.py", line 2288, in write_cache
  File "mypy/build.py", line 1476, in write_cache
  File "mypy/build.py", line 1428, in json_dumps
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/json/__init__.py", line 234, in dumps
    return cls(
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type AnyType is not JSON serializable
```

## System information

- OS:
- `python` version: 3.9.6
- `django` version: 3.2.7
- `mypy` version: 0.910
- `django-stubs` version: 1.9.0

sobolevn:
It can be an issue with custom `TypeInfo` metadata 😞 

We use it in several places:
[...]

Issue #681, opened 2021-07-29
Error messages refer to private _QuerySet class


syastrov:
# Bug report

Error messages on `master` now refer to the private `_QuerySet` class.
It's not friendly to users, and doesn't hint about how to properly write the annotations.

We should make the "repr" use either the `QuerySet` alias (if the 2 type parameters are both the same model) or the `django_stubs_ext.ValuesQuerySet` alias.

See https://github.com/python/mypy/issues/6501 for a possible hacky way to do this in the plugin.

intgr:
This has been solved as a side-effect of

* #2104

PR #725, opened 2021-10-05
Fix mypy cache of WithAnnotation types

kalekseev:
- Refs #700

This is a result of my research of that issue, the fix works for my projects but introduces other problems.

The fix is based on assumption that metadata may consist only of primitives, the fastest way to serialize dict of object is to use pickle, so I just pickle/unpickle metadata on set/get.
After that fix another problem was found, mypy doesn't support lookups of names with generics:
```
mypy/lookup.py", line 47, in lookup_fully_qualified
    assert key in names, "Cannot find component %r for %r" % (key, name)
AssertionError: Cannot find component 'WithAnnotations[project' for "django_stubs_ext.WithAnnotations[project.Model, TypedDict({'remind_after': Any, 'remind_before': Any})]"
```
The reason is https://github.com/python/mypy/blob/d1d431630d967923a9562b4dff436a509c2acb8b/mypy/lookup.py#L31 splits by '.' inside generic `[...]`. See how it complaints "Cannot find component `'WithAnnotations[project`" while it should search for `WithAnnotations[project.Model, TypedDict({'remind_after': Any, 'remind_before': Any})]`.
That problem was solved by replacing dots with `__` but I believe that will prevent users to write their own `WithAnnotations` typings...

Having said that I don't expect that PR to be merged in current state.

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -121,6 +121,7 @@ class DjangoContext:
             if "," in fullname:
                 # Remove second type arg, which might be present
                 fullname = fullname[: fullname.index(",")]
+            fullname = fullname.replace("__", ".")
 
         module, _, model_cls_name = fullname.rpartition(".")
         for model_cls in self.model_modules.get(module, set()):
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -512,9 +512,9 @@ def get_or_create_annotated_type(
         model_type = model_type.type.bases[0]
 
     if fields_dict is not None:
-        type_name = f"WithAnnotations[{model_type.type.fullname}, {fields_dict}]"
+        type_name = f"WithAnnotations[{model_type.type.fullname.replace('.', '__')}, {fields_dict}]"
     else:
-        type_name = f"WithAnnotations[{model_type.type.fullname}]"
+        type_name = f"WithAnnotations[{model_type.type.fullname.replace('.', '__')}]"
 
     annotated_typeinfo = helpers.lookup_fully_qualified_typeinfo(
         cast(TypeChecker, api), model_module_name + "." + type_name
--- a/tests/typecheck/managers/querysets/test_annotate.yml
+++ b/tests/typecheck/managers/querysets/test_annotate.yml
@@ -8,20 +8,20 @@
 
         unannotated_user = User.objects.get(id=1)
 
-        print(annotated_user.asdf) # E: "WithAnnotations[myapp.models.User, TypedDict({'foo': Any})]" has no attribute "asdf"
+        print(annotated_user.asdf) # E: "WithAnnotations[myapp__models__User, TypedDict({'foo': Any})]" has no attribute "asdf"
         print(unannotated_user.asdf) # E: "User" has no attribute "asdf"
 
         def func(user: Annotated[User, Annotations]) -> str:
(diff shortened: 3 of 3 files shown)


sobolevn:
Thanks a lot for the research! I will also try to see what we can do here.

syastrov:
It seems like my hack with making type names include `[` and `.` is causing issues. I think we can try to use `_` instead and live with the fact that error messages don't 100% match the type the user must use.
Otherwise, we could try to customize the "repr" used in error messages, using some hack. Similar to:
https://github.com/typeddjango/django-stubs/issues/681

Unfortunately I don't have time to look at this at the moment.

AustinScola:
Is there any update here or a plan to fix this? The `WithAnnotation` causes mypy to crash so often that I have resorted to clearing the entire cache first which results in really slow type checking. I would be willing to fix bug if someone could point me in the right direction.

kalekseev:
> It seems like my hack with making type names include `[` and `.` is causing issues. I think we can try to use `_` instead and live with the fact that error messages don't 100% match the type the user must use.

That was implemented in that PR, I rebased it.

sobolevn (approved):
Looks like a solid hack 😆 
Thanks, let's try this out.

AustinScola:
I'm still having problems with the caching of `WithAnnotations` on version 1.10.1

AustinScola:
Here is a slightly redacted trace if it helps:
```
  File "/usr/local/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/austin.s/src/xxx/.venvs/type_check/lib/python3.9/site-packages/mypy/__main__.py", line 34, in <module>
    console_entry()
  File "/Users/austin.s/src/xxx/.venvs/type_check/lib/python3.9/site-packages/mypy/__main__.py", line 12, in console_entry
  File "mypy/main.py", line 96, in main
  File "mypy/main.py", line 173, in run_build
  File "mypy/build.py", line 256, in _build
  File "mypy/build.py", line 2727, in dispatch
  File "mypy/build.py", line 3068, in process_graph
  File "mypy/build.py", line 3146, in process_fresh_modules
  File "mypy/build.py", line 2006, in fix_cross_refs
  File "mypy/fixup.py", line 25, in fixup_module
  File "mypy/fixup.py", line 92, in visit_symbol_table
  File "mypy/nodes.py", line 3058, in accept
  File "mypy/fixup.py", line 141, in visit_type_alias
  File "mypy/types.py", line 1065, in accept
  File "mypy/fixup.py", line 155, in visit_instance
  File "mypy/fixup.py", line 272, in lookup_fully_qualified_typeinfo
  File "mypy/lookup.py", line 47, in lookup_fully_qualified
AssertionError: Cannot find component "WithAnnotations[xxx__models__Report, TypedDict('xxx" for "django_stubs_ext.WithAnnotations[xxx__models__Report, TypedDict('xxx.api.report.RelatedInfo', {'response_count': builtins.int, 'viewed': builtins.bool, 'preview': Union[builtins.str, None]})]"
```

ashkop:
I'm also still having this issue with on version 1.13.0

PR #730, opened 2021-10-18
Make AddRelatedManagers look for "objects" on parent model

### LINKED ISSUE #652: Reverse relation is lost for abstract models with custom querysets
## What's wrong

The reverse relation to a concrete subclass of an abstract model that uses a custom queryset is not being populated by `django-stubs`.

A full set of test cases that illustrate the situation are in a commit [here](https://github.com/leamingrad/django-stubs/commit/00b7d75b42b8d757e2e2c5b6f0a1578b22f1db49) that I added on top of the latest master.

Since an example is worth a thousand words:

```python
from django.db import models

class AbstractLogQuerySet(models.QuerySet):
    pass

    pass

class AbstractLog(models.Model):
    objects = AbstractLogQuerySet.as_manager()
    class Meta:
        abstract=True

class TransactionLog(AbstractLog):
    transaction = models.ForeignKey(Transaction, on_delete=models.CASCADE, related_name="logs")

Transaction().logs
# myapp/models:13: error: "Transaction" has no attribute "logs"
```

Interestingly, the problem seems to be specific to querysets being used to make managers:
* An abstract model with the default manager works fine
* An abstract model with a custom manager works fine
* An abstract model with a custom manager and queryset fails as well

I'm happy to help with a fix, but to be honest I have no idea where to start.

## System information

I've seen this on multiple Python/Django/mypy/django-stubs versions but the test cases are written against:

- `python` version: 3.9.2
- `django` version:3.2.4
- `mypy` version: 0.910
- `django-stubs` version: HEAD

(Apologies if this is a duplicate - I looked and there seem to be similar issues but not this exact one)

RJPercival:
Previously, AddRelatedManagers would fail if a related model had inherited its `objects` field from a parent class. This would result in missing relation attributes. This is fixed by using `get()` instead of `names`; the former searches the MRO for the symbol, whereas the latter only looks for symbols declared directly on the class.

## Related issues

Closes #652

### DIFF
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -602,7 +602,6 @@
 
 
 -   case: test_foreign_key_from_superclass_inherits_correctly_when_also_inheriting_manager
-    expect_fail: yes
     main: |
         from myapp.models import MyUser, Book, Article, LibraryEntity
         reveal_type(Book().registered_by_user)  # N: Revealed type is "myapp.models.MyUser*"
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -601,6 +601,38 @@
                     pass
 
 
+-   case: test_foreign_key_from_superclass_inherits_correctly_when_also_inheriting_manager
+    expect_fail: yes
+    main: |
+        from myapp.models import MyUser, Book, Article, LibraryEntity
+        reveal_type(Book().registered_by_user)  # N: Revealed type is "myapp.models.MyUser*"
+        reveal_type(Article().registered_by_user)  # N: Revealed type is "myapp.models.MyUser*"
+
+        user = MyUser()
+        reveal_type(user.book_set) # N: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.Book]"
+        reveal_type(user.article_set) # N: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.Article]"
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+              from django.db import models
+              class MyUser(models.Model):
+                  pass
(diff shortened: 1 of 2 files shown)


tests/typecheck/fields/test_related.yml
```diff
@@ -626,6 +626,153 @@
...
+                from django.db import models
+                class Transaction(models.Model):
+                    pass
+                class AbstractLog(models.Model):
+                    class Meta:
+                        abstract=True
+                class TransactionLog(AbstractLog):
+                    transaction = models.ForeignKey(Transaction, on_delete=models.CASCADE, related_name="logs")
+
+                Transaction().logs
```
sobolevn: Can we please use `reveal_type()` here?

sobolevn: The same for all other cases of `.logs` access 🙂

sobolevn (commented):
Thanks a lot, let's see what CI will say.

RJPercival:
Looks like CI is broken @sobolevn.

tests/typecheck/fields/test_related.yml
RJPercival: I'm going to rewrite the tests to be consistent with the style of the other tests.

sobolevn:
I am also fixing the CI failure in https://github.com/typeddjango/django-stubs/pull/734 🙂

tests/typecheck/fields/test_related.yml
RJPercival: I've trimmed it down to just a single test case that fails prior to this fix and passes afterwards. It's almost identical to `test_foreign_key_from_superclass_inherits_correctly` except that it defines `objects` on `LibraryEntity`.

sobolevn (approved):
Thanks a lot for working on it! 👍

RJPercival:
No problem 🙂  Any idea when the next release is likely to be?

sobolevn:
We need to fix #725 before releasing a new one 😞

PR #731, opened 2021-10-18
Correct the type of FileField.storage

brianhelba:
This instance property can't be a callable. Although the `FileField` constructor allows a callable, it is immediately resolved to an instance of `Storage`.
See: https://github.com/django/django/blob/f5802a21c401b92764a9f3e2886144f3c5d77573/django/db/models/fields/files.py#L231-L235

---

Also, correct the type of `FieldFile.storage`
This instance property is copied directly from `FileField.storage` and should be
the same type.
See: https://github.com/django/django/blob/f5802a21c401b92764a9f3e2886144f3c5d77573/django/db/models/fields/files.py#L21


sobolevn:
Thanks!

PR #733, opened 2021-10-19
Upgrade dev-requirement for pytest-mypy-plugins

brianhelba:
This fixes broken CI.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,7 @@ coreapi==2.3.3
 gitpython==3.1.9
 pre-commit==2.13.0
 pytest==6.1.1
-pytest-mypy-plugins==1.7.0
+pytest-mypy-plugins==1.9.1
 psycopg2-binary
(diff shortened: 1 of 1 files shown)


brianhelba:
CI is breaking because of the release of [`setuptools` v58](https://setuptools.pypa.io/en/latest/history.html#v58-0-0), which dropped support for the `use_2to3` argument. This caused the long-dead `pystache` to fail to build.

`pytest-mypy-plugins==1.7.0` still requires `pystache`, whereas `pytest-mypy-plugins==1.9.1` no longer has this requirement.

brianhelba:
@sobolevn For some reason, the Python 3.8 test is failing. It looks like it might be an issue with `pytest-mypy-plugins`, but I don't understand the error message.

sobolevn:
Replaced with #734

sobolevn:
Thanks a lot for working on it! 👍

Issue #736, opened 2021-10-20
CharField(blank=True) treated as nullable


zach-waggoner:
## What's wrong

When I run mypy with this plugin enabled, it infers the type for CharFields with `blank=True` as `Optional[str]`.

## How is that should be

I believe this is incorrect, as `blank=True` only permits empty strings, not NULL. It seems like this was allegedly fixed in #39, but I'm still having the same issue. I noticed that if I remove the following lines from `DjangoContext.get_field_nullability`, it works:

```python
if not nullable and isinstance(field, CharField) and field.blank:
    return True
```

## System information

- OS:
- `python` version: 3.9.7
- `django` version: 3.2.9
- `mypy` version: 0.910
- `django-stubs` version: 1.9.0
- `django-stubs-ext` version: N/A

sobolevn:
@zach-waggoner PR is welcome!

flaeppe:
I cannot reproduce this, feel free to reopen with a breaking case.

PR #737, opened 2021-10-21
Don't treat CharFields with blank=True as nullable

zach-waggoner:
## Related issues

Refs #736

This was allegedly fixed in #39. I'm not sure when the issue got reintroduced, but I was able to resolve the issue described in #736 by removing these couple of lines.


zach-waggoner:
Will have to look at failing tests later when I have some time, or someone else can pick it up.

PR #739, opened 2021-10-22
Adjust django.core.signing.dumps and loads key type.

charettes:
An optional key can be provided to these functions.


sobolevn (approved):
Thanks! `None` is clearly broken.

PR #743, opened 2021-11-01
Implement stubs for views.decorators.common.no_append_slash

flaeppe:
[Relevant no_append_slash docs](https://docs.djangoproject.com/en/3.2/topics/http/decorators/#django.views.decorators.common.no_append_slash), decorator was introduced in `3.2`


sobolevn (approved):

sobolevn:
Thanks!

PR #749, opened 2021-11-12
Update MigrationExecutor stubs to use Sequence

biggerfisch:
There is no need for a specific List type here, using a higher type works fine and allows more working code to pass type-checking.

# I have made things!

I have a project that is having a hard time passing linting right now due to the invariant nature of `List`, despite everything working fine. Tested this fix against it locally and it seemed to pass with flying colors. `pre-commit` passes fine locally too.

Django code reference (I think?): https://github.com/django/django/blob/adb4100e58d9ea073ee8caa454bb7c885b6a83ed/django/db/migrations/executor.py#L22

## Related issues

None that I could find for this exact class, but https://github.com/typeddjango/django-stubs/pull/740 does the same thing for another section of stubs.


sobolevn (approved):
Thanks!

biggerfisch:
Thanks for merging!

PR #752, opened 2021-11-15
Fix BaseConnectionWrapper.execute_wrapper return type

ljodal:
# Fix BaseConnectionWrapper.execute_wrapper return type

This is defined as a context manager [in the Django source code][1]. When the return type is defined as an iterator mypy will complain when this is used as documented [in the Django docs][2].

[1]: https://github.com/django/django/blob/8d9827c06ce1592cca111e7eafb9ebe0153104ef/django/db/backends/base/base.py#L669-L679
[2]: https://docs.djangoproject.com/en/3.2/topics/db/instrumentation/


sobolevn (approved):
Thanks!

PR #762, opened 2021-11-30
Fix `QuerySet.create` and `Manager.create` annotation

kevinmarsh:
# I have made things!

Fix `QuerySet.create` and `Manager.create` annotation since `create` doesn't accept `*args` (only `**kwargs`)

- https://docs.djangoproject.com/en/3.2/ref/models/querysets/#create
- https://github.com/django/django/blob/stable/3.2.x/django/db/models/query.py#L446-L454

Calling `MyModel.objects.create("foo", bar=1)` raises `TypeError: create() takes 1 positional argument but 2 were given`

## Related issues

None


sobolevn (approved):
Looks reasonable!

sobolevn:
Thank you!

PR #767, opened 2021-12-08
Add ValidationError type hints

intgr:
# I have made things!

Added type hints to `ValidationError`. Some reverse engineered from Django code, others from documentation at https://docs.djangoproject.com/en/4.0/ref/forms/validation/#raising-validationerror


django-stubs/core/exceptions.pyi
```diff
@@ -24,15 +24,20 @@ class MiddlewareNotUsed(Exception): ...
...
-    params: Any = ...
-    def __init__(self, message: Any, code: Optional[str] = ..., params: Optional[Mapping[str, Any]] = ...) -> None: ...
+    error_dict: Dict[str, ValidationError] = ...
+    error_list: List[ValidationError] = ...
+    message: str = ...
+    code: Optional[str] = ...
+    params: Optional[Dict[str, Any]] = ...
+    def __init__(
+        self,
+        message: Union[str, ValidationError, Dict[str, Any], List[Any]],
```
intgr, line 37: Can be `str`, `ValidationError` or an arbitrarily nested combination of dicts/lists containing more dicts/lists/str/ValidationError.

Can this be expressed in the type system currently?

sobolevn: Not really 😞 
`Any` is good enough for now.

sobolevn (approved):
Thank you!

PR #768, opened 2021-12-08
Add more precise RunSQL, RunPython migration hints

intgr:
# I have made things!

The code and sql arguments of `django.db.migrations.RunSQL`, `django.db.migrations.RunPython` were not rigorously hinted.


sobolevn (approved):
Thank you! 👍

PR #769, opened 2021-12-09
More accurate template.Parser parse_until argument type

SmileyChris:
The `parse_until` argument type can be any iterable of strings, not just limited to a tuple.


sobolevn (approved):
Good catch! Thanks

PR #772, opened 2021-12-13
Set types for on_delete functions

### LINKED ISSUE #746: db/models/deletion.pyi protection types are unknown
# Bug report

## What's wrong

https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/deletion.pyi

Pyright complains that the types of the protection functions are not known, as they are defined like so:

```py
def CASCADE(collector, field, sub_objs, using): ...
def SET_NULL(collector, field, sub_objs, using): ...
def SET_DEFAULT(collector, field, sub_objs, using): ...
def DO_NOTHING(collector, field, sub_objs, using): ...
def PROTECT(collector, field, sub_objs, using): ...
def RESTRICT(collector, field, sub_objs, using): ...
def SET(value: Any) -> Callable: ...
```

## How is that should be

The argument types and return types of the functions should be specified so Pyright won't complain about the functions having unknown types. You can assign this issue to me and I'll create a pull request for it some time.

## System information

- OS: Linux
- `python` version: 3.9
- `django` version: 3.2
- `mypy` version: N/A
- `django-stubs-ext` version: N/A

w0rp:
Set types for on_delete functions so they can be provided as values to the on_delete attribute of ForeignKey in Pyright's strict mode.

Let me know if I need to add any tests for this. I'm not familiar with the test suite.

## Related issues

Closes #746

### DIFF
--- a/django-stubs/db/models/deletion.pyi
+++ b/django-stubs/db/models/deletion.pyi
@@ -4,14 +4,45 @@ from django.db import IntegrityError
 from django.db.models.base import Model
 from django.db.models.fields import Field
 from django.db.models.options import Options
+from django.db.models.query import QuerySet
 
-def CASCADE(collector, field, sub_objs, using): ...
-def SET_NULL(collector, field, sub_objs, using): ...
-def SET_DEFAULT(collector, field, sub_objs, using): ...
-def DO_NOTHING(collector, field, sub_objs, using): ...
-def PROTECT(collector, field, sub_objs, using): ...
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thank you!

sobolevn:
Can you please run `black` on your changes? It always complains 😒

w0rp:
I've fixed the strings with `black` now. I'll check back later and see if all of the checks pass.

sobolevn (approved):
Thanks again!

PR #773, opened 2021-12-13
Fix the get_field type for Pyright

### LINKED ISSUE #747: Options.get_field callable type is uknown
# Bug report

## What's wrong

https://github.com/typeddjango/django-stubs/blob/d5e45db79bcea6d6f4f45eacca3a2fe6e125e2e3/django-stubs/db/models/options.pyi#L114

The argument types and return type of the `Callable` for `get_field` are not specified, so Pyright complains with a `reportUnknownMemberType` error in its strictest mode.

## How is that should be

We should define the argument and return types for the Callable to be whatever they need to be, even `Any` if we can't determine the types. That will quell the error from Pyright. You can assign this issue to me, and I can create a pull request for it later.

## System information

- OS:
- `python` version: 3.9
- `django` version: 3.2
- `mypy` version: N/A
- `django-stubs-ext` version: N/A

w0rp:
This PR fixes the type of get_field so it won't report errors in Pyright's strictest type checking mode. Let me know if some tests are needed for this.

## Related issues

Closes #747


sobolevn (approved):
Ok, we can adjust to this style if needed! Thanks!

Any other places that we can change?
You can write some kind of script to convert everything at once, if needed.

w0rp:
That's just what I noticed for now.

PR #774, opened 2021-12-14
add id field to model class

eugeniyost:
# Fix <model_name> has no attribute "id" mypy error

When you try to access the id field of the model, mypy generates an error: <model_name> has no attribute 'id'
For example:
```
class Ex(models.Model):
    def __str__(self):
        return self.id
```


sobolevn:
Not all models have `id`. But all models have `pk`.

Injecting proper `pk` field is something our `mypy` plugin does for you 🤔

eugeniyost:
Most of the models contain an automatically generated id field, django supports both id and pk fields, maybe the plugin should also support both fields?

sobolevn:
> Most of the models contain an automatically generated id field

Unless you specify different `primary_key`.

> maybe the plugin should also support both fields?

This exactly what our plugin does. It checks if a model has default `id` field and adds it dynamically. The same way `django` does that.

MichaelAnckaert:
@sobolevn when you say "This exactly what our plugin does", does that mean that uses of `self.id` should not be flagged by mypy? Because I keep getting errors `"ModelName" has no attribute "id"` while I have the plugin configured.

sobolevn:
Yes, this should not be an error.

Source: https://github.com/typeddjango/django-stubs/blob/master/mypy_django_plugin/transformers/models.py#L96-L106
I've opened https://github.com/typeddjango/django-stubs/pull/777 to add a test for this.

PR #775, opened 2021-12-14
Update autoreload to Django 3.2

SmileyChris:
Updates the autoreload stub to match Django 3.2, keeping the current hints where appropriate.

(my motivation was that I was using the signals and pyright was complaining the import was invalid)


sobolevn (approved):
Thanks!

PR #776, opened 2021-12-14
Add Model._base_manager

SmileyChris:
Simple addition: `_base_manager` is missing off the model class (useful in some cases when you don't want to use the `_default_manager`)


sobolevn (approved):
Thank you!

PR #779, opened 2021-12-16
Testing that self.id don't work

eugeniyost:
# Testing that self.id don't work

Refs #774


tests/typecheck/fields/test_base.yml
```diff
@@ -180,3 +180,11 @@
         obj = MyModel()
 
         reveal_type(obj.small) # N: Revealed type is "builtins.int*"
+
+
+-  case: test_access_to_id_field_through_self_if_no_primary_key_defined
+   main: |
+        from django.db import models
+        class MyModel(models.Model):
```
sobolevn, line 188: Since I've never had models with 0 fields in Django, what is going to happen in runtime? Does it work? 🤔

eugeniyost: This works, in this model there will be one field - id. But for reproduce error, it doesn't matter how many fields there will be in the model if the field with `primary_key=True` undefined.

sobolevn: So, why does this work then? 🤔 
https://github.com/typeddjango/django-stubs/pull/777/files#diff-1f9420e92c89c34537838f68941bff315ab1c4057d5940cfb065caf8ab40b062R22

sobolevn:
@eugeniyost I've fixed the CI, can you please rebase your PR?

PR #794, opened 2021-12-28
remove generics from QuerySet type

### LINKED ISSUE #704: [Question] What is an alternative to isinstance(obj, QuerySet)?
I've recently got 

> Parameterized generics cannot be used with class or instance checks

with `django-stubs==1.9.0` for `isinstance(obj, QuerySet)` when I run mypy over the code.

What is the recommended way to do it?

jerch:
See #704 - currently it is not possible to deal with QuerySet class obj in python without getting a generics alert. By removing the `[_T, _T]` it works again without complaints.

## Related issues

Closes #704.

I have no idea about possible side effects of this change, so please someone with more knowledge about the typing internals of django-stubs would have to review this change.


sobolevn:
Maybe like this?

```python
class QuerySet(_QuerySet[_T, _T]): ...
```

Or:

```python
class _QuerySetAlias(_QuerySet[_T, _T]): ...

QuerySet = _QuerySetAlias
```

jerch:
@sobolevn Yes, trying your first variant (as it looks nicer to me). Still I wonder if the queryset base class has a model base class connection at all (have not looked up the django source). In the end it does not matter as long as the interfaces with types are announced correctly.

sobolevn (changes requested):
Thanks! We still need a test case for #704

jerch:
@sobolevn Are those explicit tests against `_QuerySet[XY, XY]` needed or can they be replaced by `QuerySet[XY]`? Would that lose test/type information?

sobolevn:
> Are those explicit tests against _QuerySet[XY, XY] needed

Yes, we need them to test our internal representation of `QuerySet`, which is needed for things like `.annotate()` to work.

jerch:
@sobolevn Not sure if I can help you with the test cases. I am getting tons of errors, if I try to run pytest locally in a venv with python 3.10 (bootstrapped similar to the CI setup), like these:
```
/home/jerch/Dokumente/github/django-stubs/tests/typecheck/test_config.yml:5: 
E   Actual:
E     ../../home/jerch/Dokumente/github/django-stubs/django-stubs/db/models/deletion.pyi:12: error: Bad number of arguments for type alias, expected: 2, given: 1 (diff)

<here like 30 more errors of the same type in other places>

E     main:3: note: Revealed type is "builtins.int*" (diff)
E     main:4: note: Revealed type is "django.contrib.auth.models.User*" (diff)
E     main:5: note: Revealed type is "django.db.models.manager.Manager[myapp.models.MyModel]" (diff)
E     myapp/models:6: note: Revealed type is "django.contrib.auth.models.User*" (diff)
E   Expected:
E     main:3: note: Revealed type is "builtins.int*" (diff)
E     main:4: note: Revealed type is "django.contrib.auth.models.User*" (diff)
E     main:5: note: Revealed type is "django.db.models.manager.Manager[myapp.models.MyModel]" (diff)
E     myapp/models:6: note: Revealed type is "django.contrib.auth.models.User*" (diff)
E     E: main:3: note: Revealed type is "builtins.int*"...
E     A: ../../home/jerch/Dokumente/github/django-stubs/django-stubs/db/models/de...
E        ^

**Comment by intgr** [comment]
I tried this change and it causes new errors in my code. For example:

```python
            MyModel.objects.annotate(
                last_status=Subquery(
                    OtherModel.objects.values("last_status")
                )
```

> error: Argument 1 to "Subquery" has incompatible type "_QuerySet[OtherModel, TypedDict({'last_status': str})]"; expected "QuerySet[Any]"

Because `_QuerySet` and `QuerySet` are now distinct types, `_QuerySet` cannot be converted to its subclass `QuerySet`.

I think some changes are needed in the plugin code as well.

jerch:
@intgr Thats the reason why I asked if `QuerySet[XY]` could replace `_QuerySet[XY, XY]`. But your use case already answers that as a clear no (see the different types in `_QuerySet[OtherModel, TypedDict({'last_status': str})]`. Sadly I cannot get the test cases working locally, thus cannot do much to get things fixed.

intgr:
This seems redundant now that #794 has been merged?

Feel free to re-open if I'm missing something.

PR #797, opened 2021-12-28
Export Datasource, Driver from gdal

corydolphin:
It looks like gdal default exports are not consistent with Django, notably missing DataSource and Driver amongst others.

https://github.com/django/django/blob/main/django/contrib/gis/gdal/__init__.py#L29

### DIFF
--- a/django-stubs/contrib/gis/gdal/__init__.pyi
+++ b/django-stubs/contrib/gis/gdal/__init__.pyi
@@ -1,3 +1,5 @@
+from django.contrib.gis.gdal.datasource import DataSource
+from django.contrib.gis.gdal.driver import Driver
 from django.contrib.gis.gdal.error import GDALException as GDALException
 from django.contrib.gis.gdal.error import SRSException as SRSException
 from django.contrib.gis.gdal.error import check_err as check_err


django-stubs/contrib/gis/gdal/__init__.pyi
```diff
@@ -1,3 +1,6 @@
+from django.contrib.gis.gdal.datasource import DataSource
```
sobolevn, line 1:
```suggestion
from django.contrib.gis.gdal.datasource import DataSource as DataSource
```

django-stubs/contrib/gis/gdal/__init__.pyi
```diff
@@ -1,3 +1,6 @@
...
+from django.contrib.gis.gdal.driver import Driver
```
sobolevn, line 2:
```suggestion
from django.contrib.gis.gdal.driver import Driver as Driver
```

sobolevn (changes requested):
Thanks! We need explicit re-export for mypy to be happy.

corydolphin:
> Thanks! We need explicit re-export for mypy to be happy.

Thanks for the quick response. Sorry I missed that -- thank you for proposing the changes inline. Updated.

sobolevn (approved):
Please, make sure `isort` and `black` are happy 🙂 

Thank you!

corydolphin:
> Please, make sure `isort` and `black` are happy 🙂
> 
> Thank you!

Fixed! Sorry for the delay and missing that.

sobolevn:
Can you please rebase your PR?

It should be fine after https://github.com/typeddjango/django-stubs/commit/ed68fbc99f372bd6884433f44eb64748dfd4f228

sobolevn:
Thanks again!

PR #798, opened 2021-12-28
Fix Feed.get_object signature

Crocmagnon:
get_object's default implementation returns None but subclasses are allowed to override this method and make it return anything.

The returned object would then be passed to other methods to publish different data for different URL parameters.

https://docs.djangoproject.com/en/4.0/ref/contrib/syndication/

By the way, I think there is room for improvement on this stub class:
* make it generic (two types: one for `item`, the other for `obj`)
* add stubs for all methods, even the "dynamic" ones


sobolevn (approved):
Thanks!

sobolevn:
> By the way, I think there is room for improvement on this stub class:

Please, go ahead!

PR #800, opened 2021-12-29
Update syndication Feed class

Crocmagnon:
Followup of #798


django-stubs/contrib/syndication/views.pyi
```diff
@@ -11,17 +12,160 @@ def add_domain(domain: str, url: str, secure: bool = ...) -> str: ...
...
 class FeedDoesNotExist(ObjectDoesNotExist): ...
 
-class Feed:
-    feed_type: Any = ...
-    title_template: Any = ...
-    description_template: Any = ...
+Item = TypeVar("Item")
+Object = TypeVar("Object")
+
+class Feed(Generic[Item, Object]):
```
Crocmagnon: I can't remember whether that's the correct way of defining a generic class

django-stubs/contrib/syndication/views.pyi
```diff
@@ -11,17 +12,160 @@ def add_domain(domain: str, url: str, secure: bool = ...) -> str: ...
...
+    feed_type: Type[SyndicationFeed] = ...
```
Crocmagnon, line 17: This should be set to a subclass (not instance) of `SyndicationFeed`, is this the right way to do it?

django-stubs/contrib/syndication/views.pyi
```diff
@@ -11,17 +12,160 @@ def add_domain(domain: str, url: str, secure: bool = ...) -> str: ...
...
-    def item_link(self, item: Model) -> str: ...
-    def item_enclosures(self, item: Model) -> List[Enclosure]: ...
-    def feed_extra_kwargs(self, obj: None) -> Dict[Any, Any]: ...
-    def item_extra_kwargs(self, item: Model) -> Dict[Any, Any]: ...
-    def get_object(self, request: WSGIRequest, *args: Any, **kwargs: Any) -> Any: ...
+    def get_object(self, request: WSGIRequest, *args: Any, **kwargs: Any) -> Optional[Object]: ...
+    # One of the following three is required
+    def title(self, obj: Object) -> str: ...
+    def title(self) -> str: ...
+    title: str = ...
```
Crocmagnon: These are tricky. The docs says one of them is required/optional, I don't know how to represent that here.

Crocmagnon: Tests fail because of the redefinitions here and below.

django-stubs/contrib/syndication/views.pyi
```diff
@@ -11,17 +12,160 @@ def add_domain(domain: str, url: str, secure: bool = ...) -> str: ...
 
 class FeedDoesNotExist(ObjectDoesNotExist): ...
 
-class Feed:
-    feed_type: Any = ...
-    title_template: Any = ...
-    description_template: Any = ...
+Item = TypeVar("Item")
```
sobolevn: We use `_` prefix to indicate that this is not a real object.

django-stubs/contrib/syndication/views.pyi
sobolevn: This one looks correct to me 🙂

django-stubs/contrib/syndication/views.pyi
sobolevn: Yes!

django-stubs/contrib/syndication/views.pyi
sobolevn: Maybe `title: Any`?

sobolevn: We use this hack when we cannot represent something with types.
Or we can use `title: Union[str, Callable[[Feed], str], Callable[[Feed, Object], str]]`

Crocmagnon: I like the Union approach, I’ll come up with something in the next couple of days 👍🏻

Crocmagnon:
Thank you for your help!

django-stubs/contrib/syndication/views.pyi
Crocmagnon: What's the recommended approach for writing unit tests for this kind of types?

sobolevn (changes requested):
Thank you! Can you please add a simple test case for it?
Because this looks quite complex, we don't want to break it in the future.

Crocmagnon:
Well, unfortunately the precise typing I was going with was not usable on a real subclass and I couldn't find a way to make it work.

Seems like mypy doesn't appreciate `Unions` of `Callable`s as potential methods. I went with Any, as it at least lets people know that these exist even if the type doesn't give much information.

How should I test it?

sobolevn (approved):
`Any` is fine as is! Thanks!

PR #801, opened 2021-12-31
Fix type for templatetag

hambro:
I have made things!

_Hi, thanks for submitting a Pull Request. We appreciate it._
No thank you for all your hard work. I'm mostly a user of the service and all I got to contribute is this tiny change. 

The type for the arg `arg` to [naturalday](https://github.com/django/django/blob/a21a63cc288ba51bcf8c227a49de6f5bb9a72cc3/django/contrib/humanize/templatetags/humanize.py#L143) is off.
`def naturalday(value, arg=None):...` passes `arg` to `defaultfilters.date(value, arg)` which is already typed to `Optional[str]`, but `arg` to `naturalday` is just `arg = None` in django-stubs ([source](https://github.com/typeddjango/django-stubs/blob/3b02222beba43bcb52313608926f0803dee597f3/django-stubs/template/defaultfilters.pyi#L57)). 

So my change is just tweaking the type of the arg.


sobolevn (approved):
Looks good, thank you!

P.S. Happy New Year!

Issue #709, opened 2021-09-08
Custom manager causes phantom errors and sometimes crash


christianbundy:
# Bug report

## What's wrong

Still working on a minimal repro, but here's my stack trace:

```
Traceback (most recent call last):
  File "/.venv/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/.venv/lib/python3.9/site-packages/mypy/__main__.py", line 11, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "mypy/main.py", line 87, in main
  File "mypy/main.py", line 165, in run_build
  File "mypy/build.py", line 179, in build
  File "mypy/build.py", line 254, in _build
  File "mypy/build.py", line 2697, in dispatch
  File "mypy/build.py", line 3021, in process_graph
  File "mypy/build.py", line 3137, in process_stale_scc
  File "mypy/errors.py", line 565, in file_messages
  File "mypy/errors.py", line 538, in format_messages
IndexError: list index out of range
```

I have a `FooManager`:

```python3
from django.db import models

class FooManager(models.Manager):

    def create_with_foo(self, *args, **kwargs) -> models.Model: # Line 7
        pass

    def save_foo(self, model: models.Model) -> None: # Line 25
        pass
```

It's often used on models like this:

```python3
objects = FooManager()
```

For some reason, three _completely unrelated_ files are getting these errors:

```
some_file.py:7: error: Name "models" is not defined  [name-defined]
    )
    ^
some_file.py:25: error: Name "models" is not defined  [name-defined]
            does_not_matter_anything_could_be_here()
```

I tried adding whitespace to the top of `some_file.py` but the line numbers didn't change. Odd. I looked through my project for files where lines 7 and 25 referenced `models` and found my `FooManager`. I added 1 line to the top of `FooManager` and the mypy errors incremented by 1. I added 100 lines to the top of `FooManager` and got this sweet stack trace.

## How is that should be

There should be no errors.

## System information

- OS:
- `python` version: 3.9.6
- `django` version: 3.1
- `mypy` version: 0.910
- `django-stubs` version: 1.8.0 
- `django-stubs-ext` version: 0.3.1

antonagestam:
I'm running into this too. I realized probably all of our models that define custom managers using `objects = BaseManager.from_queryset(...)()` reveals `Any` for `reveal_type(Models.objects)`, and therefor also for `Model.objects.{get,all,...}()`.

[...]

PR #802, opened 2022-01-03
Repro test case for phantom errors in #709

flaeppe:
I found a repro case for some(?) phantom errors mentioned in https://github.com/typeddjango/django-stubs/issues/709#issuecomment-948732864. Not at all sure about why they happen like this, or how to resolve them properly. Thought it might be useful for someone else with better insight.

Worth noting from the error is that it only seem to happen for return types imported e.g. `from typing import ...` as there's no error for the queryset methods having `-> "ModelQuerySet":` or `-> str:`

## Related issues

Refs #709

---

Here's the error output of the failing test (`from_queryset_queryset_imported_from_other_module`):

```
______________ from_queryset_queryset_imported_from_other_module _______________
/home/runner/work/django-stubs/django-stubs/tests/typecheck/managers/querysets/test_from_queryset.yml:31: 
E   Actual:
E     ...
E     main:5: note: Revealed type is "Any"          (diff)
E    main:6: note: Revealed type is "builtins.str"
E    main:7: note: Revealed type is "myapp.querysets.ModelQuerySet"
E    main:8: note: Revealed type is "myapp.querysets.ModelQuerySet[myapp.models.MyModel*]"
E     myapp/models:10: error: Name "Iterable" is not defined (diff)
E     myapp/models:10: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Iterable") (diff)
E   Expected:
E     ...
E     main:5: note: Revealed type is "typing.Iterable[Any]" (diff)
E    main:6: note: Revealed type is "builtins.str"
E    main:7: note: Revealed type is "myapp.querysets.ModelQuerySet"
E    main:8: note: Revealed type is "myapp.querysets.ModelQuerySet[myapp.models.MyModel*]"
E     E: ...te: Revealed type is "typing.Iterable[Any]"
E     A: ...te: Revealed type is "Any"
```

### DIFF
--- a/tests/typecheck/managers/querysets/test_from_queryset.yml
+++ b/tests/typecheck/managers/querysets/test_from_queryset.yml
@@ -22,6 +22,46 @@
                 class MyModel(models.Model):
                     objects = NewManager()
 
+-   case: from_queryset_queryset_imported_from_other_module
+    main: |
+        from myapp.models import MyModel
+        reveal_type(MyModel().objects)  # N: Revealed type is "myapp.models.NewManager[myapp.models.MyModel]"
+        reveal_type(MyModel().objects.get())  # N: Revealed type is "myapp.models.MyModel*"
+        reveal_type(MyModel().objects.queryset_method())  # N: Revealed type is "myapp.querysets.ModelQuerySet"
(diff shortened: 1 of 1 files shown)


flaeppe:
Superseded by #820

PR #806, opened 2022-01-05
Allow expressions in UniqueConstraint

### LINKED ISSUE #805: Expressions on UniqueConstraint aren't supported
# Bug report

## What's wrong

Django 4.0 added support for a [positional `expressions` argument](https://docs.djangoproject.com/en/4.0/ref/models/constraints/#expressions), but it doesn't look like it's supported here yet.

Example:

```
class User:
    ...
    class Meta:
        constraints = [
            models.UniqueConstraint(Upper("email"), name="email_key"),
        ]
```

Output:

```
mypy .
xxx/models.py:74: error: Too many positional arguments for "UniqueConstraint"
xxx/models.py:74: error: Argument 1 to "UniqueConstraint" has incompatible type "Upper"; expected "Sequence[str]"
xxx/migrations/0001_initial.py:111: error: Too many positional arguments for "UniqueConstraint"
xxx/migrations/0001_initial.py:112: error: Argument 1 to "UniqueConstraint" has incompatible type "Upper"; expected "Sequence[str]"
```

## How is that should be

Shouldn't output errors.

## System information

- OS: Mac OS
- `python` version: 3.10
- `django` version: 4.0
- `mypy` version: 0.930

knyghty:
## Related issues

Closes #805

I'm not sure I did it right, but both using `expressions` and `fields` works with this at least, e.g.:

```python
class MyModel(models.Model):
    ...
    class Meta:
        constraints = [
            models.UniqueConstraint(Upper("foo"), name="foo_upper_key"),
            models.UniqueConstraint(fields=["bar"], name="bar_key"),
        ]
```

### DIFF
--- a/django-stubs/db/models/constraints.pyi
+++ b/django-stubs/db/models/constraints.pyi
@@ -1,8 +1,9 @@
 import enum
-from typing import Any, Optional, Sequence, Tuple, Type, TypeVar
+from typing import Any, Optional, Sequence, Tuple, Type, TypeVar, Union
 
 from django.db.backends.base.schema import BaseDatabaseSchemaEditor
 from django.db.models.base import Model
+from django.db.models.expressions import Combinable
 from django.db.models.query_utils import Q
 
 _T = TypeVar("_T", bound="BaseConstraint")
@@ -27,9 +28,15 @@ class CheckConstraint(BaseConstraint):
     def __init__(self, *, check: Q, name: str) -> None: ...
 
 class UniqueConstraint(BaseConstraint):
+    expressions: Tuple[Combinable]
     fields: Tuple[str]
     condition: Optional[Q]
     deferrable: Optional[Deferrable]
     def __init__(
-        self, *, fields: Sequence[str], name: str, condition: Optional[Q] = ..., deferrable: Optional[Deferrable] = ...
+        self,
+        *expressions: Union[str, Combinable],
+        fields: Optional[Sequence[str]] = ...,
+        name: str,
+        condition: Optional[Q] = ...,
+        deferrable: Optional[Deferrable] = ...,
     ) -> None: ...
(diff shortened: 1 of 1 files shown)


django-stubs/db/models/constraints.pyi
```diff
@@ -27,9 +28,15 @@ class CheckConstraint(BaseConstraint):
     def __init__(self, *, check: Q, name: str) -> None: ...
 
 class UniqueConstraint(BaseConstraint):
+    expressions: Tuple[Combinable]
```
sobolevn: `Tuple[Combinable, ...]`?

knyghty: Yes, I think so, I wasn't aware of this syntax, thanks.

I assume this should also be changed for `fields`? I can do this here, but it's not completely relevant to this change.

django-stubs/db/models/constraints.pyi
```diff
@@ -27,9 +28,15 @@ class CheckConstraint(BaseConstraint):
     def __init__(self, *, check: Q, name: str) -> None: ...
 
 class UniqueConstraint(BaseConstraint):
+    expressions: Tuple[Combinable, ...]
     fields: Tuple[str]
```
sobolevn: Yes, the same for fields 🙂

knyghty: 👍🏻

sobolevn (approved):
Thanks!

linusg:
For anyone else finding this and wondering why it still doesn't work in 1.12.0: unfortunately this change was reverted/commented out in #909 :confused:

https://github.com/typeddjango/django-stubs/blob/c0414c7b025b77caf9f72dbaf9e8471e11aad085/django-stubs/db/models/constraints.pyi#L30-L48

PR #812, opened 2022-01-07
Fix annotation of ModelAdmin.get_object

rolandcrosby-check:
The type annotation in django-stubs indicated that the type of the `from_field` argument to `ModelAdmin.get_object` should be `None`. In fact, `None` is the default value of that argument. If the argument is not `None`, [it is passed to a model's `_meta.get_field` method][impl], which takes an argument of type `Union[Callable[..., Any], str`.

[impl]: https://github.com/django/django/blob/ac5cc6cf01463d90aa333d5f6f046c311019827b/django/contrib/admin/options.py#L767


sobolevn (approved):
Thanks!

PR #816, opened 2022-01-13
Fix extra_context argument type in admin views

rolandcrosby-check:
The `extra_context` argument to `ModelAdmin.changeform_view`,
`add_view`, `change_view`, `changelist_view`, `delete_view`, and
`history_view` is used to pass extra context variables to a template, so
all of these arguments should have type `Optional[Dict[str, Any]]`.
Instead, these are currently typed as `Optional[Dict[str, bool]]`,
`Optional[Dict[str, str]]`, or `None`.


sobolevn (approved):
Thank you!

PR #817, opened 2022-01-14
Make SomeModel._default_manager return a BaseManager[SomeModel] instead of BaseManager[Model]

emma-blossom:
This changes the type annotations for _default_manager (and _base_manager) such that the returned manager's model is the specific model the manager belongs to instead of django.db.models.base.Model


django-stubs/utils/functional.pyi
```diff
@@ -62,8 +62,11 @@ def partition(
...
-    def __init__(self, method: Optional[Callable] = ...) -> None: ...
-    def __get__(self, instance: Any, cls: Optional[type] = ...) -> Any: ...
-    def getter(self, method: Callable) -> classproperty: ...
+_Get = TypeVar("_Get")
+_Self = TypeVar("_Self")
+
+class classproperty(Generic[_Get]):
+    fget: Optional[Callable[[Type[_Self]], _Get]] = ...
+    def __init__(self, method: Optional[Callable[[Type[_Self]], _Get]] = ...) -> None: ...
+    def __get__(self, instance: Any, cls: Type[_Self] = ...) -> _Get: ...
```
sobolevn: From what I see `__get__` is usually an `overload` of two functions.
You can take a look here: https://github.com/python/typeshed/

tests/typecheck/managers/test_managers.yml
```diff
@@ -48,7 +48,7 @@
                 class Base(Generic[_T]):
                     def __init__(self, model_cls: Type[_T]):
                         self.model_cls = model_cls
-                        reveal_type(self.model_cls._default_manager)  # N: Revealed type is "django.db.models.manager.BaseManager[django.db.models.base.Model]"
+                        reveal_type(self.model_cls._default_manager)  # N: Revealed type is "django.db.models.manager.BaseManager[_T`1]"
```
sobolevn, line 51: Let's also test `._base_manager` of some `Model`, so we can ensure that `_T` is correct.

sobolevn (changes requested):
Thank you! Great feature, please, see my comments

tests/typecheck/managers/test_managers.yml
emma-blossom: I'm super unfamiliar with these types of tests. I hope the new test makes sense!

sobolevn (approved):
Thank you!

clouds56:
I'm using pylance and found this does not work.
```python
user: Users = Users._default_manager.get(pk=user_id)
```
shows error:
> (property) _default_manager: BaseManager[_Self@_default_manager]
Expression of type "_Self@_default_manager" cannot be assigned to declared type "Users"
  "Model*" is incompatible with "Users" Pylance(reportGeneralTypeIssues)

any ideas?

Issue #738, opened 2021-10-22
No way to achieve typing of managers built from querysets


antonagestam:
Currently there seems to be no way to get non-`Any` typing for `Model.objects` and by extension also `Model.objects.get()`, `Model.objects.all()` et al, for models that specify a custom manager built from a `QuerySet`. I'm submitting this as a tracking-issue, even though there are individual issues for some of the underlying problems, I thought it'd be nice to have a unified discussion about how to find a way forward here, whether it's a workaround or a need of a fix in the stubs or in the extension.

### Option 1

Managers built dynamically within the model class body, just seem to not work and induces `Any`.

```python
CustomQSSelf = TypeVar("CustomQSSelf", bound="CustomQS")

class CustomQS(QuerySet["A"]):
    def custom_filter(self: CustomQSSelf) -> CustomQSSelf:
        return self.filter(pk=1)

class A(Model):
    objects = BaseManager.from_queryset(CustomQS)()

reveal_type(A.objects)  # -> Any
reveal_type(A.objects.get())  # -> Any
reveal_type(A.objects.all())  # -> Any
```

### Option 2

Moving the manager construction to outside of the model body. [This induces very strange phantom errors](https://github.com/typeddjango/django-stubs/issues/709), as well as errors complaining that the "self annotations" on the QuerySet are unbound.

```python
CustomQSSelf = TypeVar("CustomQSSelf", bound="CustomQS")

class CustomQS(QuerySet["A"]):
    def custom_filter(self: CustomQSSelf) -> CustomQSSelf:
        return self.filter(pk=1)

AManager = BaseManager.from_queryset(CustomQS)

class A(Model):
    objects = AManager()
```

```
error: Type variable "...CustomQSSelf" is unbound  [valid-type]
```

### Option 3

Using `QuerySet.as_manager()`. I tried this even though it seems generally recommended against in other issues, and to have less support than `BaseManager.from_queryset()`. As expected this also yields `Any`, whether or not it's defined inside the model body.

```python
CustomQSSelf = TypeVar("CustomQSSelf", bound="CustomQS")

class CustomQS(QuerySet["A"]):
    def custom_filter(self: CustomQSSelf) -> CustomQSSelf:
        return self.filter(pk=1)

class A(Model):
    objects = CustomQS.as_manager()

reveal_type(Lease.objects.get())  # -> Any
reveal_type(Lease.objects.all())  # -> django.db.models.manager.Manager[Any]
```

[...]

PR #820, opened 2022-01-15
Set custom queryset methods as manager attrs instead of method copies

### LINKED ISSUE #709: Custom manager causes phantom errors and sometimes crash

flaeppe:
Instead of copying methods over from a `QuerySet` passed to a base manager when invoking `<BaseManager>.from_queryset`, any `QuerySet` methods are declared as attributes on the dynamically created manager.

This allows us to properly lookup any `QuerySet` method types via a `get_attribute_hook` and will thus remove disorienting phantom errors occuring from mypy trying to resolve types only existing in the module where the _original_ (and real) queryset method was declared.

Consider _the currently breaking_ case of having the following model and queryset modules

```python
# myapp/querysets.py
from django.db import models
from typing import TYPE_CHECKING, Iterable

if TYPE_CHECKING:
    from .models import MyModel

    def queryset_method(self) -> Iterable[int]:
        return [1]
```

```python
# myapp/models.py
from django.db import models
from .querysets import MyQuerySet

NewManager = BaseManager.from_queryset(MyQuerySet)
class MyModel(models.Model):
    objects = NewManager()
```

Now, to elaborate a bit more, if we expand the `NewManager` type that `django-stubs` plugin will dynamically create (without the changes seen in this PR), we get:

```python
# myapp/models.py
from django.db import models
from .querysets import MyQuerySet

...

class NewManager(BaseManager):
    def queryset_method(self) -> Iterable[int]:
        ...
```

That happens via _copying_ any method definitions from the passed in queryset (`MyQuerySet` in this example) on to the manager type. Copying becomes troublesome when queryset and model live in different modules and don't share _all_ definitions. As `Iterable` isn't imported in `myapp/models.py` mypy yields an error:

```
error: Name "Iterable" is not defined
```

Which makes sense.

Now, to the fix included here. Instead of _copying the method_, we instead declare each queryset method as a class attribute of the new manager type (in this example `NewManager`). To display the difference, this is what that will look like if we unwrap the dynamically created manager:

```python
# myapp/models.py
from django.db import models
from .querysets import MyQuerySet

...

class NewManager(BaseManager):
    queryset_method = MyQuerySet.queryset_method
```

Mypy is able to resolve that attribute properly without any definition errors. And from my (very shallow) investigation of Django's [.from_queryset](https://github.com/django/django/blob/0a17666045de6739ae1c2ac695041823d5f827f7/django/db/models/manager.py#L103-L110)(more exactly the 3-arg `type` call) implementation, this representation (if we disregard or unwrap the decorator they implement) aligns better with the truth (feel free to correct me if I'm wrong).

## Related issues

Refs #709 (Not sure the fix fully resolves this issue)
Fixes https://github.com/typeddjango/django-stubs/issues/709#issuecomment-948732864
Refs #738
Supersedes #802 

---

There was a bit of a chain reaction implementing this change, mainly in regards to reversed managers. I'm having a bit of an expectation that there's some edits there that could allow the dynamically created `<Model>_<RelatedModel>_RelatedManager` named manager types to be replaced with a proper traversal via `RelatedModel._default_manager.queryset_method`(allowing us to ditch that type creation). But I've tried to leave reversed managers with as few edits as possible and dig in to that thought another time.

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -34,7 +34,7 @@ from mypy.plugin import (
     MethodContext,
     SemanticAnalyzerPluginInterface,
 )
-from mypy.plugins.common import add_method
+from mypy.plugins.common import add_method_to_class
 from mypy.semanal import SemanticAnalyzer
 from mypy.types import AnyType, CallableType, Instance, NoneTyp, TupleType
 from mypy.types import Type as MypyType
@@ -383,7 +383,9 @@ def copy_method_to_another_class(
             return
 
         arguments, return_type = build_unannotated_method_args(method_node)
-        add_method(ctx, new_method_name, args=arguments, return_type=return_type, self_type=self_type)
+        add_method_to_class(
+            semanal_api, ctx.cls, new_method_name, args=arguments, return_type=return_type, self_type=self_type
+        )
         return
 
     method_type = method_node.type
@@ -421,7 +423,9 @@ def copy_method_to_another_class(
         argument.set_line(original_argument)
         arguments.append(argument)
 
-    add_method(ctx, new_method_name, args=arguments, return_type=return_type, self_type=self_type)
+    add_method_to_class(
+        semanal_api, ctx.cls, new_method_name, args=arguments, return_type=return_type, self_type=self_type
+    )
 
 
 def add_new_manager_base(api: SemanticAnalyzerPluginInterface, fullname: str) -> None:
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -24,7 +24,10 @@ import mypy_django_plugin.transformers.orm_lookups
 from mypy_django_plugin.django.context import DjangoContext
 from mypy_django_plugin.lib import fullnames, helpers
 from mypy_django_plugin.transformers import fields, forms, init_create, meta, querysets, request, settings
-from mypy_django_plugin.transformers.managers import create_new_manager_class_from_from_queryset_method
+from mypy_django_plugin.transformers.managers import (
+    create_new_manager_class_from_from_queryset_method,
+    resolve_manager_method,
+)
 from mypy_django_plugin.transformers.models import (
     handle_annotated_type,
     process_model_class,
@@ -302,6 +305,7 @@ class NewSemanalDjangoPlugin(Plugin):
                 mypy_django_plugin.transformers.orm_lookups.typecheck_queryset_filter,
                 django_context=self.django_context,
             )
+
         return None
 
     def get_base_class_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
@@ -330,6 +334,13 @@ class NewSemanalDjangoPlugin(Plugin):
             return partial(request.set_auth_user_model_as_type_for_request_user, django_context=self.django_context)
         if info and info.has_base(fullnames.ABSTRACT_USER_MODEL_FULLNAME) and attr_name in ("is_staff", "is_active"):
             return partial(set_auth_user_model_boolean_fields, django_context=self.django_context)
+        if (
+            info
+            and info.has_base(fullnames.BASE_MANAGER_CLASS_FULLNAME)
+            and class_name in self._get_current_manager_bases()
+        ):
+            return partial(resolve_manager_method, django_context=self.django_context)
+
         return None
 
     def get_type_analyze_hook(self, fullname: str) -> Optional[Callable[[AnalyzeTypeContext], MypyType]]:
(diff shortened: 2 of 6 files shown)


mypy_django_plugin/main.py
```diff
@@ -330,6 +334,13 @@ def get_attribute_hook(self, fullname: str) -> Optional[Callable[[AttributeConte
             return partial(request.set_auth_user_model_as_type_for_request_user, django_context=self.django_context)
         if info and info.has_base(fullnames.ABSTRACT_USER_MODEL_FULLNAME) and attr_name in ("is_staff", "is_active"):
             return partial(set_auth_user_model_boolean_fields, django_context=self.django_context)
+        if (
+            info
+            and info.has_base(fullnames.BASE_MANAGER_CLASS_FULLNAME)
+            and class_name in self._get_current_manager_bases()
+        ):
+            return partial(resolve_manager_method, django_context=self.django_context)
```
flaeppe, line 342: Might be reasonable to add a comment here. Or lift out the checks in the statement into a descriptive variable name, so that it becomes visible what case we've found

e.g.

```suggestion
        is_dynamic_manager_class = (
            info
            and info.has_base(fullnames.BASE_MANAGER_CLASS_FULLNAME)
            and class_name in self._get_current_manager_bases()
        )
        if is_dynamic_manager_class:
            return partial(resolve_manager_method, django_context=self.django_context)
```

mypy_django_plugin/lib/helpers.py
```diff
@@ -34,7 +34,7 @@
     MethodContext,
     SemanticAnalyzerPluginInterface,
 )
-from mypy.plugins.common import add_method
+from mypy.plugins.common import add_method_to_class
```
flaeppe, line 37: Justification of this change: `add_method` is mentioned as deprecated by mypy https://github.com/python/mypy/blob/2bfa3b48e671f461b9d9ad9e49dcb83c31bade37/mypy/plugins/common.py#L85-L96

Might make more sense to perhaps change that in a different PR

mypy_django_plugin/transformers/managers.py
```diff
@@ -54,9 +177,11 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
...
     new_manager_info.defn.line = ctx.call.line
     new_manager_info.metaclass_type = new_manager_info.calculate_metaclass_type()
-
-    current_module = semanal_api.cur_mod_node
-    current_module.names[ctx.name] = SymbolTableNode(GDEF, new_manager_info, plugin_generated=True)
+    # Stash the queryset fullname which was passed to .from_queryset
+    # So that our 'resolve_manager_method' attribute hook can fetch the method from that QuerySet class
+    new_manager_info.metadata.setdefault("django", {})
+    new_manager_info.metadata["django"].setdefault("from_queryset_manager", {})
+    new_manager_info.metadata["django"]["from_queryset_manager"] = derived_queryset_fullname
```
flaeppe, line 184: No need for `.setdefault` here. `TypeInfo`(`new_manager_info`) is newly created which means that `metadata` is an empty dict.

```suggestion
    new_manager_info.metadata["django"] = {"from_queryset_manager": derived_queryset_fullname}
```

sobolevn (approved):
This is amazing! Idea and execution! Thank you 👏

sobolevn:
You can add any refactoring you've mentioned in a follow up PRs. I just didn't want this to become stale 🙂

flaeppe:
> You can add any refactoring you've mentioned in a follow up PRs. I just didn't want this to become stale 🙂

Cool, I've opened #821 with those small changes

PR #821, opened 2022-01-16
Fix suggested changes from #820

flaeppe:
Fixes some additional suggested changes in #820. Just minor cleanup really.

## Related issues

Refs #820

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -309,37 +309,51 @@ class NewSemanalDjangoPlugin(Plugin):
         return None
 
     def get_base_class_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
+        # Base class is a Model class definition
         if (
             fullname in self.django_context.all_registered_model_class_fullnames
             or fullname in self._get_current_model_bases()
         ):
             return partial(transform_model_class, django_context=self.django_context)
 
+        # Base class is a Manager class definition
         if fullname in self._get_current_manager_bases():
             return add_new_manager_base_hook
 
+        # Base class is a Form class definition
         if fullname in self._get_current_form_bases():
             return transform_form_class
         return None
 
     def get_attribute_hook(self, fullname: str) -> Optional[Callable[[AttributeContext], MypyType]]:
         class_name, _, attr_name = fullname.rpartition(".")
+
+        # Lookup of a settings variable
         if class_name == fullnames.DUMMY_SETTINGS_BASE_CLASS:
             return partial(settings.get_type_of_settings_attribute, django_context=self.django_context)
 
         info = self._get_typeinfo_or_none(class_name)
+
+        # Lookup of the '.is_superuser' attribute
         if info and info.has_base(fullnames.PERMISSION_MIXIN_CLASS_FULLNAME) and attr_name == "is_superuser":
             return partial(set_auth_user_model_boolean_fields, django_context=self.django_context)
+
+        # Lookup of the 'request.user' attribute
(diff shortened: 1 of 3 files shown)


mypy_django_plugin/main.py
```diff
@@ -334,11 +334,13 @@ def get_attribute_hook(self, fullname: str) -> Optional[Callable[[AttributeConte
             return partial(request.set_auth_user_model_as_type_for_request_user, django_context=self.django_context)
         if info and info.has_base(fullnames.ABSTRACT_USER_MODEL_FULLNAME) and attr_name in ("is_staff", "is_active"):
             return partial(set_auth_user_model_boolean_fields, django_context=self.django_context)
-        if (
+
+        is_dynamic_manager_class = (
```
sobolevn: Let's refactor other conditions as well! 
At least in this hook.

flaeppe: Yeah, sounds like a good idea

flaeppe: IMO having all statements broken out to individual variables became a bit messy. I went with comments instead, WDYT?

mypy_django_plugin/transformers/managers.py
```diff
@@ -179,9 +178,7 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
     new_manager_info.metaclass_type = new_manager_info.calculate_metaclass_type()
     # Stash the queryset fullname which was passed to .from_queryset
     # So that our 'resolve_manager_method' attribute hook can fetch the method from that QuerySet class
-    new_manager_info.metadata.setdefault("django", {})
-    new_manager_info.metadata["django"].setdefault("from_queryset_manager", {})
-    new_manager_info.metadata["django"]["from_queryset_manager"] = derived_queryset_fullname
+    new_manager_info.metadata["django"] = {"from_queryset_manager": derived_queryset_fullname}
```
antonagestam, line 181: I'm guessing this slightly decreases compatibility if some other tool were to use the "django" namespace in the metadata dict?

flaeppe: Yes, it does. Though the top-level metadata key is set as `"django"` throughout `django-stubs`.

flaeppe: With a quick peek I saw that it's not always the case that `"django"` is included. Mypy mentions that it's intended for a plugin to have a top-level metadata key coinciding with the plugin name:

https://github.com/python/mypy/blob/254d41e10a353612d5edfdcb8b23b8e90508830b/mypy/plugin.py#L43-L48

Although, I suppose that the key should perhaps be `"django-stubs"` instead

antonagestam: Right, I'm probably bike-shedding. Seems highly unlikely that there will be conflicts. And if there are, it can be addressed in the future.

flaeppe: True, although to perhaps minimize custom metadata usage. It could be useful with some helper to set metadata values(?). Basically just to keep the same top-level name.

```python
def set_metadata_value(key: str, value: Any, info: TypeInfo) -> None:
    info.metadata.setdefault("django-stubs", {})
    info.metadata["django-stubs"][key] = value
```

flaeppe: For reference, SQLAlchemy's mypy plugin has a utility getter/setter for `TypeInfo` metadata: https://github.com/sqlalchemy/sqlalchemy/blob/eb716884a4abcabae84a6aaba105568e925b7d27/lib/sqlalchemy/ext/mypy/util.py#L90-L95

sobolevn (approved):
Thank you!

PR #824, opened 2022-01-20
Notify when Manager.from_queryset happens inside model class body

flaeppe:
Whenever `Manager.from_queryset` happens inside of a model class body, an error will be displayed

e.g. Doing

```python
class MyModel(models.Model):
    objects = BaseManager.from_queryset(MyQuerySet)()
```

Yields:

```
error: `.from_queryset` called from inside model class body
```

---
__Note:__ I've also included a refactoring commit for config parsing, that hopefully will make it less of a hassle to introduce new settings in the future.

## Related issues

Refs #738

### DIFF
--- a/mypy_django_plugin/config.py
+++ b/mypy_django_plugin/config.py
@@ -1,0 +1,98 @@
+import configparser
+import textwrap
+from functools import partial
+from pathlib import Path
+from typing import Any, Callable, Dict, NoReturn, Optional
+
+import tomli
+
+INI_USAGE = """
+(config)
+...
+[mypy.plugins.django_stubs]
+    django_settings_module: str (required)
+...
+"""
+TOML_USAGE = """
+(config)
+...
+[tool.django-stubs]
+django_settings_module = str (required)
+...
+"""
+INVALID_FILE = "mypy config file is not specified or found"
+COULD_NOT_LOAD_FILE = "could not load configuration file"
+MISSING_SECTION = "no section [{section}] found".format
+MISSING_DJANGO_SETTINGS = "missing required 'django_settings_module' config"
+INVALID_SETTING = "invalid {key!r}: the setting must be a boolean".format
+
+
+def exit_with_error(msg: str, is_toml: bool = False) -> NoReturn:
+    """Using mypy's argument parser, raise `SystemExit` to fail hard if validation fails.
+
+    Considering that the plugin's startup duration is around double as long as mypy's, this aims to
+    import and construct objects only when that's required - which happens once and terminates the
+    run. Considering that most of the runs are successful, there's no need for this to linger in the
+    global scope.
+    """
+    from mypy.main import CapturableArgumentParser
+
+    handler = CapturableArgumentParser(
+        prog="(django-stubs) mypy", usage=textwrap.dedent(TOML_USAGE if is_toml else INI_USAGE)
+    )
+    handler.error(msg)
+
+
+class DjangoPluginConfig:
+    __slots__ = ("django_settings_module",)
+    django_settings_module: str
+
+    def __init__(self, config_file: Optional[str]) -> None:
+        if not config_file:
+            exit_with_error(INVALID_FILE)
+
+        filepath = Path(config_file)
+        if not filepath.is_file():
+            exit_with_error(INVALID_FILE)
+
+        if filepath.suffix.lower() == ".toml":
+            self.parse_toml_file(filepath)
+        else:
+            self.parse_ini_file(filepath)
+
+    def parse_toml_file(self, filepath: Path) -> None:
+        toml_exit: Callable[[str], NoReturn] = partial(exit_with_error, is_toml=True)
+        try:
+            with filepath.open(mode="rb") as f:
+                data = tomli.load(f)
+        except (tomli.TOMLDecodeError, OSError):
+            toml_exit(COULD_NOT_LOAD_FILE)
+
+        try:
+            config: Dict[str, Any] = data["tool"]["django-stubs"]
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -66,6 +66,16 @@ We rely on different `django` and `mypy` versions:
 | 0.12.x | old semantic analyzer (<0.711), dmypy support | 2.1.x | ^3.6
 
 
+## Plugin settings
+
+An optional strictness flag is available for stronger checks:
+
+- `warn_manager_type_created_in_model_body`
```
sobolevn: We don't need to make this a setting. Since this error can be fixed with a simple `# type: ignore` comment.

mypy_django_plugin/config.py
```diff
@@ -0,0 +1,111 @@
...
+        filepath = Path(config_file)
+        if not filepath.is_file():
+            exit_with_error(INVALID_FILE)
+
+        if filepath.suffix.lower() == ".toml":
+            self.parse_toml_file(filepath)
+        else:
+            self.parse_ini_file(filepath)
+
+    def parse_toml_file(self, filepath: Path) -> None:
```
sobolevn, line 63: I am +1 on using a separate module for this, but without the new setting 🙂

sobolevn (changes requested):
Awesome, thank you!

README.md
flaeppe: Alright, I'll drop the setting

flaeppe:
I've removed the setting so that the new check is always enabled

flaeppe:
Seems like some related manager issues popped up when enforcing the check. Gonna try to dig in and see if I can resolve them

sobolevn:
Feel free to ask any questions! 👍

tests/typecheck/fields/test_related.yml
```diff
@@ -608,8 +608,12 @@
...
 
         user = MyUser()
-        reveal_type(user.book_set) # N: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.Book]"
-        reveal_type(user.article_set) # N: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.Article]"
+        reveal_type(user.book_set) # N: Revealed type is "myapp.models.MyUser_Book_RelatedManager1"
+        reveal_type(user.article_set) # N: Revealed type is "myapp.models.MyUser_Article_RelatedManager1"
+        reveal_type(user.book_set.add)  # N: Revealed type is "def (*objs: Union[myapp.models.Book*, builtins.int], *, bulk: builtins.bool =)"
+        reveal_type(user.article_set.add)  # N: Revealed type is "def (*objs: Union[myapp.models.Article*, builtins.int], *, bulk: builtins.bool =)"
+        reveal_type(user.book_set.queryset_method())  # N: Revealed type is "builtins.int"
+        reveal_type(user.article_set.queryset_method())  # N: Revealed type is "builtins.int"
```
flaeppe: I've extended this test to also check in on methods only existing on the custom queryset class. Basically to ensure the related manager is built from correct bases. I realised the bases were previously incorrect, so the related manager was built from `models.Manager` and `RelatedManager` which yielded an `attribute does not exist` error for any methods existing on the custom queryset.

I'm thinking that the asserts now should ensure a correct hierarchy. Perhaps we should also check on e.g. `user.book_set.filter` to see that the `Manager`/`BaseManager` is also a base and correct? Since now we're checking that methods from `RelatedManager` and `LibraryEntityQuerySet` are included

sobolevn: > I'm thinking that the asserts now should ensure a correct hierarchy. Perhaps we should also check on e.g. user.book_set.filter to see that the Manager/BaseManager is also a base and correct?

Sure, why not?

mypy_django_plugin/transformers/models.py
```diff
@@ -326,10 +326,8 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
                     related_manager_info = self.lookup_typeinfo_or_incomplete_defn_error(
                         fullnames.RELATED_MANAGER_CLASS
                     )  # noqa: E501
-                    # TODO: Use default manager instead of 'objects'
-                    # See: https://docs.djangoproject.com/en/dev/topics/db/queries/#using-a-custom-reverse-manager
-                    objects = related_model_info.get("objects")
-                    if not objects:
+                    default_manager = related_model_info.get("_default_manager")
+                    if not default_manager:
```
flaeppe, line 330: Changed from `objects` to `_default_manager` with https://docs.djangoproject.com/en/dev/topics/db/queries/#using-a-custom-reverse-manager as justification ("By default the RelatedManager used for reverse relations is a subclass of the default manager for that model.")

mypy_django_plugin/transformers/models.py
```diff
@@ -360,7 +358,7 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
     def try_generate_related_manager(
         self, related_model_cls: Type[Model], related_model_info: TypeInfo
     ) -> Optional[Instance]:
-        manager = related_model_cls._meta.managers_map["objects"]
+        manager = related_model_cls._meta.managers_map["_default_manager"]
```
flaeppe, line 361: Changed from `objects` to `_default_manager` with https://docs.djangoproject.com/en/dev/topics/db/queries/#using-a-custom-reverse-manager as justification ("By default the RelatedManager used for reverse relations is a subclass of the default manager for that model.")

flaeppe:
Alright, I think the related manager issues are resolved

sobolevn (approved):
Thank you!

Issue #269, opened 2019-12-12
Reverse manager should be a subclass of `_default_manager` of related model


youssefm:
# Bug report

## What's wrong

``` python
class User(Model):
    pass

class OrderManager(Manager):
    @property
    def active(self):
        return self.exclude(state='canceled')

class Order(Model):
    objects = OrderManager()
    user = models.ForeignKey(User, related_name='orders')
```

When type checking:

`count: int = user.orders.active.count()`

it should succeed, but instead we get:

`error: "RelatedManager[Order]" has no attribute "active"`

It seems like the wrong manager gets associated with `user.orders`. The real manager is a subclass of the default manager for the model - see https://github.com/django/django/blob/26cab4e8c113b873b99b747128a064ba72d40c74/django/db/models/fields/related_descriptors.py#L551.

## System information

- OS:
- `python` version: 3.6.5
- `django` version: 2.2.1
- `mypy` version: 0.750
- `django-stubs` version: tested with latest master as of 6b3b6be3c139d1165925ee02e39b8ea3135e368f

mkurnikov:
> The real manager is a subclass of the default manager for the model

Shouldn't it be a `_base_manager`? https://docs.djangoproject.com/en/3.0/topics/db/managers/#base-managers 
Which is still a `Manager` in your example. 

On other hand, from this code https://github.com/django/django/blob/26cab4e8c113b873b99b747128a064ba72d40c74/django/db/models/fields/related_descriptors.py#L519
it seems that `_default_manager` is used (which will be consistent with your example)

youssefm:
The documentation doesn't seem entirely clear, but Django's behavior is that a subclass of the default manager seems to be used. Here's some testing I did:

``` python
In [5]: isinstance(user.orders, type(Order._default_manager))
Out[5]: True

In [6]: isinstance(user.orders, type(Order._base_manager))
Out[6]: True

In [7]: type(user.orders)
Out[7]: django.db.models.fields.related_descriptors.create_reverse_many_to_one_manager.<locals>.RelatedManager

In [8]: type(user.orders).__bases__
Out[8]: (OrderManager,)

In [10]: type(Order._default_manager) in type(user.orders).__bases__
Out[10]: True

In [11]: type(Order._base_manager) in type(user.orders).__bases__
Out[11]: False
```

AdrienLemaire:
Looks like I have the same issue.

```py
class PlanManager(models.Manager):
[...]

PR #825, opened 2022-01-22
Reuse reverse managers instead of recreating

flaeppe:
Instead of recreating related managers for each pass through `AddManagers`, a subclass of `RelatedManager` is created __once__ for a related model. The model's `_default_manager` is also set as base for the related manager.

The generated reverse manager's fullname is stored in metadata on the related model. Allowing us to collect that manager, if it exists, for other related models.

There's currently only support for a reverse manager having `_default_manager` as base. (Not really sure how one would go about to support `Model().related_set("custom_manager").queryset_method`).

---

The changes here should mostly affect internals and resolved a `TODO:` found in a test case

https://github.com/typeddjango/django-stubs/blob/edec5a1c99a784208d5d8ce78237016cdaa1c54f/tests/typecheck/fields/test_related.yml#L699

 In general minimizing the amount of dynamically generated reverse managers for a model.

## Related issues

Refs #269 (Not really touching the source of that problem though..)

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -209,7 +209,11 @@ def is_annotated_model_fullname(model_cls_fullname: str) -> bool:
 
 
 def add_new_class_for_module(
-    module: MypyFile, name: str, bases: List[Instance], fields: Optional[Dict[str, MypyType]] = None
+    module: MypyFile,
+    name: str,
+    bases: List[Instance],
+    fields: Optional[Dict[str, MypyType]] = None,
+    no_serialize: bool = False,
 ) -> TypeInfo:
     new_class_unique_name = checker.gen_unique_name(name, module.names)
 
@@ -229,10 +233,14 @@ def add_new_class_for_module(
             var = Var(field_name, type=field_type)
             var.info = new_typeinfo
             var._fullname = new_typeinfo.fullname + "." + field_name
-            new_typeinfo.names[field_name] = SymbolTableNode(MDEF, var, plugin_generated=True)
+            new_typeinfo.names[field_name] = SymbolTableNode(
+                MDEF, var, plugin_generated=True, no_serialize=no_serialize
+            )
 
     classdef.info = new_typeinfo
-    module.names[new_class_unique_name] = SymbolTableNode(GDEF, new_typeinfo, plugin_generated=True)
+    module.names[new_class_unique_name] = SymbolTableNode(
+        GDEF, new_typeinfo, plugin_generated=True, no_serialize=no_serialize
+    )
     return new_typeinfo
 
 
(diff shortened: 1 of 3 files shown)


sobolevn:
> The generated reverse manager's fullname is stored in metadata on the related model

Can you please test it on a real project with cache enabled?
Because sometimes it is problematic to store complex objects in metadata, when mypy tries to dump it into a cache file.

mypy_django_plugin/transformers/models.py
```diff
@@ -297,12 +302,26 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+        manager_fullname = helpers.get_django_metadata(model_info).get("reverse_managers", {}).get(derived_from)
+        if not manager_fullname:
+            return None
+
+        symbol = self.api.lookup_fully_qualified_or_none(manager_fullname)
+        if symbol is None or not isinstance(symbol.node, TypeInfo):
+            return None
+        return symbol.node
+
+    def set_reverse_manager_info(self, model_info: TypeInfo, derived_from: str, fullname: str) -> None:
```
sobolevn, line 315: Please, make sure that we use both flags here: https://github.com/python/mypy/blob/0cec4f7a2e2c569b28f0ced916edd57903e57b7a/mypy/nodes.py#L3173-L3186

flaeppe: Hm, yeah. I'm guessing that we'd like to have `plugin_generated=True` and `no_serialize=True` set for both the new manager type we create and the reverse manager node we insert on the related model. Does that sound reasonable?

flaeppe: Also, I'm not sure if it's a bit too optimistic, in regards to cache, with the adjustment I did to look if the reverse manager attribute is already inserted into the related model's symbol table

https://github.com/typeddjango/django-stubs/pull/825/files#diff-fcf5f5a98547e722d4134f8bd410d2f21a686877f1fdcfac421373edf13ff67eR322-R324

flaeppe:
> > The generated reverse manager's fullname is stored in metadata on the related model
> 
> Can you please test it on a real project with cache enabled? Because sometimes it is problematic to store complex objects in metadata, when mypy tries to dump it into a cache file.

Alright, so I've run this branch with the [flags you mentioned](https://github.com/typeddjango/django-stubs/pull/825#discussion_r790147678) set to `True` on a real project with cache enabled. I didn't run in to any caching issues in regards to reverse managers. I should however mention that I run in to caching issues regarding `.filter`-kwargs (an internal error which I get on `master` and latest release too, on cached runs). So it might be that I should run it on a different project..

Note that I'm kind of shooting in the dark in regards to that I'm actually triggering relevant cache for reverse managers..

sobolevn:
If your project has related managers and it writes cache - it should be fine.

flaeppe:
Yeah, it does

mypy_django_plugin/lib/helpers.py
```diff
@@ -331,15 +339,15 @@ def check_types_compatible(
     api.check_subtype(actual_type, expected_type, ctx.context, error_message, "got", "expected")
 
 
-def add_new_sym_for_info(info: TypeInfo, *, name: str, sym_type: MypyType) -> None:
+def add_new_sym_for_info(info: TypeInfo, *, name: str, sym_type: MypyType, no_serialize: bool = False) -> None:
```
sobolevn, line 342: Probably this is the cause for other errors like the one with `WithAnnotations` 🤔 

We should disallow serializing it as well.

sobolevn (approved):
Thank you! Great work!

mypy_django_plugin/lib/helpers.py
flaeppe: Yeah, that might be a reason

PR #828, opened 2022-01-23
Fix .filter-kwargs lookup crash during cached runs

### LINKED ISSUE #827: Cached runs crashes on typechecking .filter-kwargs with IndexError
# Bug report

Cached runs crashes on typechecking `QuerySet.filter()`-kwargs

## What's wrong

```
error: Unexpected keyword argument "shipment_id" for "filter" of "ShipmentLineManager"  [call-arg]
                    ShipmentLine.objects.filter(
                    ^
error: INTERNAL ERROR -- Please try using mypy master on Github:
version: 0.931
  File "mypy/checkexpr.py", line 3962, in accept
  File "mypy/nodes.py", line 1713, in accept
  File "mypy/checkexpr.py", line 288, in visit_call_expr
  File "mypy/checkexpr.py", line 371, in visit_call_expr_inner
  File "mypy/checkexpr.py", line 878, in check_call_expr_with_callee_type
  File "mypy/checkexpr.py", line 938, in check_call
  File "mypy/checkexpr.py", line 1068, in check_callable_call
  File "mypy/checkexpr.py", line 758, in apply_function_plugin
  File "/app/venv/lib/python3.10/site-packages/mypy_django_plugin/transformers/orm_lookups.py", line 12, in typecheck_queryset_filter
    lookup_kwargs = ctx.arg_names[1]
IndexError: list index out of range
```

## How is that should be

`.filter`-kwargs should be properly typechecked on cached mypy runs.

## System information

- OS:
- `python` version: 3.10.1
- `django` version: 4.0.1
- `mypy` version: 0.931
- `django-stubs` version: 1.9.0 (but also failing on `master`)

flaeppe:
When copying class methods via `copy_method_to_another_class`, build arguments from the signature (`CallableType`) instead of from `FuncDef.arguments` as `FuncDef.arguments` are not serialized by mypy. Currently, methods copied on cached rounds have no arguments, due to an `AttributeError` being raised here(given that `method_node` is built from cache):

https://github.com/typeddjango/django-stubs/blob/220f0e4cf09df170427bd83ee63f7cbe84f223f5/mypy_django_plugin/lib/helpers.py#L409-L412

## Related issues

Fixes #827 

---

__Note:__ I've run this commit with cache enabled on the project where I encountered the error seen in #827, without being able to recreate it.

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -406,30 +406,31 @@ def copy_method_to_another_class(
         return_type = bind_or_analyze_type(method_type.ret_type, semanal_api, original_module_name)
     if return_type is None:
         return
-    try:
-        original_arguments = method_node.arguments[1:]
-    except AttributeError:
-        original_arguments = []
 
+    # We build the arguments from the method signature (`CallableType`), because if we were to
+    # use the arguments from the method node (`FuncDef.arguments`) we're not compatible with
+    # a method loaded from cache. As mypy doesn't serialize `FuncDef.arguments` when caching
     arguments = []
-    for arg_name, arg_type, original_argument in zip(
-        method_type.arg_names[1:], method_type.arg_types[1:], original_arguments
+    # Note that the first argument is excluded, as that's `self`
+    for pos, (arg_type, arg_kind, arg_name) in enumerate(
+        zip(method_type.arg_types[1:], method_type.arg_kinds[1:], method_type.arg_names[1:]),
+        start=1,
     ):
         bound_arg_type = bind_or_analyze_type(arg_type, semanal_api, original_module_name)
         if bound_arg_type is None:
             return
-
-        var = Var(name=original_argument.variable.name, type=arg_type)
-        var.line = original_argument.variable.line
-        var.column = original_argument.variable.column
-        argument = Argument(
-            variable=var,
-            type_annotation=bound_arg_type,
-            initializer=original_argument.initializer,
-            kind=original_argument.kind,
+        if arg_name is None and hasattr(method_node, "arguments"):
+            arg_name = method_node.arguments[pos].variable.name
+        arguments.append(
+            Argument(
+                # Positional only arguments can have name as `None`, if we can't find a name, we just invent one..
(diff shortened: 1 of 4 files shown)


mypy_django_plugin/lib/helpers.py
```diff
@@ -406,30 +406,27 @@ def copy_method_to_another_class(
...
-
-        var = Var(name=original_argument.variable.name, type=arg_type)
-        var.line = original_argument.variable.line
-        var.column = original_argument.variable.column
-        argument = Argument(
-            variable=var,
-            type_annotation=bound_arg_type,
-            initializer=original_argument.initializer,
-            kind=original_argument.kind,
+        assert arg_name is not None
```
sobolevn: I think it might be `None` for pos-only args: `arg, /`

sobolevn (commented):
Thanks! I have one concern 🙂

mypy_django_plugin/lib/helpers.py
flaeppe: Good point. That sounds correct, I think. Not sure how that argument should be built though

flaeppe: I made a fallback on `arg{pos}` if a name can't be found, for pos-only args.

mypy_django_plugin/lib/helpers.py
```diff
@@ -406,30 +406,31 @@ def copy_method_to_another_class(
...
-            variable=var,
-            type_annotation=bound_arg_type,
-            initializer=original_argument.initializer,
-            kind=original_argument.kind,
+        if arg_name is None and hasattr(method_node, "arguments"):
+            arg_name = method_node.arguments[pos].variable.name
+        arguments.append(
+            Argument(
+                # Positional only arguments can have name as `None`, if we can't find a name, we just invent one..
+                variable=Var(name=arg_name if arg_name is not None else f"arg{pos}", type=arg_type),
```
sobolevn, line 427:
```suggestion
                variable=Var(name=arg_name if arg_name is not None else str(pos), type=arg_type),
```

```suggestion
                variable=Var(name=arg_name if arg_name is not None else str(pos), type=arg_type),
```

sobolevn (approved):
Looks good to me! Thank you!

PR #833, opened 2022-01-28
Add object_list to BaseListView

### LINKED ISSUE #790: Add object_list to BaseListView
# Bug report

## What's wrong

When accessing `self.object_list` in a class inheriting from `ListView` implementation, mypy reports that the attribute does not exist:

```
$ mypy testcase.py
testcase.py:18: error: "MyListView" has no attribute "object_list"  [attr-defined]
            context["objects"] = self.object_list
                                 ^
```

Corresponding code:
```py

from typing import Any, Iterable

from django.views.generic import ListView


class MyModel(Model):
    pass


class MyListView(ListView):
    model = MyModel

    def get_context_data(self, *, object_list: Iterable[MyModel] | None = None, **kwargs: Any) -> dict[str, Any]:
        context = super().get_context_data(object_list=object_list, **kwargs)
        context["objects"] = self.object_list
        return context
```

## How is that should be

While `object_list` isn't defined at construction, `BaseListView.get()` sets it when handling a request, so in practice it is defined whenever a list view is doing its work. Therefore I think it makes sense to add it to the stub of `BaseListView`.

## System information

- OS: Ubuntu Linux
- `python` version: 3.9.7
- `django` version: 3.2.10
- `mypy` version: 0.920

mthuurne:
## Related issues

Closes #790

The Django documentation describes `object_list` as "the list of objects (usually, but not necessarily a queryset)", so I picked `Sequence` as the type.

### DIFF
--- a/django-stubs/views/generic/list.pyi
+++ b/django-stubs/views/generic/list.pyi
@@ -30,6 +30,7 @@ class MultipleObjectMixin(Generic[T], ContextMixin):
     def get_context_object_name(self, object_list: QuerySet) -> Optional[str]: ...
 
 class BaseListView(MultipleObjectMixin[T], View):
+    object_list: Sequence[T]
     def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse: ...
 
 class MultipleObjectTemplateResponseMixin(TemplateResponseMixin):


sobolevn (approved):
Thanks!

PR #835, opened 2022-01-28
`Model.__init__` supporting same typing as assigment

### LINKED ISSUE #697: Support `Model.__init__`
We've mentioned this in a previous issue, but I thought it was worth having its own tracking issue.

### What's wrong

The problem is that instantiating a model with invalid field types doesn't yield type errors.

```python
class AModel(models.Model):
    a_field = cast("models.Field[int, int]", models.IntegerField())

inst = AModel()
inst.a_field = "not an int"  # this yields an error
AModel(a_field="not an int")  # but this doesn't
```

### How is that should be

Instantiating a model with incorrect field types should yield the same errors as on assignment.

flaeppe:
## Related issues

Fixes #697 

---

There was a bit of a chain reaction, as the approach for supporting `Model.__init__` revealed a few issues with the `base_field` for `ArrayField`, AFAIK it doesn't support `Combinable`.

I'm not sure it's possible to test all cases(more exactly triggering the error message) of the implementation for removing `Combinable` for `ArrayField`'s `base_field` until support for declaring custom fields and their get/set types exists(given how the getters/setters are declared for Django's fields). But I don't have any good idea of a different implementation. Let me know if I should change something in regards to that. Or if there's a way to test it that I don't know of.

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -193,8 +193,8 @@ class DjangoContext:
             if isinstance(field, Field):
                 field_name = field.attname
                 # Try to retrieve set type from a model's TypeInfo object and fallback to retrieving it manually
-                # from django-stubs own declaration. This is to honor any explicitly declared (i.e. non-inferred)
-                # field types.
+                # from django-stubs own declaration. This is to align with the setter types declared for
+                # assignment.
                 field_set_type = get_field_set_type_from_model_type_info(
                     model_info, field_name
                 ) or self.get_field_set_type(api, field, method=method)
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -15,6 +15,7 @@ from django.db.models.lookups import Exact
 from django.db.models.sql.query import Query
 from django.utils.functional import cached_property
 from mypy.checker import TypeChecker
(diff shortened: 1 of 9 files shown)


mypy_django_plugin/django/context.py
```diff
@@ -174,10 +175,29 @@ def get_expected_types(self, api: TypeChecker, model_cls: Type[Model], *, method
...
+            return set_type
+
+        model_info = helpers.lookup_class_typeinfo(api, model_cls)
         for field in model_cls._meta.get_fields():
             if isinstance(field, Field):
                 field_name = field.attname
-                field_set_type = self.get_field_set_type(api, field, method=method)
+                # Try to retrieve set type from a model's TypeInfo object and fallback to retrieving it manually
+                # from django-stubs own declaration. This is to honor any explicitly declared (i.e. non-inferred)
+                # field types.
```
flaeppe: Just realised the last sentence is probably incorrect. It should most likely say that "it attempts to fetch the same setter types as for assignment" (since that happens earlier than the callback here)

```suggestion
                # from django-stubs own declaration. This is to align with the setter types declared for
                # assignment.
```

sobolevn (approved):
Thank you, looking good! 👍

PR #838, opened 2022-01-31
Correct return type of get_commands()

RJPercival:
According to [the docstring for this function](https://github.com/django/django/blob/6e499a28ac6ec10c2f0f77b92cab49fb8f8df419/django/core/management/__init__.py#L59-L61), "the instantiated module can be placed in the dictionary in place of the application name".

Here is an example of Django expecting that a value in the dictionary might be a `BaseCommand` rather than a `str`:
https://github.com/django/django/blob/6e499a28ac6ec10c2f0f77b92cab49fb8f8df419/django/core/management/__init__.py#L104-L111


sobolevn (approved):
Thanks!

PR #846, opened 2022-02-08
DatabaseClient: add missing class methods and reflect Django 3.2 refactor

kevinmarsh:
### What's changed in Django

The `DatabaseClient` went under a refactor in the Django 3.2 release

> `DatabaseClient.runshell()` now gets arguments and an optional dictionary with environment variables to the underlying command-line client from `DatabaseClient.settings_to_cmd_args_env()` method. Third-party database backends must implement `DatabaseClient.settings_to_cmd_args_env()` or override `DatabaseClient.runshell()`.

- https://docs.djangoproject.com/en/dev/releases/3.2/#database-backend-api
- https://github.com/django/django/pull/13622

Sample source code for reference:

- https://github.com/django/django/blob/stable/3.2.x/django/db/backends/base/client.py
- https://github.com/django/django/blob/stable/3.2.x/django/db/backends/postgresql/client.py

### What I've done

I've added the new standardized class method `DatabaseClient.settings_to_cmd_args_env` added a missing arg to `DatabaseClient.runshell` and ensured that all backends (postgres/mysql/oracle/sqlite) implement via the same interface.

These changes are compatible with Django 4.0.x and Django 3.2.x


django-stubs/db/backends/mysql/client.pyi
```diff
@@ -1,11 +1,10 @@
...
 
 class DatabaseClient(BaseDatabaseClient):
     executable_name: str = ...
     @classmethod
-    def settings_to_cmd_args(
-        cls, settings_dict: Dict[str, Optional[Union[Dict[str, Dict[str, str]], int, str]]]
-    ) -> List[str]: ...
-    def runshell(self) -> None: ...
+    def settings_to_cmd_args_env(
+        self, settings_dict: dict, parameters: Iterable[str]
```
sobolevn: `Dict[str, Any]`?

sobolevn (commented):
Thanks!

django-stubs/db/backends/mysql/client.pyi
kevinmarsh: Yea sorry was being a bit lazy, but `settings_dict` gets passed in here from from `BaseDatabaseWrapper.settings_dict` which is `Dict[str, Any]`. Have pushed the change
https://github.com/typeddjango/django-stubs/blob/dd8ca922872676990ef04a5e023a6bac72ec037f/django-stubs/db/backends/base/base.pyi#L56

tests/typecheck/test_import_all.yml
```diff
@@ -231,6 +231,7 @@
         import django.db.backends.dummy.base
         import django.db.backends.mysql
         import django.db.backends.mysql.client
+        import django.db.backends.oracle.client
```
kevinmarsh, line 234: I noticed these two files weren't being checked as part of `pytest tests/typecheck/test_import_all.yml::import_all_modules` so I added them as I was typing those files. However there's quite a few missing modules compared to Django 3.2/4.0, I didn't want to muddy up this PR, but is there an automated way to update this test for better coverage?

sobolevn: > but is there an automated way to update this test for better coverage?

No 🙂

sobolevn (approved):
Thank you!

PR #849, opened 2022-02-13
BaseDeleteView now extends FormMixin

voidus:
Sorry, accidentally closed #847 when I renamed master to main, so here it is again.

https://github.com/django/django/blob/f9ec777a826816e20e68021c0e73b5a76be650af/django/views/generic/edit.py#L248

Dropped in 4.0.2: https://github.com/django/django/commit/3a45fea0832c5910acee6e0d29f230f347a50462


CircuitCipher:
I'm converting one of my projects to be typed and I ran into this issue with DeleteView.  I submitted #1214 before I saw this pull request.  Any reason why this hasn't been merged in? I see that its been here since February

adamchainz:
I think no reason except maintainer capacity. I just rebased this in #1215, which I'll merge shortly.

PR #851, opened 2022-02-14
cache_page should be Optional[float] .

xncbf:
If None is input, the cache timeout is applied indefinitely.


sobolevn:
@xncbf thank you! Can you please take a look at what's wrong with black?

xncbf:
Sorry let me take a look.

xncbf:
@sobolevn Ready for lint test

PR #855, opened 2022-02-21
Added `widget_type` annotation to `BoundField`

ethanscorey:
See https://github.com/django/django/blob/main/django/forms/boundfield.py#L283

# Added `widget_type` annotation to `BoundField`

## Related issues

I didn't open an issue for this, since it seemed like a simple fix, but I'm happy to do so if that makes things easier. As of April 20, 2020, the `BoundField` class has a property called `widget_type` that returns the field's widget class name as a lower-case string.


django-stubs/forms/boundfield.pyi
```diff
@@ -47,6 +47,8 @@ class BoundField:
     def build_widget_attrs(
         self, attrs: Dict[str, str], widget: Optional[Widget] = ...
     ) -> Dict[str, Union[bool, str]]: ...
+    @property
+    def widget_type(self) -> AnyStr: ...
```
sobolevn: This does not seem correct. `AnyStr` is `TypeVar('AnyStr', str, bytes)`.

Maybe we should define this as `Union[str, bytes]`?

ethanscorey: I initially had it as `str`, because practically speaking, I don't think it could ever return `bytes`. However, since the attribute itself is just a call to `re.sub`, I made the return type match `re.sub`. Thinking about it again in light of your comment, perhaps `str` is the correct return type, since I don't think a class name can ever be `bytes`.

In case it's helpful, here's the source for `widget_type`:

```python
@property
    def widget_type(self):
        return re.sub(
            r"widget$|input$", "", self.field.widget.__class__.__name__.lower()
        )
```

sobolevn: Agreed 👍

ethanscorey:
Just updated to revert to using `str` as the return type.

sobolevn (approved):

PR #857, opened 2022-02-22
Make formfield_from_dbfield return forms.Field instead of models.Field

rolandcrosby-check:
In `django.contrib.admin.options`, `BaseModelAdmin` has a `formfield_from_dbfield` method that takes a `django.db.models.Field` argument and returns an optional `django.forms.Field`. The existing stub incorrectly typed the return value as `django.db.models.Field`.

I could have left the argument type as `Field` (imported from `django.db.models.fields`) but for clarity I added an import of `models` to make it explicit which `Field` is being used here; let me know if you prefer to leave the argument type as `Field` or something else.

### DIFF
--- a/django-stubs/contrib/admin/options.pyi
+++ b/django-stubs/contrib/admin/options.pyi
@@ -16,6 +16,7 @@ from typing import (
     Union,
 )
 
+from django import forms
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks!

PR #862, opened 2022-03-02
Improve forms stubs

adamchainz:
# I have made things!

* `default_renderer` may be a renderer class, instance, or `None`: https://docs.djangoproject.com/en/4.0/ref/forms/api/#django.forms.Form.default_renderer / https://github.com/django/django/blob/d8b437b1fbe3bf54822833bea5e19d2142cf3e1f/django/forms/forms.py#L120-L129
* `field_order` should be a list/sequence of field names, or `None`: https://docs.djangoproject.com/en/4.0/ref/forms/api/#django.forms.Form.field_order / https://github.com/django/django/blob/d8b437b1fbe3bf54822833bea5e19d2142cf3e1f/django/forms/forms.py#L131-L151
* `cleaned_data` should be a dict/mapping of field names to values, as returned by `clean()`: https://docs.djangoproject.com/en/4.0/ref/forms/api/#django.forms.Form.cleaned_data . I similarly updated `clean()` to allow arbitrary mappings, since it may be legitimate to use some special mapping class.
* `clean()` does not need to return a mapping - https://docs.djangoproject.com/en/4.0/ref/forms/validation/#validating-fields-with-clean : “If your form inherits another that doesn’t return a `cleaned_data` dictionary in its `clean()` method (doing so is optional)...”
* `is_multipart` returns a `bool`: https://docs.djangoproject.com/en/4.0/ref/forms/api/#django.forms.Form.is_multipart
* `hidden_fields` and `visible_fields` aren't documented, but they return lists of `BoundField` instances: https://github.com/django/django/blob/d8b437b1fbe3bf54822833bea5e19d2142cf3e1f/django/forms/forms.py#L501-L513

## Related issues

No related issues - just the above research.

Thanks for maintaining django-stubs 😎


sobolevn (approved):
Thank you! 👍

PR #864, opened 2022-03-10
Fixed #822

### LINKED ISSUE #822: fields type in bulk_update
# Bug report

## What's wrong
The `fields` argument in `bulk_update` is typed as `Sequence[str]`: https://github.com/typeddjango/django-stubs/blob/a9e41a8ba97820cb6be7d85083dce23cab3ba2c1/django-stubs/db/models/query.pyi#L62

This prevents using `Set[str]`, as it is not a subtype of `Sequence` (thats my actual issue with it).

## How is that should be
Looking at the django source `bulk_update` transforms the `fields` argument by a list comprehension early, thus any iterable should work here. I suggest to widen the type info to `Iterable[str]`.

himanshu-balasamanta:
Closes #822


django-stubs/db/models/manager.pyi
```diff
@@ -51,7 +51,7 @@ class BaseManager(Generic[_T]):
     def bulk_create(
         self, objs: Iterable[_T], batch_size: Optional[int] = ..., ignore_conflicts: bool = ...
     ) -> List[_T]: ...
-    def bulk_update(self, objs: Iterable[_T], fields: Sequence[str], batch_size: Optional[int] = ...) -> int: ...
+    def bulk_update(self, objs: Iterable[_T], fields: Iterator[str], batch_size: Optional[int] = ...) -> int: ...
```
sobolevn: Why do you use `Iterator` here and `Iterable` below?

himanshu-balasamanta: Corrected it now.

PR #866, opened 2022-03-10
Fixes #830

### LINKED ISSUE #830: Col.__init__ has wrong type for argument `target`
# Bug report

## What's wrong

This is the current `Col` type declaration:
https://github.com/typeddjango/django-stubs/blob/515e382d4a5eaa938e1949852df6349d69c132df/django-stubs/db/models/expressions.pyi#L189-L190

Looking at django sources, `target` is meant to be a `Field` type, not `str`.

## How is that should be

Make `target` a `Field` type.

himanshu-balasamanta:


sobolevn:
Fixed in https://github.com/typeddjango/django-stubs/pull/909

PR #868, opened 2022-03-12
Fixes #822

### LINKED ISSUE #822: fields type in bulk_update

himanshu-balasamanta:
Changed type of field from Sequence to Iterable.


himanshu-balasamanta:
Could you re-run the tests for this commit?

sobolevn:
Sure!

sobolevn (approved):
Thanks!

PR #871, opened 2022-03-14
Fixes #870

### LINKED ISSUE #870: Mypy version 0.940 released, causes tests to fail
On [11th March 2022, Mypy version 0.940](https://pypi.org/project/mypy/#history) was released, causing django-stubs tests run by pytest to fail in Github Actions started on or after 11th Match.
We either need to use version 0.931 of mypy for local testing and Github actions, or find a way around the error `Cannot determine type of "Any"` being thrown due to mypy version 0.940 on running the pytest command.

himanshu-balasamanta:
Changed the version requirement for mypy back to 0.931 in requirements.txt

Not sure if better methods of solving the issue is possible, just did what I saw that worked.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ pytest-mypy-plugins==1.9.3
 psycopg2-binary
 -e ./django_stubs_ext
 -e .
+mypy==0.931
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ with open("README.md") as f:
     readme = f.read()
 
 dependencies = [
-    "mypy==0.931",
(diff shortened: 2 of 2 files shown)


sobolevn (changes requested):
Please, also change `setup.py`

himanshu-balasamanta:
Changed. Kindly approve the workflow to be started if all looks good.

sobolevn (approved):

himanshu-balasamanta:
Hi, just to be sure.... Earlier mypy >=0.931 was allowed, now >=0.930 and < 0.940 is allowed.
This means 0.930 is included too now which it wasn't before.

PR #876, opened 2022-03-16
Fixes #875

### LINKED ISSUE #875: pip install -e django-stubs in virtual environment may sometimes throw errors
# Bug report

## What's wrong

on running pip install -e django-stubs, the following error is obtained

`WARNING: The user site-packages directory is disabled.
    error: can't create or remove files in install directory`
    
    The following error occurred while trying to add or remove files in the
    installation directory:
    
        [Errno 2] No such file or directory: '/usr/lib/python3.8/site-packages/test-easy-install-16599.write-test'
    
    The installation directory you specified (via --install-dir, --prefix, or
    the distutils default setting) was:
    
        /usr/lib/python3.8/site-packages/
    
    This directory does not currently exist.  Please create it and try again, or
    choose a different installation directory (using the -d or --install-dir
    option).

## How is that should be

Installing collected packages: django-stubs
  Running setup.py develop for django-stubs
Successfully installed django-stubs

## System information

- OS:
- `python` version: 3.8.10
- `django` version: 4.0.3
- `mypy` version: 0.931
- `django-stubs` version: editable version from source code

himanshu-balasamanta:

Workaround implemented from https://github.com/pypa/pip/issues/7953

Similar workaround is used in [django repo setup.py file](https://github.com/django/django/blob/main/setup.py).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,11 @@
 import os
+import site
 from distutils.core import setup
 from typing import List
 
 from setuptools import find_packages
+
+site.ENABLE_USER_SITE = True
 
 
 def find_stub_files(name: str) -> List[str]:


himanshu-balasamanta:
Closes the issue https://github.com/typeddjango/django-stubs/issues/875

setup.py
```diff
@@ -1,9 +1,12 @@
 import os
+import site
 from distutils.core import setup
 from typing import List
 
 from setuptools import find_packages
 
+site.ENABLE_USER_SITE = True
```
sobolevn, line 8: Sorry, I don't understand what is going on.

This is the first time I see something like this in `setup.py` in 10+ years of maintaining python projects.

I am 99% sure that it is not needed.

sobolevn (changes requested):

sobolevn:
This looks to be a problem with your local env. So, closing. Please, feel free to reopen / explain why this is required.

setup.py
himanshu-balasamanta: Take a look at line 8 - line 10 in [django setup.py](https://github.com/Himanshu-Balasamanta/django/blob/main/setup.py#L8) .
[This issue](https://github.com/pypa/pip/issues/7953) is opened in pip src code repo, so it most likely depends on the version of pip you use, in my case, 20.0.2
Also I think as the flag is set in the src code repo for django, the issue isn't just with my local env.

PR #879, opened 2022-03-17
Fix missing args on SlugField.__init__

andersonsouza:
This PR includes the args `max_length`, `min_length`, `strip` and `empty_value` on `forms.SlugField.__init__`'.

These args are accepted as `kwargs` on `SlugField` and passed to CharField using `super().__init__(**kwargs)`, where they are defined.

### DIFF
--- a/django-stubs/forms/fields.pyi
+++ b/django-stubs/forms/fields.pyi
@@ -378,6 +378,10 @@ class SlugField(CharField):
     def __init__(
         self,
         allow_unicode: bool = ...,
+        max_length: Optional[Any] = ...,
+        min_length: Optional[Any] = ...,
+        strip: bool = ...,
+        empty_value: Optional[str] = ...,
         required: bool = ...,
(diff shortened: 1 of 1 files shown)


django-stubs/forms/fields.pyi
```diff
@@ -378,6 +378,10 @@ class SlugField(CharField):
     def __init__(
         self,
         allow_unicode: bool = ...,
+        max_length: Optional[Any] = ...,
```
sobolevn, line 381: `Any` or `int`?

andersonsouza: It accepts any value and tries to cast to `int`. So, weird things like `max_length=Decimal("10.3")` are acceptable.

Also, the stub for `CharField` defines it as `Any` and inherited classes should not use a more specific type.

sobolevn (approved):

sobolevn:
Thank you!

PR #882, opened 2022-03-24
Set type parameters for Field in get_fields()

w0rp:
This fixes an issue with Pyright strict type checking where Pyright complains that the type parameters aren't known.

## Related issues

I haven't created an issue for this, I just proposed the change through GitHub after noticing the problem in my codebase.


sobolevn (approved):
Thanks!

PR #897, opened 2022-03-29
Extend return type of `django.views.generic.list.ListView.get_queryset` to allow any iterable

codeinthehole:
Which can return either a `QuerySet` or an iterable. See, e.g.
https://github.com/django/django/blob/83c803f161044fbfbfcd9a0c94ca93dc131be662/django/views/generic/list.py#L26-L27

This has been true for the generic `ListView` class since at least Django 1.7:
https://github.com/django/django/blob/1.6/django/views/generic/list.py#L24-L27

Happy to write some tests if you can point my in the right direction for how to do that.


django-stubs/views/generic/list.pyi
```diff
@@ -18,7 +18,7 @@ class MultipleObjectMixin(Generic[T], ContextMixin):
     paginator_class: Type[Paginator] = ...
     page_kwarg: str = ...
     ordering: Sequence[str] = ...
-    def get_queryset(self) -> QuerySet[T]: ...
+    def get_queryset(self) -> Union[QuerySet[T], Iterable[Any]]: ...
```
sobolevn: `Iterable[Any]` or `Iterable[T]`?

sobolevn: One more thing, we try to use less `Union` return types, where possible.
This will force our users to always check what exactly was returned. Which is not very nice.

flaeppe:
I will close this with the argument found in #2174, that the method named `get_queryset` returns a `QuerySet`

PR #899, opened 2022-03-30
Improve MultiValueDict/QueryDict dict() return

intgr:
According to Django documentation and code, the `MultiValueDict.dict()` method only returns singular values for each entry, never a list.

* https://docs.djangoproject.com/en/dev/ref/request-response/#django.http.QueryDict.dict
* https://github.com/django/django/blob/3de787a70b6c140a3b1289a8ce12b17ebcb8a200/django/utils/datastructures.py#L215-L217


sobolevn (approved):

sobolevn:
Thanks!

sobolevn (approved):

sobolevn:
Thank you!

sterliakov:
It is not completely true:(

```python
from django.utils.datastructures import MultiValueDict

d = MultiValueDict()
d.setlist('foo', [])
print(d.dict())
# {'foo': []}
```

intgr:
@sterliakov Thanks, that's a good point I didn't realize.

Most Django users probably only use `MultiValueDict` via the `QueryDict` subclass used in `request.GET`, etc parameters. In this usage, my change is useful, because there's no way to pass a query param value as an empty list.

For my use cases, this practical benefit seems to outweigh the theoretical correctness of the union return.

I'm wondering if there are any reasonable compromises or alternatives that could be implemented?

Or maybe Django developers would be willing to change this behavior.

intgr:
EDIT: This comment was tangentially related to the `dict()` method, I opted to move this discussion back to https://github.com/typeddjango/django-stubs/pull/909#discussion_r843949589.

Issue #906, opened 2022-04-01
AssertionError on Union with custom QuerySet/Manager


ljodal:
# Bug report

## What's wrong

Mypy crashes with an assertion error if a variable has a union type of two different, custom querysets. I have a branch with a test case to reproduce the issue [here](https://github.com/ljodal/django-stubs/tree/union-queryset-crash). I haven't had time to look at fixing this yet, but I think it's a simple case of adding support for this case.

```
Traceback (most recent call last):
  File "mypy/checkexpr.py", line 3978, in accept
  File "mypy/nodes.py", line 1670, in accept
  File "mypy/checkexpr.py", line 2087, in visit_member_expr
  File "mypy/checkexpr.py", line 2105, in analyze_ordinary_member_access
  File "mypy/checkmember.py", line 130, in analyze_member_access
  File "mypy/checkmember.py", line 152, in _analyze_member_access
  File "mypy/checkmember.py", line 320, in analyze_union_member_access
  File "mypy/checkmember.py", line 147, in _analyze_member_access
  File "mypy/checkmember.py", line 231, in analyze_instance_member_access
  File "mypy/checkmember.py", line 385, in analyze_member_var_access
  File "mypy/checkmember.py", line 614, in analyze_var
  File "/Users/sigurd/Developer/django-stubs/mypy_django_plugin/transformers/managers.py", line 114, in resolve_manager_method
    assert isinstance(manager_instance, Instance)
```

## How is that should be

It should correctly resolve the type or complain if one of the union types are missing the method. Alternatively it could fall back to `Any` for now.

## System information

- OS:
- `python` version: 3.9.10
- `django` version: 3.2.12
- `mypy` version: 0.942
- `django-stubs` version: latest master
- `django-stubs-ext` version: latest master

flaeppe:
Closed via #907

ljodal:
It would be nice to actually handle this case, not just emit an error. Maybe this issue should be updated or should I rather open a new one @flaeppe?

flaeppe:
Yes, please open a new one!

PR #907, opened 2022-04-01
Emit error instead of raising on union custom QuerySet

ljodal:
Emit an error and return AnyType instead of crashing when encountering a custom method on an enum QuerySet/Manager. This at least makes it possible to ignore the type error, but I'll also look into actually adding support for this at some point.

## Related issues

- Refs #906

### DIFF
--- a/tests/typecheck/managers/querysets/test_union_type.yml
+++ b/tests/typecheck/managers/querysets/test_union_type.yml
@@ -4,13 +4,16 @@
         from django.db.models import QuerySet
         from myapp.models import User, Order
 
-        instance: Union[User, Order] = Order()
+        instance: Union[Order, User] = User()
 
         model_cls = type(instance)
 
-        reveal_type(model_cls.objects)  # N: ...
-        reveal_type(model_cls.objects.all())  # N: ...
-        reveal_type(model_cls.objects.my_method())  # N: ...
+        reveal_type(model_cls) # N: Revealed type is "Union[Type[myapp.models.Order], Type[myapp.models.User]]"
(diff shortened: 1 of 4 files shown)


tests/typecheck/managers/querysets/test_union_type.yml
```diff
@@ -0,0 +1,36 @@
...
+
+        instance: Union[Order, User] = User()
+
+        model_cls = type(instance)
+
+        reveal_type(model_cls) # N: Revealed type is "Union[Type[myapp.models.Order], Type[myapp.models.User]]"
+        reveal_type(model_cls.objects)  # N: Revealed type is "Union[myapp.models.OrderManager[myapp.models.Order], myapp.models.UserManager[myapp.models.User]]"
+        model_cls.objects.my_method()  # E: Unable to resolve return type of queryset/manager method "my_method"
+
+        # TODO: This also fails and somehow uses the first item in the union
```
sobolevn: Do you want to work on this task here? Or in another PR?

sobolevn (approved):
Thank you!

tests/typecheck/managers/querysets/test_union_type.yml
ljodal: I was planning on adding a separate issue for that, so I can remove that line from this PR and add a new test case specifically for that. I haven't really figure out what's causing that issue, as it doesn't appear to be django-stubs' fault, at least not in an obvious way.

ljodal: Removed it now, so feel free to merge this. I'll continue in separate branches for the remaining issues

sobolevn (approved):

Issue #873, opened 2022-03-15
MultipleObjectMixin and SingleObjectMixin has incompatible types for queryset and context_object_name


PetrDlouhy:
# Bug report

## What's wrong

I have view class in [django-hordak](https://github.com/adamcharnock/django-hordak/blob/1.10.0/hordak/views/accounts.py#L76) defined as:
```python
class AccountTransactionsView(LoginRequiredMixin, SingleObjectMixin, ListView):
    ...
```

Which gives the following mypy errors:
```
# mypy .
hordak/views/accounts.py:86: error: Definition of "context_object_name" in base class "SingleObjectMixin" is incompatible with definition in base class "MultipleObjectMixin"
hordak/views/accounts.py:86: error: Definition of "queryset" in base class "SingleObjectMixin" is incompatible with definition in base class "MultipleObjectMixin"
Found 2 errors in 1 file (checked 56 source files)
```

## How is that should be

Types for `queryset` and `context_object_name` should be made less strict to be compatible similar to solution of issue #152

## System information

- OS: Ubuntu
- `python` version: 3.9.7
- `django` version: 4.0.2
- `mypy` version: 0.931
- `django-stubs` version: 1.9.0
- `django-stubs-ext` version: 0.3.1

sterliakov:
Probably I misunderstand something, but it looks like unsafe code. `SingleObjectMixin.queryset` is a queryset to `get_object` from, it should be `QuerySet[Account]` (if it is "view listing transactions for account"). `MultipleObjectMixin.queryset` is queryset you operate on (e.g. render in `ListView`) - it should be `QuerySet[Transaction]`. These two bases are really incompatible.

flaeppe:
As previously mentioned, `SingleObjectMixin` and `ListView` classes are, typing wise, incompatible. Currently over the `get_context_object_name` method

PR #909, opened 2022-04-03
Large update

### LINKED ISSUE #744: RemoteUserBackend.configure_user uses User instead of AbstractUser
# Bug report

## What's wrong

`contrib.auth.backends.RemoteUserBackend.configure_user` is typed to take a `User: user`  argument and likewise return a `User`.

## How is that should be

I believe this should be `AbstractBaseUser`.
It could also see using a `TypVar` to ensure the return type is the same subtype of `AbstractBaseUser` as the passed type.
Ideally it would even be the configured user model, but I am not sure if that is feasible.

## System information

- OS: Linux
- `python` version: Python 3.9.7
- `django` version: 3.1
- `mypy` version: 0.812
- `django-stubs` version: 1.8.0 (it is the same in master)

### LINKED ISSUE #834: Return type for `force_bytes()` on protected types is incorrect
# Bug report

## What's wrong

When checking the following code:
```py
from django.utils.encoding import force_bytes

b = force_bytes(123)
reveal_type(b)
```
mypy reports:
```
testcase.py:4: note: Revealed type is "builtins.int*"
```

## How is that should be

Since the omitted `strings_only` argument is `False` by default, `force_bytes()` will convert the given `int` to ` str` and then convert that `str` to `bytes`.

When I explicitly pass `strings_only=False`, mypy does report `bytes` as the returned type.

The stubs look like this:
```py
@overload
def force_bytes(s: _PT, encoding: str = ..., strings_only: Literal[True] = ..., errors: str = ...) -> _PT: ...
@overload
def force_bytes(s: Any, encoding: str = ..., strings_only: bool = ..., errors: str = ...) -> bytes: ...
```
I thought that perhaps the problem is that mypy can't know the default value for `strings_only`, but adding `False` as an explicit default doesn't solve it.

I also tried exchanging the order of the two overload definitions, but while that fixes the testcase, it breaks when `strings_only=True` is passed, so that's not a solution.

Omitting the default is not an option either, because `strings_only` isn't a keyword-only argument and the `encoding` argument before it has a default value.

So I'm currently out of ideas for how this can be annotated in a way that makes mypy do the right thing.

## System information

- OS: Ubuntu Linux 18.04
- `python` version: 3.9.7
- `django` version: 4.0.1
- `mypy` version: 0.931
- `django-stubs` version: current master (5d3768e5)
- `django-stubs-ext` version:  current master (5d3768e5)

### LINKED ISSUE #848: PasswordValidator Protocol methods do not match
# Bug report

## What's wrong

```
class PasswordValidator(Protocol):
    def password_changed(self, password: str, user: Optional[_UserModel] = ...): ...
```

## How is that should be

The methods of all actual Validators look like this.
```
    def validate(self, password: str, user: Optional[_UserModel] = ...) -> None: ...
    def get_help_text(self) -> str: ...
```

I am not sure why this does not turn up as error for the given functions where it is used as parameters. E.g. `contrib.auth.password_validation.validate_password` uses the `validate` method of the argument / return value from `get_default_password_validators` typed `PasswordValidator`.

## System information

- `django-stubs` version: [master](https://github.com/typeddjango/django-stubs/blob/8f97bf880d3022772581ea4cf8b5bf5297a27bad/django-stubs/contrib/auth/password_validation.pyi)

### LINKED ISSUE #895: utils.timesince.timesince should accept an optional integer `depth`
# Bug report

## What's wrong

My code looks like this:

```python
from django.utils.timesince import timesince

timesince(d, now, depth=1)
```

When I run mypy, I see the following error:

```
error: Unexpected keyword argument "depth" for "timesince"
env/lib/python3.10/site-packages/django-stubs/utils/timesince.pyi:7: note: "timesince" defined here
```

It looks like the `depth` argument was added at django 3.2: django/django@8fa9a6d29efe2622872b4788190ea7c1bcb92019

The django-stubs definition doesn’t include the `depth` argument:
https://github.com/typeddjango/django-stubs/blob/4e3f9d6f06d6253f28638a52b795ad00569fe147/django-stubs/utils/timesince.pyi#L7-L9

## How it should be

I guess the timesince definition should look like this:

```python
def timesince( 
     d: date, now: Optional[date] = ..., reversed: bool = ..., time_strings: Optional[Dict[str, str]] = ..., depth: Optional[int] = ...
 ) -> str: ...
```

The same issue applies to utils.timesince.timeuntil.

## System information

- `python` version: 3.10
- `django` version: 3.2
- `mypy` version: 0.941
- `django-stubs` version: 1.10.0

sterliakov:
# Major fix in different areas

I found out that `django.forms` were typed very poorly and forked to improve it. Then I wrote a [small script](https://github.com/sterliakov/stubmerge/blob/main/retypeme.py) to check this stubs with real code (targeting 3.2 because stubs are created for this version): it a) reports all incompatible stubs and b) allows me to run `mypy` on Django itself. Django code style is far from type safety, of course, but "Incompatible return value", for instance, is often a bad sign. 

Major steps:
- Sync outdated and deprecated code (SafeText, OrderedDict)
- Fix wrong signatures (where some arguments were renamed/added/removed)
- Fix wrong decorators (missing `@property`)
- Replace `None` as return type or variable type where needed (often within generator functions)
- Define common type aliases and reuse them (more DRY)
- Where possible, replace `Any` with more specific type.
- Replace `WSGIRequest` with `HttpRequest` in most files (not related to handlers).
- When callable is assigned as instance attribute, use protocol instead of `Callable` to avoid mypy confusion (false-positives)

I also added a few new test cases (and will add more for forms and views, they are not really covered now).

## Related issues

- Closes #848 
- Closes #834 
- Closes #895 
- Closes #744
- Refs #873


sobolevn:
@sterliakov thanks a lot! This is just amazing.

Do you think that we can implement your script as a part of our CI?

sterliakov:
We can add it to CI to ensure that errors count does never increase (like coverage). To do so, we can just dump `mypy` output to version controlled file. It's also a good way to see which errors are removed or added by commit. I can try to implement it if you'd like.

sobolevn (approved):
I've looked through all 300+ files, most changes look ok to me.
I am not going to nitpick in this PR, because we might end up with stale work and lots of merge conflicts.

So, I am going to merge this.

However, I have several questions beforehand:
1. Which Django version do we target with this PR? Do we need to update docs? 
2. Can you please add your script to the CI in the next PR?

Thanks a lot for such amazing work. Project lives because of contributions like this! 👍

sterliakov:
1. It targets Django 3.2 (I mentioned that in the first message), but we can discuss and switch to 4.0 if needed (with this script it's fairly easy to detect incompatible changes).
2. Okay, will do it soon. 
3. After merging this many people will have new errors detected - probably we should bump minor version?

Also please tell me if I can help with some urgent issues (not only with stubs, but with plugin too).

tilsche:
This is great work. It fixes two issues we had encountered.
It is much more pedantic, which I consider a benefit.

There are just two specific issues:
- `assertInHTML` uses `haystack: SafeString`. That means I cannot pass a `str`, and that feels wrong. `assertHTML(Not)Equal` uses `str` and the underlying `assert_and_parse_html` also uses `str`. Before, it used `haystack: SafeText`, which apparently I could pass a `str` to.
- `RequestFactory` has a generic type parameter that is unbound by the constructor. Therefore `f = RequestFactory()` always needs a type annotation. From skimming over the implementation, shouldn't it just be `class RequestFactory(_RequestFactory[WSGIRequest]):`? (Similarly for `AsyncRequestFactory`)

sterliakov:
- `assertInHTML` should use `str`, I'll patch it soon, thanks! I don't completely understand your idea: `str` was never compatible with `SafeText` (there used to be an alias `SafeString = SafeText`; now direction is reversed following Django deprecation notices, it should not affect any real code). 
- `(Async)RequestFactory` will be fixed too, thanks! The following solution should work (looks a bit overcomplicated, but I can't find better solution to explicitly point out returned request/response types):
```python
class _RequestFactory(Generic[_T]):
    ...  # methods here
class _AsyncRequestFactory(_RequestFactory[_T]):
    ...  # methods here
class AsyncRequestFactory(_AsyncRequestFactory[ASGIRequest]): ...
class Client(ClientMixin, _RequestFactory[_MonkeyPatchedHttpResponseBase[WSGIRequest]]):
    ...  # methods here
class AsyncClient(ClientMixin, _AsyncRequestFactory[_MonkeyPatchedHttpResponseBase[ASGIRequest]]):
    ...  # methods here
```

tilsche:
> * `assertInHTML` should use `str`, I'll patch it soon, thanks! I don't completely understand your idea: `str` was never compatible with `SafeText` (there used to be an alias `SafeString = SafeText`; now direction is reversed following Django deprecation notices, it should not affect any real code).

To clarify: Yes, I agree it should be `str`.

django-stubs/db/migrations/operations/models.pyi
```diff
@@ -1,58 +1,86 @@
...
+    @property
+    def migration_name_fragment(self) -> str: ...
 
-class AlterUniqueTogether(FieldRelatedOptionOperation):
+class AlterUniqueTogether(AlterTogetherOptionOperation):
     option_name: str = ...
-    unique_together: Collection[Sequence[str]] = ...
-    def __init__(self, name: str, unique_together: Optional[Collection[Sequence[str]]]) -> None: ...
+    unique_together: _ListOrTuple[Tuple[str, str]] = ...
+    def __init__(self, name: str, unique_together: _ListOrTuple[Tuple[str, str]]) -> None: ...
```
intgr, line 76: This isn't correct, Django 4.0 generates migrations using the `set` type, e.g.

```python
        migrations.AlterUniqueTogether(
            name='blabla',
            unique_together={('field1', 'field2')},
        ),
```

Causes mypy error:

> Argument "unique_together" to "AlterUniqueTogether" has incompatible type "Set[Tuple[str, str, str]]"; expected "Union[List[Tuple[str, str]], Tuple[Tuple[str, str], ...], Tuple[]]" [arg-type]

Same issue affects `AlterIndexTogether`.

django-stubs/db/models/query_utils.pyi
```diff
@@ -1,34 +1,53 @@
...
-    def as_sql(self, compiler: Optional[SQLCompiler] = ..., connection: Optional[BaseDatabaseWrapper] = ...) -> Any: ...
-
 class Q(tree.Node):
     AND: str = ...
     OR: str = ...
     conditional: bool = ...
-    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
-    def __or__(self, other: Any) -> Q: ...
-    def __and__(self, other: Any) -> Q: ...
+    def __init__(self, *args: Any, _connector: Optional[Any] = ..., _negated: bool = ..., **kwargs: Any) -> None: ...
```
intgr, line 48: Arguably this is correct, but it confuses mypy when passing dynamically generated dict as an argument: 

```
filter = Q(**{f"{self.field_name}__{self.lookup_expr}": value})
```

Causes mypy to say:

> error: Argument 1 to "Q" has incompatible type "**Dict[str, List[str]]"; expected "bool"

Given that these underscore arguments are used in Django internally and not public usage, maybe revert this change?

django-stubs/db/utils.pyi
```diff
@@ -17,28 +20,27 @@ class InternalError(DatabaseError): ...
...
-    def all(self) -> List[BaseDatabaseWrapper]: ...
+    def create_connection(self, alias: str) -> BaseDatabaseWrapper: ...
     def close_all(self) -> None: ...
 
 class ConnectionRouter:
     def __init__(self, routers: Optional[Iterable[Any]] = ...) -> None: ...
     @property
     def routers(self) -> List[Any]: ...
-    def db_for_read(self, model: Type[Model], **hints: Any) -> str: ...
-    def db_for_write(self, model: Type[Model], **hints: Any) -> str: ...
```
intgr, line 41: This removal is incorrect. See Django code:

```python
class ConnectionRouter:
    ...
    def _router_func(action):
        def _route_db(self, model, **hints):
            ...
        return _route_db
    db_for_read = _router_func("db_for_read")
    db_for_write = _router_func("db_for_write")
```

django-stubs/utils/datastructures.pyi
```diff
@@ -15,61 +16,117 @@ from typing import (
...
+    def get(self, key: _K, default: _Z = ...) -> Union[_V, _Z]: ...
+    def getlist(self, key: _K, default: _Z = ...) -> Union[List[_V], _Z]: ...
     def setlist(self, key: _K, list_: List[_V]) -> None: ...
+    def setdefault(self, key: _K, default: _V = ...) -> _V: ...
     def setlistdefault(self, key: _K, default_list: Optional[List[_V]] = ...) -> List[_V]: ...
     def appendlist(self, key: _K, value: _V) -> None: ...
+    def items(self) -> Iterator[Tuple[_K, Union[_V, List[object]]]]: ...  # type: ignore
     def lists(self) -> Iterable[Tuple[_K, List[_V]]]: ...
-    def dict(self) -> Dict[_K, _V]: ...
+    def dict(self) -> Dict[_K, Union[_V, List[object]]]: ...
```
intgr, line 94: As explained in https://github.com/typeddjango/django-stubs/pull/899#issuecomment-1090184768, this is technically correct, but for most use cases, a distinct step backward.

intgr:
Thanks, most of this looks like a significant improvement.

I ran this over my code. There were a few new issues introduced, I added review comments for those.

django-stubs/utils/datastructures.pyi
```diff
@@ -15,61 +16,117 @@ from typing import (
...
     def copy(self: _D) -> _D: ...
+    def __getitem__(self, key: _K) -> Union[_V, List[object]]: ...  # type: ignore
```
intgr, line 96: This is also related to the change of `MultiValueDict.dict()`: technically correct, but for most use cases, a step backward.

The [Django tutorial contains this example](https://docs.djangoproject.com/en/4.0/intro/tutorial04/):
```python
def vote(request, question_id):
    question = get_object_or_404(Question, pk=question_id)
    try:
        selected_choice = question.choice_set.get(pk=request.POST['choice'])
        #                                         ^^^^^^^^^^^^^^^^^^^^^^^^^
    except (KeyError, Choice.DoesNotExist):
        ...
```

But now accessing `.get(pk=request.POST['choice'])` will cause mypy errors like:

> Incompatible type for lookup 'pk': (got "Union[str, List[object]]", expected "Union[str, int]") [misc]

What's your proposed solution for this common usage pattern?

One could add a guard like `if isinstance(request.POST['choice'], str):` or `assert`. But that gets very unwieldy and as explained in https://github.com/typeddjango/django-stubs/pull/899#issuecomment-1090184768, the list return cannot really occur with `request.GET` and `request.POST`.

intgr: One option would be to keep `MultiValueDict` as you have changed, but override `QueryDict` back to the old behavior of assuming no empty lists, for the `request.POST/GET` use case. Is that a compromise you could live with?

Although that way this issue would still affect `request.FILES`.

django-stubs/db/models/query_utils.pyi
sterliakov: Thanks! I agree that it's much more important.

django-stubs/db/migrations/operations/models.pyi
sterliakov: Ough. I have `migrations` excluded from typechecking so haven't noticed this issue. Thank you for reporting! Will fix soon. I have not respected `except TypeError: return option_together` in `models.options.normalize_together()`.

django-stubs/db/models/query_utils.pyi
tilsche: I ran into the issue as well. I wasn't sure my kind of usage was representative.

django-stubs/utils/datastructures.pyi
sterliakov: Overriding `QueryDict` to old behavior could be a better choice, also I don't insist on preserving this change if you think it's troublesome and doesn't really help. 

However, I should also note that there is `QueryDict.get` interface, that works exactly like `dict.get`, returning `None or default` if key is missing or corresponds to empty list. 

Also I have one more solution, which may be a massive overhead but should provide very descriptive interface: define in stubs fake `_ImmutableQueryDict` with get methods that behave as expected (`getitem: (_K) -> _V`; it is perfectly valid assuming that this querydict was created by django and is coming from `request.GET(POST)`), modifying methods disallowed (`__setitem__: (_K, _V) -> NoReturn` etc.) and `copy: () -> QueryDict`. This way we can additionally detect almost all disallowed operations (with `--warn-unreachable`) while keeping edge-cases checked for mutable copy. It does not eliminate issue with `request.FILES`, but for this using `request.FILES.get(key)` is almost always convenient.

intgr: Hmm, yes, the `_ImmutableQueryDict` option may work. Although it looks like there are some situations where mutation is considered legal in Django:

```python
class HttpRequest:
    def __init__(self):
        self.GET = QueryDict(mutable=True)
        self.POST = QueryDict(mutable=True)
```

Apparently this was introduced in https://github.com/django/django/pull/2778 and the reason given is "GET and POST on a vanilla HttpRequest object to be mutable QueryDicts (mutable because the Django tests, and probably many third party tests were expecting it)."

Not sure if this is relevant any more in 2022.

django-stubs/test/client.pyi
```diff
@@ -28,45 +44,68 @@ class FakePayload:
...
+
 def encode_multipart(boundary: str, data: Dict[str, Any]) -> bytes: ...
 def encode_file(boundary: str, key: str, file: Any) -> List[bytes]: ...
 
-class RequestFactory:
+# fake to distinguish WSGIRequest and ASGIRequest
+
+_R = TypeVar("_R", bound=HttpRequest)
+
+class _MonkeyPatchedHttpResponseBase(Generic[_R], HttpResponseBase):
```
intgr, line 69: Should this derive from `HttpResponse` instead?

I have a test that accesses `resp.content` and it works as expected. The `content` attribute is defined in `HttpResponse` not `HttpResponseBase`.

sterliakov: No: it can be `FileResponse` that inherits from `HttpResponseBase`.

intgr: Looks like the `__iter__` and `getvalue()` APIs are "the right way" to access the response content in a universal manner -- both `StreamingHttpResponse` and `HttpResponse` define these methods.

But since these methods aren't defined on the level of `HttpResponseBase` either, this is still an issue.

Should we add `__iter__` and `getvalue()` to `HttpResponseBase` since presumably all subclasses are expected to implement these methods?

django-stubs/utils/datastructures.pyi
sterliakov: Interesting, I haven't noticed that before. `HttpRequest.{GET, POST}` are really mutable, while `{W, A}SGIRequest.{GET, POST}` are not. As far as I understand (quick `grep 'HttpRequest('`), it's used directly only in tests. We don't use `{W, A}SGIRequest` in stubs because it depends on deploy method, not on code itself, so we have to make `HttpRequest.GET` immutable too in stubs. I see two options:

- add `_CommonHttpRequest(HttpRequest)` with immutable `GET/POST` and use it in stubs instead of `HttpRequest`. Will require some work and is hard to maintain.
- make `HttpRequest.{GET, POST}` immutable. It will break code that instantiates `HttpRequest` manually.

I think that the latter is a reasonable choice, but I'm not sure about it. Is there any scenario when somebody can intentionally create `HttpRequest` manually and modify its `GET/POST` data?

django-stubs/test/client.pyi
sterliakov: Yeah, it should be great solution, I don't see any drawbacks. I'll add it to the patch.

django-stubs/utils/datastructures.pyi
sterliakov: I found one more solution which looks very promising. `mypy` [does not 100% support this currently](https://github.com/python/mypy/issues/1020), but the same can be achieved with a tweak (requiring `type: ignore` on `__init__` in stub, because `mypy` doesn't consider it valid, but working outside like a charm):
```python
class _ImmutableQueryDict(QueryDict): ...  # declaring immutability here
class HttpRequest:
    GET: _ImmutableQueryDict
    POST: _ImmutableQueryDict
    # Magic happens here:
    def __init__(self: _MutableHttpRequest) -> None: ...  # type: ignore
    # If mypy issue #1020 is resolved, the following (better) will work:
    # def __new__(cls, *args, **kwargs) -> _MutableHttpRequest: ...
class _MutableHttpRequest(HttpRequest):
    GET: QueryDict  # type: ignore
    POST: QueryDict  # type: ignore
```

And now
```python

request = HttpRequest()
reveal_type(request)  # N: Revealed type is "django.http.request._MutableHttpRequest"
reveal_type(request.GET)  # N: Revealed type is "django.http.request.QueryDict"
request.GET['foo'] = 'bar'

def mk_request() -> HttpRequest:
    return HttpRequest()

req = mk_request()
reveal_type(req)  # N: Revealed type is "django.http.request.HttpRequest"
req.GET['foo'] = 'bar'
```

This way any code that does for some reason `request = HttpRequest()` will get mutable version while all `request: HttpRequest` will have immutable dict.

intgr: > If mypy issue `#1020` is resolved, the following (better) will work

Hmm, this issue https://github.com/python/mypy/issues/1020 is closed. Wasn't it solved by https://github.com/python/mypy/pull/7188 already?

intgr: Anyway this solution is a bit hacky, but seems fine to me. Of course in the end it's up to django-stubs reviewers.

If your changes are in a good enough shape, I would gladly test out your branch (or maybe open a draft PR already).

django-stubs/urls/resolvers.pyi
```diff
@@ -93,7 +102,7 @@ class URLResolver:
     _reverse_dict: MultiValueDict
     def __init__(
         self,
-        pattern: Any,
+        pattern: LocalePrefixPattern,
```
intgr, line 105: This also allows RegexPattern, RoutePattern. I have opened a PR for this: #941

django-stubs/core/exceptions.pyi
```diff
@@ -25,27 +30,37 @@ class MiddlewareNotUsed(Exception): ...
...
+_MsgTypeBase = Union[str, ValidationError]
+# Yeah, it's really ugly, but __init__ checks with isinstance()
+_MsgType = Union[
+    _MsgTypeBase,
+    Dict[str, _MsgTypeBase],
+    List[_MsgTypeBase],
+    Dict[str, str],
+    Dict[str, ValidationError],
+    List[str],
+    List[ValidationError],
```
intgr, line 44: This is not correct. `ValidationError` accepts arbitrarily nested data structures as the `message` argument.

For example this correct code from `django-rest-framework` produces a false positive error: https://github.com/encode/django-rest-framework/blob/3.13.1/tests/test_fields.py#L2428-L2433

Mypy does not support recursive type aliases, so it's impossible to describe this type accurately AFAIK.

For now I think it's preferable to revert than to have false positives. I have opened PR #943 for that. If you have any thoughts, please comment there.

PR #915, opened 2022-04-08
issue-914: ValidationError.error_dict type fix

### LINKED ISSUE #914: ValidationError error_code type is incorrect as of 1.10 release (with Django 4...)
# Bug report

## What's wrong

Version 1.10.1 release has a type error that 1.9.0 did not.

#767 added ValidationError type hints but seems to be wrong!

In particular https://github.com/typeddjango/django-stubs/pull/767/files#diff-3d6bcffaa859b600adce0172d5aeff213edea0b62ff8aa2812193a69e7cccad0R30 is incorrect in that the value type for an error_dict appears to be an error_list, not a bare ValidationError.

```
>       self.assertEqual(
            ex.exception.error_dict["__all__"].code,
            "one_mapped_pull_per_integration",
        )
E       AttributeError: 'list' object has no attribute 'code'

.../integrations/pull_test.py:198: AttributeError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> .../integrations/pull_test.py(198)test_innate_mapped_value()
-> self.assertEqual(
(Pdb) pp ex.exception.error_dict           
{'__all__': [ValidationError(['Unique constraint violated: one mapped pull per integration'])]}
```

## How it should be

If you look at the code for ValidationError and look at the git blame, this code has not changed in over 7 years:

https://github.com/django/django/blame/0b31e024873681e187b574fe1c4afe5e48aeeecf/django/core/exceptions.py#L146-L151

In this case the error_dict clearly may contain a list of errors.

## System information

- OS:
- `python` version: 3.9.12
- `django` version: 4.0.3
- `mypy` version: 0.942
- `django-stubs` version: 1.10.1

jeking3:
# ValidationError.error_dict

In 1.10, specifically in #767, ValidationError typing was added.  I am unable to use it in my project because it is wrong, so I cannot update to 1.10 and I am stuck on 1.9 instead.  In 1.10 unfortunately the error_dict type is not correct:

https://github.com/typeddjango/django-stubs/blob/master/django-stubs/core/exceptions.pyi#L47:
```
class ValidationError(Exception):
    error_dict: Dict[str, ValidationError] = ...
```

It is shown to be a dict with values being the same as an error_list [here](https://github.com/django/django/blame/0b31e024873681e187b574fe1c4afe5e48aeeecf/django/core/exceptions.py#L146-L151) and [here](https://github.com/django/django/blame/0b31e024873681e187b574fe1c4afe5e48aeeecf/django/core/exceptions.py#L184-L190).  Furthermore this code has been relatively stable (with respect to error_dict) for at least 7 years so this fix would be valid on a large set of django versions.

As such, I modified error_dict to have a value type of List[ValidationError] instead of a bare ValidationError:
```
class ValidationError(Exception):
    error_dict: Dict[str, List[ValidationError]] = ...
```

## Related issues

This closes #914


sobolevn (approved):

jeking3:
@sobolevn 1.10.2 any time soon? :)

intgr:
Thanks! Sorry about creating the bug :)

PR #919, opened 2022-04-12
Add missing type for setup_databases

XF-FW:
Hello there,

`django.test.utils` was seemingly missing the type for `setup_databases`. 

This change resolves my issue locally. The type was copied directly from `django.test.runner`.

Let me know, if I need to do anything else.
Cheers.

### DIFF
--- a/django-stubs/test/utils.pyi
+++ b/django-stubs/test/utils.pyi
@@ -12,6 +12,7 @@ from typing import (
     List,
     Mapping,
     Optional,
+    Protocol,
     Set,
     Tuple,
     Type,
@@ -139,12 +140,28 @@ def tag(*tags: str): ...
 _Signature = str
 _TestDatabase = Tuple[str, List[str]]
 
+class TimeKeeperProtocol(Protocol):
+    @contextmanager
+    def timed(self, name: Any) -> Iterator[None]: ...
+    def print_results(self) -> None: ...
+
 def dependency_ordered(
     test_databases: Iterable[Tuple[_Signature, _TestDatabase]], dependencies: Mapping[str, List[str]]
 ) -> List[Tuple[_Signature, _TestDatabase]]: ...
 def get_unique_databases_and_mirrors(
     aliases: Optional[Set[str]] = ...,
 ) -> Tuple[Dict[_Signature, _TestDatabase], Dict[str, Any]]: ...
+def setup_databases(
+    verbosity: int,
+    interactive: bool,
+    *,
+    time_keeper: Optional[TimeKeeperProtocol] = ...,
+    keepdb: bool = ...,
+    debug_sql: bool = ...,
+    parallel: int = ...,
+    aliases: Optional[Mapping[str, Any]] = ...,
+    **kwargs: Any
+) -> List[Tuple[BaseDatabaseWrapper, str, bool]]: ...
 def teardown_databases(
     old_config: Iterable[Tuple[Any, str, bool]], verbosity: int, parallel: int = ..., keepdb: bool = ...
 ) -> None: ...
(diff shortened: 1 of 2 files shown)


django-stubs/test/utils.pyi
```diff
@@ -145,6 +145,7 @@ def dependency_ordered(
 def get_unique_databases_and_mirrors(
     aliases: Optional[Set[str]] = ...,
 ) -> Tuple[Dict[_Signature, _TestDatabase], Dict[str, Any]]: ...
+def setup_databases(**kwargs: Any) -> List[Tuple[BaseDatabaseWrapper, str, bool]]: ...
```
sobolevn: It does not look correct: https://github.com/django/django/blob/3.2/django/test/utils.py#L159-L160

XF-FW: Updated.

XF-FW:
Not sure about a few things:

- I've created `StringConvertableProtocol`, since the name parameter on `timed` can be anything that can be converted into string. It's only used here: https://github.com/django/django/blob/3.2/django/test/utils.py#L867 . It might be a bit overkill, perhaps just `str` would be preferable? 

- I've kept the positional args / named args separator (the `*`), on the signature, but I'm not sure if that's correct.

- On `aliases`, the type for the Mapping value should be `BaseDatabaseWrapper`, but I'm not sure if I should include it for a couple of reasons: 
  - Its value is not used anywhere as far as I could see.
  - It's typed as `Mapping[str, Any]` in some places and `Dict[str, BaseDatabaseWrapper]` in others. For example:
    - `Mapping[str, Any]`: https://github.com/typeddjango/django-stubs/blob/master/django-stubs/utils/connection.pyi#L4
    - `Dict[str, BaseDatabaseWrapper]`: https://github.com/typeddjango/django-stubs/blob/master/django-stubs/test/testcases.pyi#L221

- The typecheck_tests fail due to some issue when fetching django 4.0. I've tried to see what was wrong, but I'm not completely sure. Adding `tags=True` to the fetch on the `git_helpers` fixes it, but it fetches all tags, which is likely not the expected outcome. I would gladly fix it, if you gave me some pointers.

Apologies for all these questions in such a small PR.
Thanks!

sobolevn:
- `StringConvertableProtocol` is not a thing. `Any` is good enough. Everything has `__str__` method.
- `I've kept the positional args / named args separator (the *), on the signature, but I'm not sure if that's correct.` ✅ 
- `The typecheck_tests fail due to some issue when fetching django 4.0. I've tried to see what was wrong, but I'm not completely sure. Adding tags=True to the fetch on the git_helpers fixes it, but it fetches all tags, which is likely not the expected outcome. I would gladly fix it, if you gave me some pointers.` We need an exact commit hash / version tag for django 3.2

XF-FW:
Removed `StringConvertableProtocol`.

Yeah, I've checked that, but that's correct. [For reference](https://github.com/django/django/releases/tag/4.0).
On `scripts.git_helpers.checkout_django_branch` it clones and fetches from the `stable/{django_version}.x`. I'm not sure that makes sense. I think [here](https://github.com/typeddjango/django-stubs/blob/master/scripts/git_helpers.py#L29), it should be either:

- Change 1st line to `repo.remote("origin").fetch(commit_sha, progress=ProgressPrinter(), depth=100)` if we want to use the commit for the specified tag to run the test.
- Change 2nd line to `repo.git.checkout(branch)` if we want to use the stable branch to run the test.

Both work locally.

Let me know which option should be the correct one and if I should open a new PR for it or if fixing it in this one suffices.
Thank you.

sobolevn:
Let's try `repo.git.checkout(branch='3.2')` first! 👍

XF-FW:
Done.

If it passes, I guess the last thing remaining is writing a couple of tests. I don't think I'll have the time during the rest of the week, but I should be able to write them during the weekend.

Edit: I've noticed, you've ran the workflow, but I think I still hadn't pushed this last commit, which fixes the issue.

sobolevn (approved):

Issue #924, opened 2022-04-16
Form.files should be a MultiValueDict


rolandcrosby-check:
# Bug report
In #909, [the `BaseForm` class's `files` attribute](https://github.com/typeddjango/django-stubs/blob/f69e0639c74fee0e8ad3d165790ac9804ea4a3f2/django-stubs/forms/forms.pyi#L21) (and everything related to it, like the `files` argument to [`Widget.value_from_datadict`](https://github.com/typeddjango/django-stubs/blob/f69e0639c74fee0e8ad3d165790ac9804ea4a3f2/django-stubs/forms/widgets.pyi#L59)) were updated to [`_FilesT`](https://github.com/typeddjango/django-stubs/blob/f69e0639c74fee0e8ad3d165790ac9804ea4a3f2/django-stubs/forms/utils.pyi#L10), which is defined as `Mapping[str, Iterable[File]]`. Unless there are counterexamples in the Django codebase that I haven't seen, I'm pretty sure this can be more specific than `Mapping` - it appears to be a `MultiValueDict` (which is what this was typed as in previous versions of django-stubs).

Is it safe to assume that this is a `MultiValueDict` everywhere, or can users only rely on it being a `Mapping`?

sterliakov:
(disclaimer: it was my commit)

Using `Iterable` was definitely a mistake, yes, it should have been `Mapping[str, File]`. 

I will revert this for now in #925 and use `MultiValueDict[str, UploadedFile]` for now again. I see the point that `Mapping[str, Iterable[File]]` is absolutely invalid (it doesn't return iterables as values) and `Mapping[str, UploadedFile]` doesn't work with multiple files upload (no `.getlist` method). However, there's problem with `MultiValueDict[str, UploadedFile]` too: it invalidates all test cases where we use `MyForm({}, {'file_field': UploadedFile(...)})`. 

I think this case needs discussion and mainainer's final decision. 

@sobolevn I'm sure you're one of maintainers, please resolve this or ping somebody else :)

rolandcrosby-check:
@sterliakov thanks for taking a look (and for all the great enhancements you made!)

PR #925, opened 2022-04-17
Recover after #909

### LINKED ISSUE #893: Changed the semantics of Annotated.
# Bug report

## What's wrong

`typing.Annotated` is being reported incorrectly by changing their semantics.  I have a small dataclass like:

```python
@dataclass(unsafe_hash=True)
class RatingComposite(LoweringComposite):
    require_picture: Annotated[bool, ExtendedBooleanType(labels={True: "Always", False: "Never")] = True
    max_value: Annotated[int, IntegerType(min_value=1, max_value=10)] = 5
```

We're using [xotl.plato](https://merchise-autrement.gitlab.io/xotl.plato/) to parse and dump this kind of structures to JSON.  However, this code fails to type-check with:

```
error: Invalid type comment or annotation
note: Suggestion: use ExtendedBooleanType[...] instead of ExtendedBooleanType(...)
error: Incompatible types in assignment (expression has type "bool", variable has type "WithAnnotations[builtins.bool]")

error: Invalid type comment or annotation
note: Suggestion: use IntegerType[...] instead of IntegerType(...)
error: Incompatible types in assignment (expression has type "int", variable has type "WithAnnotations[builtins.int]")
``` 

Without `django-stubs-ext` mypy would accept that annotations and their default values.
 

## How is that should be

- The extension should not impose any syntax to the annotation.  The purpose of `Annotated` is to annotate with 3rd party objects and django-stub-ext should not change the semantics of it.  If it doesn't know about `ExtendedBooleanType`  or `IntegerType`, it should completly ignore them.  Notice the second example is quite similar to the use cases described in [PEP 593](https://peps.python.org/pep-0593/#motivating-examples).

- The final erased type `WithAnnotations[builtins.bool]` should only be introduced for the cases `django-stub-ext` knows about; in this case I'd expect the final type would be just `bool`.

## System information

- OS: Linux (mostly, Manjaro and also Debian-based containers in a CI pipeline)
- `python` version: 3.10.2
- `django` version: 4.0.3 (but is actually not Django related code)
- `mypy` version: 0.941

### LINKED ISSUE #916: `typing.Protocol` is not supported in Python < 3.8
# Bug report

## What's wrong

On Python 3.7, importing `django_stubs_ext` causes `ImportError: cannot import name 'Protocol' from 'typing' (/usr/local/lib/python3.7/typing.py)`.

This issue only surfaced in django-stubs 1.10.0 and django-stubs-ext 0.4.0.

The workaround in the meantime is to stay on 1.9.0 and 0.3.0.

```
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/commands/check.py", line 69, in handle
    databases=options['databases'],
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 423, in check
    databases=databases,
  File "/usr/local/lib/python3.7/site-packages/django/core/checks/registry.py", line 76, in run_checks
    new_errors = check(app_configs=app_configs, databases=databases)
  File "/usr/local/lib/python3.7/site-packages/django/core/checks/templates.py", line 29, in check_string_if_invalid_is_string
    for conf in settings.TEMPLATES:
  File "/usr/local/lib/python3.7/site-packages/django/conf/__init__.py", line 82, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python3.7/site-packages/django/conf/__init__.py", line 69, in _setup
  File "/usr/local/lib/python3.7/site-packages/django/conf/__init__.py", line 170, in __init__
  File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/testsite/testsite/settings.py", line 4, in <module>
    import django_stubs_ext
  File "/usr/local/lib/python3.7/site-packages/django_stubs_ext/__init__.py", line 5, in <module>
    from .types import AnyAttrAllowed as AnyAttrAllowed
  File "/usr/local/lib/python3.7/site-packages/django_stubs_ext/types.py", line 1, in <module>
ImportError: cannot import name 'Protocol' from 'typing' (/usr/local/lib/python3.7/typing.py)
```

## How is that should be

If Python 3.7 compatibility is desired, we shuld use `Protocol` from `typing_extensions` instead.

## System information

- OS:
- `python` version: 3.7
- `django` version: 3.2
- `mypy` version: 0.910
- `django-stubs` version: 1.10.0

sterliakov:
# Fix revealed in #909 problems

## Note on `QueryDict` support

Now `http.request.QueryDict` mutability is checked with plugin. When it is instantiated with `mutable=True` or obtained from `query_dict.copy()`, it is mutable, otherwise &mdash; immutable. `mypy` will report error on the line where immutable version is modified and report any following code as unreachable. 

"Convenient" signatures for `__getitem__: (str) -> str` and `dict: () -> dict[str, str]` are guaranteed only for immutable instances.

`HttpRequest.GET` and `HttpRequest.POST` are immutable in most cases. The only exception: when request is directly instantiated, it has mutable parts.

```python

request = HttpRequest()
reveal_type(request)  # N: Revealed type is "django.http.request._MutableHttpRequest"
reveal_type(request.GET)  # N: Revealed type is "django.http.request.QueryDict"
request.GET['foo'] = 'bar'

def mk_request() -> HttpRequest:
    return HttpRequest()

req = mk_request()
reveal_type(req)  # N: Revealed type is "django.http.request.HttpRequest"
req.GET['foo'] = 'bar'  # E: This QueryDict is immutable.
x = 1  # E: Statement is unreachable
```

# Related issues
Refs #924 
Fixes #916 
Fixes #893


tests/typecheck/test_request.yml
```diff
@@ -69,3 +69,41 @@
...
+    main: |
+        from django.http.request import HttpRequest
+
+        request = HttpRequest()
+        reveal_type(request)  # N: Revealed type is "django.http.request._MutableHttpRequest"
+        reveal_type(request.GET)  # N: Revealed type is "django.http.request.QueryDict"
+        request.GET['foo'] = 'bar'
+
+        def mk_request() -> HttpRequest:
+            return HttpRequest()
```
sobolevn, line 102: Sorry, I fail to understand how this works.

None of these calls use `.copy` or `mutable=True`. Why is there any difference?

sterliakov: My bad, probably I haven't explained that good enough.

We assume that anything annotated as `HttpRequest` is referring to `{W,A}SGIRequest` and thus has immutable `GET` and `POST`. However, this assumption doesn't hold for `HttpRequest` itself that has mutable dicts. So when we instantiate `HttpRequest` directly, overridden `__new__` returns fake `_MutableHttpRequest` class, so the first part of case is actually operating on it (see `reveal_type` there). The second part works with `HttpRequest` and assumes it's immutable.

sobolevn: Do you think that having `_MutableHttpRequest` is worth it?
In my experience, we almost never do such things. 

I fear that it complicates our code quite a bit.

intgr:
Thanks! I can confirm that this PR fixes all the regressions I saw in #909, and introduces no new regressions.

(I would have tested sooner, but I was not made aware of this PR -- feel free to ping me directly next time).

tests/typecheck/test_request.yml
intgr: @sobolevn For background, the reason of this complexity is not distinguishing between mutable/immutable `QueryDict`, but about what types `QueryDict` methods can return.

1. As streliakov points out in [this comment](https://github.com/typeddjango/django-stubs/pull/899#issuecomment-1086718345), the `dict()` and `__getitem__()` methods of `QueryDict` and `MultiValueDict` may return `str` as well as empty `list` (but never a list with length>0).

   IMO this is crazy behavior on Django's part -- it should throw KeyError -- but nothing we can do about that. This isn't mentioned in official Django documentation, but is mentioned in `MultiValueDict` docstring: https://github.com/django/django/blob/main/django/utils/datastructures.py#L78-L82
4. However, the vast majority of code that uses `QueryDict` is within the request lifecycle of Django. In these cases, the empty lists cannot appear. So `dict()` and `__getitem__()` can safely be assumed to only have `str` values.
5. Large amounts of code rely on the # 2 assumption, even example code from Django's official tutorial ([explained here](https://github.com/typeddjango/django-stubs/pull/909#discussion_r843949589)).
6. The `QueryDict` attributes of Django request are created with `mutable=False`

So there are three options:
* We could have one `QueryDict` class and these methods so they always return `Union[str, List]`. This is the situation after #909. But this causes false positive errors in a significant amount of idiomatic Django code, since most code assumes list return can not appear in `request.GET["key"]`
* Do the opposite, have one `QueryDict` class, and hint these methods as returning `str`. This is **not strictly correct** typing, although the list return is a rare fringe case.
* A mix of the above, which is what this PR does. In Django request lifecycle, use the magic `_ImmutableQueryDict` type, which only holds `str` values. In other cases, assume mutable `QueryDict` may also return empty list for values.

I would prefer the second or third option out of these three.

sobolevn (approved):
Ok then, thanks a lot to both of you! 👍

django-stubs/http/request.pyi
```diff
@@ -114,7 +155,46 @@ class QueryDict(MultiValueDict[str, str]):
...
     def __setitem__(self, key: Union[str, bytes], value: Union[str, bytes]) -> None: ...
+    def __delitem__(self, key: Union[str, bytes]) -> None: ...
+    def setlist(self, key: Union[str, bytes], list_: Iterable[Union[str, bytes]]) -> None: ...
+    def setlistdefault(self, key: Union[str, bytes], default_list: Optional[List[str]] = ...) -> List[str]: ...
+    def appendlist(self, key: Union[str, bytes], value: Union[str, bytes]) -> None: ...
+    # Fake signature (because *args is used in source, but it fails with more that 1 argument)
+    @overload
+    def pop(self, key: Union[str, bytes], /) -> str: ...
+    @overload
+    def pop(self, key: Union[str, bytes], default: Union[str, _Z] = ..., /) -> Union[str, _Z]: ...
```
intgr, line 166: django-stubs still supports Python 3.7, but the `/` positional-only parameter syntax is supported only in Python 3.8+.

Opened PR #945 to fix this.

PR #926, opened 2022-04-17
More CI jobs

sterliakov:
# Add typechecking Django to CI flow

## Idea

This setup (ab)uses `git-notes` to store metadata under specific ref without cluttering commit history. When PR is raised (workflow started) or something is pushed to `master` (including merge commit), `reapply` workflow job does the following:

1. If already on master branch, compare with `HEAD^`, otherwise compare with `master`. 
2. Try to retrieve cache corresponding to this reference. Failure => first run, go to 4. 
3. Build new cache for current revision and compare.
4. Write new cache as note for current commit. 

This job has `continue-on-failure: true`, so this workflow fail doesn't prevent other from execution. It's primarily interesting for manual review.

## Cache structure

```python
class MetaDict(TypedDict):
    apply_errors_count: int
    mypy_errors_count: int
    apply_stdout: list[str]  # List of rows
    mypy_stdout: list[str]  # List of rows
```

## Comparison algorithm

1. Outputs of scripts as sets. Removed errors are printed (no failure), added errors are printed and marked as failure.
2. Process `mypy` stdout to produce `dict` where keys are rows with lineno replaced with `.` and values are `list` of lineno's. 
3. Removed keys are printed (no failure).
4. Common keys are compared by length of lineno's list: if it decreased, print difference (no error), if same - no output, if increased - print difference and mark as failure. Length is used instead of exact match, because some shifts can occur and messages are unique enough.
6. New keys are printed and marked as failure.

## Exit  codes

- New errors added to `reapply`: 2
- New `mypy` errors: 1
- Success: 0

P.S. I'm sorry for bash, tell me if it is real problem (e.g. Windows compatibility is important) - I can try rewriting in python.


scripts/compare_to_master.sh
```diff
@@ -0,0 +1,42 @@
+#! /usr/bin/env bash
```
sobolevn, line 1: Can you please run `shellcheck -S style` on this file?

.github/workflows/test.yml
```diff
@@ -23,10 +23,38 @@ jobs:
...
+          fetch-depth: 2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          pip install -U pip setuptools wheel
+          pip install -r ./requirements.txt
+          pip install 'Django>=3.2,<4.0' isort mypy types-pytz types-PyYAML types-docutils types-geoip2
```
sobolevn, line 47: Why do we override deps here?

.github/workflows/test.yml
```diff
@@ -23,10 +23,38 @@ jobs:
...
+      - name: Check against Django code
+        run: ./scripts/compare_to_master.sh
+      - name: Save cache as notes
+        run: git -c 'user.email=django_stubs_test@example.com' -c 'user.name=Cache writing bot' notes --ref cache_history add -f -F .custom_cache/.apply_errors $cur_hash
+      - name: Push notes
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: refs/notes/cache_history
```
sobolevn, line 56: Do we have any other options? Pushing to a branch seems like not the best idea to store cache 🤔

.pre-commit-config.yaml
```diff
@@ -4,21 +4,30 @@ default_language_version:
...
       - id: end-of-file-fixer
+  - repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.9.0
+    hooks:
+      - id: python-check-blanket-type-ignore
+        name: no bare type_ignores (pyi)
+        types: [pyi]
+      - id: python-check-blanket-type-ignore
+        name: no bare type_ignores (python)
+        types: [python]
```
sobolevn, line 23: Can we unify these two?

CONTRIBUTING.md
```diff
@@ -77,6 +77,28 @@ To execute the script run:
...
+python ./scripts/compare_errors.py --check-only
+```
+
+To verify changes against master branch, run
+
+```bash
+python ./scripts/compare_to_master.py
+```
+
+NOTE: this doesn't work with uncommited changes, so stash or commit them in advance.
```
sobolevn, line 92: This should raise an error then. I will forget about it in like 5 hours 😆

scripts/compare_errors.py
```diff
@@ -0,0 +1,189 @@
...
+    from typing_extensions import TypedDict
+else:
+    from typing import TypedDict
+
+
+CURRENT_DIR = (Path(__file__) / ".." / "..").resolve()
+ERRORS_FILE = CURRENT_DIR / ".custom_cache" / ".apply_errors"
+TEMP_ERRORS_FILE = CURRENT_DIR / ".custom_cache" / ".apply_errors_temp"
+
+ERRORS_FILE.parent.mkdir(parents=True, exist_ok=True)
```
sobolevn: Let's not perform any side-effects on import.

scripts/compare_errors.py
```diff
@@ -0,0 +1,189 @@
...
+class MetaDict(TypedDict):
+    apply_errors_count: int
+    mypy_errors_count: int
+    apply_stdout: List[str]
+    mypy_stdout: List[str]
+
+
+def make_meta() -> MetaDict:
+    parser = reapply_types.make_parser()
+    args = parser.parse_args(["--ignore-errors", "decorators", "--no-color"])
```
sobolevn: Why do we need `--ignore-errors decorators` here?

scripts/compare_errors.py
```diff
@@ -0,0 +1,189 @@
...
+    try:
+        os.chdir(temp_dir / "..")
+        errors, fatal, code = run(["-p", "django", "--config-file", str((CURRENT_DIR / "mypy.ini").resolve())])
+    finally:
+        shutil.rmtree(temp_dir)
+
+    try:
+        mypy_errors_count = int(
+            re.search(
+                r"Found (\d+) errors in \d+ files \(checked \d+ source files\)",
```
sobolevn: Let's put all `re` contants somewhere on the top-level and compile them.

.pre-commit-config.yaml
```diff
@@ -32,19 +41,19 @@ repos:
...
       - id: flake8
   - repo: local
     hooks:
-      - id: mypy
+      - id: main_mypy
         name: mypy
         entry: mypy
         language: system
         types: [ python ]
         exclude: "scripts/|django_stubs_ext/"
```
sobolevn, line 54: `scripts` is excluded now. Maybe we should enable it?

sobolevn (commented):
Thanks a lot! Can you please also attach some output examples?

sterliakov:
I'm back with this idea, will fix and reopen this on success.

PR #928, opened 2022-04-19
Fix incorrect instance arg for BaseInlineFormSet.__init__

SmileyChris:
BaseInlineFormSet's `__init__` method incorrectly has:
```python
    instance: Optional[_M] = ...,
```
The instance should actually be the parent model, not the formset's `_M` typevar.

The easy solution would be just to make this `Optional[Model]`, but here's the correct fix by adding the parent model class info
to the BaseInlineFormSet generic.


sobolevn (approved):
Thanks!

PR #943, opened 2022-04-28
Fix ValidationError false positive on nested inputs

intgr:
# I have made things!

Regression from PR #909.

As previously discussed in https://github.com/typeddjango/django-stubs/pull/767#discussion_r764773385, `ValidationError` accepts arbitrarily nested data structures as the `message` argument. Mypy does not currently support recursive type aliases, so AFAIK it's impossible to accurately describe the correct type of this. In any case, the current hints are not correct.

For example this correct code from `django-rest-framework` produces a false positive error: https://github.com/encode/django-rest-framework/blob/3.13.1/tests/test_fields.py#L2428-L2433

I don't have any better ideas other than to revert to my version from #767 that uses `Any`.

### DIFF
--- a/django-stubs/core/exceptions.pyi
+++ b/django-stubs/core/exceptions.pyi
@@ -1,8 +1,6 @@
 import sys
 from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
 
-from django.forms.utils import ErrorDict
-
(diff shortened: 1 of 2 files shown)


sobolevn (approved):

sobolevn:
Thank you!

PR #944, opened 2022-04-28
Fix typecheck default Django version

intgr:
# I have made things!

Currently running `typecheck_tests.py` without arguments fails with "KeyError: '3.0'"


scripts/typecheck_tests.py
```diff
@@ -16,6 +16,7 @@
     "3.2": "0153a63a674937e4a56d9d5e4ca2d629b011fbde",
     "4.0": "67d0c4644acfd7707be4a31e8976f865509b09ac",
 }
+DEFAULT_DJANGO_VERSION = "4.0"
```
sobolevn: I think that 3.2 is still the default 🤔

sobolevn (approved):
Thank you!

PR #945, opened 2022-04-28
Restore Python 3.7 compatibility

intgr:
django-stubs still supports Python 3.7, but the `/` positional-only parameter syntax is supported only in Python 3.8+.

Not sure why this error wasn't caught in CI. But [djangorestframework-stubs CI](https://github.com/typeddjango/djangorestframework-stubs/runs/6214943709?check_suite_focus=true) fails with:
```
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/django-stubs/http/request.pyi:164: error: invalid syntax
```

## Related issues

Regression in #925.


sobolevn (approved):
Probably we don't test this on 3.7 🤔 
Thanks!

PR #949, opened 2022-05-05
Improve type hints of URL conf & include()

intgr:
# I have made things!

The type of `urlpatterns` list is `List[Union[URLPattern, URLResolver]]` -- `path()` can return either type, depending on overloaded arguments.

* https://docs.djangoproject.com/en/dev/ref/urls/#django.urls.include
* https://docs.djangoproject.com/en/4.0/ref/urls/


intgr:
This may require users to add explicit type hints, if they use `include(urlpatterns_list)`, if the list itself contains both views and further `include()`. But if they're using django-stubs, I don't think they will mind.

Without an explicit hint, mypy infers the type as `List[object]`, as there is no common base class between `URLPattern` and `URLResolver`, but correct would be `List[Union[URLPattern, URLResolver]]`.

Generated error, for google-ability:

> urls.py: error: Argument 1 to "include" has incompatible type "List[object]"; expected "Union[Union[str, Module, Sequence[Union[URLPattern, URLResolver]]], Tuple[Union[str, Module, Sequence[Union[URLPattern, URLResolver]]], str]]"

django-stubs/contrib/admindocs/urls.pyi
```diff
@@ -1,3 +1,5 @@
-from typing import Any, List
+from typing import List, Union
 
-urlpatterns: List[Any] = ...
+from django.urls import URLPattern, URLResolver
+
+urlpatterns: List[Union[URLPattern, URLResolver]] = ...
```
sobolevn: `Union[URLPattern, URLResolver]` might be a good candidate for a type alias. What do you think?

django-stubs/contrib/admindocs/views.pyi
```diff
@@ -17,5 +18,7 @@ class TemplateDetailView(BaseAdminDocsView): ...
 
 def get_return_data_type(func_name: Any): ...
 def get_readable_field_data_type(field: Union[Field, str]) -> str: ...
-def extract_views_from_urlpatterns(urlpatterns: Any, base: str = ..., namespace: Optional[Any] = ...): ...
+def extract_views_from_urlpatterns(
+    urlpatterns: Sequence[Union[URLPattern, URLResolver]], base: str = ..., namespace: Optional[str] = ...
```
sobolevn: Looks like `Iterable` to me: https://github.com/django/django/blob/7119f40c9881666b6f9b5cf7df09ee1d21cc8344/django/contrib/admindocs/views.py#L462

django-stubs/contrib/admindocs/views.pyi
```diff
@@ -17,5 +18,7 @@ class TemplateDetailView(BaseAdminDocsView): ...
...
+): ...
```
sobolevn: `-> List[Tuple[Callable, re.Pattern]]`?

Docs: `Each object in the returned list is a two-tuple: (view_func, regex)`
https://github.com/django/django/blob/7119f40c9881666b6f9b5cf7df09ee1d21cc8344/django/contrib/admindocs/views.py#L459

sobolevn (commented):
Thanks!

django-stubs/contrib/admindocs/urls.pyi
intgr: Sure. Although the hard question is, where to put it and what to call it.

Right now I opted for `_AnyURL = Union[URLPattern, URLResolver]` in urls/__init__.pyi, but that doesn't seem too nice.

django-stubs/contrib/admindocs/views.pyi
intgr: Hmm, looks outdated though, the implementation has 4-tuples:

```python
    views = []
    ...
            views.append((p.callback, base + str(p.pattern), namespace, p.name))
    ...
    return views
```

django-stubs/contrib/admindocs/urls.pyi
intgr: `URLPatternOrResolver`?

sobolevn: I actually like `_AnyURL` 🤔

sobolevn (approved):
Thanks! Awesome work.

PR #953, opened 2022-05-11
Support passing extra classes to monkeypatch

### LINKED ISSUE #946: Patch View to be generic to support djangorestframework-stubs?
This isn't actually needed for django-stubs, but djangorestframework-stubs uses generic views, so could we add it to the classes monkeypatched by django_stubs_ext? If so I'd be happy to create a PR for that

henribru:
# I have made things!

## Related issues
Closes #946

### DIFF
--- a/django_stubs_ext/tests/test_monkeypatching.py
+++ b/django_stubs_ext/tests/test_monkeypatching.py
@@ -7,11 +7,14 @@ from _pytest.fixtures import FixtureRequest
 from _pytest.monkeypatch import MonkeyPatch
 from django.db.models import Model
 from django.forms.models import ModelForm
+from django.views import View
 from typing_extensions import Protocol
 
 import django_stubs_ext
 from django_stubs_ext import patch
 from django_stubs_ext.patch import _need_generic, _VersionSpec
+
+extra_classes = [View]
 
 
 class _MakeGenericClasses(Protocol):
@@ -30,11 +33,14 @@ def make_generic_classes(
         for el in _need_generic:
             with suppress(AttributeError):
                 delattr(el.cls, "__class_getitem__")
+        for cls in extra_classes:
+            with suppress(AttributeError):
+                delattr(cls, "__class_getitem__")
 
     def factory(django_version: Optional[_VersionSpec] = None) -> None:
         if django_version is not None:
             monkeypatch.setattr(patch, "VERSION", django_version)
-        django_stubs_ext.monkeypatch()
+        django_stubs_ext.monkeypatch(extra_classes)
 
     request.addfinalizer(fin)
     return factory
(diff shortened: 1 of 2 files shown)


django_stubs_ext/tests/test_monkeypatching.py
```diff
@@ -7,12 +7,15 @@
...
 from django.db.models import Model
 from django.forms.models import ModelForm
+from django.views import View
 from typing_extensions import Protocol
 
 import django_stubs_ext
 from django_stubs_ext import patch
 from django_stubs_ext.patch import _need_generic, _VersionSpec
 
+extra_classes = [View]
```
sobolevn: Let's create a separate `def test_` for this

sobolevn (changes requested):
Thanks!

django_stubs_ext/tests/test_monkeypatching.py
henribru: Done!

django_stubs_ext/tests/test_monkeypatching.py
```diff
@@ -26,15 +27,21 @@ def make_generic_classes(
     request: FixtureRequest,
     monkeypatch: MonkeyPatch,
 ) -> _MakeGenericClasses:
+    _extra_classes: list[type] = []
```
sobolevn, line 32: `= []` is mutable, this is a big no-no for a default value. Can we please use `= None`?

sobolevn (changes requested):

django_stubs_ext/tests/test_monkeypatching.py
henribru: This line isn't a default value, it's for collecting the classes to "unpatch" at the end of the test. I assume this comment was meant for https://github.com/typeddjango/django-stubs/blob/ff9f7bbc9bc98541c8de5d752dd59a4a79610ef0/django_stubs_ext/tests/test_monkeypatching.py#L40 or https://github.com/typeddjango/django-stubs/blob/ff9f7bbc9bc98541c8de5d752dd59a4a79610ef0/django_stubs_ext/django_stubs_ext/patch.py#L61 instead? The mutable default here isn't a huge problem in my opinion because they're typed as `Iterable` and not `list`, so any operation inside the function that would mutate them would cause Mypy to complain anyway. I've changed them now though.

django_stubs_ext/tests/test_monkeypatching.py
```diff
@@ -17,7 +18,7 @@
 class _MakeGenericClasses(Protocol):
     """Used to represent a type of ``make_generic_classes`` fixture."""
 
-    def __call__(self, django_version: Optional[_VersionSpec] = None) -> None:
+    def __call__(self, django_version: Optional[_VersionSpec] = None, extra_classes: Iterable[type] = []) -> None:
```
sobolevn:
```suggestion
    def __call__(self, django_version: Optional[_VersionSpec] = None, extra_classes: Optional[Iterable[type]] = None) -> None:
```

henribru: Woops, I missed that one, thanks

sobolevn (approved):
Thank you!

PR #955, opened 2022-05-11
Fix type for related fields in values_list querysets

ljodal:
When fetching a related field in a values_list queryset Django will
return the object's primary key, not model instances as was previously
what the mypy plugin assumed.


sobolevn (approved):
Thanks! Good catch.

PR #956, opened 2022-05-13
Sync django.contrib.gis.gdal with upstream

mjakob:
Dear maintainers,

I was missing `GDALRaster` in `django.contrib.gis.gdal` and found some other missing exports compared to the upstream Django project (3.2/4.0) that I have added in this PR:
- Add `Envelope`, `OGRGeometry`, `OGRGeomType`, and `GDALRaster`.
- Add explicit re-exports for `DataSource` and `Driver`.

## Related issues

#797 added `DataSource` and `Driver` but it seems like the re-exports disappeared.


sobolevn (approved):
Thanks!

PR #964, opened 2022-05-25
Fix false positive ModelAdmin.list_display_links=None

### LINKED ISSUE #963: False positive: list_display_links can be None
As noted in the [Django docs](https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display_links), `list_display_links` can be None.

Currently this raised an error:
```
Incompatible types in assignment (expression has type "None", base class "ModelAdmin" defined the type as "Union[List[Union[str, Callable[[Any], str]]], Tuple[Union[str, Callable[[Any], str]], ...], Tuple[]]")  [assignment]
```

https://github.com/typeddjango/django-stubs/blob/4a90ad9edc280ddc3e4b3abc1f7e108e94992a0e/django-stubs/contrib/admin/options.pyi#L153

AllexVeldman:
list_display_links can be set to None to not generate any links in the list display.

https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display_links

closes #963


django-stubs/contrib/admin/options.pyi
```diff
@@ -150,7 +150,7 @@ _DisplayT = _ListOrTuple[Union[str, Callable[[_ModelT], str]]]
 
 class ModelAdmin(BaseModelAdmin[_ModelT]):
     list_display: _DisplayT = ...
-    list_display_links: _DisplayT = ...
+    list_display_links: Union[_DisplayT, None] = ...
```
```suggestion
    list_display_links: Optional[_DisplayT] = ...
```

sobolevn (approved):

PR #966, opened 2022-05-25
Fix AlterModelManagers managers annotation to be [(str, Manager), ...]

### LINKED ISSUE #965: migrations.AlterModelManagers.__init__ managers annotation doesn't match what Django generates
# Bug report

## What's wrong

We have the following migration that was auto-generated by `makemigrations`:

```python
import psqlextra.manager.manager


class Migration(migrations.Migration):

    dependencies = [
        ("organisation", "0002_auto_20220321_1406"),
    ]

    operations = [
        migrations.AlterModelManagers(
            name="organisation",
            managers=[
                ("objects", psqlextra.manager.manager.PostgresManager()),
            ],
        ),
    ]
```

django-stubs==1.11.0 changed the annotation on the `managers` argument from `typing.Any`[^1] to `typing.Sequence[Manager]`[^2], leading to the following mypy error:

```
myproject/organisation/migrations/0003_alter_organisation_managers.py:17: error: List item 0 has incompatible type "Tuple[str, Any]"; expected "Manager[Any]"
```

[^1]: https://github.com/typeddjango/django-stubs/blob/v1.5.0/django-stubs/db/migrations/operations/models.pyi#L65-L67
[^2]: https://github.com/typeddjango/django-stubs/blob/v1.11.0/django-stubs/db/migrations/operations/models.pyi#L90-L92

Django's v4.0.4 tests[^3] show that the `managers` argument is `typing.Sequence[typing.Tuple[str, Manager]]`:

```python
        operation = migrations.AlterModelManagers(
            "Pony",
            managers=[
                ("food_qs", FoodQuerySet.as_manager()),
                ("food_mgr", FoodManager("a", "b")),
                ("food_mgr_kwargs", FoodManager("x", "y", 3, 4)),
            ],
        )
```

[^3]: https://github.com/django/django/blob/4.0.4/tests/migrations/test_operations.py#L3792-L3799 

## How is that should be

The typing stub for `AlterModelManagers` should look like this:

```python
class AlterModelManagers(ModelOptionOperation):
    managers: Sequence[Tuple[str, Manager]] = ...
    def __init__(self, name: str, managers: Sequence[Tuple[str, Manager]]) -> None: ...
```

## System information

- OS: Fedora Linux 35
- `python` version: 3.10.4
- `django` version: 4.0.4
- `mypy` version: 0.930
- `django-stubs` version: 1.11.0

alexdutton:
Django generates `AlterModelManagers` with `Sequence[Tuple[str, Manager]]` arguments — not `Sequence[Manager]` — as shown by their tests[^1].

[^1]: https://github.com/django/django/blob/4.0.4/tests/migrations/test_operations.py#L3792-L3799

Closes #965.


intgr:
Thanks, looks good to me.

sobolevn (approved):
Thanks!

alexdutton:
I've just force-pushed to fix the missing `)` in my original commit, so now the checks pass.

PR #967, opened 2022-05-25
Create regression test for related managers 2022

aleksanb:
RelatedManagers do not get set up for me anymore with the latest plugin rewrites.
I took another stab at making a test case.

https://github.com/typeddjango/django-stubs/pull/902#issuecomment-1137017925

### DIFF
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -640,6 +640,46 @@
                   pass
 
 
+-   case: test_related_managers_for_subclass_with_multiple_inheritance
+    main: |
+        from myapp import models
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+
+                class TimestampedModel(models.Model):
+                    id = models.UUIDField(primary_key=True, editable=False)
+                    created_at = models.DateTimeField(auto_now_add=True, db_index=True)
+                    updated_at = models.DateTimeField(auto_now=True, db_index=True)
+
+                    class Meta:
+                        abstract = True
+
+                class ReprMixin:
+                    def __str__(self) -> str:
+                        return self.__repr__()
+
+                    def __repr__(self) -> str:
+                        return 'yolo'
+
+                class User(ReprMixin, TimestampedModel):
+                    name = models.TextField()
+
+                class Booking(ReprMixin, TimestampedModel):
+                    renter = models.ForeignKey(User, on_delete=models.PROTECT)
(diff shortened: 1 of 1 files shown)


tests/typecheck/fields/test_related.yml
```diff
@@ -640,6 +640,46 @@
...
+                class Booking(ReprMixin, TimestampedModel):
+                    renter = models.ForeignKey(User, on_delete=models.PROTECT)
+                    owner = models.ForeignKey(User, on_delete=models.PROTECT, related_name='bookingowner_set')
+
+                    class Meta(TimestampedModel.Meta):
+                        ordering = ['-created_at']
+
+                def process_user(user: User):
+                    reveal_type(user.bookingowner_set)  # N: Revealed type is 'django.db.models.manager.RelatedManager[myapp.models.Booking]'
+                    reveal_type(user.booking_set)  # N: Revealed type is 'django.db.models.manager.RelatedManager[myapp.models.Booking]'
```
intgr, line 681: We could consider using `typing_extensions.assert_type()` instead of `reveal_type()`, which is less verbose.

But would need to ensure that testing dependencies are `typing-extensions>=4.2.0` and `mypy>=0.950` then.

sobolevn: I think that it won't work with our `pytest-mypy-plugin` lib 🤔

tests/typecheck/fields/test_related.yml
```diff
@@ -640,6 +640,46 @@
...
+
+                class Booking(ReprMixin, TimestampedModel):
+                    renter = models.ForeignKey(User, on_delete=models.PROTECT)
+                    owner = models.ForeignKey(User, on_delete=models.PROTECT, related_name='bookingowner_set')
+
+                    class Meta(TimestampedModel.Meta):
+                        ordering = ['-created_at']
+
+                def process_user(user: User):
+                    reveal_type(user.bookingowner_set)  # N: Revealed type is 'django.db.models.manager.RelatedManager[myapp.models.Booking]'
```
sobolevn, line 680:
```suggestion
                    reveal_type(user.bookingowner_set)  # N: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.Booking]"
```

tests/typecheck/fields/test_related.yml
sobolevn, line 681:
```suggestion
                    reveal_type(user.booking_set)  # N: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.Booking]"
```

aleksanb:
Closed in favor of https://github.com/typeddjango/django-stubs/pull/993.

PR #970, opened 2022-05-25
Fix BaseModelFormSet.save_m2m: accept `self`

blueyed:
Before it would result in the following when calling it:

> Attribute function "save_m2m" with type "Callable[[], None]" does not accept self argument  [misc]

Originally added in https://github.com/typeddjango/django-stubs/pull/909/files#diff-3201a55e0a4a5182b8e53740898a55b982ca9c47806c4d4d7be224c1c68bcb39R152

/cc @sterliakov (Thanks for this great update btw!)


django-stubs/forms/models.pyi
```diff
@@ -150,7 +150,7 @@ class BaseModelFormSet(Generic[_M, _ModelFormT], BaseFormSet[_ModelFormT]):
     def save_existing(self, form: _ModelFormT, instance: _M, commit: bool = ...) -> _M: ...
     def delete_existing(self, obj: _M, commit: bool = ...) -> None: ...
     saved_forms: List[_ModelFormT] = ...
-    save_m2m: Callable[[], None] = ...
+    save_m2m: Callable[[self], None] = ...
```
sobolevn: Myabe we should re-write it as `def save_m2m`?

blueyed: ok, done.

Note that it is a method that will only be there when `commit=False` is used before, but I think that cannot be reflected really.

django-stubs/forms/models.pyi
```diff
@@ -150,7 +150,7 @@ class BaseModelFormSet(Generic[_M, _ModelFormT], BaseFormSet[_ModelFormT]):
     def save_existing(self, form: _ModelFormT, instance: _M, commit: bool = ...) -> _M: ...
     def delete_existing(self, obj: _M, commit: bool = ...) -> None: ...
     saved_forms: List[_ModelFormT] = ...
-    save_m2m: Callable[[self], None] = ...
+    def save_m2m(self) -> None: ...
```
sobolevn, line 153: Context: I think that confusion came from these lines: https://github.com/django/django/blob/8c0886b068ba4e224dd78104b93c9638b860b398/django/forms/models.py#L781-L785

But, this is indeed a instance-level method. So, `self` is there.

sobolevn (approved):

PR #973, opened 2022-05-25
Upgrade mypy support to 0.960

DamianZaremba:
# I have made things!

## Related issues

- Closes #972

- Updates test_model_field_classes_from_existing_locations to account
  for the behaviour change in https://github.com/python/mypy/pull/12663
- Bumps the version of django-stubs for a new release

PR #939, opened 2022-04-28
Bump mypy 0.950

DamianZaremba:
# I have made things!

This change bumps the dependencies specifier to enable support for mypy 0.950.

The test cases have also been updated to reflect the expected revealed types change post https://github.com/python/mypy/pull/12459.

## Related issues

Closes #938

PR #979, opened 2022-06-03
Move mypy version upper bound to a [compatible-mypy] extra

andersk:
Due to a bug in mypy 0.940 (#870), we made two changes in #871:

* pinned `mypy==0.931` in `requirements.txt` (for running our tests);
* bounded `mypy<0.940` in `setup.py` (for downstream users).

After the mypy bug was quickly fixed upstream in 0.941, our `setup.py` bound has been repeatedly raised but not removed (#886, #939, #973). The only changes in those commits have been to the precise wording of error messages expected in our tests.  Those wording changes don’t impact compatibility for downstream users, so it should be safe to go back to allowing them to upgrade mypy independently.

Since mypy doesn’t yet guarantee backwards compatibility in the plugin API (although in practice it has rarely been an issue), add a `django-stubs[compatible-mypy]` extra for users who prefer a known-good version of mypy even if it’s a little out of date.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ dependencies = [
 ]
 
 extras_require = {
-    "compatible-mypy": ["mypy<0.970"],
+    "compatible-mypy": ["mypy>=0.930,<0.970"],
 }
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ with open("README.md") as f:
     readme = f.read()
 
 dependencies = [
-    "mypy>=0.930,<0.970",
+    "mypy>=0.930",
     "django",
     "django-stubs-ext>=0.4.0",
     "tomli",
@@ -29,6 +29,10 @@ dependencies = [
     "types-pytz",
     "types-PyYAML",
 ]
+
+extras_require = {
+    "compatible-mypy": ["mypy<0.970"],
+}
 
 setup(
     name="django-stubs",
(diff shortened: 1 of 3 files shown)


sobolevn:
Since mypy does not provide any kind of backward-compat, I think it is safe to leave this bound.

🤔

andersk:
They don’t *guarantee* plugin compatibility at this time, but they are making some effort—for example, python/mypy#11932. Plugin compatibility isn’t what motivated this upper bound in the first place, and doesn’t seem to be a frequent problem in practice.

Meanwhile, for a project that’s [interested](https://github.com/zulip/zulip/pull/18777) in using django-stubs, I’d be more comfortable knowing that it won’t unnecessarily hold back other dependencies (as it currently would, since #973 hasn’t been released). I already pin all transitive dependencies with a lock file that I regularly update on my terms, so I can always decide to hold back mypy myself if something doesn’t work.

Would you be more comfortable moving the upper bound to an extra, perhaps `django-stubs[compatible-mypy]`?

sobolevn:
Hm, I like the `django-stubs[compatible-mypy]` idea!
I always miss good extra deps opportunities 😆

andersk:
Cool, updated as such.

setup.py
```diff
@@ -30,6 +30,10 @@ def find_stub_files(name: str) -> List[str]:
     "types-PyYAML",
 ]
 
+extras_require = {
+    "compatible-mypy": ["mypy<0.970"],
```
sobolevn:
```suggestion
    "compatible-mypy": ["mypy>=0.930,<0.970"],
```

sobolevn (approved):
Thank you!

setup.py
andersk: Not strictly necessary as `extras_require` lists are always merged with `install_requires`, but if you prefer this, sure.

Issue #959, opened 2022-05-16
Release schedule


christianbundy:
Hey there! Is there a release schedule for this project? I'll often bump into bugs and then see the fix merged, but we'll often go a while without a new release. Is there a regular cadence that we're aiming to make releases on, or anything that I can do to help reduce the maintenance burden of cutting a new release?

sobolevn:
No 🙂 

I am the only one doing releases in my free time.

I think that there's only one way we can automate this: https://github.com/semantic-release/semantic-release

But, I am not sure that this is a good idea.

christianbundy:
Is this something that I can help with? Totally understand if you aren't comfortable adding permissions,  but I've maintained loads of open source projects (mostly in the Node.js ecosystem) and would love to help with the "create a tag and push a release" chore.

sobolevn:
I am totally fine with automating this process 🙂 
I think that we can push to PyPI from GitHub itself.

SmileyChris:
For a bit more control, you can have the action to push to pypi trigger only when you create a github release rather than implicitly when a tag is created: https://www.caktusgroup.com/blog/2021/02/11/automating-pypi-releases/

intgr:
I notice that Kurnikov has also made releases in the past. Since Sobolev is busy, any chance @mkurnikov you have time to make a release? This is holding up mypy 0.950 adoption (released on 27 April).

mkurnikov:
Hi, I've made a `1.11.0` release from the latest master, please check that everything works as expected.

mkurnikov:
I think new PyPI releases could be created on every release tag in the repo, it will simplify release maintenance a little bit. I do not have enough time and experience with GitHub actions to be able to do it though. 
So, @christianbundy, if you'd like to create such a workflow, I can add the PyPI token to the Github Secrets later. 

Also, feel free to tag me if you need a release. I do not have enough context to review code anymore, but I can manage that much.

christianbundy:
Excellent. I know I'll be pretty busy with work this week, but I'll ping you when I've got something put together.

mkurnikov:
I'll try to implement it myself this week, feel free to take it over if
it's not yet done by the end of Sunday.

mkurnikov:
Aaand I probably won't have time for that, sorry, please take over. Tag me later if you need anything from me.

terencehonles:
@christianbundy / @mkurnikov I added a simple workflow in #980

intgr:
[...]

PR #980, opened 2022-06-06
Add GitHub release action to upload to PyPI & create GitHub release

terencehonles:
Creating a release action as mentioned in https://github.com/typeddjango/django-stubs/issues/959

I have tested everything except the PyPI part (commented out in https://github.com/terencehonles/django-stubs/commit/78924b806a77da623530a964cc49117132694f00 and https://github.com/terencehonles/django-stubs/commit/cc1e839df2107fb0c0be5f650aa922f1144a5d59) by pushing the tags[`test-tag-failure`](https://github.com/terencehonles/django-stubs/tree/test-tag-failure) & [`test-tag-success`](https://github.com/terencehonles/django-stubs/tree/test-tag-success) in my fork which created the relase https://github.com/terencehonles/django-stubs/releases/tag/test-tag-success

## Related issues

- Refs https://github.com/typeddjango/django-stubs/issues/959

I can open the PR in the other repos if this looks good.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,6 +67,6 @@ jobs:
             await github.rest.issues.create({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              title: `release-ci failure on ${new Date().toDateString()}`,
+              title: `Release failure on ${new Date().toDateString()}`,
               body: `Details: https://github.com/${context.repo.owner}/${context.repo.repo}/actions/workflows/release.yml`,
             })
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,72 @@
+name: Release
+
+on:
+  push:
+    tags:
+    - '*'
+
+jobs:
+  build:
+    if: github.repository == 'typeddjango/django-stubs'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.9'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install -U pip
+          python -m pip install -U setuptools twine wheel
+
+      - name: Build package
+        run: |
+          python setup.py --version
+          python setup.py sdist bdist_wheel
+
+      - name: Publish package to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_API_TOKEN }}
+
+          print_hash: true
+
+      - name: Create release
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const tagName = context.ref.replace(/^refs\/tags\//, '');
+            const release = await github.rest.repos.createRelease({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              tag_name: tagName,
+              name: `Release ${tagName.replace(/^v/, '')}`,
+            });
+
+            if (release.status < 200 || release.status >= 300) {
+              core.setFailed(`Could not create release for tag '${tagName}'`);
+              return;
+            }
+
+  # https://github.community/t/run-github-actions-job-only-if-previous-job-has-failed/174786/2
+  create-issue-on-failure:
+    name: Create an issue if release failed
+    runs-on: ubuntu-latest
+    needs: [build]
+    if: ${{ github.repository == 'typeddjango/django-stubs' && always() && needs.build.result == 'failure' }}
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/github-script@v6
+        with:
+          script: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+  build:
+    if: github.repository == 'typeddjango/django-stubs'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.x'
```
sobolevn: Let's pin it. To whatever. 3.9?

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Build package
+        run: |
+          python setup.py --version
+          python setup.py sdist bdist_wheel
+          twine check dist/*
+
+      - name: Publish package to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
```
sobolevn, line 33: Can you please add a comment: why `__token__`?

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+          password: ${{ secrets.PYPI_API_TOKEN }}
+
+          print_hash: true
+          verify_metadata: false  # check done above
```
sobolevn: Why don't we do this check here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Create release
```
sobolevn, line 38: Let's create an issue if this fails. https://github.com/sobolevn/git-secret/blob/master/.github/workflows/release-ci.yml#L73-L91

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          python setup.py --version
+          python setup.py sdist bdist_wheel
+          twine check dist/*
+
+      - name: Publish package to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_API_TOKEN }}
```
terencehonles, line 34: This secret will need to be set up in the repo or in the org (Org would be best to allow each of the stub packages to use it)

sobolevn (changes requested):
Thank you, looking cool!

.github/workflows/release.yml
terencehonles: The username is just `__token__` for an API token, not sure if you want to have that comment in the file.

See: https://packaging.python.org/en/latest/specifications/pypirc/#using-a-pypi-token

.github/workflows/release.yml
terencehonles: Done

.github/workflows/release.yml
terencehonles: Done

.github/workflows/release.yml
terencehonles: Done, not sure it's as needed because the example you gave is only being triggered on `schedule`, but it doesn't seem like it would hurt.

terencehonles:
I re-ran the changes triggering the two actions:

https://github.com/terencehonles/django-stubs/actions/runs/2449869912
https://github.com/terencehonles/django-stubs/actions/runs/2449869932 (the create issue part "worked" but failed because I don't have issues enabled on my fork)

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            await github.rest.issues.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              title: `release-ci failure on ${new Date().toDateString()}`,
```
sobolevn:
```suggestion
              title: `Release failure on ${new Date().toDateString()}`,
```

sobolevn (approved):
Awesome! I will try it tomorrow.
Thanks for working on this!

intgr:
It would be great to have something similar for [djangorestframework-stubs](https://github.com/typeddjango/djangorestframework-stubs). If the test is successful, is anyone interested in porting this over to there?

.github/workflows/release.yml
terencehonles: Lol, I guess I forgot to include this here. I had fixed it when I was putting this request after failing to create the release but realized it should be a separate job so any failure would trigger the issue and I guess I backed out the change. Thanks for making the fix

terencehonles:
> It would be great to have something similar for [djangorestframework-stubs](https://github.com/typeddjango/djangorestframework-stubs). If the test is successful, is anyone interested in porting this over to there?

I can do it

> I can open the PR in the other repos if this looks good.

PR #984, opened 2022-06-07
Improve admin filter types

adamchainz:
# I have made things!

- `lookups()` can return any iterable, as covered [in its docs](https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_filter) and visible in [implementation](https://github.com/django/django/blob/49b470b9187b6be60e573fed08c8f4a87f133750/django/contrib/admin/filters.py#L86-L89)
- `request` and `queryset` can be properly specified.

## Related issues

n/a


sobolevn (approved):
Thank you!

adamchainz:
Speedy merge!

PR #988, opened 2022-06-09
Enable GitHub auto generated release notes for GitHub releases

terencehonles:
I missed this option in #980 and figured it would be nice to have.

You can see what this would generate by clicking the "create release" button and then selecting a tag and then there's a "generate release notes" above the release notes text box. This will use that description by default and if something needs to be edited the release can be edited after it's created.

## Related issues

- Refs #980


sobolevn (approved):
Good idea!

Issue #958, opened 2022-05-15
Alias argument to QuerySet.using() typed as _SpecialForm


ljodal:
# Bug report

## What's wrong

For some custom managers in our codebase we get an incorrect typing error for calls to `.using()`:

```
error: Argument 1 to "using" of "SomeManager" has incompatible type "str"; expected "_SpecialForm"  [arg-type]
```

Stepping through the plugin this appears to stem from [this](https://github.com/typeddjango/django-stubs/blob/e221f0e0533c98417ed726cba764b5f6282442f8/mypy_django_plugin/transformers/managers.py#L281-L288) call to [`helpers.copy_method_to_another_class`](https://github.com/typeddjango/django-stubs/blob/master/mypy_django_plugin/lib/helpers.py#L379-L438). That calls [`bind_or_anaylyze_type`](https://github.com/typeddjango/django-stubs/blob/e221f0e0533c98417ed726cba764b5f6282442f8/mypy_django_plugin/lib/helpers.py#L366) for each argument, which again ends up calling [`api.lookup_fully_qualified_or_none("django.db.models.query.Optional")`](https://github.com/typeddjango/django-stubs/blob/e221f0e0533c98417ed726cba764b5f6282442f8/mypy_django_plugin/lib/helpers.py#L371). This is where the `_SpecialCase` comes into play, because the type of the `alias` parameter is an instance of `UnboundType` up to this point.

Has anyone else seen this? I've been unable to create an isolated reproducer and I'm not familiar enough with mypy to understand what's really going on here. Would love to get this fixed though, as it's causing us use a lot of `# type: ignore` comments! I'll keep digging, but would love some help to get this fixed.

## How is that should be

The type signature should be kept as `def (self: _QS, alias: Optional[str]) -> _QS`.

## System information

- OS: macOS 12.3.1 and Ubuntu 20.04
- `python` version: 3.9.9 an 3.10.2
- `django` version: 3.2.13
- `mypy` version: 0.950
- `django-stubs` version: latest master (`e221f0e0533c98417ed726cba764b5f6282442f8`)
- `django-stubs-ext` version: latest master (`e221f0e0533c98417ed726cba764b5f6282442f8`)

PR #991, opened 2022-06-12
Fix manager types scope

sterliakov:
Fixed wrong scope of managers methods when manager is imported from another file.

On master branch the following is enough to reproduce the issue:
```
from django.db import models
from django.contrib.sites.managers import CurrentSiteManager


class MyModel(models.Model):
    site = models.ForeignKey(Site, on_delete=models.CASCADE)

    on_site = CurrentSiteManager()
```

It results in weird `mypy` error:
```
myproject/myapp/models.py:6: error: Name "Optional" is not defined
myproject/myapp/models.py:6: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Optional")
```

The changes are self-explanatory, but here's a short explanation:
* Before this PR all types were resolved as if manager is defined in current module. However, this assumption doesn't always hold. 
* After passing `original_module_name`, two other tests began to fail. It happened because `api.anal_type` was expected, but ran only if `module_name is None`. So we want to call it even if module name was given, but only if initial solution (with `lookup_fully_qualified`) fails to return something meaningful.
* While researching this issue, I found out that with incremental mode some tests were wrongly passing. After cleaning cache this changed, so incremental mode is a bad idea for testing.
* Added regression test for this particular issue.

## Related issues
This issue was discovered in [this SO question](https://stackoverflow.com/questions/72585301/mypy-complaining-about-name-optional-is-not-defined-without-the-use-of-optiona/72587601#72587601).

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -2,8 +2,7 @@
 allow_redefinition = True
 check_untyped_defs = True
 ignore_missing_imports = True
-# Incremental mode causes false-negatives
-incremental = False
+incremental = True
 strict_optional = True
 show_traceback = True
 warn_no_return = False
--- a/mypy.ini
+++ b/mypy.ini
@@ -2,7 +2,8 @@
 allow_redefinition = True
 check_untyped_defs = True
 ignore_missing_imports = True
-incremental = True
+# Incremental mode causes false-negatives
+incremental = False
 strict_optional = True
 show_traceback = True
 warn_no_return = False
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,5 +1,10 @@
 # Regular configuration file (can be used as base in other projects, runs in CI)
 
+# NOTE: this config file is not used by pytest locally.
+# See comment in mypy.ini.dev for explanation.
+
+# WARNING: when changing this file, consider doing the same with mypy.ini.dev
+
 [mypy]
 allow_redefinition = True
 check_untyped_defs = True
--- a/mypy.ini.dev
+++ b/mypy.ini.dev
@@ -1,10 +1,11 @@
-# Regular configuration file (can be used as base in other projects, runs in CI)
+# Configuration file to use during development
 
 [mypy]
 allow_redefinition = True
 check_untyped_defs = True
 ignore_missing_imports = True
-incremental = True
+# Avoid caching between test runs
+incremental = False
 strict_optional = True
 show_traceback = True
 warn_no_return = False
--- a/mypy.ini.dev
+++ b/mypy.ini.dev
@@ -1,5 +1,14 @@
 # Configuration file to use during development
 
+# Changes of plugin code are not detected by mypy, so
+# incremental mode can be harmful during development
+# (mypy cache is not always invalidated).
+# However, incremental mode has to be supported by django-stubs,
+# thus another config (mypy.ini) is used in CI.
+# Incremental mode is recommended for regular usage.
+
+# WARNING: when changing this file, consider doing the same with mypy.ini
+
 [mypy]
 allow_redefinition = True
 check_untyped_defs = True
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,7 +46,7 @@ jobs:
           pip install -r ./requirements.txt
 
       - name: Run tests
-        run: pytest
+        run: pytest --mypy-ini-file=mypy.ini
 
   typecheck:
     runs-on: ubuntu-latest
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,6 +70,12 @@ To execute the unit tests, simply run:
 pytest
 ```
 
+If you get some unexpected results or want to be sure that tests run is not affected by previous one, remove `mypy` cache:
+
+```bash
+rm -r .mypy_cache
+```
+
 We also test the stubs against Django's own test suite. This is done in CI but you can also do this locally.
 To execute the script run:
 
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -369,11 +369,10 @@ def bind_or_analyze_type(t: MypyType, api: SemanticAnalyzer, module_name: Option
     That should hopefully give a bound type."""
     if isinstance(t, UnboundType) and module_name is not None:
         node = api.lookup_fully_qualified_or_none(module_name + "." + t.name)
-        if node is None:
-            return None
-        return node.type
-    else:
-        return api.anal_type(t)
+        if node is not None and node.type is not None:
+            return node.type
+
+    return api.anal_type(t)
 
 
 def copy_method_to_another_class(
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -212,7 +212,11 @@ class AddManagers(ModelClassInitializer):
             # replace self type with new class, if copying method
             if isinstance(sym.node, FuncDef):
                 helpers.copy_method_to_another_class(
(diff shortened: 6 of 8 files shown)


mypy.ini
```diff
@@ -2,7 +2,8 @@
 allow_redefinition = True
 check_untyped_defs = True
 ignore_missing_imports = True
-incremental = True
+# Incremental mode causes false-negatives
+incremental = False
```
sobolevn: We had this flag to support mypy in incremental mode as the first-class-citizen. It is important for:
- large projects that use `--incremental` to speed up their checks
- IDEs that use mypy plugins

I am not sure that removing it is a good idea.

sterliakov: Got it, okay. Reverted this and added explicit mention in CONTRIBUTING.md

sobolevn: Can you please post unexpected results here? What were they?

sterliakov: The unexpected result was quite simple: I started from this regression test and confirmed that it fails. Then I attempted first fix (added `module_name`) and rerun test suite - this reg.test passed as expected and two others failed. Then I reverted everything and... got fully passing test suite, including reg.test - which is obviously wrong. Removing mypy cache helped to make reg. test fail again, so incremental mode is harmful during development. Maybe we should have different configs for CI and local development?

sobolevn: > Maybe we should have different configs for CI and local development?

Yes, this is a good option!

mypy.ini.dev
```diff
@@ -0,0 +1,21 @@
+# Configuration file to use during development
```
sobolevn, line 1: Can you please drop a few lines about why do we need this file?

sobolevn (commented):
👍

sobolevn (approved):
Awesome work!

ljodal:
In multiple of our repos this commit _introduces_ a phantom warning, saying `<...>/models.py:106: error: Name "Sequence" is not defined`. Any ideas why? We're using mypy==0.950 in both repos, but I've also tested with 0.961

sterliakov:
Same here. Will investigate this tomorrow. I'm surprised it was not caught by test suite.
I should've checked this... 

UPD: this `Sequence` comes from `django.db.models.query`, line 106, referring to annotation of `select_for_update` method.

ljodal:
Yeah, I found that as well, but it doesn't help that much. I'm 99% sure the phantom warning comes from calling `api.anal_type()`, which didn't happen before this change, but I'm not familiar enough with mypy to understand what goes wrong. When debugging https://github.com/typeddjango/django-stubs/issues/958 I made some changes similar to this and ended up with similar symptoms. It's clear to me that the `copy_method_to_another_class` is doing _something_ weird, but I don't understand what.

flaeppe:
Yeah, I think @ljodal is right here. The problem is most likely running `api.anal_type(<Sequence>)` while processing a `<...>/models.py` that doesn't import a `Sequence` and gets hold of that `Sequence` value, from copying method `django.db.models.query._QuerySet.select_for_update` to `<...>/models.py`.

I can't track down where in mypy the error will be emitted from. But supposedly from somewhere in: https://github.com/python/mypy/blob/203bd64d8c14e8ddbed7911eb6f8f8aa61d728de/mypy/semanal.py#L639

Be aware of that the code emitting the error is from the other call site of `copy_method_to_another_class`: https://github.com/typeddjango/django-stubs/blob/1a29ad4f974915301ce3fbaa33dde71d9bd5d1a9/mypy_django_plugin/transformers/managers.py#L281-L288

---

Half side note:

IMO that `copy_method_to_another_class` is quite brutal. Spontaneously I'm thinking that it shouldn't be _that_ complex to "follow along" the dynamic call to `builtins.type` with 3 arguments([here](https://github.com/django/django/blob/6f63e0ce8e36fbdb2d0c4dfd77ed3cf02e467f49/django/db/models/manager.py#L111-L118)). IIRC the `copy_method_to_another_class` is almost reimplementing what the MRO should/can help out with? But I haven't been able to figure out if or how the copying can be replaced using the MRO instead.

It's just that, runtime, that's kind of what Django does. "Only" thing being that the class is dynamically created..

ljodal:
> Half side note:
> 
> IMO that `copy_method_to_another_class` is quite brutal. Spontaneously I'm thinking that it shouldn't be _that_ complex to "follow along" the dynamic call to `builtins.type` with 3 arguments([here](https://github.com/django/django/blob/6f63e0ce8e36fbdb2d0c4dfd77ed3cf02e467f49/django/db/models/manager.py#L111-L118)). IIRC the `copy_method_to_another_class` is almost reimplementing what the MRO should/can help out with? But I haven't been able to figure out if or how the copying can be replaced using the MRO instead.

The problem with doing this through MRO, other than it not matching the actual runtime types, is that Django is skipping some methods when copying/redefining queryset methods on managers. So if we simply say that the generated manager inherits from the queryset we would expose methods that doesn't actually exist on the manager at runtime. Methods can be explicitly marked as skipped or be skipped because they're already implemented by the manager: https://github.com/django/django/blob/b2eff16806057095c7dd3daa9402ad615e51627f/django/db/models/manager.py#L95-L102

I'm wondering if we're doing something wrong further upstream. I'm really curious about why this "binding" stage is required and if we should rather be deferring to the next mypy iteration at some point maybe 🤔

flaeppe:
> ... is that Django is skipping some methods when copying/redefining queryset methods on managers. So if we simply say that the generated manager inherits from the queryset we would expose methods that doesn't actually exist on the manager at runtime. Methods can be explicitly marked as skipped or be skipped because they're already implemented by the manager ...

I'm pretty sure we're not honoring skipping methods currently: https://github.com/typeddjango/django-stubs/blob/926661ab4ae50ad4796ebe9f15a269f7a18a45c4/mypy_django_plugin/transformers/managers.py#L215-L288

But I see your point with the difficulty of resolving that anyways, with the MRO.

---

Worth noting is also that the only copied methods through `copy_method_to_another_class` are manager methods:

https://github.com/typeddjango/django-stubs/blob/926661ab4ae50ad4796ebe9f15a269f7a18a45c4/mypy_django_plugin/transformers/managers.py#L249-L251

Methods coming from the queryset are resolved differently:

https://github.com/typeddjango/django-stubs/blob/926661ab4ae50ad4796ebe9f15a269f7a18a45c4/mypy_django_plugin/transformers/managers.py#L215-L233

The queryset approach was implemented to resolve _exactly_ these kind of "phantom errors", but for custom queryset methods, a while back through #820 (initial issue in #709)

ljodal:
Oh, that's a very good point. I was planning on testing out the same plug-in callback logic for all queryset methods to see if that resolved https://github.com/typeddjango/django-stubs/issues/958 (which is a similar "something is wrong with the types after copying" issue), but never got around to that. That would probably be worth a try to see if it maybe resolves both issues

ljodal:
> Worth noting is also that the only copied methods through `copy_method_to_another_class` are manager methods:
> 
> 
> 
> https://github.com/typeddjango/django-stubs/blob/926661ab4ae50ad4796ebe9f15a269f7a18a45c4/mypy_django_plugin/transformers/managers.py#L249-L251
> 
> 
> 
> Methods coming from the queryset are resolved differently:
> 
> 
> 
> https://github.com/typeddjango/django-stubs/blob/926661ab4ae50ad4796ebe9f15a269f7a18a45c4/mypy_django_plugin/transformers/managers.py#L215-L233
> 
> 
> 
> The queryset approach was implemented to resolve _exactly_ these kind of "phantom errors", but for custom queryset methods, a while back through #820 (initial issue in #709)

I think you're slightly off here, methods from _custom_ querysets are not copied, methods from the default `QuerySet` class are copied

PR #992, opened 2022-06-13
Improve admin filter lookups() return type

adamchainz:
Follow-up to #984. I missed there that `lookups()` is allowed to return `None`, as shown in [an example in the docs](https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_filter) and [an `is None` check](https://github.com/django/django/blob/49b470b9187b6be60e573fed08c8f4a87f133750/django/contrib/admin/filters.py#L86-L89) in the implementation.


sobolevn (approved):
Thank you!

Issue #969, opened 2022-05-25
Some related managers not getting set up


aleksanb:
See discussion in https://github.com/typeddjango/django-stubs/pull/902 as well.

I still have most related managers missing after the latest upgrades to the type checker code.

The patch we used to have to apply to an older version of django-mypy was this one: https://github.com/HyreAS/django-stubs/commit/2ec73849bf919cbf49550884e89d2314459bb1f7.

But i'm not quite sure as to how i would apply this to the current state of the code.

Some related managers do get set up, so it seems like django-mypy just gives up at a certain point and stops adding more related managers.

Our models are on the form

class MyModel(SomeMixin, SomeAbstractClassThatItselfInheritsFromModelsModel):
  mode_fields = ...

I'll try to create a repro, but haven't had any luck making a reduced test case actually fail on ci before.

Update: the repro does not fail: https://github.com/typeddjango/django-stubs/pull/967 :(

This is a dump of what i get with some extra logging added:

```
no manager <ManyToOneRel: hyre__payment.salesorderrow> django.db.models.manager.RelatedManager TypeInfo(
  Name(django.db.models.manager.RelatedManager)
  Bases(django.db.models.manager.Manager[_T`1])
  Mro(django.db.models.manager.RelatedManager, django.db.models.manager.Manager, django.db.models.manager.BaseManager, builtins.object)
  Names(
    add
    clear
    related_val (builtins.tuple[builtins.int, ...])
    remove
    set))
```

When adding logging here:

<img width="944" alt="bilde" src="https://user-images.githubusercontent.com/1413365/170237675-dc16422a-2a3c-4882-819c-11557743702e.png">

Unsure where to go from here.

flaeppe:
Closed via #993

PR #993, opened 2022-06-13
 Add fallback related manager in final iteration of AddRelatedManagers 

### LINKED ISSUE #981: Default manager is not setup with `django-modeltranslations`
# Bug report

## What's wrong

`_default_manager` is not being setup in plugin when using `django-modeltranslations` as it replaces the default manager and base manager to be `modeltranslation.manager.MultilingualManager` which causes https://github.com/typeddjango/django-stubs/blob/24a3b22c60448b79647dfab8c12cb7d29da4ba65/mypy_django_plugin/transformers/models.py#L285-L291 to fail and never setup a default manager. This also affects the generation of related managers of foreign keys

Instead of giving up assigning a generic manager in `AddDefaultManagerAttribute` fixes the issue though not sure if this is the right thing to do:
```python
class AddDefaultManagerAttribute(ModelClassInitializer):
    def run_with_model_cls(self, model_cls: Type[Model]) -> None:
        if "_default_manager" in self.model_classdef.info.names:
            return None

        default_manager_cls = model_cls._meta.default_manager.__class__
        default_manager_fullname = helpers.get_class_fullname(default_manager_cls)

        try:
            default_manager_info = self.lookup_typeinfo_or_incomplete_defn_error(default_manager_fullname)
            # Check if default manager could be a generated manager
            base_manager_fullname = helpers.get_class_fullname(default_manager_cls.__bases__[0])
            generated_manager_info = self.get_generated_manager_info(default_manager_fullname, base_manager_fullname)
            if generated_manager_info is None:
                # Manager doesn't appear to be generated. Giving up by setting up a generic manager
                default_manager_info = self.lookup_typeinfo_or_incomplete_defn_error("django.db.models.manager.Manager")

        default_manager = Instance(default_manager_info, [Instance(self.model_classdef.info, [])])
        self.add_new_node_to_model_class("_default_manager", default_manager)
```

## How is that should be

`_default_manager` should be setup

## System information

- OS: docker python:3.10 image
- `python` version: 3.10
- `django` version: 3.2.13
- `mypy` version: 0.950
- `django-stubs` version: 1.11.0 and 24a3b22c60448b79647dfab8c12cb7d29da4ba65

aleksanb:
If a django model has a Manager class that cannot be resolved statically
(if it is generated in a way where we cannot import it, like `objects =
my_manager_factory()`), we fallback to the default related manager, so
you at least get a base level of working type checking.

Closes https://github.com/typeddjango/django-stubs/issues/981, https://github.com/typeddjango/django-stubs/issues/969

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -14,6 +14,7 @@ from mypy.types import Type as MypyType
 from mypy.types import TypedDictType, TypeOfAny
 
 from mypy_django_plugin.django.context import DjangoContext
+from mypy_django_plugin.errorcodes import MANAGER_MISSING
 from mypy_django_plugin.lib import fullnames, helpers
 from mypy_django_plugin.lib.fullnames import ANNOTATIONS_FULLNAME, ANY_ATTR_ALLOWED_CLASS_FULLNAME, MODEL_CLASS_FULLNAME
 from mypy_django_plugin.lib.helpers import add_new_class_for_module
@@ -341,6 +342,7 @@ class AddRelatedManagers(ModelClassInitializer):
                 continue
 
             if isinstance(relation, (ManyToOneRel, ManyToManyRel)):
+                related_manager_info = None
                 try:
                     related_manager_info = self.lookup_typeinfo_or_incomplete_defn_error(
                         fullnames.RELATED_MANAGER_CLASS
@@ -352,6 +354,26 @@ class AddRelatedManagers(ModelClassInitializer):
                     if not self.api.final_iteration:
                         raise exc
                     else:
+                        if related_manager_info:
+                            """
+                            If a django model has a Manager class that cannot be
+                            resolved statically (if it is generated in a way
+                            where we cannot import it, like `objects = my_manager_factory()`),
+                            we fallback to the default related manager, so you
+                            at least get a base level of working type checking.
+
+                            See https://github.com/typeddjango/django-stubs/pull/993
+                            for more information on when this error can occur.
+                            """
+                            self.add_new_node_to_model_class(
+                                attname, Instance(related_manager_info, [Instance(related_model_info, [])])
+                            )
+                            related_model_fullname = related_model_cls.__module__ + "." + related_model_cls.__name__
+                            self.ctx.api.fail(
+                                f"Couldn't resolve related manager for relation {relation.name!r} (from {related_model_fullname}.{relation.field}).",
+                                self.ctx.cls,
+                                code=MANAGER_MISSING,
+                            )
                         continue
 
                 # Check if the related model has a related manager subclassed from the default manager
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -640,6 +640,42 @@
                   pass
 
 
+-   case: test_related_managers_when_manager_is_dynamically_generated_and_cannot_be_imported
+    main: |
+        from myapp import models
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+
+                class User(models.Model):
+                    name = models.TextField()
+
+                def DynamicManager() -> models.Manager:
+                    class InnerManager(models.Manager):
+                        def some_method(self, arg: str) -> None:
+                            return None
+
+                    return InnerManager()
+
+                class Booking(models.Model):
+                    renter = models.ForeignKey(User, on_delete=models.PROTECT)
+                    owner = models.ForeignKey(User, on_delete=models.PROTECT, related_name='bookingowner_set')
+
+                    objects = DynamicManager()
+
+                def process_booking(user: User):
+                    reveal_type(user.bookingowner_set)
+                    reveal_type(user.booking_set)
+    out: |
+        myapp/models:3: error: Couldn't resolve related manager for relation 'booking' (from myapp.models.Booking.myapp.Booking.renter).
+        myapp/models:3: error: Couldn't resolve related manager for relation 'bookingowner_set' (from myapp.models.Booking.myapp.Booking.owner).
+        myapp/models:20: note: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.Booking]"
+        myapp/models:21: note: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.Booking]"
+
 -   case: foreign_key_relationship_for_models_with_custom_manager
     main: |
         from myapp.models import Transaction
--- a/mypy_django_plugin/errorcodes.py
+++ b/mypy_django_plugin/errorcodes.py
@@ -1,3 +1,4 @@
 from mypy.errorcodes import ErrorCode
 
 MANAGER_UNTYPED = ErrorCode("django-manager", "Untyped manager disallowed", "Django")
+MANAGER_MISSING = ErrorCode("django-manager-missing", "Couldn't resolve related manager for model", "Django")


mypy_django_plugin/transformers/models.py
```diff
@@ -345,15 +345,19 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
-                    default_manager = related_model_info.names.get("_default_manager")
-                    if not default_manager:
-                        raise helpers.IncompleteDefnException()
                 except helpers.IncompleteDefnException as exc:
                     if not self.api.final_iteration:
                         raise exc
                     else:
                         continue
 
+                default_manager = related_model_info.names.get("_default_manager")
```
sobolevn: Any chance we can test that? 🤔

sobolevn (commented):
Thanks!

mypy_django_plugin/transformers/models.py
aleksanb: I don't know if this change is correct though, i see there's some test diff as well.

aleksanb: And i've been unable to create a test-case that fails on main but passes here.

aleksanb: What should happen if there's no default manager? Do we need to run any of the code further down as well?

mypy_django_plugin/transformers/models.py
```diff
@@ -345,15 +345,19 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
                     related_manager_info = self.lookup_typeinfo_or_incomplete_defn_error(
                         fullnames.RELATED_MANAGER_CLASS
                     )  # noqa: E501
-                    default_manager = related_model_info.names.get("_default_manager")
```
aleksanb: I guess the reason we raise an exception here if it's not found is that the code is supposed to run multiple iterations over a project to actually finish?

aleksanb: But it never runs again.

aleksanb:
https://gitter.im/mypy-django/Lobby?at=62a86e2ef8daa71e07cbe011. Perhaps @flaeppe can chime in, I see you've written some of this code.

flaeppe:
> https://gitter.im/mypy-django/Lobby?at=62a86e2ef8daa71e07cbe011. Perhaps @flaeppe can chime in, I see you've written some of this code.

Be aware of that the traceback you've linked points to a manager in a package that isn't exporting types (i.e. there's no type hints in djmoney)

I would suspect that's why there's some misbehaviour.

aleksanb:
> Be aware of that the traceback you've linked points to a manager in a package that isn't exporting types (i.e. there's no type hints in djmoney)
> 
> I would suspect that's why there's some misbehaviour.

yes this seems to be the case!
Do you know if there's something i can do to surpress this error?

aleksanb:
The weird thing is that moneyed, the package, doesn't actually export any models that subclass django's models.Model.

flaeppe:
django-money is a Django extension of pymoneyed. pymoneyed has no stuff related to Django. While pymoneyed is typed and django-money isn't.

I could have a quick peek on the code later on, if you want

aleksanb:
I tried adding even more debug logging:
bilde

which gives

```
cls <class 'hyre.payment.models.models.SalesOrderRow'>
meta hyre__payment.salesorderrow
default_manager hyre__payment.SalesOrderRow.objects
dmclass <class 'djmoney.models.managers.money_manager.<locals>.MoneyManager'>
the default manager is <class 'djmoney.models.managers.money_manager.<locals>.MoneyManager'>
CHECK: adding default manager for <class 'hyre.payment.models.models.SalesOrderRow'> djmoney.models.managers.money_manager.<locals>.MoneyManager
lookin up typeinfo for  djmoney.models.managers.money_manager.<locals>.MoneyManager
fail No 'djmoney.models.managers.money_manager.<locals>.MoneyManager' found info is none
```

It thinks the default manager for some reason is MoneyManager.

aleksanb:
> I could have a quick peek on the code later on, if you want


That would be nice. I cannot share much of the code i'm working on, but i'm really not sure as to why this interaction with Money changes what the default_manager is.

aleksanb:
There we have it!
When i remove all my MoneyField fields from my model,
```
class SalesOrderRow(ReprMixin, TimestampedModel):
    # base_amount = MoneyField(max_digits=10, decimal_places=2, default_currency='NOK')
    # vat_amount = MoneyField(max_digits=10, decimal_places=2, default_currency='NOK')
    # total_amount = MoneyField(max_digits=10, decimal_places=2, default_currency='NOK')
    ... # other fields down here
```

i get
```
cls <class 'hyre.payment.models.models.SalesOrderRow'>
meta hyre__payment.salesorderrow
default_manager hyre__payment.SalesOrderRow.objects
dmclass <class 'django.db.models.manager.Manager'>
the default manager is <class 'django.db.models.manager.Manager'>
CHECK: adding default manager for <class 'hyre.payment.models.models.SalesOrderRow'> django.db.models.manager.Manager
lookin up typeinfo for  django.db.models.manager.Manager
SUCCES django.db.models.manager.Manager[hyre.payment.models.models.SalesOrderRow]
```

And the project sets up relatedmanagers correctly.

flaeppe:
Interesting, could it be some compatibility problem in `django-money`? I've used that package quite a lot, but haven't had any issues similar to yours.

I looked quickly through `django-money` and found these two things below, could that be related? Could your codebase perhaps be affected by it?

- https://github.com/django-money/django-money/blob/8ce85afd51f238a9f1793d9de1ffee4d299191fe/djmoney/_compat.py#L12-L17

- https://github.com/django-money/django-money/blob/c6c80d438e0f6138de37fc108917996407d98dca/docs/changes.rst#note-about-automatic-model-manager-patches

Also, are you relying on having `django-money` in your `INSTALLED_APPS`? Otherwise I'd suggest trying to remove it from there and see if that'd resolve your problem as well (if I'm reading stuff correctly that would remove triggering the "auto-patching" of default managers)

flaeppe:
Nevermind, seems that `apps.ready` won't affect it. It's not possible to turn off when I looked a bit further.

https://github.com/django-money/django-money/blob/c6c80d438e0f6138de37fc108917996407d98dca/djmoney/models/fields.py#L331-L344

In any case, I don't think this is an issue for `django-stubs`? And adding a fallback to always add a default manager, which I think changed here, feels like it could introduce quite a bit of false positives(?)

Might be reasonable if it we changed this here to emit an error if the manager couldn't be resolved and then attach a default manager. That way it's possible to explicitly ignore any error but still get all the django defaults of a related manager.

mypy_django_plugin/transformers/models.py
```diff
@@ -345,15 +345,19 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
                     if not self.api.final_iteration:
                         raise exc
                     else:
                         continue
 
+                default_manager = related_model_info.names.get("_default_manager")
+                if not default_manager:
+                    self.add_new_node_to_model_class(
+                        attname, Instance(related_manager_info, [Instance(related_model_info, [])])
+                    )
```
flaeppe: What if we move this piece inside the `else` block on line 351. Together with emitting an error? I think that feels most correct, as the model's manager can't be resolved, an error is displayed, and the default manager type attached still includes defaults from Django.

Emitting an error there feels quite helpful IMO, for debugging etc.

aleksanb:
This seems to be it exactly (django-money creating unreferenceable default managers).

> In any case, I don't think this is an issue for django-stubs? And adding a fallback to always add a default manager, which I think changed here, feels like it could introduce quite a bit of false positives(?)

Largest downside is that one cannot use django-money toghether with django-stubs then, even though it is django-stubs that is behaving unexpectedly by patching all default managers.

sobolevn:
> Might be reasonable if it we changed this here to emit an error if the manager couldn't be resolved and then attach a default manager. That way it's possible to explicitly ignore any error but still get all the django defaults of a related manager.

Seems reasonable. But, this will require a complete test suite! 👍

flaeppe:
> > Might be reasonable if it we changed this here to emit an error if the manager couldn't be resolved and then attach a default manager. That way it's possible to explicitly ignore any error but still get all the django defaults of a related manager.
> 
> Seems reasonable. But, this will require a complete test suite! 👍

I think, in theory, that's quite simple to trigger? Just put a random name or whatever that can't be resolved as a model's manager?

Problematic part might be that since django's setup is run, it might validate that all managers are correct..

mypy_django_plugin/transformers/models.py
aleksanb: This seems like a fair approach.
Do we have any prior art on emitting errors/warnings for certain nodes?

aleksanb:
> I think, in theory, that's quite simple to trigger? Just put a random name or whatever that can't be resolved as a model's manager?

I can try making a failing test where the manager is the return value of some function perhaps, so the symbol doesn't exist in scope?

mypy_django_plugin/transformers/models.py
aleksanb: ctx.api.fail it seems.

flaeppe:
> This seems to be it exactly (django-money creating unreferenceable default managers).
> 
> > In any case, I don't think this is an issue for django-stubs? And adding a fallback to always add a default manager, which I think changed here, feels like it could introduce quite a bit of false positives(?)
> 
> Largest downside is that one cannot use django-money toghether with django-stubs then, even though it is django-stubs that is behaving unexpectedly by patching all default managers.

It still kind of makes sense though. Since `django-money` isn't typed and `django-stubs` requires types. But I agree it's unfortunate that it does so.

flaeppe:
> > I think, in theory, that's quite simple to trigger? Just put a random name or whatever that can't be resolved as a model's manager?
> 
> I can try making a failing test where the manager is the return value of some function perhaps, so the symbol doesn't exist in scope?

That sounds like an approach that could work 👍 Might also work to monkey-patch the model class via that function (e.g. after model declaration), then we're very close to imitate what `django-money` does.

aleksanb:
```
hyre/damage/models.py:805: error: Couldn't resolve related manager, fallbacking to a stock django related manager.  [django-manager-missing]
```

sample error.

aleksanb:
All right! Updated with testcase (that catches the error!) and comments and descriptions, and separate error code so that i can surpress django-manager-missing on my side.

tests/typecheck/fields/test_related.yml
```diff
@@ -640,6 +640,51 @@
...
+        from myapp import models
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+                class TimestampedModel(models.Model):
+                    id = models.UUIDField(primary_key=True, editable=False)
```
sobolevn: Let's simplify this case. Do we need `id` to be explicit? Or can we remove it?

tests/typecheck/fields/test_related.yml
```diff
@@ -640,6 +640,51 @@
...
+                    created_at = models.DateTimeField(auto_now_add=True, db_index=True)
+                    updated_at = models.DateTimeField(auto_now=True, db_index=True)
```
sobolevn: Do we need this field?

tests/typecheck/fields/test_related.yml
```diff
@@ -640,6 +640,51 @@
...
+                    class Meta:
+                        abstract = True
+
+                class User(TimestampedModel):
+                    name = models.TextField()
+
+                def DynamicManager():
```
sobolevn: Right now `DynamicManager` is not typed checked at all. Do we need `-> None`?

sobolevn (commented):
Thank you! Great work!

tests/typecheck/fields/test_related.yml
sobolevn: Also, do we need boolean args here?

mypy_django_plugin/transformers/models.py
```diff
@@ -352,6 +354,25 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+                            # cannot be imported when we try to set up the
+                            # default related managers in
+                            # AddDefaultManagerAttribute), we fallback to a
+                            # default related manager, so you at least get a
+                            # base level of working type checking.
+                            self.add_new_node_to_model_class(
+                                attname, Instance(related_manager_info, [Instance(related_model_info, [])])
+                            )
+                            self.ctx.api.fail(
+                                f"Couldn't resolve related manager for relation {relation.name}, constructing a default one instead.",
```
flaeppe: Perhaps it's a bit over the top to mention that a default is used?

```suggestion
                                f"Couldn't resolve related manager for relation {relation.name}.",
```

mypy_django_plugin/transformers/models.py
```diff
@@ -352,6 +354,25 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+                            # If a django model has a Manager class that cannot
+                            # be resolved on runtime (like django-money
+                            # monkeypatches all default managers after models
+                            # already have been declared, and generates this
+                            # manager as the return value of a function, so it
+                            # cannot be imported when we try to set up the
+                            # default related managers in
+                            # AddDefaultManagerAttribute), we fallback to a
+                            # default related manager, so you at least get a
+                            # base level of working type checking.
```
flaeppe: IMO, involving `django-money` in this comment is a bit too much.

```suggestion
                            # If a django model has a Manager class that cannot
                            # be resolved, we fallback to the default related manager,
                            # so you at least get a base level of working type checking.
```

I mean, if `django-money` changes behaviour this comment becomes obsolete.

tests/typecheck/fields/test_related.yml
```diff
@@ -640,6 +640,51 @@
...
+                        ordering = ['-created_at']
+
+                    objects = DynamicManager()
+
+                def process_booking(user: User):
+                    reveal_type(user.bookingowner_set)
+                    reveal_type(user.booking_set)
+    out: |
+        myapp/models:9: error: Couldn't resolve related manager for relation booking, constructing a default one instead.
+        myapp/models:9: error: Couldn't resolve related manager for relation bookingowner_set, constructing a default one instead.
```
flaeppe: Thought: I'm wondering if it could be valuable to try to include the path/fullname to the relation declaration as well?

It's good that the name of the reverse relation is there. But maybe it's also nice if e.g. `myapp.models.Booking.x` was in the message too?

e.g.

```suggestion
        myapp/models:9: error: Couldn't resolve related manager for relation booking, constructing a default one instead. (myapp.models.Booking.renter)
        myapp/models:9: error: Couldn't resolve related manager for relation bookingowner_set, constructing a default one instead. (myapp.models.Booking.owner)
```

I think we have that context/data available when emitting, right? Otherwise just skip this.

mypy_django_plugin/transformers/models.py
flaeppe:
```suggestion
                                f"Couldn't resolve related manager for relation {relation.name!r}, constructing a default one instead.",
```

Should wrap the name inside of: `''`. e.g. `'bookingowner_set'`

flaeppe (commented):
Nice work 👍

I added a few thoughts.

flaeppe:
I also think merging this PR would close #981 as that similarly uses a package (`django-modeltranslations`) that isn't exporting types.

mypy_django_plugin/transformers/models.py
sobolevn: Not `'`, but `"`

aleksanb:
Updated with better comments, warnings, simpler test case etc. @sobolevn.

sobolevn (approved):
LGTM, great work! 👍

sobolevn:
@flaeppe do you have any feedback? :)

flaeppe (approved):
LGTM too. Nice work 👍

aleksanb:
Wohoo!

What's the release schedule nowadays?
is it easy to cut a new release for pypi?

aleksanb:
Hm, it seems like you cannot surpress mypy errors from extensions through `--disable-error-code django-manager-missing` and friends:
```
➜  app git:(mypy-wip-again-june) ✗ make mypy
usage: mypy [-h] [-v] [-V] [more options; see below]
            [-m MODULE] [-p PACKAGE] [-c PROGRAM_TEXT] [files ...]
mypy: error: Invalid error code(s): django-manager-missing
```

aleksanb:
https://github.com/python/mypy/issues/12987

mannyanebi:
Following this to see how this gets resolved. Having this `mypy: error: Invalid error code(s): django-manager-missing` can't seem to suppress it

PR #994, opened 2022-06-14
Add Template class to the django template backend

SmileyChris:
`django.template.backends.django` is missing a definition of its `Template` class.


sobolevn (changes requested):
Please, take a look at:

```
/home/runner/work/django-stubs/django-stubs/django-stubs/template/backends/django.pyi:15: error: Return type "django.template.backends.django.Template" of "from_string" incompatible with return type "django.template.base.Template" in supertype "BaseEngine"
```

django-stubs/template/backends/django.pyi
```diff
@@ -9,6 +12,17 @@ class DjangoTemplates(BaseEngine):
     engine: Engine = ...
     def __init__(self, params: Dict[str, Any]) -> None: ...
     def get_templatetag_libraries(self, custom_libraries: Dict[str, str]) -> Dict[str, str]: ...
+    def from_string(self, template_code: str) -> Template: ...
+    def get_template(self, template_name: str) -> Template: ...
+
+class Template:
+    template: _BaseTemplate
+    backend: BaseEngine
```
smithdc1, line 20: Not `DjangoTemplates`?

django-stubs/template/backends/django.pyi
```diff
@@ -9,6 +12,17 @@ class DjangoTemplates(BaseEngine):
...
+
+    def __init__(self, template: _BaseTemplate, backend: BaseEngine) -> None: ...
+    @property
+    def origin(self) -> _Origin: ...
+    def render(self, context: Optional[Dict[str, Any]] = None, request: Optional[_HttpRequest] = None) -> str: ...
```
smithdc1, line 25: `context` could be a `Context` too? 

https://github.com/typeddjango/django-stubs/blob/f4407155a56a50f70872d793ae365bd7d1a8691e/django-stubs/template/base.pyi#L75

smithdc1 (commented):
> /home/runner/work/django-stubs/django-stubs/django-stubs/template/backends/django.pyi:15: error: Return type "django.template.backends.django.Template" of "from_string" incompatible with return type "django.template.base.Template" in supertype "BaseEngine"

It looks like the `BaseEngine` has the incorrect `Template` class. 

I'm new to typing but maybe we need a protocol here that can be used for the various `Template` classes in the `backend` files.

``` python
class TemplateProto(Protocol):
    def render(
        self, 
        context: Optional[Union[Context, Dict[str, Any]]], 
        request: Optional[HttpRequest]
    ) -> Union[SafeString, TemplateDoesNotExist]: ...
```

UnknownPlatypus:
Superseded by #1302

PR #998, opened 2022-06-16
Set type of default `django.core.cache.cache` to `BaseCache`

flaeppe:
The previous type `ConnectionProxy` is just a proxy class, revealing `Any` for _all_ cache methods instead of exposing the structure of the underlying protocol (`BaseCache`).

The typing here could probably be more advanced, although I think this would be sufficient as some sort of improvement from `ConnectionProxy`.


flaeppe:
Perhaps slightly off topic, but should I include the `async def`s from `BaseCache` which can be found at `HEAD` on Django? They're not due until Django 4.1 but perhaps there's no harm in including them now anyways?

sobolevn:
> They're not due until Django 4.1 but perhaps there's no harm in including them now anyways?

I am a bit old-fashioned on these topics 😉 
Let's wait a bit more.

django-stubs/core/cache/backends/base.pyi
```diff
@@ -18,22 +18,22 @@ class BaseCache:
...
-    def add(self, key: Any, value: Any, timeout: Any = ..., version: Optional[Any] = ...) -> bool: ...
-    def get(self, key: Any, default: Optional[Any] = ..., version: Optional[Any] = ...) -> Any: ...
-    def set(self, key: Any, value: Any, timeout: Any = ..., version: Optional[Any] = ...) -> None: ...
-    def touch(self, key: Any, timeout: Any = ..., version: Optional[Any] = ...) -> bool: ...
-    def delete(self, key: Any, version: Optional[Any] = ...) -> None: ...
+    def make_key(self, key: Any, version: Optional[int] = ...) -> str: ...
+    def add(self, key: Any, value: Any, timeout: Any = ..., version: Optional[int] = ...) -> bool: ...
+    def get(self, key: Any, default: Optional[Any] = ..., version: Optional[int] = ...) -> Any: ...
+    def set(self, key: Any, value: Any, timeout: Any = ..., version: Optional[int] = ...) -> None: ...
+    def touch(self, key: Any, timeout: Any = ..., version: Optional[int] = ...) -> bool: ...
```
sobolevn: Is `timeout` really `Any`?

tests/typecheck/core/test_cache.yml
```diff
@@ -0,0 +1,23 @@
+-   case: test_default_cache_exposes_base_cache_methods
```
sobolevn: Thanks, testing things is great. But, this is a bit of an overkill.

We generally test dynamic / plugin related / complex generic code. Regular annotations are not tested: they will just make things harder.

flaeppe: Alrighty, I'll drop it.

django-stubs/core/cache/backends/base.pyi
flaeppe: Nope, looking closer now, I see that they resolve the default arg that is set as `object()`:

https://github.com/django/django/blob/2a2bde52f31e09e95ce616e8e6bc0ffeb68f76c9/django/core/cache/backends/base.py#L85-L95

Looks like it should be an `Optional[int]`. Let me change that.

sobolevn (approved):
Thanks!

AngellusMortis:
The async interface for `BaseCache` was added in 4.0, not 4.1. This is now causing mypy errors for 4.0.5 saying `aget` / `aset` does not exist. 

https://github.com/django/django/blob/4.0.5/django/core/cache/backends/base.py#L149

4.1 adds the async ORM interface. 4.0 was cache interface.

PR #999, opened 2022-06-16
Emit error and set fallback type for managers that can't be resolved

### LINKED ISSUE #981: Default manager is not setup with `django-modeltranslations`

flaeppe:
I extended the work done in #993 by also emitting an error when the type for a manager, in general, can't be determined. In addition to only related managers.

There I also incorrectly stated (https://github.com/typeddjango/django-stubs/pull/993#issuecomment-1155592098) that it'd close #981. However, the changes here is what should actually close #981.

A small question here is if the errors always comes in pair? e.g. if a model's manager can't be resolved, all related managers (always based on that?) can't be resolved.
I've assumed that it's not true here, in case there would be some edge case where a _related_ manager can't be resolved, while the regular manager for that model __can__.

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -5,7 +5,7 @@ from django.db.models.fields import DateField, DateTimeField, Field
 from django.db.models.fields.related import ForeignKey
 from django.db.models.fields.reverse_related import ManyToManyRel, ManyToOneRel, OneToOneRel
 from mypy.checker import TypeChecker
-from mypy.nodes import ARG_STAR2, Argument, Context, FuncDef, TypeInfo, Var
+from mypy.nodes import ARG_STAR2, Argument, AssignmentStmt, Context, FuncDef, NameExpr, TypeInfo, Var
 from mypy.plugin import AnalyzeTypeContext, AttributeContext, CheckerPluginInterface, ClassDefContext
 from mypy.plugins import common
 from mypy.semanal import SemanticAnalyzer
@@ -234,7 +234,7 @@ class AddManagers(ModelClassInitializer):
     def run_with_model_cls(self, model_cls: Type[Model]) -> None:
         manager_info: Optional[TypeInfo]
 
-        encountered_incomplete_manager_def = False
+        incomplete_manager_defs = set()
         for manager_name, manager in model_cls._meta.managers_map.items():
             manager_class_name = manager.__class__.__name__
             manager_fullname = helpers.get_class_fullname(manager.__class__)
@@ -247,7 +247,7 @@ class AddManagers(ModelClassInitializer):
                 if manager_info is None:
                     # Manager doesn't appear to be generated. Track that we encountered an
                     # incomplete definition and skip
-                    encountered_incomplete_manager_def = True
+                    incomplete_manager_defs.add(manager_name)
                     continue
                 _, manager_class_name = manager_info.fullname.rsplit(".", maxsplit=1)
 
@@ -275,9 +275,36 @@ class AddManagers(ModelClassInitializer):
 
                 self.add_new_node_to_model_class(manager_name, custom_manager_type)
 
-        if encountered_incomplete_manager_def and not self.api.final_iteration:
-            #  Unless we're on the final round, see if another round could figuring out all manager types
+        if incomplete_manager_defs and not self.api.final_iteration:
+            #  Unless we're on the final round, see if another round could figure out all manager types
             raise helpers.IncompleteDefnException()
+        elif self.api.final_iteration:
+            for manager_name in incomplete_manager_defs:
+                # We act graceful and set the type as the bare minimum we know of
+                # (Django's default) before finishing. And emit an error, to allow for
+                # ignoring a more specialised manager not being resolved while still
+                # setting _some_ type
+                django_manager_info = self.lookup_typeinfo(fullnames.MANAGER_CLASS_FULLNAME)
+                assert django_manager_info is not None
+                self.add_new_node_to_model_class(
+                    manager_name, Instance(django_manager_info, [Instance(self.model_classdef.info, [])])
+                )
+                # Find expression for e.g. `objects = SomeManager()`
+                manager_expr = [
+                    expr
+                    for expr in self.ctx.cls.defs.body
+                    if (
+                        isinstance(expr, AssignmentStmt)
+                        and isinstance(expr.lvalues[0], NameExpr)
+                        and expr.lvalues[0].name == manager_name
+                    )
+                ]
+                manager_fullname = f"{self.model_classdef.fullname}.{manager_name}"
+                self.api.fail(
+                    f"Couldn't resolve manager type for {manager_fullname!r}",
+                    manager_expr[0] if manager_expr else self.ctx.cls,
+                    code=MANAGER_MISSING,
+                )
 
 
 class AddDefaultManagerAttribute(ModelClassInitializer):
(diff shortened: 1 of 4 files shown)


mypy_django_plugin/transformers/models.py
```diff
@@ -275,9 +275,36 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+            #  Unless we're on the final round, see if another round could figure out all manager types
             raise helpers.IncompleteDefnException()
+        elif self.api.final_iteration:
+            for manager_name in incomplete_manager_defs:
+                # We act graceful and set the type as the bare minimum we know of
+                # (Django's default) before finishing. And emit an error, to allow for
+                # ignoring a more specialised manager not being resolved while still
+                # setting _some_ type
+                django_manager_info = self.lookup_typeinfo(fullnames.MANAGER_CLASS_FULLNAME)
+                assert django_manager_info is not None
```
sobolevn: Error message here would help in unexpected case.

mypy_django_plugin/transformers/models.py
```diff
@@ -275,9 +275,36 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+                    for expr in self.ctx.cls.defs.body
+                    if (
+                        isinstance(expr, AssignmentStmt)
+                        and isinstance(expr.lvalues[0], NameExpr)
+                        and expr.lvalues[0].name == manager_name
+                    )
+                ]
+                manager_fullname = f"{self.model_classdef.fullname}.{manager_name}"
+                self.api.fail(
+                    f"Couldn't resolve manager type for {manager_fullname!r}",
```
sobolevn:
```suggestion
                    f'Could not resolve manager type for "{manager_fullname!r}"',
```

tests/typecheck/managers/test_managers.yml
```diff
@@ -442,3 +442,83 @@
...
+            content: |
+                from django.db import models
+
+                def LocalManager() -> models.Manager:
+                    """
+                    Returns a manager instance of an inlined manager type that can't
+                    be resolved.
+                    """
+                    class InnerManager(models.Manager):
+                        def some_method(self, arg: str) -> None:
```
sobolevn: Do we really need this method here? Or just dynamic class is fine?

mypy_django_plugin/transformers/models.py
sobolevn: Mypy uses `"` to highlight output :)

sobolevn (commented):
Thank you! This looks consistent!

tests/typecheck/managers/test_managers.yml
flaeppe: Nah, shouldn't be needed.

mypy_django_plugin/transformers/models.py
flaeppe: Good point, didn't think about that

mypy_django_plugin/transformers/models.py
flaeppe: I tried to come up with some helpful message here, let me know if you have any better suggestion.

mypy_django_plugin/transformers/models.py
```diff
@@ -243,13 +243,11 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
-                manager_info = self.get_generated_manager_info(manager_fullname, base_manager_fullname)
-                if manager_info is None:
+                if manager_fullname not in self.get_generated_manager_mappings(base_manager_fullname):
                     # Manager doesn't appear to be generated. Track that we encountered an
                     # incomplete definition and skip
-                    encountered_incomplete_manager_def = True
-                    continue
-                _, manager_class_name = manager_info.fullname.rsplit(".", maxsplit=1)
+                    incomplete_manager_defs.add(manager_name)
+                continue
```
flaeppe, line 250: This changeset triggered some presumably premature assumption that dynamically generated managers can be looked up and have a type/typeinfo during semantic analysis (where they are prepared). That shouldn't be possible until the type checker has run. Which, I think, only happens after semantic analysis.

That's why I dropped the continuation if `manager_info` was found, I don't think it's possible for it to happen.

tests/typecheck/fields/test_related.yml
```diff
@@ -649,10 +649,12 @@
         -   path: myapp/models.py
             content: |
                 from django.db import models
+                from django.db.models.manager import BaseManager
                 class TransactionQuerySet(models.QuerySet):
                     pass
+                TransactionManager = BaseManager.from_queryset(TransactionQuerySet)
                 class Transaction(models.Model):
-                    objects = TransactionQuerySet.as_manager()
```
sobolevn, line 692: This is also important. Can we still be sure that this example works just fine?

sobolevn (commented):
👍

tests/typecheck/fields/test_related.yml
flaeppe: As far as I am aware there's no support for .as_manager, it'd basically need a corresponding approach to what .from_queryset has gotten. (Which I don't think exists?)

This triggered the newly introduced error, as the dynamically created type that .as_manager produces can't be found during type checking.

But I don't think we're introducing anything new here other than highlighting that .as_manager doesn't work. With an error

flaeppe: Hm, actually I think we are improving here. Since one can now ignore the error and get a manager type, instead of a silent Any.

sobolevn: Let's add an explicit test case for this then! 👍

sobolevn:
I will add a test case myself, thanks!

flaeppe:
Cool, a new test case sounds reasonable!

PR #1011, opened 2022-06-21
Improved type of template.loader.get_template()

smithdc1:
[`get_template()`](https://github.com/django/django/blob/de74a74b4b889c986cabab837bb03e2113880afe/django/template/loader.py#L5) calls `get_template()` on each `engine` and returns the [`Template`](https://github.com/django/django/blob/de74a74b4b889c986cabab837bb03e2113880afe/django/template/backends/django.py#L32) for that engine. 

That is, this is the `Template` class from the relevant backend files rather than the `Template` in `base.py`. 

This is also intended to fix the issue seen here. https://github.com/typeddjango/django-stubs/pull/994#pullrequestreview-1005806821


django-stubs/template/loader.pyi
```diff
@@ -4,8 +4,9 @@ from django.http.request import HttpRequest
 from django.template.exceptions import TemplateDoesNotExist as TemplateDoesNotExist  # noqa: F401
 
 from . import engines as engines  # noqa: F401
+from .backends.base import EngineTemplate
 
-def get_template(template_name: str, using: Optional[str] = ...) -> Any: ...
+def get_template(template_name: str, using: Optional[str] = ...) -> Union[EngineTemplate, TemplateDoesNotExist]: ...
```
sobolevn: As far as I can tell, `TemplateDoesNotExist` is not returned, but is raised: https://github.com/django/django/blob/d80a2585532ef15184afed6cf4ec1358198a894d/django/template/loader.py#L19

django-stubs/template/backends/base.pyi
```diff
@@ -9,8 +12,15 @@ class BaseEngine:
     def __init__(self, params: Mapping[str, Any]) -> None: ...
     @property
     def app_dirname(self) -> Optional[str]: ...
-    def from_string(self, template_code: str) -> Template: ...
-    def get_template(self, template_name: str) -> Optional[Template]: ...
+    def from_string(self, template_code: str) -> EngineTemplate: ...
+    def get_template(self, template_name: str) -> Optional[EngineTemplate]: ...
```
sobolevn: Looks like this exception is not returned, but raised: https://github.com/django/django/blob/d80a2585532ef15184afed6cf4ec1358198a894d/django/template/loader.py#L15

django-stubs/template/loader.pyi
smithdc1: So I learnt something here!

My understanding, having done a bit more reading, is that exceptions are not typed and therefore the type here should just be `EngineTemplate`?

sobolevn: Yes :)

sobolevn: If you want to type exceptions, take a look at (unrelated) project: https://github.com/dry-python/returns

django-stubs/template/backends/base.pyi
```diff
@@ -9,8 +12,15 @@ class BaseEngine:
...
     def app_dirname(self) -> Optional[str]: ...
-    def from_string(self, template_code: str) -> Template: ...
-    def get_template(self, template_name: str) -> Optional[Template]: ...
+    def from_string(self, template_code: str) -> EngineTemplate: ...
+    def get_template(self, template_name: str) -> EngineTemplate: ...
     @property
     def template_dirs(self) -> Tuple[str]: ...
     def iter_template_filenames(self, template_name: str) -> Iterator[str]: ...
+
+class EngineTemplate(Protocol):
```
sobolevn: What about other props?
Like `template`, `origin`, and `backend`?

Examples:
- https://github.com/django/django/blob/de74a74b4b889c986cabab837bb03e2113880afe/django/template/backends/django.py#L48-L64
- https://github.com/django/django/blob/de74a74b4b889c986cabab837bb03e2113880afe/django/template/backends/jinja2.py#L55

django-stubs/template/backends/base.pyi
```diff
@@ -9,8 +12,15 @@ class BaseEngine:
...
+    def render(
+        self,
+        context: Optional[Union[Context, Dict[str, Any]]] = ...,
+        request: Optional[HttpRequest] = ...,
+    ) -> Union[SafeString, TemplateDoesNotExist]: ...
```
smithdc1: This should just be `SafeString`

django-stubs/template/backends/base.pyi
smithdc1: As well as the Django and Jinja classes, I think it could also be `dummy.Template` which only has `render()`?

https://github.com/django/django/blob/6f63e0ce8e36fbdb2d0c4dfd77ed3cf02e467f49/django/template/backends/dummy.py#L43

sobolevn: Good point! 👍

sobolevn:
```suggestion
class _EngineTemplate(Protocol):
```

sobolevn (approved):
Thank you! 👍

PR #1012, opened 2022-06-22
BaseContext.__copy__

smithdc1:
Currently if a copy of `Context` is made (or indeed any other subclass of `BaseContext`)  the type returned is `BaseContext` rather than the type of the current instance. 

``` python
from django.template import Context

c = Context()
a = c.__copy__()
a.update({}) # error: "BaseContext" has no attribute "update"
```

I think ideally we'd use the new `Self` type here? But that's not yet implemented in `mypy`?


django-stubs/template/context.pyi
```diff
@@ -1,12 +1,13 @@
...
-from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Type, Union
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Type, TypeVar, Union
 
 from django.http.request import HttpRequest
 from django.template.base import Node, Origin, Template
 from django.template.defaulttags import IfChangedNode
 from django.template.loader_tags import IncludeNode
 
 _ContextValues = Union[Dict[str, Any], "Context"]
+ContextCopy = TypeVar("ContextCopy", bound="BaseContext")
```
sobolevn:
```suggestion
_ContextCopy = TypeVar("_ContextCopy", bound="BaseContext")
```

This needs to be protected, not public.

smithdc1: Thank you for the super quick review and feedback!

django-stubs/template/context.pyi
```diff
@@ -18,7 +19,7 @@ class ContextDict(dict):
 
 class BaseContext(Iterable[Any]):
     def __init__(self, dict_: Any = ...) -> None: ...
-    def __copy__(self) -> BaseContext: ...
+    def __copy__(self: ContextCopy) -> ContextCopy: ...
```
sobolevn:
```suggestion
    def __copy__(self: _ContextCopy) -> _ContextCopy: ...
```

sobolevn (approved):
Thank you! Similar fixes are welcome :)

PR #1015, opened 2022-06-22
Update RunSQL type to include parametrized sql

lachtanek:
`RunSQL` supports parametrized SQLs by passing list of 2-tuples, where first item is the SQL and second item is the list of parameters. Technically, the 2-tuples could also be 2-item lists, but afaik there's no way to type that, so I just went with strictly tuple there.

## Related issues
No issue I've found related to this, probably not used that frequently


sobolevn (approved):
Thank you!

PR #1016, opened 2022-06-22
Add missing type for TestCase.captureOnCommitCallbacks

### LINKED ISSUE #933: Missing TestCase.captureOnCommitCallbacks
Django 3.2 added `TestCase.captureOnCommitCallbacks()` context manager but this typing package is missing annotations for it.

https://github.com/django/django/blob/stable/3.2.x/django/test/testcases.py#L1260

Checked 1.10.1, master.

lachtanek:
Since version 3.2, Django has added `captureOnCommitCallbacks`, see [docs](https://docs.djangoproject.com/en/4.0/topics/testing/tools/#django.test.TestCase.captureOnCommitCallbacks)

## Related issues
Fixes #933


django-stubs/test/testcases.pyi
```diff
@@ -183,6 +183,7 @@ class TransactionTestCase(SimpleTestCase):
 class TestCase(TransactionTestCase):
     @classmethod
     def setUpTestData(cls) -> None: ...
+    def captureOnCommitCallbacks(cls, *, using: Optional[str] = ..., execute: bool = False): ...
```
sobolevn:
```suggestion
    def captureOnCommitCallbacks(cls, *, using: Optional[str] = ..., execute: bool = ...): ...
```

lachtanek: fixed, thanks, wasn't sure whether to include the default value or not

sobolevn (approved):

PR #1018, opened 2022-06-23
Set generic type of SimpleCookie

w0rp:
Pyright complains about `response.cookies` because the generic type isn't known. `str` may or may not be the correct type to use here. Something should be set here.

## Related issues

There are no related issues. I just saw this in my editor and wanted to point it out here.

### DIFF
--- a/django-stubs/template/response.pyi
+++ b/django-stubs/template/response.pyi
@@ -17,7 +17,7 @@ class ContentNotRenderedError(Exception): ...
 class SimpleTemplateResponse(HttpResponse):
     content: Any = ...
     closed: bool
-    cookies: SimpleCookie
+    cookies: SimpleCookie[str]
     status_code: int
     rendering_attrs: Any = ...
     template_name: _TemplateForResponseT = ...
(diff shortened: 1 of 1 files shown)


django-stubs/template/response.pyi
```diff
@@ -17,7 +17,7 @@ class ContentNotRenderedError(Exception): ...
 class SimpleTemplateResponse(HttpResponse):
     content: Any = ...
     closed: bool
-    cookies: SimpleCookie
+    cookies: SimpleCookie[str]
```
sobolevn, line 20: Or `bytes`? Can you please double check that?

w0rp: I _think_ it's meant to be `str`.

I also just noticed there are other instances of `SimpleCookie` in the codebase without type parameterts, such as in `test/client.pyi`. Maybe someone more familiar with cookies will come along and fix these types.

sobolevn: Yes, it is `str` as per docs: https://docs.djangoproject.com/en/4.0/ref/request-response/#django.http.HttpRequest.COOKIES

sobolevn (approved):
Thank you! Please, feel free to fix other cookies as well!

PR #1020, opened 2022-06-23
Fix nullability of blank charfields in values and values_list

ljodal:
Char fields with `blank=True` set should not be considered nullable in the context of `values()` and `values_list()` querysets.

I'm also not a huge fan of the way these fields are made optional in the constructur to the model classes, it feels like it would be better to mark the arguments as having a default value, rather than allow sending in `None`, but I'd rather keep this fix small and look at the overall problem at a later point.

### DIFF
--- a/tests/typecheck/managers/querysets/test_values.yml
+++ b/tests/typecheck/managers/querysets/test_values.yml
@@ -124,3 +124,21 @@
                     pass
                 class Book(models.Model):
                     authors = models.ManyToManyField(Author, related_name='books')
+
+-   case: queryset_values_blank_charfield
+    main: |
+        from myapp.models import Blog
+        values = Blog.objects.values('text').get()
+        reveal_type(values)  # N: Revealed type is "TypedDict({'text': builtins.str})"
+        reveal_type(values["text"])  # N: Revealed type is "builtins.str"
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+
+                class Blog(models.Model):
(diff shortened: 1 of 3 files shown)


tests/typecheck/managers/querysets/test_values.yml
```diff
@@ -124,3 +124,21 @@
                     pass
                 class Book(models.Model):
                     authors = models.ManyToManyField(Author, related_name='books')
+
+-   case: queryset_values_blank_charfield
+    main: |
+        from myapp.models import Blog
+        values = Blog.objects.values('text').get()
```
sobolevn, line 131: Forgive me my lack of knowledge here, but what will happen in runtime?

`None` -> `''`?

ljodal: I'm not sure if I'm fully understanding what you're asking, but setting `blank=True` on a field does not affect the nullability of that field in the database. All it does is enable a Django-level check to either allow or disallow empty strings to be saved. That means that in the database the field will not accept null values unless `null=True`. An empty string will be stored as just that, and empty string, and when querying the database it will always return a string value, either empty or not.

This plugin will type blank CharFields as `Optional[str]`, even if that's never gonna happen at runtime because the database (at least a strictly type checking one like Postgres, not sure about SQLite) will never allow you to store null values in the column. So that's what I'm fixing here, making the type of the returned columns non-optional as that's the runtime behavior

sobolevn: Thank you, I get it now! 👍

sobolevn (approved):
Any further `blank=` and `null=` improvements are very welcome!

PR #1025, opened 2022-06-27
Implement support for `<QuerySet>.as_manager()`

### LINKED ISSUE #324: Make as_manager() in queryset generic instead of using Any

flaeppe:
I've reorganised and extended dynamic manager class generation to support `<QuerySet>.as_manager()`.

This also renames the dynamically generated manager types to align with what those class names will be during runtime, set by Django. See here: https://github.com/django/django/blob/b2eff16806057095c7dd3daa9402ad615e51627f/django/db/models/manager.py#L109-L110
That means that `reveal_type(<Model>.<manager>)` should output the same type name as `type(<Model>.<manager>)` (unless there's collisions etc etc.)

## Related issues
Closes: #324

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,9 +145,7 @@ And then use `AuthenticatedHttpRequest` instead of the standard `HttpRequest` fo
 
 ### My QuerySet methods are returning Any rather than my Model
 
-`QuerySet.as_manager()` is not currently supported.
-
-If you are using `MyQuerySet.as_manager()`, then your `Manager`/`QuerySet` methods will all not be linked to your model.
+If you are using `MyQuerySet.as_manager()`:
 
 Example:
 
@@ -163,12 +161,12 @@ class MyModel(models.Model):
     objects = MyModelQuerySet.as_manager()
 
 
-def use_my_model():
-    foo = MyModel.objects.get(id=1) # This is `Any` but it should be `MyModel`
-    return foo.xyz # No error, but there should be
+def use_my_model() -> int:
+    foo = MyModel.objects.get(id=1) # Should now be `MyModel`
+    return foo.xyz # Gives an error
 ```
 
-There is a workaround: use `Manager.from_queryset` instead.
+Or if you're using `Manager.from_queryset`:
 
 Example:
 
@@ -188,9 +186,9 @@ class MyModel(models.Model):
     objects = MyModelManager()
 
 
-def use_my_model():
-    foo = MyModel.objects.get(id=1)
-    return foo.xyz  # Gives an error
+def use_my_model() -> int:
+    foo = MyModel.objects.get(id=1) # Should now be `MyModel`
+    return foo.xyz # Gives an error
 ```
 
 ### How do I annotate cases where I called QuerySet.annotate?
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -24,6 +24,7 @@ from mypy_django_plugin.lib import fullnames, helpers
 from mypy_django_plugin.transformers import fields, forms, init_create, meta, querysets, request, settings
 from mypy_django_plugin.transformers.functional import resolve_str_promise_attribute
 from mypy_django_plugin.transformers.managers import (
+    create_new_manager_class_from_as_manager_method,
     create_new_manager_class_from_from_queryset_method,
     resolve_manager_method,
 )
@@ -301,11 +302,15 @@ class NewSemanalDjangoPlugin(Plugin):
 
     def get_dynamic_class_hook(self, fullname: str) -> Optional[Callable[[DynamicClassDefContext], None]]:
         # Create a new manager class definition when a manager's '.from_queryset' classmethod is called
-        if fullname.endswith("from_queryset"):
-            class_name, _, _ = fullname.rpartition(".")
+        class_name, _, method_name = fullname.rpartition(".")
+        if method_name == "from_queryset":
             info = self._get_typeinfo_or_none(class_name)
             if info and info.has_base(fullnames.BASE_MANAGER_CLASS_FULLNAME):
                 return create_new_manager_class_from_from_queryset_method
+        elif method_name == "as_manager":
+            info = self._get_typeinfo_or_none(class_name)
+            if info and info.has_base(fullnames.QUERYSET_CLASS_FULLNAME):
+                return create_new_manager_class_from_as_manager_method
         return None
 
 
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -429,8 +429,8 @@ def create_new_manager_class_from_as_manager_method(ctx: DynamicClassDefContext)
                 with_unique_name=True,
             )
         except helpers.IncompleteDefnException:
-            if not ctx.api.final_iteration:
-                ctx.api.defer()
+            if not semanal_api.final_iteration:
+                semanal_api.defer()
             return
 
         populate_manager_from_queryset(new_manager_info, queryset_info)
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -15,7 +15,8 @@ from mypy.nodes import (
     TypeInfo,
     Var,
 )
-from mypy.plugin import AttributeContext, DynamicClassDefContext, SemanticAnalyzerPluginInterface
+from mypy.plugin import AttributeContext, DynamicClassDefContext
+from mypy.semanal import SemanticAnalyzer
 from mypy.semanal_shared import has_placeholder
 from mypy.types import AnyType, CallableType, Instance, ProperType
 from mypy.types import Type as MypyType
@@ -150,7 +151,6 @@ def get_method_type_from_reverse_manager(
 
 
 def resolve_manager_method_from_instance(instance: Instance, method_name: str, ctx: AttributeContext) -> MypyType:
-
     api = helpers.get_typechecker_api(ctx)
     method_type = get_method_type_from_dynamic_manager(
         api, method_name, instance
@@ -164,8 +164,10 @@ def resolve_manager_method(ctx: AttributeContext) -> MypyType:
     A 'get_attribute_hook' that is intended to be invoked whenever the TypeChecker encounters
     an attribute on a class that has 'django.db.models.BaseManager' as a base.
     """
-    # Skip (method) type that is currently something other than Any
+    # Skip (method) type that is currently something other than Any of type `implementation_artifact`
     if not isinstance(ctx.default_attr_type, AnyType):
+        return ctx.default_attr_type
+    elif ctx.default_attr_type.type_of_any != TypeOfAny.implementation_artifact:
         return ctx.default_attr_type
 
     # (Current state is:) We wouldn't end up here when looking up a method from a custom _manager_.
@@ -197,12 +199,12 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
         return
 
     # Don't redeclare the manager class if we've already defined it.
-    manager_node = semanal_api.lookup_current_scope(ctx.name)
-    if manager_node and isinstance(manager_node.node, TypeInfo):
+    manager_sym = semanal_api.lookup_current_scope(ctx.name)
+    if manager_sym and isinstance(manager_sym.node, TypeInfo):
         # This is just a deferral run where our work is already finished
         return
 
-    new_manager_info = create_manager_info_from_from_queryset_call(ctx.api, ctx.call, ctx.name)
+    new_manager_info = create_manager_info_from_from_queryset_call(semanal_api, ctx.call, ctx.name)
     if new_manager_info is None:
         if not ctx.api.final_iteration:
             ctx.api.defer()
@@ -212,8 +214,17 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
     helpers.add_new_manager_base(semanal_api, new_manager_info.fullname)
 
 
+def register_dynamically_created_manager(fullname: str, manager_name: str, manager_base: TypeInfo) -> None:
+    manager_base.metadata.setdefault("from_queryset_managers", {})
+    # The `__module__` value of the manager type created by Django's
+    # `.from_queryset` is `django.db.models.manager`. But we put new type(s) in the
+    # module currently being processed, so we'll map those together through metadata.
+    runtime_fullname = ".".join(["django.db.models.manager", manager_name])
+    manager_base.metadata["from_queryset_managers"][runtime_fullname] = fullname
+
+
 def create_manager_info_from_from_queryset_call(
-    api: SemanticAnalyzerPluginInterface, call_expr: CallExpr, name: Optional[str] = None
+    api: SemanticAnalyzer, call_expr: CallExpr, name: Optional[str] = None
 ) -> Optional[TypeInfo]:
     """
     Extract manager and queryset TypeInfo from a from_queryset call.
@@ -247,30 +258,48 @@ def create_manager_info_from_from_queryset_call(
     else:
         manager_name = f"{base_manager_info.name}From{queryset_info.name}"
 
-    try:
-        new_manager_info = create_manager_class(api, base_manager_info, name or manager_name, call_expr.line)
-    except helpers.IncompleteDefnException:
-        return None
+    # Always look in global scope, as that's where we'll declare dynamic manager classes
+    manager_sym = api.globals.get(manager_name)
+    if (
+        manager_sym is not None
+        and isinstance(manager_sym.node, TypeInfo)
+        and manager_sym.node.has_base(base_manager_info.fullname)
+        and manager_sym.node.metadata.get("django", {}).get("from_queryset_manager") == queryset_info.fullname
+    ):
+        # Reuse an identical, already generated, manager
+        new_manager_info = manager_sym.node
+    else:
+        # Create a new `TypeInfo` instance for the manager type
+        try:
+            new_manager_info = create_manager_class(
+                api=api,
+                base_manager_info=base_manager_info,
+                name=manager_name,
+                line=call_expr.line,
+                with_unique_name=name is not None and name != manager_name,
+            )
+        except helpers.IncompleteDefnException:
+            return None
 
-    popuplate_manager_from_queryset(new_manager_info, queryset_info)
-
-    manager_fullname = ".".join(["django.db.models.manager", manager_name])
-
-    base_manager_info = new_manager_info.mro[1]
-    base_manager_info.metadata.setdefault("from_queryset_managers", {})
-    base_manager_info.metadata["from_queryset_managers"][manager_fullname] = new_manager_info.fullname
+        populate_manager_from_queryset(new_manager_info, queryset_info)
+        register_dynamically_created_manager(
+            fullname=new_manager_info.fullname,
+            manager_name=manager_name,
+            manager_base=base_manager_info,
+        )
 
     # Add the new manager to the current module
     module = api.modules[api.cur_mod_id]
-    module.names[name or manager_name] = SymbolTableNode(
-        GDEF, new_manager_info, plugin_generated=True, no_serialize=False
-    )
+    if name is not None and name != new_manager_info.name:
+        # Unless names are equal, there's 2 symbol names that needs the manager info
+        module.names[name] = SymbolTableNode(GDEF, new_manager_info, plugin_generated=True)
 
+    module.names[new_manager_info.name] = SymbolTableNode(GDEF, new_manager_info, plugin_generated=True)
     return new_manager_info
 
 
 def create_manager_class(
-    api: SemanticAnalyzerPluginInterface, base_manager_info: TypeInfo, name: str, line: int
+    api: SemanticAnalyzer, base_manager_info: TypeInfo, name: str, line: int, with_unique_name: bool
 ) -> TypeInfo:
 
     base_manager_instance = fill_typevars(base_manager_info)
@@ -280,17 +309,24 @@ def create_manager_class(
     if any(has_placeholder(type_var) for type_var in base_manager_info.defn.type_vars):
         raise helpers.IncompleteDefnException
 
-    manager_info = helpers.create_type_info(name, api.cur_mod_id, bases=[base_manager_instance])
+    if with_unique_name:
+        manager_info = helpers.add_new_class_for_module(
+            module=api.modules[api.cur_mod_id],
+            name=name,
+            bases=[base_manager_instance],
+        )
+    else:
+        manager_info = helpers.create_type_info(name, api.cur_mod_id, bases=[base_manager_instance])
+
     manager_info.line = line
     manager_info.type_vars = base_manager_info.type_vars
     manager_info.defn.type_vars = base_manager_info.defn.type_vars
     manager_info.defn.line = line
-    manager_info.metaclass_type = manager_info.calculate_metaclass_type()
 
     return manager_info
 
 
-def popuplate_manager_from_queryset(manager_info: TypeInfo, queryset_info: TypeInfo) -> None:
+def populate_manager_from_queryset(manager_info: TypeInfo, queryset_info: TypeInfo) -> None:
     """
     Add methods from the QuerySet class to the manager.
     """
@@ -318,7 +354,7 @@ def popuplate_manager_from_queryset(manager_info: TypeInfo, queryset_info: TypeI
             helpers.add_new_sym_for_info(
                 manager_info,
                 name=name,
-                sym_type=AnyType(TypeOfAny.special_form),
+                sym_type=AnyType(TypeOfAny.implementation_artifact),
             )
 
     # For methods on BaseManager that return a queryset we need to update
@@ -330,5 +366,103 @@ def popuplate_manager_from_queryset(manager_info: TypeInfo, queryset_info: TypeI
         helpers.add_new_sym_for_info(
             manager_info,
             name=method_name,
-            sym_type=AnyType(TypeOfAny.special_form),
+            sym_type=AnyType(TypeOfAny.implementation_artifact),
         )
+
+
+def create_new_manager_class_from_as_manager_method(ctx: DynamicClassDefContext) -> None:
+    """
+    Insert a new manager class node for a
+
+    ```
+    <manager name> = <QuerySet>.as_manager()
+    ```
+    """
+    semanal_api = helpers.get_semanal_api(ctx)
+    # Don't redeclare the manager class if we've already defined it.
+    manager_node = semanal_api.lookup_current_scope(ctx.name)
+    if manager_node and manager_node.type is not None:
+        # This is just a deferral run where our work is already finished
+        return
+
+    manager_sym = semanal_api.lookup_fully_qualified_or_none(fullnames.MANAGER_CLASS_FULLNAME)
+    assert manager_sym is not None
+    manager_base = manager_sym.node
+    if manager_base is None:
+        if not semanal_api.final_iteration:
+            semanal_api.defer()
+        return
+
+    assert isinstance(manager_base, TypeInfo)
+
+    callee = ctx.call.callee
+    assert isinstance(callee, MemberExpr)
+    assert isinstance(callee.expr, RefExpr)
+
+    queryset_info = callee.expr.node
+    if queryset_info is None:
+        if not semanal_api.final_iteration:
+            semanal_api.defer()
+        return
+
+    assert isinstance(queryset_info, TypeInfo)
+
+    manager_class_name = manager_base.name + "From" + queryset_info.name
+    current_module = semanal_api.modules[semanal_api.cur_mod_id]
+    existing_sym = current_module.names.get(manager_class_name)
+    if (
+        existing_sym is not None
+        and isinstance(existing_sym.node, TypeInfo)
+        and existing_sym.node.has_base(fullnames.MANAGER_CLASS_FULLNAME)
+        and existing_sym.node.metadata.get("django", {}).get("from_queryset_manager") == queryset_info.fullname
+    ):
+        # Reuse an identical, already generated, manager
+        new_manager_info = existing_sym.node
+    else:
+        # Create a new `TypeInfo` instance for the manager type
+        try:
+            new_manager_info = create_manager_class(
+                api=semanal_api,
+                base_manager_info=manager_base,
+                name=manager_class_name,
+                line=ctx.call.line,
+                with_unique_name=True,
+            )
+        except helpers.IncompleteDefnException:
+            if not ctx.api.final_iteration:
+                ctx.api.defer()
+            return
+
+        populate_manager_from_queryset(new_manager_info, queryset_info)
+        register_dynamically_created_manager(
+            fullname=new_manager_info.fullname,
+            manager_name=manager_class_name,
+            manager_base=manager_base,
+        )
+
+        # So that the plugin will reparameterize the manager when it is constructed inside of a Model definition
+        helpers.add_new_manager_base(semanal_api, new_manager_info.fullname)
+
+    # Whenever `<QuerySet>.as_manager()` isn't called at class level, we want to ensure
+    # that the variable is an instance of our generated manager. Instead of the return
+    # value of `.as_manager()`. Though model argument is populated as `Any`.
+    # `transformers.models.AddManagers` will populate a model's manager(s), when it
+    # finds it on class level.
+    var = Var(name=ctx.name, type=Instance(new_manager_info, [AnyType(TypeOfAny.from_omitted_generics)]))
+    var.info = new_manager_info
+    var._fullname = f"{current_module.fullname}.{ctx.name}"
+    var.is_inferred = True
+    # Note: Order of `add_symbol_table_node` calls matters. Depending on what level
+    # we've found the `.as_manager()` call. Point here being that we want to replace the
+    # `.as_manager` return value with our newly created manager.
+    assert semanal_api.add_symbol_table_node(
+        ctx.name, SymbolTableNode(semanal_api.current_symbol_kind(), var, plugin_generated=True)
+    )
+    # Add the new manager to the current module
+    assert semanal_api.add_symbol_table_node(
+        # We'll use `new_manager_info.name` instead of `manager_class_name` here
+        # to handle possible name collisions, as it's unique.
+        new_manager_info.name,
+        # Note that the generated manager type is always inserted at module level
+        SymbolTableNode(GDEF, new_manager_info, plugin_generated=True),
+    )
--- a/tests/typecheck/managers/querysets/test_as_manager.yml
+++ b/tests/typecheck/managers/querysets/test_as_manager.yml
@@ -1,0 +1,214 @@
+-   case: declares_manager_type_like_django
+    main: |
+        from myapp.models import MyModel
+        reveal_type(MyModel.objects)  # N: Revealed type is "myapp.models.ManagerFromMyQuerySet[myapp.models.MyModel]"
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+
+                class MyQuerySet(models.QuerySet):
+                    ...
+
+                class MyModel(models.Model):
+                    objects = MyQuerySet.as_manager()
+
+-   case: includes_django_methods_returning_queryset
+    main: |
+        from myapp.models import MyModel
+        reveal_type(MyModel.objects.none)  # N: Revealed type is "def () -> myapp.models.MyQuerySet[myapp.models.MyModel]"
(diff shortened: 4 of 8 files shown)


README.md
```diff
@@ -157,11 +155,11 @@ class MyModel(models.Model):
   objects = MyModelQuerySet.as_manager()
 
 def use_my_model():
```
sobolevn:
```suggestion
def use_my_model() -> int:
```

Otherwise this function won't be checked.

sobolevn (commented):
Looks good to me. Awesome work.
But, I want to have another look from someone :)

flaeppe:
Did you have anyone in mind for a second pair of eyes? Or did you mean for me to ask someone?

sobolevn:
@mkurnikov or @intgr or @syastrov

intgr:
Sorry, mypy plugin code is entirely foreign to me and I don't have time right now to delve into it.

ljodal:
Seems like you managed to implement support for creating these managers inline, does that mean that it's technically possible to support `Manager.from_queryset(QuerySet)` inline as well?

Oh and I'd be happy to review this, but I'm kinda swamped for the next few days so I might not get to it until Friday.

flaeppe:
> Seems like you managed to implement support for creating these managers inline, does that mean that it's technically possible to support `Manager.from_queryset(QuerySet)` inline as well?
> 
> Oh and I'd be happy to review this, but I'm kinda swamped for the next few days so I might not get to it until Friday.

Yes, it should be technically possible (I'm assuming class level/definition, when you say _inline_) __as long as we're not talking about the "full" call `Manager.from_queryset(QuerySet)()`__. But I opted out of supporting it here, since I thought it felt a bit strange. Reasoning being that unwrapping the non-dynamic version would look like something below:

```python
class MyModel(models.Model):
    class ManagerFromMyQuerySet(...):
        ...
    
    objects = ManagerFromMyQuerySet()
```

Felt more natural to keep force it outside of the model class definition.

ljodal:
> > Seems like you managed to implement support for creating these managers inline, does that mean that it's technically possible to support `Manager.from_queryset(QuerySet)` inline as well?
> > Oh and I'd be happy to review this, but I'm kinda swamped for the next few days so I might not get to it until Friday.
> 
> [...] **as long as we're not talking about the "full" call `Manager.from_queryset(QuerySet)()`**. [...]

This is what I wanted. All our models currently do `objects = MyManager.from_queryset(MyQuerySet)()`, so I'm having to go through all of them and move the manager definitions outside of the models, which feels a bit pointless, but oh well.

I'd forgotten about that extra call to initialise the model. I assume that causes mypy to not call `get_dynamic_class_hook` on the plugin and that's the problem with supporting that? Wonder if it would be possible to still support this by using different hooks. Something to dive into when I have time I guess

flaeppe:
> ... I assume that causes mypy to not call get_dynamic_class_hook on the plugin and that's the problem with supporting that? ...

Yes, exactly, for clarity, the reason it wont work with inlining the full call, seen below, is that it _wont trigger the `get_dynamic_class_hook`_

```python
class MyModel(models.Model):
    objects = Manager.from_queryset(MyQuerySet)()
```

> ... This plugin hook is called for every assignment to a simple name where right hand side is a function call ...

Ref: https://mypy.readthedocs.io/en/stable/extending_mypy.html#current-list-of-plugin-hooks (docs for `get_dynamic_class_hook`)

The more fortunate case for `.as_manager` is that it returns a manager instance, while `.from_queryset` returns a manager type.

sobolevn:
There are conflicts right now 😉

flaeppe:
> There are conflicts right now 😉

Hehe, yeah. Think it's resolved now

mypy_django_plugin/transformers/managers.py
```diff
@@ -175,6 +176,67 @@ def resolve_manager_method(ctx: AttributeContext) -> MypyType:
...
+            # the method is set as Any. Figuring out the type is the job of the
+            # 'resolve_manager_method' attribute hook, which comes later.
+            #
+            # class BaseManagerFromMyQuerySet(BaseManager):
+            #    queryset_method: Any = ...
+            #
+            helpers.add_new_sym_for_info(
+                manager_info,
+                name=name,
+                sym_type=AnyType(TypeOfAny.special_form),
```
```suggestion
                sym_type=AnyType(TypeOfAny.special_form),
```
```suggestion
                sym_type=AnyType(TypeOfAny.implementation_detail),
```

mypy_django_plugin/transformers/managers.py
```diff
@@ -175,6 +176,67 @@ def resolve_manager_method(ctx: AttributeContext) -> MypyType:
...
+    # For methods on BaseManager that return a queryset we need to update the
+    # return type to be the actual queryset subclass used. This is done by
+    # adding the methods as attributes with type Any to the manager class,
+    # similar to how custom queryset methods are handled above. The actual type
+    # of these methods are resolved in resolve_manager_method.
+    for name in MANAGER_METHODS_RETURNING_QUERYSET:
+        helpers.add_new_sym_for_info(
+            manager_info,
+            name=name,
+            sym_type=AnyType(TypeOfAny.special_form),
```
```suggestion
            sym_type=AnyType(TypeOfAny.special_form),
```
```suggestion
            sym_type=AnyType(TypeOfAny.implementation_detail),
```

mypy_django_plugin/transformers/managers.py
flaeppe: We resolve these anys into queryset methods, but we're only looking for `isinstance(..., AnyType)`. Should we be more specific and look for `implementation_artifact` as well?

```python
if isinstance(ctx.default_attr_type, AnyType) and ctx.default_attr_type.type_of_any == TypeOfAny.implementation_artifact
```

https://github.com/typeddjango/django-stubs/blob/2a6f4647f083b2a9cd728b7e5698dc76c59b5449/mypy_django_plugin/transformers/managers.py#L158

What do you think?

sobolevn: Yes, seems like a good idea. Let's add a specific test that explicit `Any` stays as it is:

`def some(self) -> Any:`

flaeppe: We have a test case `from_queryset_can_resolve_explicit_any_methods` doing that already. I think that should suffice

https://github.com/typeddjango/django-stubs/blob/2a6f4647f083b2a9cd728b7e5698dc76c59b5449/tests/typecheck/managers/querysets/test_from_queryset.yml#L333-L360

mypy_django_plugin/main.py
```diff
@@ -303,11 +304,16 @@ def get_type_analyze_hook(self, fullname: str) -> Optional[Callable[[AnalyzeType
 
     def get_dynamic_class_hook(self, fullname: str) -> Optional[Callable[[DynamicClassDefContext], None]]:
         # Create a new manager class definition when a manager's '.from_queryset' classmethod is called
-        if fullname.endswith("from_queryset"):
+        if fullname.endswith(".from_queryset"):
             class_name, _, _ = fullname.rpartition(".")
```
ljodal:
```suggestion
        class_name, _, method_name = fullname.rpartition(".")
        if method_name == "from_queryset":
```
Maybe do this first and just check the method name, since that's needed anyway?

mypy_django_plugin/main.py
```diff
@@ -303,11 +304,16 @@ def get_type_analyze_hook(self, fullname: str) -> Optional[Callable[[AnalyzeType
...
             info = self._get_typeinfo_or_none(class_name)
             if info and info.has_base(fullnames.BASE_MANAGER_CLASS_FULLNAME):
                 return create_new_manager_class_from_from_queryset_method
+        elif fullname.endswith(".as_manager"):
+            class_name, _, _ = fullname.rpartition(".")
```
ljodal:
```suggestion
        elif method_name == ".as_manager":
```
Slightly less duplication this way at least. Would also like to move the `_get_typeinfo_or_none` call out, but I guess there's a significant overhead of doing that for everything

mypy_django_plugin/transformers/managers.py
```diff
@@ -183,112 +247,182 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
...
 
-    class_def_context = ClassDefContext(cls=new_manager_info.defn, reason=ctx.call, api=semanal_api)
-    self_type = fill_typevars(new_manager_info)
-    assert isinstance(self_type, Instance)
+    symbol_kind = semanal_api.current_symbol_kind()
+    # Annotate the module variable as `<Variable>: Type[<NewManager[Any]>]` as the model
+    # type won't be defined on variable level.
+    var = Var(
+        name=ctx.name,
+        type=TypeType(Instance(new_manager_info, [AnyType(TypeOfAny.from_omitted_generics)])),
```
ljodal: Why is this `Any`? Shouldn't we take the model from either the queryset or the manager (and potentially validate that they match)?

mypy_django_plugin/transformers/managers.py
```diff
@@ -183,112 +247,182 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
...
+    )
+    var.info = new_manager_info
+    var._fullname = f"{semanal_api.cur_mod_id}.{ctx.name}"
+    var.is_inferred = True
+    # Note: Order of `add_symbol_table_node` calls matter. Case being if
+    # `ctx.name == new_manager_info.name`, then we'd _only_ like the type and not the
+    # `Var` to exist..
```
ljodal: Took me a while to understand that in the case of `ctx.name == new_manager_info.name` the second call will basically override the changes made by the first call. So maybe explicitly state that?

mypy_django_plugin/transformers/managers.py
```diff
@@ -183,112 +247,182 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
...
-    new_manager_info.metaclass_type = new_manager_info.calculate_metaclass_type()
-    # Stash the queryset fullname which was passed to .from_queryset
-    # So that our 'resolve_manager_method' attribute hook can fetch the method from that QuerySet class
-    new_manager_info.metadata["django"] = {"from_queryset_manager": derived_queryset_fullname}
+    queryset_info = queryset_sym.node
+    assert isinstance(queryset_info, TypeInfo)
 
     if len(ctx.call.args) > 1:
         expr = ctx.call.args[1]
         assert isinstance(expr, StrExpr)
```
ljodal: Maybe an edge case, but it's also valid to explicitly call this with `None` or a dynamic string. And I don't like that we crash on type errors. Maybe just return if the type is not None or a string, mypy will emit a type error then anyway, and emit an error if it's not `None` or a literal string (as we're then unable to resolve the name)?

I guess that's not really related to this PR, but would be nice to fix this when you're rewriting this anyway

ljodal (commented):
The changes look very nice, I've just added some minor nits here and there. I am however seeing this when running this on our largest Django project at work:

```
project/users/forms/normal.py: error: INTERNAL ERROR: maximum semantic analysis iteration count reached
```

I'm not sure what's causing this (we're subclassing some Django builtin forms etc), but I assume it's a `defer()` call somewhere. I can dig into that tomorrow if you need help reproducing it.

mypy_django_plugin/transformers/managers.py
flaeppe: I'm not sure, I'm kinda split on if we should display a model here. Since this is when we're outside of a model class definition, e.g.

```python
MyManager = Manager.from_queryset(MyQuerySet)
```

What I'm thinking about here is what the manager type is during runtime, model won't be set until the model metaclass is called. Which is basically inside model class definition. And, while I have never used it like that, same manager could be used on multiple models..

But perhaps you're right, maybe we should require model arguments for both queryset and manager to match. Since we're basically merging those two arguments into one and the model argument exists for manager and queryset types.

Also, I think `AddManagers` will change the model argument if called on a different model? But perhaps that should also change?

mypy_django_plugin/main.py
flaeppe: I don't know how much overhead it is. But since it is some, I just kept them inlined.

This is the work that'll happen:
https://github.com/python/mypy/blob/fc335cb16315964b923eb1927e3aad1516891c28/mypy/lookup.py#L12

mypy_django_plugin/transformers/managers.py
flaeppe: Sure, we can do that

mypy_django_plugin/transformers/managers.py
flaeppe: Good point. I might as well update it here.

flaeppe:
> I'm not sure what's causing this (we're subclassing some Django builtin forms etc), but I assume it's a defer() call somewhere. I can dig into that tomorrow if you need help reproducing it.

Thanks for the review. It'd be very helpful if you could find a repro, I have no idea where to start looking for it really.

mypy_django_plugin/transformers/managers.py
flaeppe: I realised that a `TypeError` is raised from `builtins.type` if the name value in the 3-arg call isn't `str`. So I changed to fallback on the default value unless `class_name` is a hard coded `StrExpr`.

mypy_django_plugin/transformers/managers.py
flaeppe: Let me know if it's more obvious with the updated comment

mypy_django_plugin/transformers/managers.py
ljodal: I was thinking it would make sense to inspect the generics of the manager and querysets (if possible). If they are different an error message should probably be emitted and in other case use the class from either?

And I'd think that if `AddManagers` sees a model argument that's not compatible with the one on the manager then an error should be emitted?

I'm okay with not adding this in this PR, but that's at least how I would expect this to behave. I see that it's really an edge case and probably not a big problem.

flaeppe: Yeah, I've been looking in to exactly that. And it seems possible if we for `.from_queryset` traverse `.bases` (as those are `types.Instance`, we'll get access to `Instance.args`) and compare model arguments, then on mismatch emit an error.

This should probably happen for both `.from_queryset` and `AddManagers`, as you say.

flaeppe: There's a case left to decide on, when model arguments aren't populated for `Queryset` or `Manager`. Either:

- We emit an error and enforce population of the model arg or,
- We then allow it to be any model, thus just populate it with whatever we got

I'm leaning towards the latter, since one can switch on enforcing via mypy config: `disallow_any_generics = true`.

For the first case, we could then establish some sort of "auto-fill". Depending on if 1 of the model arguments are populated, if both are missing `.from_queryset` results in `ManagerFromQueryset[Any]` while `AddManagers` just populates it with the model it's processing.

This is basically the table:

1. Manager and queryset model argument populated -> error if mismatch for both `.from_queryset` and `AddManagers`
2. Only manager model argument populated -> Populate queryset model argument with manager model argument
3. Only queryset model argument populated -> Populate manager model argument with queryset model argument
4. No argument populated -> `.from_queryset` sets `Any` and `AddManagers` populates with the model it's processing.

I think this aligns with when using Django builtins e.g. `BaseManager.from_queryset(CustomQuerySet)`, since otherwise we'll enforce people to write custom managers to populate the argument, even though they only have/need a queryset defined. Like this:

```python
class SuperFluousManager(BaseManager[MyModel]):
    ...
```

What do you think?

ljodal:
I dug a bit into the mypy code yesterday and from what I could see the problem isn't guaranteed to be in the file it reports, so I think this is related to something else. The curious thing is that we don't have any `.as_manager()` calls in our codebase anymore, so either it's from a dependency or it's related to the other changes in this PR.

I'll try to extract just the "refactoring" bit of this PR and see if it reproduces with just that at least, but I'm struggling to understand what causes it. It does happen during processing of the majority of our models though.

Running `mypy -v` shows `Processing SCC of size 748`, where SCC stands for strongly connected cluster or something along those lines and 748 is the number of modules it's processing. So it might be a symptom of our codebase being highly entangled (which it is 😅), but the cluster size is the same before this patch so it hasn't changed that.

ljodal:
I commented out the new `create_new_manager_class_from_as_manager_method` method in `main.py` and it still crashes, so the problem is somehow related to the refactoring of the `from_queryset` logic 🤔 Guess I'll have to try to step through that code and see where it goes wrong

mypy_django_plugin/transformers/managers.py
ljodal: Yeah, that sounds exactly like what I'd expect 👍

flaeppe: I've been looking in to this for a bit and think we should push for it to a future PR, as it's quite an isolated addition but generates quite a bit of lines. And the changes we have here are already large enough.

flaeppe:
Interesting, could it be related to the new bailing out when `.from_queryset` isn't called at module level? https://github.com/typeddjango/django-stubs/blob/c2322264035aa951e51e28906c3322711da1c3d8/mypy_django_plugin/transformers/managers.py#L274-L277

mypy_django_plugin/transformers/managers.py
flaeppe: I've pushed progress to a PR here: https://github.com/flaeppe/django-stubs/pull/29

ljodal:
I'm pretty sure it's not related to that. I've had that line there unrelated to this and that worked fine. I'll try to pick the PR apart and test it bit by bit next week to see if I can figure it out.

ljodal:
I think it might be related to the name changes for generated managers. I tried to do something similar while working on #1045 and quickly got the same problem in a test repo. Should probably be solvable, but maybe we could try without those changes and do it in a separate PR?

flaeppe:
> I think it might be related to the name changes for generated managers. I tried to do something similar while working on #1045 and quickly got the same problem in a test repo. Should probably be solvable, but maybe we could try without those changes and do it in a separate PR?

Sure, we could drop the name change.

I ran in to it too when fiddling with https://github.com/flaeppe/django-stubs/pull/29

Got stuck in AddManagers with a metadata key that didn't match the runtime manager name for a dynamic manager.

I think you should get a trace to the model definition it gets stuck on? Anything special happening with it?

ljodal:
> I think you should get a trace to the model definition it gets stuck on? Anything special happening with it?

Yeah, I get a trace, but it ends in a random form class (for our user model, which I guess subclasses a few `django.contrib.auth` forms). It's literally 100s of lines long, if not 1000s, and I couldn't find anything useful there. I tried to check if any single model was processed more than 20 times (which appears to be the iteration limit) in a row and that wasn't the case. Max was 10

ljodal:
Yeah, it's related to the manager name. Changing that logic back fixes the issue. I wonder if it might be because the name already exists in the module?

``` diff
diff --git a/mypy_django_plugin/transformers/managers.py b/mypy_django_plugin/transformers/managers.py
index 6aa00c8..3e9b8fe 100644
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -318,20 +318,20 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
     symbol_kind = semanal_api.current_symbol_kind()
     # Annotate the module variable as `<Variable>: Type[<NewManager[Any]>]` as the model
     # type won't be defined on variable level.
-    var = Var(
-        name=ctx.name,
-        type=TypeType(Instance(new_manager_info, [AnyType(TypeOfAny.from_omitted_generics)])),
-    )
-    var.info = new_manager_info
-    var._fullname = f"{semanal_api.cur_mod_id}.{ctx.name}"
-    var.is_inferred = True
+    # var = Var(
+    #     name=ctx.name,
+    #     type=TypeType(Instance(new_manager_info, [AnyType(TypeOfAny.from_omitted_generics)])),
+    # )
+    # var.info = new_manager_info
+    # var._fullname = f"{semanal_api.cur_mod_id}.{ctx.name}"
+    # var.is_inferred = True
     # Note: Order of `add_symbol_table_node` calls matter. Case being if
     # `ctx.name == new_manager_info.name`, then we'd like the type to override the
     # `Var`, so the `Var` won't exist in the end.
-    assert semanal_api.add_symbol_table_node(ctx.name, SymbolTableNode(symbol_kind, var, plugin_generated=True))
+    # assert semanal_api.add_symbol_table_node(ctx.name, SymbolTableNode(symbol_kind, var, plugin_generated=True))
     # Insert the new manager dynamic class
     assert semanal_api.add_symbol_table_node(
-        new_manager_info.name, SymbolTableNode(symbol_kind, new_manager_info, plugin_generated=True)
+        ctx.name, SymbolTableNode(symbol_kind, new_manager_info, plugin_generated=True)
     )
```

flaeppe:
> ... I wonder if it might be because the name already exists in the module?

Yep, that might be it. I added a test which didn't resolve that properly.

flaeppe:
I added a test for that for `.as_manager`, but forgot about doing it for `.from_queryset`

https://github.com/flaeppe/django-stubs/blob/c2322264035aa951e51e28906c3322711da1c3d8/tests/typecheck/managers/querysets/test_as_manager.yml#L60-L80

flaeppe:
@ljodal I pushed an update resolving name collisions with `.from_queryset`. Do you have a possibility to test with that? See if it helps?

ljodal:
Still crashing :(

ljodal:
The problem appears to be the `Var`, not the name of the class. This works:

```diff
diff --git a/mypy_django_plugin/transformers/managers.py b/mypy_django_plugin/transformers/managers.py
index d7369d3..41281d6 100644
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -336,21 +336,12 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
             class_name=manager_class_name,
         )
 
-    symbol_kind = semanal_api.current_symbol_kind()
-    # Annotate the module variable as `<Variable>: Type[<NewManager[Any]>]` as the model
-    # type won't be defined on variable level.
-    var = Var(
-        name=ctx.name,
-        type=TypeType(Instance(new_manager_info, [AnyType(TypeOfAny.from_omitted_generics)])),
-    )
-    var.info = new_manager_info
-    var._fullname = f"{current_module.fullname}.{ctx.name}"
-    var.is_inferred = True
-    # Note: Order of `add_symbol_table_node` calls matter. Case being if
-    # `ctx.name == new_manager_info.name`, then we'd like the type to override the
-    # `Var`, so the `Var` won't exist in the end.
-    assert semanal_api.add_symbol_table_node(ctx.name, SymbolTableNode(symbol_kind, var, plugin_generated=True))
     # Insert the new manager dynamic class
+    symbol_kind = semanal_api.current_symbol_kind()
+    if ctx.name != new_manager_info.name:
+        assert semanal_api.add_symbol_table_node(
+            ctx.name, SymbolTableNode(symbol_kind, new_manager_info, plugin_generated=True)
+        )
     assert semanal_api.add_symbol_table_node(
         new_manager_info.name, SymbolTableNode(symbol_kind, new_manager_info, plugin_generated=True)
     )
```

---

Edit: I realised that the problem reproduces in my test repo: https://github.com/ljodal/django-stubs-regression-test

flaeppe:
I see, I was under the impression that the dynamic class hook is only called for a `Var` and thus that `ctx.name` could always be overwritten with another `Var`. But maybe there's more to it.

flaeppe:
@ljodal I hit the reproduction in your test repo until I specified the config file for mypy (`--config-file=pyproject.toml`). With explicit config file I can't reproduce the issue.

__Edit:__ nevermind, I haven't had coffee yet. It's the other way around... Haha, config should be `setup.cfg`..

flaeppe:
Right, think we should change to your edit, not inserting a `Var` but instead the type info two times.

---

Out of interest I looked around a bit more in https://github.com/ljodal/django-stubs-regression-test and realised that if I change these two things:

- Remove the `.from_queryset` call inside the class body at: https://github.com/ljodal/django-stubs-regression-test/blob/0c620d46fafa3850fa8f607802e8ec1bbe75da69/project/app/models.py#L22, and
- Fully type the signature for the `custom_method` method at https://github.com/ljodal/django-stubs-regression-test/blob/0c620d46fafa3850fa8f607802e8ec1bbe75da69/project/app/models.py#L34

I don't reach the maximum iteration error. But the `Var` has to mess something up in the module symbol table or something.

ljodal:
Yeah, this kinda seems like a bug in mypy, but hard to say exactly what goes wrong here

flaeppe:
Do you still reach the maximum iteration roof if you try out the latest commit here?

mypy_django_plugin/transformers/managers.py
```diff
@@ -183,112 +247,198 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
...
+    var = Var(name=ctx.name, type=Instance(new_manager_info, [AnyType(TypeOfAny.from_omitted_generics)]))
+    var.info = new_manager_info
+    var._fullname = f"{current_module.fullname}.{ctx.name}"
+    var.is_inferred = True
+    # Note: Order of `add_symbol_table_node` calls matters. Depending on what level
+    # we've found the `.as_manager()` call. Point here being that we want to replace the
+    # `.as_manager` return value with our newly created manager.
+    assert semanal_api.add_symbol_table_node(
+        ctx.name, SymbolTableNode(semanal_api.current_symbol_kind(), var, plugin_generated=True)
+    )
```
ljodal, line 460:
```suggestion
    symbol_kind = semanal_api.current_symbol_kind()
     if ctx.name != new_manager_info.name:
         # Unless names are equal, there's 2 symbol names that needs the manager info
         assert semanal_api.add_symbol_table_node(
             ctx.name, SymbolTableNode(symbol_kind, new_manager_info, plugin_generated=True)
         )
```

Should we do the same here for good measure? I assume if it's a problem with `from_queryset()` it likely is with `as_manager()` as well?

mypy_django_plugin/transformers/managers.py
```diff
@@ -183,112 +247,198 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
...
+    assert semanal_api.add_symbol_table_node(
+        # We'll use `new_manager_info.name` instead of `manager_class_name` here
+        # to handle possible name collisions, as it's unique.
+        new_manager_info.name,
+        # Note that the generated manager type is always inserted at module level
+        SymbolTableNode(GDEF, new_manager_info, plugin_generated=True),
```
ljodal, line 467:
```suggestion
        SymbolTableNode(semanal_api.current_symbol_kind(), new_manager_info, plugin_generated=True),
```

mypy_django_plugin/transformers/managers.py
flaeppe: Yeah, lets not gamble

ljodal:
> Do you still reach the maximum iteration roof if you try out the latest commit here?

Nope, it works now 🎉

mypy_django_plugin/transformers/managers.py
flaeppe: Hm, just realised that this will be incorrect. Since `.as_manager` returns an _instance_ and not a type (which `.from_queryset` does). So in one way or another, we'll need a _node_ with a set type.

mypy_django_plugin/transformers/managers.py
flaeppe: The reason behind `GDEF` here is to get a fullname on the type that `AddManagers` can do lookup on. And `AddManagers` expects the dynamically generated types to be on module level. Otherwise we'll get types looking like `myapp.models.MyModel.ManagerFromMyQuerySet` and `AddManagers` won't find them

ljodal: Does it actually work though? Cause I observed that when I did something similar in #1045 that it would fail to look up the name because it's added inside the class.

Since this isn't actually returning a type I'm wondering if it would be better to implement this support as I've done with `.from_queryset()` in #1045 (do it in the model hook, that way mypy doesn't care about the generated type) rather than through the `dynamic_class_hook`?

flaeppe: It works I think? At least through the tests that I added. I also tried to break it similarly to what happened with `.from_queryset` in https://github.com/ljodal/django-stubs-regression-test, but never got it breaking.

Perhaps it might be preferable to align with #1045 but either this PR or #1045 merges first. Instead of merging them together, as that changeset feels a bit large.

ljodal: Yeah, absolutely agree, they should be merged separately. Sounds like a good plan and if it doesn't break it's probably fine :)

flaeppe:
I just pushed a last test I felt was needed. I feel satisfied from my end, unless anyone else has some additional input?

@sobolevn @ljodal

mypy_django_plugin/transformers/managers.py
```diff
@@ -154,9 +155,11 @@ def resolve_manager_method(ctx: AttributeContext) -> MypyType:
     A 'get_attribute_hook' that is intended to be invoked whenever the TypeChecker encounters
     an attribute on a class that has 'django.db.models.BaseManager' as a base.
     """
-    # Skip (method) type that is currently something other than Any
+    # Skip (method) type that is currently something other than Any of type `implementation_artifact`
     if not isinstance(ctx.default_attr_type, AnyType):
         return ctx.default_attr_type
+    elif ctx.default_attr_type.type_of_any != TypeOfAny.implementation_artifact:
+        return ctx.default_attr_type
```
ljodal, line 171: I'm seeing some issues due to this line (after merging this and #1045, so might be my fault), why did you add this?

flaeppe: Idea was to guarantee only processing Anys from the dynamic manager class creation (read: methods). So we won't accidentally do something with stuff that isn't manager/queryset methods

Also: https://github.com/typeddjango/django-stubs/pull/1025#discussion_r910734539

ljodal: Ah, then I guess I'm doing something in one of my branches where I'm not correctly setting the Any type :)

ljodal:
There's something weird going on with manager `get_or_create`:


```python
from django.db import models


class UserQuerySet(models.QuerySet["User"]):
    pass


class User(models.Model):
    objects = UserQuerySet.as_manager()


def test() -> None:
    reveal_type(User.objects.get_or_create)
    reveal_type(User.objects.get_or_create())
    user, created = User.objects.get_or_create()
    reveal_type(user)
    reveal_type(created)
```

The second last line returns `Any` for some reason, everything until that point looks fine. Works fine when using `models.Manager.from_queryset(UserQuerySet)`

---

Edit: Looks like it might be related to caching. I'm seeing it consistently in actual code, but the example above is suddenly working now

flaeppe:
Hm, I'm unable to recreate it with your example above. Cached or non-cached.

ljodal:
Yeah, I don't think I have anything better right now and don't think I have time to try to reproduce it either. I think it's fine to merge this though, we can always bug fix later and it's possible to use `from_queryset` as well to work around the problem

ljodal (approved):

sobolevn:
Please, take a look at the merge conflicts.

aweaver6:
bump, i'm also having this issue

sobolevn:
Can you please rebase this? I am open to get it merged

flaeppe:
I think there's some rework to get done here before merging is possible. Gonna have to get to that first, not sure when I'll be able to though. (I think it's #1045 we have to align changes here with)

orsinium:
Is there anything I can help with to have it merged? About 30% of mypy violations we have in our code base are caused by this issue 🙃 We use custom QuerySet manager a lot.

orsinium:
@flaeppe the PR seems quite clean and seemingly solving the issue. I can try the patch on our code base if you want to be double sure, though. What kind of rework do you see as blocking? Maybe, I can help with that. And if it's just a matter of rebasing, I can do that for you too.

flaeppe:
@orsinium the changes here depends the version previous to #1045. Basically `.from_queryset(...)` and `.as_manager()` shares(d) some helpers/functionality to annotate types. What I did here was to extract the common bits so that it could be reused to annotate both methods.

Although #1045 had to adjust the same logic. So the pieces I extracted conflicts with #1045, basically they look different now. So one kind of have to go find them again and see if/how it can be extracted/shared.

I'll have a look now to see if I'll figure it out

orsinium:
Thank you! Lemmino if you need any help.

flaeppe:
- @sobolevn I've rebased and aligned the implementation here with #1045 (as I've spoken of earlier)
- @orsinium It would be very helpful if you try this branch out on your project(s) (I use `.from_queryset` in my projects, so I don't really have much test surface for this implementation)
- @ljodal It would be really great if you could have a second (read 42:nd) look here, see if you find any discrepancies from my realignment with #1045 or whatnot

tests/typecheck/managers/querysets/test_as_manager.yml
```diff
@@ -0,0 +1,214 @@
...
+                    def returns_int_sequence(self) -> Sequence[int]:
+                      return [1]
+
+                class MyModel(models.Model):
+                    objects = ModelQuerySet.as_manager()
+
+-   case: handles_call_outside_of_model_class_definition
+    main: |
+        from myapp.models import MyModel, MyModelManager
+        reveal_type(MyModelManager)  # N: Revealed type is "myapp.models.ManagerFromModelQuerySet[Any]"
```
ljodal, line 110: I guess this really should produce a `Manager[MyModel]`, but if I remember correctly you were thinking of fixing that in a separate PR?

mypy_django_plugin/transformers/managers.py
```diff
@@ -212,8 +214,17 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
     helpers.add_new_manager_base(semanal_api, new_manager_info.fullname)
 
 
+def register_dynamically_created_manager(fullname: str, manager_name: str, manager_base: TypeInfo) -> None:
+    manager_base.metadata.setdefault("from_queryset_managers", {})
+    # The `__module__` value of the manager type created by Django's
+    # `.from_queryset` is `django.db.models.manager`. But we put new type(s) in the
+    # module currently being processed, so we'll map those together through metadata.
+    runtime_fullname = ".".join(["django.db.models.manager", manager_name])
+    manager_base.metadata["from_queryset_managers"][runtime_fullname] = fullname
```
ljodal, line 223: Shouldn't these be under the `django` key?

ljodal (approved):
Looks good 👍 I've tested it on our codebase and it introduces no new errors. Also converted a few `Manager.from_queryset(MyQuerySet)` calls and saw no new errors

mypy_django_plugin/transformers/managers.py
flaeppe: Yeah, probably. They weren't previously so I kept them the same. Will probably enforce cache clear to work properly. But that might be completely fine too.

tests/typecheck/managers/querysets/test_as_manager.yml
flaeppe: Yeah, exactly. I opened a branch for that in my fork, it's quite a bit of code to get it up and running I think. Probably more suitable for a subsequent PR I'd say

sobolevn (approved):
Thank you so much!
I propose to merge this PR and adjust any parts later on.

orsinium:
> It would be very helpful if you try this branch out on your project(s)

Tested the new version, and it works. For posterity, I faced a few issues along the way, but I'm not sure if they are related to this PR:

1. The minimal version for mypy in constraints might be wrong, I had to upgrade to the latest version.
    ```
    pyproject.toml:1: error: Error importing plugin "mypy_django_plugin.main": cannot import name 'has_placeholder' from 'mypy.semanal_shared'
    ```
3. There are bunch of new errors related to settings:
    ```
    'Settings' object has no attribute 'PAYMENTS_SERVICE_BASE_URL'  [misc]
    Import cycle from Django settings module prevents type inference for 'POD'  [misc]
    ```

But again, this can be unrelated, I'll investigate it later.

sobolevn:
> The minimal version for mypy in constraints might be wrong

I've bumped it recently, please check if that's now correct.

orsinium:
> I've bumped it recently, please check if that's now correct.

You're right, the constraint is correct. I think I've messed up with environments somehow. It failed with mypy 0.961 but the constraint is already `mypy>=0.980`. So, that's on my side.

ljodal:
> 2. There are bunch of new errors related to settings:
>    ```
>    'Settings' object has no attribute 'PAYMENTS_SERVICE_BASE_URL'  [misc]
>    Import cycle from Django settings module prevents type inference for 'POD'  [misc]
>    ```

This one is due to this: https://github.com/typeddjango/django-stubs/pull/1163. Basically you most likely import something in the settings file that imports `django.conf.settings`

PR #1026, opened 2022-06-27
Form types for Django 4.0.

smithdc1:
Work in progress to add type for changes to Django forms introduced in 4.0


django-stubs/forms/utils.pyi
```diff
@@ -13,26 +15,54 @@ _FilesT = MultiValueDict[str, UploadedFile]
...
+        renderer: Optional[BaseRenderer] = ...,
+    ) -> None: ...
     def as_data(self) -> Dict[str, List[ValidationError]]: ...
     def get_json_data(self, escape_html: bool = ...) -> Dict[str, Any]: ...
-    def as_json(self, escape_html: bool = ...) -> str: ...
-    def as_ul(self) -> str: ...
-    def as_text(self) -> str: ...
 
-class ErrorList(UserList):
+class ErrorList(UserList, List, RenderableErrorMixin):
```
smithdc1: I expect the tests to fail here, as this class now inherits from both `UserList` and `List`. 

https://github.com/django/django/blob/90d2f9f41671ef01c8e8e7b5648f95c9bf512aae/django/forms/utils.py#L138

🤷

django-stubs/forms/renderers.pyi
```diff
@@ -9,6 +9,8 @@ from django.template.base import Template
 def get_default_renderer() -> BaseRenderer: ...
 
 class BaseRenderer:
+    form_template_name: str = ...
+    formset_template_name: str = ...
```
smithdc1: So these are some of the 4.1 changes. In 4.0 they are on `BaseForm`.

I'm not sure which version is targeted at any given time given the project does not aim to support multiple versions of Django at once?

smithdc1 (commented):
Couple of questions / comments where any input would be much appreciated!

django-stubs/forms/renderers.pyi
sobolevn: We do not target 4.1 yet.

django-stubs/forms/utils.pyi
sobolevn: `List[Something]` might be good enough.

smithdc1: Looks like that will do. Thank you!

django-stubs/forms/renderers.pyi
smithdc1: Thanks -- I've rolled back the 4.1 changes. This now targets 4.0.

django-stubs/forms/utils.pyi
```diff
@@ -49,7 +48,7 @@ class ErrorDict(dict, RenderableErrorMixin):
     def as_data(self) -> Dict[str, List[ValidationError]]: ...
     def get_json_data(self, escape_html: bool = ...) -> Dict[str, Any]: ...
 
-class ErrorList(UserList, List, RenderableErrorMixin):
+class ErrorList(List[Union[ErrorList, Sequence[Union[str, Exception]]]], RenderableErrorMixin):
```
sobolevn: It should be either `Sequence` or `List`.
List is just a mutable sequence (kind of).

PR #1028, opened 2022-06-28
Resolve all queryset methods on managers as attributes

### LINKED ISSUE #958: Alias argument to QuerySet.using() typed as _SpecialForm

### LINKED ISSUE #1022: error: Name "Sequence" is not defined
# Bug report

There's a discussion in gh-709 about a new bug, which seems to output nonsense `error: Name "Sequence" is not defined` errors. @ljodal has suggested that the regression was introduced in https://github.com/typeddjango/django-stubs/pull/991, and they've reproduced the issue in https://github.com/typeddjango/django-stubs/pull/1021 and https://github.com/ljodal/django-stubs-regression-test.

## System information

- OS:
- `python` version: 3.10.x
- `django` version: 3.2.x
- `mypy` version: 0.962

ljodal:
This changes the logic for resolving queryset methods on managers from copying the methods over to adding them as attributes and then resolving the type of those attributes in the `get_attribute_hook` callback method, similar to how it is done for custom manager methods.

This might have introduced some new errors, but I'm not sure. The test suite passes, but I'm seeing some new errors in our project. Those might be from changes made since 1.11 that we're currently using though.

## Related issues

- Closes: #958 
- Closes: #1022


tests/typecheck/managers/querysets/test_from_queryset.yml
```diff
@@ -384,3 +384,37 @@
                 MyManager = BaseManager.from_queryset(MyQuerySet)
                 class MyModel(models.Model):
                     objects = MyManager()
+
+
+-   case: from_queryset_custom_auth_user_model
```
flaeppe, line 395: Think we should disable cache for this test also

```suggestion
-   case: from_queryset_custom_auth_user_model
    disable_cache: true
```

mypy_django_plugin/transformers/managers.py
```diff
@@ -56,11 +59,23 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
     assert isinstance(method_type, CallableType)
+
+    variables = method_type.variables
+    ret_type = method_type.ret_type
+
+    # If the return type is a typevar or unbound type that appears to be a
+    # queryset we change the return type to be the actual queryset
+    if isinstance(ret_type, (UnboundType, TypeVarType)) and ret_type.name == "_QS":
+        ret_type = Instance(queryset_info, manager_type.args)
+        variables = []
```
flaeppe: In case this is merged, perhaps we should mention through a TODO that parts of the code before calling copy_method_to_another_class is replaced by this approach?

tests/typecheck/managers/querysets/test_from_queryset.yml
ljodal: Are you sure? As this is set up right now it'll run the tests both with and without caching, doesn't it add some value to test that this works _with_ caching as well?

mypy_django_plugin/transformers/managers.py
ljodal: I'm not sure if I understand what you want here? You mean adding a note down where the attribute is added that the type is resolved in `resolve_manager_method`?

tests/typecheck/managers/querysets/test_from_queryset.yml
flaeppe: I'm not sure there's a consistent way of testing with cache? At least not in a single test suite run.

But I'm quite sure this should run with cache disabled, considering regression etc etc, I'd otherwise, personally, call the test flaky. Since it shouldn't allow the test suite to pass on some runs

ljodal: Good point, I've committed your suggestion now :)

mypy_django_plugin/transformers/managers.py
flaeppe: Not really, my thought was that this approach overrides some work that is done in/together with the copying approach.

I'm just suspecting there's some dead code to remove there? Or perhaps I'm wrong?

mypy_django_plugin/transformers/managers.py
```diff
@@ -232,17 +248,21 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
...
+    # methods are handled above. The actual type of these methods are resolved
+    # in resolve_manager_method.
+
+    # Find all methods that are defined on BaseManager
     manager_method_names = []
     for manager_mro_info in new_manager_info.mro:
         if manager_mro_info.fullname == fullnames.BASE_MANAGER_CLASS_FULLNAME:
-            for name, sym in manager_mro_info.names.items():
+            for name in manager_mro_info.names:
                 manager_method_names.append(name)
```
flaeppe: Perhaps a micro-optimisation and out of scope. But `manager_method_names` could be a set, right?

```python
manager_method_names = set()
for manager_mro_info in new_manager_info.mro:
    if manager_mro_info.fullname == fullnames.BASE_MANAGER_CLASS_FULLNAME:
        for name in manager_mro_info.names:
             manager_method_names.add(name)
}
```

One could also make a comprehension out of it. Though I'm not sure if that's more, or less, readable..

```python
manager_method_names = {
    *manager_mro_info.names
    for manager_mro_info in new_manager_info.mro
    if manager_mro_info.fullname == fullnames.BASE_MANAGER_CLASS_FULLNAME
}
```

mypy_django_plugin/transformers/managers.py
flaeppe: Haha, I completely missed that the copying was replaced... Nvm this.

mypy_django_plugin/transformers/managers.py
ljodal: Haha, I just did basically the same, but decided that it was less readable, so I changed it back.

I'm wondering if it would be better to have a simple predefined list of methods here though. It's fairly easy to define that list and it might be a better approach than trying to detect which methods return querysets. What do you think of that? Then we can use that same list to update the return types in `resolve_manager_method` and avoid potentially affecting custom queryset methods

flaeppe: > What do you think of that?

I'd agree with that (I've had similar thoughts too). It's more difficult to understand what that `"_QS"` is there for, in addition to this list..

flaeppe: I mean, as long as there's a/some test cases testing that whole list. That should suffice for coverage and future changes.

I added a case close to that some time ago

https://github.com/typeddjango/django-stubs/blob/fe2d2287a1e02d35e012d39fedfe8ab0bde19308/tests/typecheck/managers/querysets/test_from_queryset.yml#L350-L386

Asserting on all django manager builtins.

flaeppe:
> The test suite passes, but I'm seeing some new errors in our project.

For some additional reference, I just tried running this on one of our projects without issues.

mypy_django_plugin/transformers/managers.py
ljodal: I decided to implement this and think the code is much more readable and understandable now. It's just 20 methods that we need to modify. I updated (and sorted) the `from_queryset_includes_methods_returning_queryset` test as well, it was missing `.extra()` :)

.github/workflows/test.yml
```diff
@@ -46,7 +47,7 @@ jobs:
           pip install -r ./requirements.txt
 
       - name: Run tests
-        run: pytest --mypy-ini-file=mypy.ini
+        run: pytest --mypy-ini-file="${{ matrix.mypy_ini_file }}"
```
sobolevn: Hm, good idea. Let's try this. You can create a new PR with just this change. It will be helpful!

mypy_django_plugin/transformers/managers.py
```diff
@@ -24,13 +24,39 @@
 from mypy_django_plugin import errorcodes
 from mypy_django_plugin.lib import fullnames, helpers
 
+MANAGER_METHODS_RETURNING_QUERYSET = (
```
sobolevn: nit

```suggestion
MANAGER_METHODS_RETURNING_QUERYSET: Final = (
```

mypy_django_plugin/transformers/managers.py
```diff
@@ -56,11 +82,24 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
 
     assert isinstance(method_type, CallableType)
+
+    variables = method_type.variables
+    ret_type = method_type.ret_type
+
+    # For methods on the manager that return a queryset we need to override the
+    # return type to be the actual queryset class, not the base QuerySet that's
+    # used by the typing stubs.
+    if method_name in MANAGER_METHODS_RETURNING_QUERYSET:
```
sobolevn, line 95: Maybe `MANAGER_METHODS_RETURNING_QUERYSET` can be a `frozenset`?

sobolevn (commented):
Thank you! Looks promising

mypy_django_plugin/transformers/managers.py
flaeppe: Agreed, it looks way way better. Django's queryset API, thus also stubs, won't change that often that there's a need to figure them out programmatically..

.github/workflows/test.yml
ljodal: Yeah, I removed it as it's no longer needed to reproduce the problem after adding `disable_cache: true` to the test case. If we're going to add this I think it makes sense to rename the files as well, but I stole this from #1017 so I think I'll leave it up to the author there to decide if he wants to create a PR for that.

mypy_django_plugin/transformers/managers.py
ljodal: Fixed this now, though I don't think it matters much performance wise. Can't be bothered to run any tests though 😇

tests/typecheck/managers/querysets/test_from_queryset.yml
```diff
@@ -384,3 +385,38 @@
...
+    disable_cache: true
```
sobolevn, line 397: Let's add a comment why do we need this.
And a link to all the discussions somewhere above this test case.

mypy_django_plugin/transformers/managers.py
```diff
@@ -19,18 +19,47 @@
...
 from mypy.types import AnyType, CallableType, Instance, ProperType
 from mypy.types import Type as MypyType
-from mypy.types import TypeOfAny, TypeVarType, UnboundType, get_proper_type
+from mypy.types import TypeOfAny
+from typing_extensions import Final
 
 from mypy_django_plugin import errorcodes
 from mypy_django_plugin.lib import fullnames, helpers
 
+MANAGER_METHODS_RETURNING_QUERYSET: Final[FrozenSet[str]] = frozenset(
```
```suggestion
MANAGER_METHODS_RETURNING_QUERYSET: Final = frozenset(
```

mypy_django_plugin/transformers/managers.py
ljodal: I bothered anyway and frozenset wins. A lookup there, wether it's a match or not, is about 11 nsec on my laptop, while for tuple it varies from 9 nsec if it's the first item that matches to 100 nsec if it's the last or the item is not in the list.

mypy_django_plugin/transformers/managers.py
```diff
@@ -1,4 +1,4 @@
-from typing import Optional, Union
+from typing import FrozenSet, Optional, Union
```
```suggestion
from typing import Optional, Union
```

tests/typecheck/managers/querysets/test_from_queryset.yml
ljodal: Done ✅

mypy_django_plugin/transformers/managers.py
```diff
@@ -19,18 +19,47 @@
...
+        "select_related",
+        "union",
+        "using",
+    )
+)
+
 
 def get_method_type_from_dynamic_manager(
-    api: TypeChecker, method_name: str, manager_type_info: TypeInfo
+    api: TypeChecker, method_name: str, manager_type: Instance
```
```suggestion
    api: TypeChecker, method_name: str, manager_instance: Instance
```

Maybe? Just thought it looked off with the `manager_type.type`

flaeppe (approved):
LGTM 👍

sobolevn (approved):
Thank you, great work!

christianbundy:
Thanks a lot! I'm excited to try this when a new version is released.

Issue #114, opened 2019-07-26
Don't make psycopg2 a dependency


kylebebak:
Installing `pscyopg2` can be difficult on many platforms.

Making this a dep for type stubs seems **really** excessive to me.

sobolevn:
One can also use `psycopg2-binary` and they might conflict.

kylebebak:
Awesome @mkurnikov 

Thanks for the quick response, and thanks for these awesome type stubs =)

PR #1029, opened 2022-06-28
Support cursor.execute(psycopg2.sql.Composable)

andersk:
In addition to `str`, PostgreSQL cursors accept the `psycopg2.sql.Composable` type, which is useful for guarding against SQL injections when building raw queries that can’t be parameterized in the normal way (e.g. interpolating identifiers).

In order to avoid reintroducing a dependency on `psycopg2`, we define a `Protocol` that matches `psycopg2.sql.Composable`.

Documentation: https://www.psycopg.org/docs/sql.html
Related: https://github.com/python/typeshed/pull/7494

### DIFF
--- a/django-stubs/db/backends/utils.pyi
+++ b/django-stubs/db/backends/utils.pyi
@@ -4,7 +4,21 @@ import types
 from contextlib import contextmanager
 from decimal import Decimal
 from logging import Logger
-from typing import Any, Dict, Generator, Iterator, List, Mapping, Optional, Sequence, Tuple, Type, Union, overload
+from typing import (
+    Any,
+    Dict,
+    Generator,
+    Iterator,
+    List,
+    Mapping,
+    Optional,
+    Protocol,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+    overload,
+)
 from uuid import UUID
 
 if sys.version_info < (3, 8):
@@ -13,6 +27,14 @@ else:
     from typing import Literal
 
 logger: Logger
+
+# Protocol matching psycopg2.sql.Composable, to avoid depending psycopg2
+class _Composable(Protocol):
+    def as_string(self, context: Any) -> str: ...
+    def __add__(self, other: _Composable) -> _Composable: ...
+    def __mul__(self, n: int) -> _Composable: ...
+
+_ExecuteQuery = Union[str, _Composable]
 
 # Python types that can be adapted to SQL.
 _SQLType = Union[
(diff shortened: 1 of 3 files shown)


PIG208:
Strictly speaking, should we also account for other backends that might not support `Composable`?

andersk:
@PIG208 Erring on the side of accepting things that might or might not be accepted at runtime is a fine way for type stubs to account for that (and probably the only way in this case).

django-stubs/db/backends/utils.pyi
```diff
@@ -13,6 +27,14 @@ else:
 
 logger: Any
 
+# Protocol matching psycopg2.sql.Composable, to avoid depending psycopg2
+class _Composable(Protocol):
+    def as_string(self, context: Any) -> str: ...
+    def __add__(self, other: _Composable) -> _Composable: ...
+    def __mul__(self, n: int) -> _Composable: ...
+
+_ExecuteQuery = str | _Composable
```
sobolevn: We use `Union`, unlike `typeshed`.
```suggestion
_ExecuteQuery = Union[str, _Composable]
```

tests/typecheck/db/test_connection.yml
```diff
@@ -4,6 +4,13 @@
       with connection.cursor() as cursor:
         reveal_type(cursor)  # N: Revealed type is "django.db.backends.utils.CursorWrapper"
         cursor.execute("SELECT %s", [123])
+-   case: raw_connection_psycopg2_composable
```
sobolevn, line 9:
```suggestion


-   case: raw_connection_psycopg2_composable
```

tests/typecheck/db/test_connection.yml
```diff
@@ -4,6 +4,13 @@
...
+    main: |
+      from django.db import connection
+      from psycopg2.sql import SQL, Identifier
+      with connection.cursor() as cursor:
+        reveal_type(cursor)  # N: Revealed type is "django.db.backends.utils.CursorWrapper"
```
sobolevn: I think that this part can be removed. We do not test `cursor` type here, do we?

sobolevn (commented):
Thank you! 👍

andersk:
Updated.

django-stubs/db/backends/utils.pyi
```diff
@@ -13,6 +27,14 @@ else:
 
 logger: Any
 
+# Protocol matching psycopg2.sql.Composable, to avoid depending psycopg2
+class _Composable(Protocol):
```
sobolevn, line 32: Can we depend on `types-psycopg2`? What do you think?
Refs https://github.com/typeddjango/django-stubs/issues/114

andersk: A risk in that approach is that typeshed packages are obsoleted when the corresponding upstream packages add type annotations. In fact, psycopg3 (psycopg on PyPI) already has upstream type annotations, and we’ll want to support its version of `Composable`, which satisfies the same protocol.

sobolevn: Ok, thanks!

sobolevn (approved):
🤖 👾 ✨

sobolevn:
Can you please resolve a conflict? I've merged something that broke this file `django-stubs/db/backends/utils.pyi`

andersk:
No problem, resolved.

sobolevn (approved):
Thanks again 😊

Issue #960, opened 2022-05-17
"SelfType is unbound" error with custom QuerySet definitions


RJPercival:
# Bug report

## What's wrong

When defining a custom `QuerySet` with methods that return the `QuerySet`, `self` and the return type should both be a `TypeVar` so that if the `QuerySet` is subclassed, the methods defined in the base class return instances of the subclass, e.g.
```py
from django.db import models
from typing import TypeVar

SelfType = TypeVar("SelfType", bound="BaseQuerySet")
ModelType = TypeVar("ModelType", bound="BaseModel")

class BaseQuerySet(models.QuerySet[ModelType]):
   def filter_foo(self: SelfType) -> SelfType:
       return self.filter(foo=True)

class FooQuerySet(BaseQuerySet["FooModel"]):
    pass

FooManager = models.Manager.from_queryset(FooQuerySet)

class BarQuerySet(BaseQuerySet["BarModel"]):
    pass

BarManager = models.Manager.from_queryset(BarQuerySet)

class BaseModel(models.Model):
    class Meta:
        abstract = True
    
    foo = models.BooleanField()

class FooModel(BaseModel):
    objects = FooManager()

class BarModel(BaseModel):
    objects = BarManager()
```

In standard Python code, this approach works. However, with django-stubs, the following error is printed:
```
error: Type variable "SelfType" is unbound  [valid-type]
note: (Hint: Use "Generic[SelfType]" or "Protocol[SelfType]" base class to bind "SelfType" inside a class)
note: (Hint: Use "SelfType" in function signature to bind "SelfType" inside a function)
```

## How is that should be

There should not be an error about the type variable being unbound. `reveal_type(FooModel.objects.filter_foo())` should print `FooQuerySet` and `reveal_type(BarModel.objects.filter_foo())` should print `BarQuerySet`.

## System information

- OS:
- `python` version: 3.8.3
- `django` version: 3.2.13
- `mypy` version: 0.950
- `django-stubs` version: 1.9.0
- `django-stubs-ext` version: 0.3.1

ljodal:
Haven't seen this issue before now, but I'm pretty sure I know where this issue stems from and it's on my list to fix it. Mostly hoping to get a few other PRs merged first though, to avoid too many merge conflicts.

If you're interested it's in `resolve_manager_method()` where we resolve the types of queryset methods that have been copied to managers. This doesn't currently deal with typevars very well (or generics at all really)

PR #1030, opened 2022-06-28
Fix plugin

sterliakov:
Bug fixes:

* This pull requests fixes a bug introduced in #991 that throws "Sequence is not defined".  `bind_or_analyze_type` should not check `node.type`, returning it anyway (`None` too). It make sense, because if lookup succeeded, we have nothing to do to resolve the type and should defer instead.
* Fixed a crash when manager, defined from `.from_queryset` in class body, is used in lookup. It resulted in internal error before. The solution is to always replace this manager with `Any` in such case. Rejected alternative was using default manager (`models.Manager`) instead. It looks worse, because manager could override existing methods. IMO, we'd rather say "this was declared in unexpected place, so we give up and do not check calls to it" (at least because such behavior is documented in README). Surprisingly, this was not reported before.
* Fixed false-positive with staticmethod of queryset (this was not reported as issue)

Other changes:
* Changed handling of managers with missing type arguments. To avoid manual construction of new managers, now the following:
```python
class MyManager(models.Manager): ...
```
is interpreted as
```python
_T_co = typing.TypeVar('_T_co', covariant=True)
class MyManager(models.Manager[_T_co]): ...
```
Using explicit `Any` (`class MyManager(models.Manager[typing.Any])`) preserves `Any` and prevents reparametrization (including subclasses). 

* <s>Python code is reformatted to 88 chars line length. I don't really insist on this style, but it is really inconvenient to read 120 chars with 2-3 columns layout (and it is useful to see two files or code + test). 88 is black's default and looks better, I think.  Stub files remain formatted to 120 chars.</s>
* `UserManager` class is made covariant by type variable (so that model subclass (`ChildUser`) can safely override manager with `UserManager[ChildUser]`)
* Separator in error message "Couldn't resolve manager method..." is changed to the hyphen, because the dot was extremely confusing, rendering as `Couldn't resolve related manager for relation 'appointment' (from appointments.models.Appointment.appointments.Appointment.owner)` (message from my project). 
* <s>Removed `black` from `requirements.txt`, because we don't use it outside of pre-commit, so it is not a real dependency.</s>


flaeppe:
Sounds like this includes quite a bit of changes discussed very recently in:

- #1022
- #1028 
- #958 
- #1017 

To get a better view could, at least, the linting changes be broken out to a separate PR?

sobolevn (commented):
> Python code is reformatted to 88 chars line length. I don't really insist on this style, but it is really inconvenient to read 120 chars with 2-3 columns layout (and it is useful to see two files or code + test). 88 is black's default and looks better, I think. Stub files remain formatted to 120 chars.

I agree, but, please, let's keep our history consistent and our diffs as minimal as possible.

I don't like this style, but I've learned that things above ^ are more important (I did it the hard way).
So, let's keep things as they are.

> Removed black from requirements.txt, because we don't use it outside of pre-c

Sounds good, but let's do it in a separate PR.

sterliakov:
Okay, I'll split this into separate PR's now, one problem per each. 

Any ideas where does the difference with 3.8 and 3.10 originate? One failing test that wants more annotations, I don't quite get the reason. 3.10 (local) runs without failures.

sterliakov:
Okay, I'll split this into separate PR's now, one problem per each. 

Any ideas where does the difference with 3.8 and 3.9+ originate? One failing test that wants more annotations, I don't quite get the reason. 3.10 (local) runs without failures.

flaeppe:
I tried these changes out locally and got into trouble when chaining queryset/manager methods from Django's manager with custom queryset methods. What I tried on one of our projects was:

```python
MyModel.objects.select_for_update().custom_queryset_method()
```

Resulted in these sort of errors:

```
error: "_QuerySet[MyModel, MyModel]" has no attribute "custom_queryset_method"  [attr-defined]
```

But I feel like this should already be tested in our suite here?

Worth noting is that the manager is created with this approach (not sure if this will be full a repro though):

```python
from django.db import models

    def custom_queryset_method(self) -> MyQuerySet:
        ...


MyManager = BaseManager.from_queryset(MyQuerySet)
class MyModel(models.Model):
    objects = MyManager()
```

sterliakov:
@flaeppe This is not a repro, unfortunately. `N: Revealed type is "project.app.models.MyQuerySet"`, works. Isn't it affected by incremental mode (reproduces without cache)? If not, could you try to extract the real reason?

flaeppe:
> @flaeppe This is not a repro, unfortunately. `N: Revealed type is "project.app.models.MyQuerySet"`, works. Isn't it affected by incremental mode (reproduces without cache)? If not, could you try to extract the real reason?

Yeah, it's not cached and not incremental. Happens each run. __But__ currently it seems that it's only `.select_for_update` that results a signature looking like: (haven't had time to check _all_ manager methods yet though)

```python
reveal_type(MyModel.objects.select_for_update)  # Revealed type is "def (nowait: builtins.bool =, skip_locked: builtins.bool =, of: typing.Sequence[builtins.str] =, no_key: builtins.bool =) -> django.db.models.query._QuerySet[MyModel, MyModel]"
```

sterliakov:
One question: *how*? 

```
note: Revealed type is "def (nowait: builtins.bool =, skip_locked: builtins.bool =, of: typing.Sequence[builtins.str] =, no_key: builtins.bool =) -> project.app.models.MyQuerySet[project.app.models.MyModel]"
```

sterliakov:
Could you run `pytest -k regression_manager_scope_foreign` for the failing CI test?) Maybe you're lucky with your setup and will break that test too...

flaeppe:
> One question: _how_?
> 
> ```
> note: Revealed type is "def (nowait: builtins.bool =, skip_locked: builtins.bool =, of: typing.Sequence[builtins.str] =, no_key: builtins.bool =) -> project.app.models.MyQuerySet[project.app.models.MyModel]"
> ```

I didn't know just yet. But now I found a repro. Problematic part, similar to https://github.com/typeddjango/django-stubs/pull/1017#issuecomment-1166575528, is `AUTH_USER_MODEL`:

```
-   case: from_queryset_select_for_update
    disable_cache: true
    main: |
        reveal_type(MyModel.objects.select_for_update)  # N: Revealed type is "def (nowait: builtins.bool =, skip_locked: builtins.bool =, of: typing.Sequence[builtins.str] =, no_key: builtins.bool =) -> myapp.managers.ModelQuerySet[myapp.models.MyModel]"
    custom_settings: |
        AUTH_USER_MODEL = "users.User"
        INSTALLED_APPS = ("django.contrib.auth", "django.contrib.contenttypes", "users", "myapp")
    files:
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                from django.db import models
                from .managers import ModelQuerySet

                NewManager = BaseManager.from_queryset(ModelQuerySet)
                class MyModel(models.Model):
                    created_by = models.ForeignKey(to=settings.AUTH_USER_MODEL, on_delete=models.PROTECT)
                    objects = NewManager()
        -   path: myapp/managers.py
            content: |
                from django.db import models

                if TYPE_CHECKING:
                    from .models import MyModel

                class ModelQuerySet(models.QuerySet["MyModel"]):
                    def queryset_method(self) -> ModelQuerySet:
                        return self.all()
        -   path: users/__init__.py
        -   path: users/models.py
            content: |
                from django.contrib.auth.models import AbstractBaseUser
                from django.db import models
                class User(AbstractBaseUser):
                    email = models.EmailField(unique=True)
                    USERNAME_FIELD = "email"
```

This is a similar setup to what failed in the project I ran it on.

sterliakov:
Well, I figured it out! `bind_or_analyze_type` was not implemented properly initially, we're doing wrong things. When analyzing sequence, we get `SymbolTableNode` from lookup. It is absolutely fine that it has `node.type is None`: we're not always interested in `node.type`. We need `node.node`, which is a `Var` (in "regular cases" like `builtins.bool`) or `TypeInfo` (or something else probably, need to check these). The latter should be used in absolutely another way: we need to check `t.args` as well and construct new instance. So something like this behaves as expected:

```python

def bind_or_analyze_type(
    t: MypyType, api: SemanticAnalyzer, module_name: Optional[str] = None
) -> Optional[MypyType]:
    """Analyze a type. If an unbound type, try to look it up in the given module name.

    That should hopefully give a bound type."""
    if isinstance(t, UnboundType) and module_name is not None:
        node = api.lookup_fully_qualified_or_none(module_name + "." + t.name)
        if node is not None:
            if isinstance(node.node, TypeInfo):
                return Instance(
                    node.node,
                    [bind_or_analyze_type(a, api, module_name) for a in t.args],
                )
            elif isinstance(node.node, Var):
                return node.node.type

    # If lookup failed or type was bound, analyze type. May be `None` too.
    return api.anal_type(t)
```

(and this is not related to `AUTH_USER_MODEL`, fortunately)

sterliakov:
I'll try to find out tomorrow why does one test fail in CI but not locally. Any ideas are welcome, since I've been trying to reproduce it a few hours without success.

mypy_django_plugin/lib/helpers.py
```diff
@@ -153,10 +155,7 @@ def parse_bool(expr: Expression) -> Optional[bool]:
 
 
 def has_any_of_bases(info: TypeInfo, bases: Iterable[str]) -> bool:
-    for base_fullname in bases:
-        if info.has_base(base_fullname):
-            return True
-    return False
+    return any(map(info.has_base, bases))
```
flaeppe: Does this really need its own helper method? I'd be fine if `any(map(info.has_base, bases))` was inlined. Name `has_any_of_bases` doesn't really reveal what it does either, IMO

mypy_django_plugin/lib/helpers.py
```diff
@@ -369,9 +373,20 @@ def bind_or_analyze_type(t: MypyType, api: SemanticAnalyzer, module_name: Option
     That should hopefully give a bound type."""
     if isinstance(t, UnboundType) and module_name is not None:
         node = api.lookup_fully_qualified_or_none(module_name + "." + t.name)
-        if node is not None and node.type is not None:
-            return node.type
-
+        if node is not None:
```
flaeppe: Can we rename `node` to `sym` or `symbol` instead?

```suggestion
        if symbol is not None:
```

mypy_django_plugin/transformers/models.py
```diff
@@ -250,61 +256,63 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+                continue
 
-                custom_model_manager_name = manager.model.__name__ + "_" + manager_class_name
-                try:
-                    custom_manager_type = self.create_new_model_parametrized_manager(
-                        custom_model_manager_name, base_manager_info=manager_info
-                    )
-                except helpers.IncompleteDefnException:
-                    continue
+            if self._is_declared_inside_class(manager_name):
```
flaeppe:
```suggestion
            elif self._is_declared_inside_class(manager_name):
```

mypy_django_plugin/transformers/models.py
```diff
@@ -250,61 +256,63 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
-                # (Django's default) before finishing. And emit an error, to allow for
-                # ignoring a more specialised manager not being resolved while still
-                # setting _some_ type
-                django_manager_info = self.lookup_typeinfo(fullnames.MANAGER_CLASS_FULLNAME)
-                assert (
-                    django_manager_info is not None
-                ), f"Type info for Django's {fullnames.MANAGER_CLASS_FULLNAME} missing"
-                self.add_new_node_to_model_class(
-                    manager_name, Instance(django_manager_info, [Instance(self.model_classdef.info, [])])
+            if not self._should_subclass_manager(manager_info):
```
flaeppe:
```suggestion
            elif not self._should_subclass_manager(manager_info):
```

mypy_django_plugin/transformers/models.py
```diff
@@ -250,61 +256,63 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+
+        if incomplete_manager_defs:
+            if not self.api.final_iteration:
+                # Unless we're on the final round, see if another round could figure out
+                # all manager types
+                raise helpers.IncompleteDefnException()
+            else:
+                self._report_incomplete_defs(incomplete_manager_defs)
+
+    def _report_incomplete_defs(self, incomplete_defs: Set[str]) -> None:
```
flaeppe, line 287: I wouldn't really mind if `_report_incomplete_defs` is inlined instead of branched out in a separate method. But perhaps that's just me.

```suggestion
        if incomplete_manager_defs and not self.api.final_iteration:
            # Unless we're on the final round, see if another round could figure out
            # all manager types
            raise helpers.IncompleteDefnException()
        elif self.api.final_iteration:
```

tests/typecheck/managers/test_managers.yml
```diff
@@ -521,3 +500,145 @@
...
+                    @staticmethod
+                    def do_something(x: int = 1) -> str:
+                        return 'something'
+
+                MyManager = models.Manager.from_queryset(MyQS)
+
+                class MyModel(models.Model):
+                    objects = MyManager()
+
+-   case: from_queryset_select_for_update
```
flaeppe: IMO this test should be moved to `test_from_queryset.yml`

mypy_django_plugin/transformers/managers.py
```diff
@@ -139,6 +143,15 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
     """
     semanal_api = helpers.get_semanal_api(ctx)
 
+    # Do not interfere with `fail_if_manager_type_created_in_model_body`:
+    # stop adding new manager, if we are inside, and fall back to any
+    if semanal_api.scope.classes:
```
flaeppe, line 193: Not sure if/what is more or less preferable

```suggestion
    if semanal_api.is_class_scope():
```

mypy_django_plugin/transformers/models.py
```diff
@@ -362,91 +369,97 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
-                # TODO: Support other reverse managers than `_default_manager`
-                default_reverse_manager_info = self.get_reverse_manager_info(
-                    model_info=related_model_info, derived_from="_default_manager"
-                )
-                if default_reverse_manager_info:
-                    self.add_new_node_to_model_class(
-                        attname, Instance(default_reverse_manager_info, []), no_serialize=True
-                    )
-                    continue
+            if not isinstance(relation, (ManyToOneRel, ManyToManyRel)):
```
flaeppe:
```suggestion
            elif not isinstance(relation, (ManyToOneRel, ManyToManyRel)):
```

tests/typecheck/fields/test_related.yml
```diff
@@ -692,6 +694,8 @@
         myapp/models:9: error: `.from_queryset` called from inside model class body
         myapp/models:11: note: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.TransactionLog]"
         myapp/models:13: note: Revealed type is "Any"
+        myapp/models:14: note: Revealed type is "Any"
+        myapp/models:15: note: Revealed type is "Any"
```
flaeppe: I think perhaps the test name should change? `foreign_key_relationship_for_models_with_custom_manager_unsolvable` -> `types_manager_as_any_when_manager_cant_be_resolved` (feel free to come up with another name). The test feels slightly outdated? I can't figure out what the expected outcome of the test is.

I also think the `Note, that we cannot resolve dynamic calls for custom managers:` should be removed..

mypy_django_plugin/main.py
```diff
@@ -244,6 +270,24 @@ def get_method_hook(self, fullname: str) -> Optional[Callable[[MethodContext], M
 
         return None
 
+    def get_customize_class_mro_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
+        """
+        Reparametrize `Manager`'s with type variable if they miss type argument
+
+        We now always assume that ``class MyManager(Manager)`` is a mistake and
+        should be interpreted as ``class MyNanager(Manager[_T])`` and not be
```
flaeppe:
```suggestion
        should be interpreted as ``class MyManager(Manager[_T])`` and not be
```

tests/typecheck/managers/querysets/test_from_queryset.yml
```diff
@@ -96,9 +96,9 @@
 -   case: from_queryset_returns_intersection_of_manager_and_queryset
     main: |
         from myapp.models import MyModel, NewManager
-        reveal_type(NewManager())  # N: Revealed type is "myapp.models.NewManager"
+        reveal_type(NewManager())  # N: Revealed type is "myapp.models.NewManager[<nothing>]"
         reveal_type(MyModel.objects)  # N: Revealed type is "myapp.models.NewManager[myapp.models.MyModel]"
-        reveal_type(MyModel.objects.get())  # N: Revealed type is "Any"
+        reveal_type(MyModel.objects.get())  # N: Revealed type is "myapp.models.MyModel"
```
flaeppe, line 150: Interesting that we previously expected `Any` from `objects.get`..

mypy_django_plugin/transformers/managers.py
```diff
@@ -43,19 +43,23 @@ def get_method_type_from_dynamic_manager(
...
-            return definition.type
+            return definition
         elif isinstance(definition, Decorator):
-            return definition.func.type
+            return definition.func
         return None
 
-    method_type = get_funcdef_type(queryset_info.get_method(method_name))
-    if method_type is None:
+    meth = get_funcdef(queryset_info.get_method(method_name))
```
```suggestion
    method_node = get_funcdef(queryset_info.get_method(method_name))
```

mypy_django_plugin/transformers/models.py
```diff
@@ -151,85 +153,89 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+                    error_context = field_sym.node
+                self.api.fail(
+                    f"Cannot find model {field.related_model!r} referenced in field {field.name!r}",
+                    ctx=error_context,
                 )
-                self.add_new_node_to_model_class(field.attname, Instance(field_info, [set_type, get_type]))
+                self.add_new_node_to_model_class(field.attname, AnyType(TypeOfAny.explicit))
+                continue
+
+            if related_model_cls._meta.abstract:
```
flaeppe:
```suggestion
            elif related_model_cls._meta.abstract:
```

mypy_django_plugin/lib/helpers.py
sterliakov: Right, done!

mypy_django_plugin/transformers/models.py
sterliakov: IMO, `run_with_model_cls` grows too large, and this part (>10 lines) can be extracted logically and with just a single argument. Too long methods are harder to debug (I had to put 5 breakpoints to `AddRelatedManagers.run_with_model_cls` just to find out where things go wrong working on another issue, it would be easier with shorter methods). I'd rather leave this extracted.

tests/typecheck/managers/querysets/test_from_queryset.yml
sterliakov: Yeah, it was a surprising failure...

mypy_django_plugin/transformers/models.py
sterliakov: I applied all `if -> elif` changes, could you elaborate on why is it important? Is it performance or readability issue?

mypy_django_plugin/transformers/managers.py
sterliakov: Not the best idea: `scope.classes` will match function scope too, while `is_class_scope` checks [exactly](https://github.com/python/mypy/blob/master/mypy/semanal.py#L5164=), so inside function we'll be confused. I doubt it is important and has a use case, but logically `scope.classes` is preferable.

mypy_django_plugin/transformers/models.py
flaeppe: Readability, perhaps personally. But I realised I'm looking at it trying to find cases entering multiple if-branches. But then realise they all do continue

tests/typecheck/fields/test_related.yml
sterliakov: Augh, I was misinterpreting the initial meaning of the test. It should be split into two tests, see now this test (renamed to `reverse_for_manager_defined_in_class_body_falls_back_to_default_relatedmanager`) + new test in `test_managers.yml` (`manager_defined_in_class_body_replaced_with_any`)

mypy_django_plugin/transformers/managers.py
flaeppe: Alright, yeah I bet allowing function scope might perhaps create some other troubles anyways. So doing `is_class_scope` would still be supporting functions halfly, probably.

Lets stick with `scope.classes`

mypy_django_plugin/transformers/models.py
sterliakov: Okay, it makes sense, thanks!

mypy_django_plugin/transformers/models.py
flaeppe: Alright, sure, that makes sense

sterliakov:
I'm out of ideas. Repeating CI stages within `python:3.8.13-slim` docker container doesn't help. This failure does not reproduce locally. Is there any way to debug such inconsistent behavior?

flaeppe:
> I'm out of ideas. Repeating CI stages within `python:3.8.13-slim` docker container doesn't help. This failure does not reproduce locally. Is there any way to debug such inconsistent behavior?

Don't think this is helpful for triggering the error locally.

But it seems that the `Need type annotation for "on_site"` error is (only?) triggered from https://github.com/python/mypy/blob/203bd64d8c14e8ddbed7911eb6f8f8aa61d728de/mypy/checker.py#L3178-L3204 in our case (`self.msg.need_annotation_for_var`), which is called if `Var.is_inferred` is set, I think: https://github.com/python/mypy/blob/203bd64d8c14e8ddbed7911eb6f8f8aa61d728de/mypy/checker.py#L2418-L2423

flaeppe:
I'm able to reproduce the error consistently when moving and running the test outside of pytest..

e.g.

With `mypy --config-file mypy.ini tests/run_it.py`

```python
# tests/run_it.py
from django.db import models
from django.contrib.sites.managers import CurrentSiteManager

class MyModel(models.Model):
    site = models.ForeignKey(Site, on_delete=models.CASCADE)
    on_site = CurrentSiteManager()

reveal_type(MyModel.on_site)
```

sterliakov:
Thanks, it helped a lot! I already tried that, but haven't removed mypy cache and everything passed. This is an incremental mode artifact: first run (or `--no-incremental`) succeeds, but subsequent calls fail.

sterliakov:
The local repro above is not a real repro: it uses this file as `models.py`, but does not add it to `INSTALLED_APPS`. Running through pdb revealed that the model `MyModel` is not resolved at all on subsequent runs, because django context is initialized only with apps listed in `INSTALLED_APPS` from `scripts/django_test_settings.py`, so this is not an example of intended plugin usage. I think that disabling cache for this test is a valid solution (we do already have such tests).

tests/typecheck/managers/querysets/test_from_queryset.yml
```diff
@@ -145,9 +145,9 @@
 -   case: from_queryset_returns_intersection_of_manager_and_queryset
     main: |
         from myapp.models import MyModel, NewManager
-        reveal_type(NewManager())  # N: Revealed type is "myapp.models.NewManager"
+        reveal_type(NewManager())  # N: Revealed type is "myapp.models.NewManager[<nothing>]"
```
sobolevn, line 148: This does not seem right. `NewManager = ModelBaseManager.from_queryset(ModelQuerySet)`, and `ModelQuerySet` is defined as `class ModelQuerySet(models.QuerySet):`

I think that it should be `NewManager[Any]`, because this is how other generics work in Python.

sterliakov: It refers to `UninhabitedType`. This `<nothing>` shows up relatively often, it is not related to this plugin. Here it happens, because there is not enough info to resolve TypeVar, like in [this example](https://mypy-play.net/?mypy=latest&python=3.10&gist=c6b167dcd29ce776d7a1de01dab74a9f). I doubt it is important, because managers are not intended to be instantiated outside of model. More interesting question is why there is no "E: expected type annotations for ..." here, I'll try to figure it out.

ljodal:
This branch causes a crash in our Django project, with this stack trace:

```
  File "mypy/checkexpr.py", line 3978, in accept
  File "mypy/nodes.py", line 1670, in accept
  File "mypy/checkexpr.py", line 2087, in visit_member_expr
  File "mypy/checkexpr.py", line 2105, in analyze_ordinary_member_access
  File "mypy/checkmember.py", line 130, in analyze_member_access
  File "mypy/checkmember.py", line 147, in _analyze_member_access
  File "mypy/checkmember.py", line 231, in analyze_instance_member_access
  File "mypy/checkmember.py", line 385, in analyze_member_var_access
  File "mypy/checkmember.py", line 614, in analyze_var
  File "/Users/sigurd/Developer/django-stubs-sterliakov/mypy_django_plugin/transformers/managers.py", line 175, in resolve_manager_method
    return resolve_manager_method_from_instance(instance=ctx.type, method_name=method_name, ctx=ctx)
  File "/Users/sigurd/Developer/django-stubs-sterliakov/mypy_django_plugin/transformers/managers.py", line 148, in resolve_manager_method_from_instance
    method_type = get_method_type_from_dynamic_manager(
  File "/Users/sigurd/Developer/django-stubs-sterliakov/mypy_django_plugin/transformers/managers.py", line 88, in get_method_type_from_dynamic_manager
    assert isinstance(method_type, CallableType)
```

I haven't had time to dive into exactly why yet. It might very well be something with the code, but I don't think mypy should crash in the case of invalid types.

---

Edit: Some additional context. The offending code something like this:

```python

class CustomQuerySet(models.QuerySet):
    def method(self) -> "QuerySet[MyModel]":
        from other_app.models import OtherModel
        return (
            self.filter(foo=OtherModel.objects.untyped_method_from_custom_queryset()[:1])
            .another_custom_method()
        )

**Comment by sterliakov** [comment]
The crash is definitely a serious problem, yeah. Confirmed, investigating now.

**Comment by sterliakov** [comment]
@ljodal Should work now.

**Comment by ljodal** [comment]
Yeah, I can confirm that it's no longer crashing 🎉

It's introducing a lot of new errors, but that's because a lot of our managers are missing the generic type for the models they type :)

File: django-stubs/db/models/manager.pyi [hunk 15]
```diff
@@ -144,4 +144,4 @@ class ManagerDescriptor:
 
 class EmptyManager(Manager):
     def __init__(self, model: Type[Model]) -> None: ...
-    def get_queryset(self) -> QuerySet[Model]: ...
+    def get_queryset(self) -> QuerySet[Model]: ...  # type: ignore[override]
```
ljodal: Rather than using a `type: ignore` comment I think you can just change this definition to this:

```python
class EmptyManager(Manager[T]):
    def __init__(self, model: Type[T]) -> None: ...
```

mypy_django_plugin/transformers/fields.py
```diff
@@ -206,7 +207,7 @@ def transform_into_proper_return_type(ctx: FunctionContext, django_context: Djan
 
     assert isinstance(outer_model_info, TypeInfo)
 
-    if helpers.has_any_of_bases(default_return_type.type, fullnames.RELATED_FIELDS_CLASSES):
+    if any(map(default_return_type.type.has_base, fullnames.RELATED_FIELDS_CLASSES)):
```
ljodal, line 210: I personally prefer the previous version here as it's more explicit and thus easier to read

mypy_django_plugin/transformers/models.py
```diff
@@ -151,85 +153,88 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
 
 class AddManagers(ModelClassInitializer):
     def has_any_parametrized_manager_as_base(self, info: TypeInfo) -> bool:
-        for base in helpers.iter_bases(info):
-            if self.is_any_parametrized_manager(base):
-                return True
-        return False
+        return any(map(self.is_any_parametrized_manager, helpers.iter_bases(info)))
 
     def is_any_parametrized_manager(self, typ: Instance) -> bool:
```
ljodal:
```suggestion
    def is_manager_with_implicit_any_parameter(self, typ: Instance) -> bool:
```

mypy_django_plugin/transformers/models.py
```diff
@@ -151,85 +153,88 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+            self.add_new_node_to_model_class(field.attname, Instance(field_info, [set_type, get_type]))
 
 
 class AddManagers(ModelClassInitializer):
     def has_any_parametrized_manager_as_base(self, info: TypeInfo) -> bool:
-        for base in helpers.iter_bases(info):
-            if self.is_any_parametrized_manager(base):
-                return True
-        return False
+        return any(map(self.is_any_parametrized_manager, helpers.iter_bases(info)))
```
ljodal:
```suggestion
        return any(map(self.is_manager_with_implicit_any_parameter, helpers.iter_bases(info)))
```

mypy_django_plugin/transformers/models.py
```diff
@@ -151,85 +153,88 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+
+            is_nullable = self.django_context.get_field_nullability(field, None)
+            set_type, get_type = get_field_descriptor_types(
+                field_info, is_set_nullable=is_nullable, is_get_nullable=is_nullable
+            )
+            self.add_new_node_to_model_class(field.attname, Instance(field_info, [set_type, get_type]))
 
 
 class AddManagers(ModelClassInitializer):
     def has_any_parametrized_manager_as_base(self, info: TypeInfo) -> bool:
```
ljodal:
```suggestion
    def has_manager_with_implicit_any_parameter_as_base(self, info: TypeInfo) -> bool:
```

mypy_django_plugin/transformers/models.py
```diff
@@ -151,85 +153,88 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
-            new_manager_info.names[name] = new_sym
+        Ending up here could for instance be due to having a custom _Manager_
+        that is not built from a custom QuerySet. Another example is a
+        related manager.
+        Don't interfere with dynamically generated manager classes
+        """
 
-        return custom_manager_type
+        is_dynamic = manager.metadata.get("django", {}).get("from_queryset_manager")
+        return self.has_any_parametrized_manager_as_base(manager) and not is_dynamic
```
ljodal:
```suggestion
        return self.has_manager_with_implicit_any_parameter_as_base(manager) and not is_dynamic
```

mypy_django_plugin/transformers/models.py
```diff
@@ -254,56 +259,62 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
-                    custom_manager_type = self.create_new_model_parametrized_manager(
-                        custom_model_manager_name, base_manager_info=manager_info
-                    )
-                except helpers.IncompleteDefnException:
-                    continue
+                continue
+            elif self._is_declared_inside_class(manager_name):
+                continue
+            elif not self._should_subclass_manager(manager_info):
+                self.add_new_node_to_model_class(manager_name, self.model_parametrize(manager_info))
```
ljodal, line 266: I think we should emit an error if a manager is explicitly typed with `Any`. I think this behaviour could be very surprising. I can easily see someone writing this and expecting a reusable manager:

```python
class MyManager(models.Manager[Any]):
    def has_items() -> bool:
        return self.exists()
```

When used that will result in this:

```python
class MyModel(models.Model):
    objects = MyManager()

reveal_type(MyModel.objects.get())  # note: Revealed type is "Any"
```

I don't really see any use case for this, so I think we should nudge people in the right direction with a useful message (maybe a hint if that's possible?). Maybe something like this:

> A manager typed with "Any" will not have types for the model it's used on. You might want to use a "TypeVar" bound to "Model" instead (ie. "TypeVar('T', bound=models.Model)")

Not super happy with that wording, but something like that would probably be useful

tests/typecheck/managers/test_managers.yml
```diff
@@ -394,46 +395,25 @@
         -   path: myapp/models.py
             content: |
                 from django.db import models
-                class MyModelManager(models.Manager):
-
-                    def create(self, **kwargs) -> 'MyModel':
-                          return super().create(**kwargs)
```
ljodal, line 400: I think it makes sense to keep this test, but change it to expect error messages to be emitted here

ljodal (commented):
Awesome work! I really like the approach to making managers implicitly parametrized 🎉 

As has already been mentioned it would be great if this PR could be split into separate PRs, that makes it much easier to review the PR. Based on what I've seen I'd suggest as least three separate PRs:

1: Manager changes
2: Settings changes
3: Various minor code style changes

django-stubs/db/models/manager.pyi
ljodal: Decided to open a separate PR for this, as it's not really related to your changes here: #1039

mypy_django_plugin/transformers/models.py
flaeppe: Are you sure we should deny the use of `Any`? We then also have to be careful so the error isn't emitted for our own plugin insertions of `Any`, etc.

> ... I can easily see someone writing this and expecting a reusable manager ...

While that might be true, it also feels like someone misunderstanding generics, in general, with mypy.

mypy_django_plugin/main.py
```diff
@@ -60,6 +62,30 @@ def add_new_manager_base_hook(ctx: ClassDefContext) -> None:
...
+    should_reparametrize = (
+        manager.node.bases
+        and manager.node.bases[0].args
+        and isinstance(manager.node.bases[0].args[0], AnyType)
+        and manager.node.bases[0].args[0].type_of_any != TypeOfAny.explicit
+        and manager.node.bases[0].type.defn.type_vars
+    )
+    if not should_reparametrize:
+        return
+
```
flaeppe:
```suggestion
    preserve_any = not (
        manager.node.bases
        and manager.node.bases[0].args
        and isinstance(manager.node.bases[0].args[0], AnyType)
        and manager.node.bases[0].args[0].type_of_any != TypeOfAny.explicit
        and manager.node.bases[0].type.defn.type_vars
    )
    if preserve_any:
        return

    # Reparametrize
```

mypy_django_plugin/main.py
```diff
@@ -60,6 +62,30 @@ def add_new_manager_base_hook(ctx: ClassDefContext) -> None:
...
 
+def reparametrize_manager_base_hook(ctx: ClassDefContext) -> None:
+    manager = ctx.api.lookup_fully_qualified_or_none(ctx.cls.fullname)
+    if manager is None or manager.node is None:
+        return
+    assert isinstance(manager.node, TypeInfo)
+
+    should_reparametrize = (
+        manager.node.bases
+        and manager.node.bases[0].args
```
flaeppe: What if I position a different type as first base?

```python
from django.db import models

class MySuperMixin:
    ...

class MyManager(MySuperMixin, models.Manager):
    ...
```

And since we've already "filtered" that the class we get inherits `BaseManager` in `get_customize_class_mro_hook`, we should be able to traverse all bases and assert that we find it.

mypy_django_plugin/main.py
```diff
@@ -60,6 +62,30 @@ def add_new_manager_base_hook(ctx: ClassDefContext) -> None:
     helpers.add_new_manager_base(ctx.api, ctx.cls.fullname)
 
 
+def reparametrize_manager_base_hook(ctx: ClassDefContext) -> None:
```
flaeppe: Have we tested reparametrization works when inheriting

- `django.db.models.Manager`, and
- `django.db.models.manager.BaseManager`

?

flaeppe:
> Other changes:
> 
> * Changed handling of managers with missing type arguments. To avoid manual construction of new managers, now the following:
> 
> ```python
> class MyManager(models.Manager): ...
> ```
> 
> is interpreted as
> 
> ```python
> _T_co = typing.TypeVar('_T_co', covariant=True)
> class MyManager(models.Manager[_T_co]): ...
> ```
> 
> Using explicit `Any` (`class MyManager(models.Manager[typing.Any])`) preserves `Any` and prevents reparametrization (including subclasses).

If we start doing this for managers, should we do the same for `QuerySet`?

mypy_django_plugin/transformers/models.py
ljodal: > Are you sure we should deny the use of `Any`? We then also have to be careful so the error isn't emitted for our own plugin insertions of `Any`, etc.

That should be a problem though? We just need to emit the error only for explicit `Any`

> > ... I can easily see someone writing this and expecting a reusable manager ...
> 
> 
> 
> While that might be true, it also feels like someone misunderstanding generics, in general, with mypy.

Yeah, but that's exactly who I'd target with an error (or as noted above ideally a note) nudging in the right direction.

ljodal:
I'm kinda torn about doing the same reparametrizing for querysets. For mangers doing this avoids the need for calling `copy_method_to_another_class` with all the weird bugs that entails. The trade off is that you get some surprising warnings referring to a `_T` you didn't define. I'm not sure it's worth the trade off for QuerySets where this isn't really needed.

Ideally I'd say we should only do this for managers defined in third party code, and require own managers to specify a parameter, by that seems infeasible.

mypy_django_plugin/transformers/models.py
sterliakov: It is not a problem to emit an error on explicit `Any`, but I think that it is extremely rare use case. In general, any `mypy` output is considered a problem, while `Any` type arg is absolutely valid. I doubt that typing library plugin should teach users to use generics, it is up to documentation, isn't it? We are still doing many things to lower required knowledge: if I were designing this plugin initially, I'd never allow `class MyManager(models.Manager)` to be reparametrized, because it is easily expressible with TypeVar (and this is a common pattern).

mypy_django_plugin/main.py
sterliakov: Good catch! Will fix now

sterliakov:
I wanted to do the same for querysets in separate PR. For me, the main advantage of doing this for querysets is proper type representation. Now we have very inconsistent handling of it (shortened version of `from_queryset_queryset_imported_from_other_module`):

```yml
    main: |
        reveal_type(MyModel.objects.filter(id=1).queryset_method()) # N: Revealed type is "myapp.querysets.ModelQuerySet"
        reveal_type(MyModel.objects.filter(id=1)) # N: Revealed type is "myapp.querysets.ModelQuerySet[myapp.models.MyModel]"
    installed_apps:
        - myapp
    files:
        -   path: myapp/__init__.py
        -   path: myapp/querysets.py
            content: |
                from django.db import models
                if TYPE_CHECKING:
                    from .models import MyModel

                class ModelQuerySet(models.QuerySet["MyModel"]):
                    def queryset_method(self) -> "ModelQuerySet":
                        return self.filter()
        -   path: myapp/models.py
            content: |
                from django.db import models
                from .querysets import ModelQuerySet

                NewManager = BaseManager.from_queryset(ModelQuerySet)
                class MyModel(models.Model):
                    objects = NewManager()
```

Note how two querysets referring to the same object are represented. (And the parametrized version is, strictly speaking, invalid, because `ModelQuerySet` is not generic)

mypy_django_plugin/transformers/models.py
sterliakov: Maybe just inline this? It is used in single place, and function name doesn't read better than `any(map(...))` directly.

mypy_django_plugin/transformers/fields.py
sterliakov: And I like this one, it's just a matter of taste..

sobolevn (approved):
Looks ready to me. Do others agree? :)

flaeppe:
I think some comments marked as resolved hasn't been changed yet?

e.g. https://github.com/typeddjango/django-stubs/pull/1030#discussion_r912325274

sterliakov:
Not ready, sorry: thanks to [this comment](https://github.com/typeddjango/django-stubs/pull/1030#discussion_r912325569) a critical bug was found. 

```
class MyModel(models.Model):
    objects = BaseManager()
reveal_type(MyModel.objects)
```

This produces `need type annotation for "objects"` for some reason, but revealed type is fine. I spent a day trying to find out, but probably need some more time.

ljodal:
> Looks ready to me. Do others agree? :)

Other than what's said above I'm not aware of any issues caused by this, but as I've noted previously it would be great if this was split up into multiple PRs (one for each fix). That makes it much easier to review and also if there are any issues it's much easier to identify what caused them

sterliakov:
That problem was related to metadata, I think it may be nice to get rid of manager metadata and rely on subclassing solely (instead of `_get_current_manager_bases` check if resolved type info inherits from BaseManager).

mypy_django_plugin/transformers/querysets.py
```diff
@@ -30,6 +30,14 @@ def _extract_model_type_from_queryset(queryset_type: Instance) -> Optional[Insta
 
 
 def determine_proper_manager_type(ctx: FunctionContext) -> MypyType:
+    """
+    This hook parametrizes manager on model with model itself, so that code::
+
+        class MyModel(models.Model):
+            objects = MyManager()
+
+    has `objects` replaced with `MyManager[MyModel]` type.
```
flaeppe:
```suggestion
    This hook populates manager model argument for a manager initialized at model class level. code::

        class MyModel(models.Model):
            objects = MyManager()

    gets a type for `objects` like `MyManager[MyModel]`.
```

mypy_django_plugin/transformers/managers.py
```diff
@@ -306,3 +320,53 @@ def fail_if_manager_type_created_in_model_body(ctx: MethodContext) -> MypyType:
...
+        return
+    assert isinstance(base_manager.node, TypeInfo)
+
+    tvars = tuple(base_manager.node.defn.type_vars)
+    parent_manager.args = tvars
+    manager.node.type_vars = []
+    manager.node.defn.type_vars = list(tvars)
+    manager.node.add_type_vars()
+    if not ctx.api.final_iteration:
+        ctx.api.defer()
```
flaeppe: Is this hook called on every iteration? If so, then we're deferring on every iteration, right? A first reaction is that it feels unnecessary

mypy_django_plugin/main.py
```diff
@@ -302,7 +324,12 @@ def get_type_analyze_hook(self, fullname: str) -> Optional[Callable[[AnalyzeType
             return partial(handle_annotated_type, django_context=self.django_context)
 
     def get_dynamic_class_hook(self, fullname: str) -> Optional[Callable[[DynamicClassDefContext], None]]:
-        # Create a new manager class definition when a manager's '.from_queryset' classmethod is called
+        """
+        Create a new manager class from manager's `.from_queryset` method
+
+        Create a new manager class definition when a manager's `.from_queryset`
+        classmethod is called
```
flaeppe: We should keep 1 of these phrases

```suggestion
        Create a new manager class definition when a manager's `.from_queryset`
        classmethod is called
```

mypy_django_plugin/transformers/fields.py
```diff
@@ -93,11 +93,12 @@ def fill_descriptor_types_for_related_field(ctx: FunctionContext, django_context
         related_model_type = Instance(related_model_info, [])  # type: ignore
 
     related_model_to_set_info = helpers.lookup_class_typeinfo(typechecker_api, related_model_to_set)
+    related_model_to_set_type: Union[MypyType, Instance]
```
```suggestion
    related_model_to_set_type: MypyType
```

Since Instance is a type

mypy_django_plugin/transformers/managers.py
sterliakov: Should be better now

sobolevn:
There are merge conflicts now 😞 
Can you please resolve them?

sterliakov:
If we allow `.from_queryset` inside class body, the following dies forever:

```
-   case: test_queryset_in_model_class_body_subclass
    main: |
        reveal_type(MyModel.objects)  # N: Revealed type is "myapp.models.BaseManagerFromBaseQuerySet"
        reveal_type(MyModel.objects.get())  # N: Revealed type is "myapp.models.MyModel"
    installed_apps:
        - myapp
    files:
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                from django.db import models

                class BaseManager(models.Manager["BaseModel"]):
                    pass

                class BaseQuerySet(models.QuerySet["BaseModel"]):
                    def custom(self) -> "BaseQuerySet":
                        pass

                    objects = BaseManager.from_queryset(BaseQuerySet)()

                class MyModel(BaseModel):
                    pass
```

Because we create model-parametrized manager for `BaseModel` in class and cannot subclass it later, so in `MyModel` manager remains attached to `BaseModel` and `MyModel.objects.get()` is `BaseModel`. 

This conflict looks quite difficult. I'll try to resolve this, but now I don't see any solution - maybe this PR will be abandoned. Should I submit separate PR with just fix for "Sequence is not defined" to fix this popular issue and release?

ljodal:
> If we allow `.from_queryset` inside class body, the following dies forever:
> 
> ```
> -   case: test_queryset_in_model_class_body_subclass
>     main: |
>         from myapp.models import MyModel
>         reveal_type(MyModel.objects)  # N: Revealed type is "myapp.models.BaseManagerFromBaseQuerySet"
>         reveal_type(MyModel.objects.get())  # N: Revealed type is "myapp.models.MyModel"
>     installed_apps:
>         - myapp
>     files:
>         -   path: myapp/__init__.py
>         -   path: myapp/models.py
>             content: |
>                 from django.db import models
> 
>                 class BaseManager(models.Manager["BaseModel"]):
>                     pass
> 
>                 class BaseQuerySet(models.QuerySet["BaseModel"]):
>                     def custom(self) -> "BaseQuerySet":
>                         pass
> 
>                 class BaseModel(models.Model):
>                     objects = BaseManager.from_queryset(BaseQuerySet)()
> 
>                 class MyModel(BaseModel):
>                     pass
> ```
> 
> Because we create model-parametrized manager for `BaseModel` in class and cannot subclass it later, so in `MyModel` manager remains attached to `BaseModel` and `MyModel.objects.get()` is `BaseModel`.

Hmm, I didn't really consider that case, but technically it shouldn't really a problem. The _class_ we create has to be generic (just like it is at runtime) and the _type_ of `BaseModel.objects` should be `BaseManagerFromBaseQuerySet[BaseModel]`. The plugin can then correct that in the subclass to set the type to `BaseManagerFromBaseQuerySet[MyModel]`. There's a bunch of other issues with generic querysets/managers though, so I think it's be better to fix that in a separate PR?

I think @flaeppe has an implementation of the latter (or something very closely related here): https://github.com/flaeppe/django-stubs/pull/29

> This conflict looks quite difficult. I'll try to resolve this, but now I don't see any solution - maybe this PR will be abandoned. Should I submit separate PR with just fix for "Sequence is not defined" to fix this popular issue and release?

That error is already fixed (or should be, I've not seen it since) in https://github.com/typeddjango/django-stubs/commit/84eff755661c3c498c36e66b7ff0b6cc8fc3d4bb, but there's at least a few other things from this PR that would be good to get merged. I don't have a full overview of exactly what's in here at the moment, but the reparametrization of manager classes is a really interesting approach and we can use it to get rid of the copying of methods (instead of trying to fix it)

On a general level my experience is that it's much easier to both review and keep alive smaller PRs, so if you're able to I'd recommend trying to split this up into separate PRs.

sterliakov:
Unfortunately I'm 100% unable to split this into multiple chunks, because there was black change (reverted later) and trying to play with history will surely result in something broken/inconsistent. 

> There's a bunch of other issues with generic querysets/managers though, so I think it's be better to fix that in a separate PR?

Could you elaborate on that? After merging this is the only problem arising (I don't talk about inconsistent QuerySet parametrization now, it is a subject of separate PR and needs later discussion anyway).

I'll try to recover this information from generic types now, maybe it is easier than seemed initially for me.

flaeppe:
Without having looked deeper. I'm fairly certain that AddManagers should be able to resolve the missing parametrization when subclassing. Or have I missed something?

It currently traverses parents mapping managers at least

ljodal:
> > There's a bunch of other issues with generic querysets/managers though, so I think it's be better to fix that in a separate PR?
> 
> Could you elaborate on that? After merging this is the only problem arising (I don't talk about inconsistent QuerySet parametrization now, it is a subject of separate PR and needs later discussion anyway).
> 
> I'll try to recover this information from generic types now, maybe it is easier than seemed initially for me.

There's at least #960, #1049, and #1046. Not entirely sure where all of them stem from, but at least #960 is because of the way we resolve types of methods copied to managers. We also incorrectly parametrize managers/querysets that are already parametrized. For example in this case:

```python
class MyQuerySet(QuerySet["MyModel"]):
    my_custom_method(self) -> "MyQuerySet": pass
```

When this is combined with a manager the return type is changed to `MyQuerySet[MyModel]` which isn't correct (though not causing any problems either it seems).

Basically the `resolve_manager_method` hook has to be expanded to be much smarter, but it's not really related to what you've been fixing here so I think it'd be better to take it in a separate PR. Much easier to to find (and fix) regressions with `git bisect` when the commits are small :)

sterliakov:
I'm dumb, sorry. This failure is not related to declaration inside body. This is specific to my original implementation:

```
-   case: inheritance_with_parents_manager_parametrized_with_parent
    main: |
        from myapp.models import ParentOfMyModel, MyModel
        reveal_type(ParentOfMyModel.objects.first()) # N: Revealed type is "Union[myapp.models.ParentOfMyModel, None]"
        reveal_type(MyModel.objects.first()) # N: Revealed type is "Union[myapp.models.MyModel, None]"
    installed_apps:
        - myapp
    files:
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                from django.db import models

                class MyManager(models.Manager['ParentOfMyModel']):
                    pass

                class ParentOfMyModel(models.Model):
                    objects = MyManager()

                class MyModel(ParentOfMyModel):
                    pass
```

This is failing as well.

```
E   Actual:
E     ...
E     main:3: note: Revealed type is "Union[myapp.models.ParentOfMyModel, None]" (diff)
E   Expected:
E     ...
E     main:3: note: Revealed type is "Union[myapp.models.MyModel, None]" (diff)
E     E: ... "Union[myapp.models.MyModel, None]"
E     A: ... "Union[myapp.models.ParentOfMyModel, None]"
```

My implementation never reparametrizes managers that already have model parameters. Idk how this was missing in the test suite. So I'm going to fix this, thanks for the new test! It should be not this hard, I hope.

sterliakov:
After some investigation I think that the behavior I reported in previous comment is a correct one (and consistent with regular typing in python). Consider the following scenario:

```
from django.db.models import Manager, Model
from typing import TypeVar

_M = TypeVar('_M', bound=Parent)

class MyManagerFixed(Manager[Parent]):
    def get(self, **kwargs: Any) -> Parent:
        return Parent.objects.first()  # I wanna ALWAYS return a Parent, because this is intended only for use on Parent
class MyManagerDynamic(Manager[_M]):
    pass
    
class Parent(Model):
    fixed_objects = MyManagerFixed()
    dynamic_objects = MyManagerDynamic()
     
class Child(Parent):
    fixed_objects = MyManagerFixed()
    dynamic_objects = MyManagerDynamic()
```

Now this is 100% valid on Parent, but very suspicious on Child: mypy thinks that `fixed_objects` returns Parent instances despite being used on Child. This is valid (since `Child <: Parent`) and does not violate `get` definition on MyManagerFixed. If we allow reparametrization with child models, then it is equivalent to `class MyManagerFixed[_M]`. This violates at least "one and preferably only one", plus does not allow to express the concept of "manager for exactly this model", and IMO the current way is more obvious. Also this will create an invalid manager class: it'll be `Manager[Child]` with `get: (**kwargs: Any) -> Parent`, and `mypy` should complain about this. There's nothing wrong with explicitly declaring manager as parametrized with bound TypeVar. 

This contradicts with the fact that on runtime managers are really model-parametrized, though. IMO the most cheap solution here is to require using TypeVar's as Manager parameter unless other behavior is required. Plus, if we decide to reparametrize, we need to verify that manager is used really on model subtype only, and this is really tricky (mind TypeVar's, aliases, NewType's, recursive things which are supported now, variadic generics, ...). And this will require copying all methods to new plugin-generated Manager subclass parametrized with TypeVar or Child model, because we should not change real objects created by user (I mean we cannot just take `MyManagerFixed` and redefine everything in it to make our plugin happy, but should create another instance instead).

So I ask for maintainer's opinion here: I can either resolve conflicts and prepare this to be merged again (if my arguments convinced you) or try to implement reparametrization here. There are some conflicts now, but the idea of this PR does not really interfere with these changes.

UnknownPlatypus:
Closing as superseded: everything here has since landed through other pr's:
- implicit-generic manager reparametrization via #1169 (extended to querysets in #3217 and fields in #3325)
- `from_queryset` in class body support in #1045
- `UserManager` generics in #2384. 
- the queryset-staticmethod false positive was fixed in #3501 recently, I've also added a regression test inspired by this pr to ensure no later regressin

Thanks!

PR #1033, opened 2022-06-29
Populate model argument for dynamically created managers

### LINKED ISSUE #1032: Problem with manager created using .from_queryset() and imported from other module
# Bug report

When a manager is created using `.from_queryset()` and imported from another module, it's type is `Any`, but only when the `QuerySet` subclass has its generic specified as `["MyModel"]`.

## What's wrong

I took [this test case](https://github.com/typeddjango/django-stubs/blob/214b0c7439d67b423815515c49eda8ecf0d62bfa/tests/typecheck/managers/querysets/test_from_queryset.yml#L25-L72) and adapted it to move the `NewManager = BaseManager.from_queryset(ModelQuerySet)` line into `myapp/querysets.py` and imported `NewManager` into `myapp/models.py` instead of `ModelQuerySet`.

See this diff: [from_queryset-extra-tests.txt](https://github.com/typeddjango/django-stubs/files/9005556/from_queryset-extra-tests.txt)

Applying this to the current `master` results in the following two tests failing:
- `from_queryset_using_basemanager_with_manager_imported_from_other_module`
- `from_queryset_using_manager_with_manager_imported_from_other_module`

If I change the following in `myapp/querysets.py` for these tests then they pass:

```python
-                class ModelQuerySet(models.QuerySet["MyModel"]):
+                class ModelQuerySet(models.QuerySet):
```

Weird.

As far as I could work out `get_dynamic_class_hook` is being called for `myapp.models.NewManager`, but then `get_function_hook`/`get_method_hook` doesn't seem to get called for methods on the manager in the failure case.  

## How is that should be

When `NewManager` is created using `.from_queryset()` and is imported from another module it shouldn't resolve to `Any` when `Queryset["MyModel"]` is used as the subclass, but should be the same as when only `QuerySet` is used as the subclass or `NewManager` is created in `myapp/models.py`.

## System information

- OS:
- `python` version: 3.9.5
- `django` version: 3.2.13
- `mypy` version: 0.961

flaeppe:
`AddManagers` will now parametrize and populate model argument for dynamically created managers

## Related issues

Closes #1032

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -243,26 +243,25 @@ class AddManagers(ModelClassInitializer):
             except helpers.IncompleteDefnException as exc:
                 # Check if manager is a generated (dynamic class) manager
                 base_manager_fullname = helpers.get_class_fullname(manager.__class__.__bases__[0])
-                if manager_fullname not in self.get_generated_manager_mappings(base_manager_fullname):
+                generated_managers = self.get_generated_manager_mappings(base_manager_fullname)
+                if manager_fullname not in generated_managers:
                     # Manager doesn't appear to be generated. Track that we encountered an
                     # incomplete definition and skip
                     incomplete_manager_defs.add(manager_name)
-                continue
+                    continue
 
-            if manager_name not in self.model_classdef.info.names:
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Awesome!

PR #1036, opened 2022-06-30
Improved return type of render_to_string().

smithdc1:
[render_to_string()](https://github.com/django/django/blob/bb2c5f69f47466fa52f3cf2727d10b3ebd79a4da/django/template/loader.py#L52) finds an [`_EngineTemplate`](https://github.com/typeddjango/django-stubs/blob/214b0c7439d67b423815515c49eda8ecf0d62bfa/django-stubs/template/backends/base.pyi#L21) and calls `render()` on it. Therefore it returns a `SafeString`.


sobolevn (approved):
Thanks!

PR #1039, opened 2022-07-02
Fix type stubs for EmptyManager

ljodal:
This should be parametrized based on the passed model. This also removes the need for overriding the `get_queryset` method.


sobolevn (approved):
I think it should work with our `django_stubs_ext` as is. Correct me if I am wrong.

Thanks!

PR #1040, opened 2022-07-02
Made BaseContext __setitem__ and __getitem__ consistant.

smithdc1:
I found that I was unable to "get" a Node from a `RenderContext`. 

Currently `__setitem__` and `__getitem__` of BaseContext are inconsistant. Making the types for these two functions solves this issue for me.


django-stubs/template/context.pyi
```diff
@@ -23,9 +23,9 @@ class BaseContext(Iterable[Any]):
...
     def push(self, *args: Any, **kwargs: Any) -> ContextDict: ...
     def pop(self) -> ContextDict: ...
-    def __setitem__(self, key: Union[Node, str], value: Any) -> None: ...
+    def __setitem__(self, key: Union[int, str, Node], value: Any) -> None: ...
     def set_upward(self, key: str, value: Union[int, str]) -> None: ...
-    def __getitem__(self, key: Union[int, str]) -> Any: ...
+    def __getitem__(self, key: Union[int, str, Node]) -> Any: ...
     def __delitem__(self, key: Any) -> None: ...
     def __contains__(self, key: str) -> bool: ...
     def get(self, key: str, otherwise: Optional[Any] = ...) -> Optional[Any]: ...
```
sobolevn: What about `get`? Right now it only works with `str`, but `__getitem__` works with `int` and `Node` as well.

smithdc1: Looking at this again, I thought maybe it's not just `get` either. 

It's really the type of the context `key` and should be consistent across the whole class?

sobolevn (approved):
Thanks!

PR #1041, opened 2022-07-02
Don't create dynamic classes for non-dynamic managers

ljodal:
We don't need to create custom managers when the manager isn't dynamically created. This changes the logic to just set the type of the manager on the class.

There's a small regression/limitation with this approach, as demonstrated by the `custom_manager_without_typevar_returns_any_model_types` test case. If the manager is _not_ typed with generics (ie. `MyManager(Manager)` vs `MyManager(Manager[T])` the manager will return querysets with `Any` as the model type instead because of implicit generics.

That can be solved in one of two ways: either through "reparametrization" as implemented in #1030 or possibly through setting the methods as attributes with type `Any` as we do for dynamic managers.


ljodal:
I'm gonna close this for now. Realized this was only happening when the manager class was missing generics. Hopefully this can be revisited once #1030 is merged.

PR #1042, opened 2022-07-02
Run CI checks on all pushes

ljodal:
This makes it easier to work in branches in forks, as the workflows will
run before creating a PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,11 +1,6 @@
 name: test
 
-on:
-  push:
-    branches:
-      - master
-  pull_request:
-  workflow_dispatch:
+on: [push, workflow_dispatch]
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -1,11 +1,6 @@
 name: test
 
-on:
-  push:
-    branches:
-      - master
-  pull_request:
-  workflow_dispatch:
+on: [push, workflow_dispatch]
```
sobolevn:
```suggestion
on: [push, pull_request, workflow_dispatch]
```

sobolevn (commented):
Please, take a look that this PR does not have attached jobs.

ljodal:
Thanks! I guess it behaves differently than I'm used to when opening PRs across repos. Seems to be working now

sobolevn (commented):
In this configuration all PRs of people with write access will run twice: as `push` and as `pull_request`.

Can you please check: are there any existing GitHub solutions to fix that?

ljodal:
Ah, that's what I suspected. I'm pretty sure there's no fix for that, so I'll just close this for now. I checked some random open source projects and most seem to be using some variant of the existing config here

PR #1044, opened 2022-07-02
Improve typing for unresolved managers

ljodal:
This changes the logic when encountering an unresolvable manager class. Instead of adding it as a `Manager` we create a subclass of `Manager` that has `fallback_to_any=True` set. Similarly a `QuerySet` class is created that also has fallbacks to `Any`. This allows calling custom methods on the manager and querysets without getting type errors.

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -12,6 +12,7 @@ from mypy.semanal import SemanticAnalyzer
 from mypy.types import AnyType, Instance
 from mypy.types import Type as MypyType
 from mypy.types import TypedDictType, TypeOfAny
+from mypy.typevars import fill_typevars
 
 from mypy_django_plugin.django.context import DjangoContext
 from mypy_django_plugin.errorcodes import MANAGER_MISSING
@@ -19,7 +20,10 @@ from mypy_django_plugin.lib import fullnames, helpers
 from mypy_django_plugin.lib.fullnames import ANNOTATIONS_FULLNAME, ANY_ATTR_ALLOWED_CLASS_FULLNAME, MODEL_CLASS_FULLNAME
 from mypy_django_plugin.transformers import fields
 from mypy_django_plugin.transformers.fields import get_field_descriptor_types
-from mypy_django_plugin.transformers.managers import create_manager_info_from_from_queryset_call
+from mypy_django_plugin.transformers.managers import (
+    MANAGER_METHODS_RETURNING_QUERYSET,
+    create_manager_info_from_from_queryset_call,
+)
 
 
 class ModelClassInitializer:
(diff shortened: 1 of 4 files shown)


mypy_django_plugin/transformers/models.py
```diff
@@ -83,6 +85,81 @@ def get_generated_manager_info(self, manager_fullname: str, base_manager_fullnam
...
+        and methods. This is used for the manager returned by the method above.
+        """
+
+        name = "UnknownQuerySet"
+
+        # Check if we've already created a fallback queryset class for this
+        # module, and if so reuse that.
+        queryset_info = self.lookup_typeinfo(f"{self.model_classdef.info.module_name}.{name}")
+        if queryset_info and queryset_info.metadata.get("django", {}).get("any_fallback_queryset"):
+            return queryset_info
```
flaeppe, line 150: While _highly_ unlikely, not saying you'll have to fix this, there could be name collisions and `add_new_class_for_current_module` guarantees unique names. So `add_new_class_for_current_module` _has a risk_ of inserting `{self.model_classdef.info.module_name}.{name}1`

But I suppose that'd just lead to always creating and not reusing info.

There's also a risk of finding and using the colliding symbol, if it happens to be a `TypeInfo`, but yeah.

---

Same goes for the manager fallback creation

mypy_django_plugin/transformers/models.py
```diff
@@ -83,6 +85,81 @@ def get_generated_manager_info(self, manager_fullname: str, base_manager_fullnam
...
+            helpers.add_new_sym_for_info(manager_info, name=method_name, sym_type=AnyType(TypeOfAny.special_form))
+
+        manager_info.metadata["django"] = django_metadata = {
+            "any_fallback_manager": True,
+            "from_queryset_manager": fallback_queryset.fullname,
+        }
+        if not related_manager:
+            django_metadata["from_queryset_managers"] = {
+                f"django.db.models.manager.{manager_info.fullname}": manager_info.fullname,
+            }
```
ljodal: I _think_ this bit can be removed, as we're not going to try to do a dynamic lookup of this manager?

tests/typecheck/fields/test_related.yml
```diff
@@ -680,7 +680,7 @@
 
                 # Note, that we cannot resolve dynamic calls for custom managers:
                 class Transaction(models.Model):
-                    objects = BaseManager.from_queryset(TransactionQuerySet)
+                    objects = BaseManager.from_queryset(TransactionQuerySet)()
```
flaeppe: IMO this test is missing a `reveal_type(Transaction.objects)`. I don't think `reveal_type(Transaction.objects.custom())` by itself tells enough of the story(?)

mypy_django_plugin/transformers/models.py
flaeppe: This only happens on final iteration, right? And the `UnknownManager` won't ever match a runtime manager either.

I'm thinking the case is like:

Runtime manager type is something like: `manager.from.third.party.ThirdPartyManager`, and we'll insert an `{current_module.fullname}.UnknownManager` in place of it. In general, if the case is that a runtime manager won't ever have the possibility to match what we generate there's no need to store any metadata, I think.

---

Side note, if you decide to keep it; `f"django.db.models.manager.{manager_info.name}"` should be the metadata value

flaeppe (commented):
Could we also add a test case having two different unresolvable managers, in the same module, that'll also reveal their model attribute type?

mypy_django_plugin/transformers/models.py
ljodal: I'm aware of the chance for a collision, but the only effect should be that a new manager is created every time rather than being reused. The metadata should ensure that we don't accidentally use a manager/queryset that's _not_ created here?

flaeppe: Yeah, true. The metadata will ensure we don't use something else, didn't think of that.

mypy_django_plugin/transformers/models.py
ljodal: Yes, it's only on the final iteration, and I'm gonna remove this. Wasn't entirely clear to me what this was used for when I wrote it, but it's clear now that this is not needed :)

ljodal:
@flaeppe isn't that what this test is checking?

https://github.com/typeddjango/django-stubs/blob/2cca99a672efb647205096cee17a6d7c3103853b/tests/typecheck/managers/test_managers.yml#L446-L543

ljodal:
I've found a bug with this implementation, so hold off on merging until I have that under control :)

Edit: Fixed now ✅

ljodal:
@sobolevn Any plans to merge this and the few other PRs that should be ready? We're currently running with this branch at work and it's been working fine (both 1.11 and 1.12 is unusable for us due to phantom type errors :().

I also have a few other things on my list to fix (like #960), but I'll get too many merge conflicts so I'm holding off for now

sobolevn:
Sorry for the delay, I was on vacation. Can you please resolve merge conflicts?

ljodal:
> Sorry for the delay, I was on vacation. Can you please resolve merge conflicts?

No worries, I assumed as much. The conflicts should be resolved now :)

sobolevn (approved):
Thank you!

PR #1045, opened 2022-07-03
Add support for inline `from_queryset` in model classes

ljodal:
This adds support for calling `<Manager>.from_queryset(<QuerySet>)()` inline in models, for example like this:

    class MyModel(models.Model):
        objects = MyManager.from_queryset(MyQuerySet)()

This is done by inspecting the class body in the `transform_class_hook`

There's a few broken tests I haven't figured out yet, but this mostly appears to be working just fine.

## Related issues

 - This overlaps a bit with #1025


sobolevn (approved):
Looks great! Awesome work!

PR #1050, opened 2022-07-05
Monkeypatch some more generics.

ngnpope:
Adds monkey patching of the following generics:

- `django.core.files.utils.FileProxyMixin`: Fixes `django.core.files.base.ContentFile` and `django.core.files.File`.
- `django.db.models.fields.related.ForeignKey`: `__class_getitem__` is being added for Django 4.1


sobolevn (approved):
LGTM! Thanks!

PR #1062, opened 2022-07-18
Use a more general type for django.core.mail.send_mass_mail's datatuple

jhhayashi:
The implementation works with all `Iterable`s, the current typing in this library is a `List`, and the docs recommend using a datatuple.

We can expand the datatuple's type to include all valid args

https://github.com/django/django/blob/main/django/core/mail/__init__.py#L112
https://docs.djangoproject.com/en/4.0/topics/email/#send-mass-mail

### DIFF
--- a/django-stubs/core/mail/__init__.pyi
+++ b/django-stubs/core/mail/__init__.pyi
@@ -23,7 +23,7 @@ def send_mail(
     html_message: Optional[str] = ...,
 ) -> int: ...
 def send_mass_mail(
-    datatuple: List[Tuple[str, str, str, List[str]]],
+    datatuple: Tuple[Tuple[str, str, str, List[str]], ...],
     fail_silently: bool = ...,
     auth_user: Optional[str] = ...,
     auth_password: Optional[str] = ...,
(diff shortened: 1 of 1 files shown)


django-stubs/core/mail/__init__.pyi
```diff
@@ -23,7 +23,7 @@ def send_mail(
     html_message: Optional[str] = ...,
 ) -> int: ...
 def send_mass_mail(
-    datatuple: List[Tuple[str, str, str, List[str]]],
+    datatuple: Tuple[Tuple[str, str, str, List[str]], ...],
```
sobolevn: What about `Sequence`?

sobolevn: Or `Iterable`?

jhhayashi: I figured we should use what's in the docs rather than what currently works with the implementation. Happy to make this more general if we're okay with relying on implementation details

sobolevn: Yes, we should be more "realistic" if we can.

jhhayashi: Makes sense, thanks

sobolevn (approved):
Thanks!

jhhayashi:
Thanks for the quick review!

PR #1069, opened 2022-07-22
Improve type annotation of DiscoverRunner.

PIG208:
# I have made things!

The return type of DiscoverRunner.get_resultclass should
be Optional[Type[TextTestResult]] given that it may also return PDBDebugResult:
https://github.com/django/django/blob/0dd29209091280ccf34e07c9468746c396b7778e/django/test/runner.py#L958-L962

`DicoverRunner.run_tests` accepts `None` for `extra_tests` because
that's already the default value for it:
https://github.com/django/django/blob/0dd29209091280ccf34e07c9468746c396b7778e/django/test/runner.py#L1030


sobolevn (approved):
Thanks!

PR #1071, opened 2022-07-22
Improve stubs for contrib.postgres.operations.

PIG208:
This adds a bunch of missing operations in contrib.postgres.operations.

Documentation: https://docs.djangoproject.com/en/4.0/ref/contrib/postgres/operations/#managing-collations-using-migrations
Source code: https://github.com/django/django/blob/2fac0a18081dcc77fc860c801e5d727dc90435b3/django/contrib/postgres/operations.py


django-stubs/contrib/postgres/operations.pyi
```diff
@@ -1,9 +1,29 @@
...
+    def database_forwards(
+        self, app_label: str, schema_editor: BaseDatabaseSchemaEditor, from_state: ProjectState, to_state: ProjectState
+    ) -> None: ...
+    def database_backwards(
+        self, app_label: str, schema_editor: BaseDatabaseSchemaEditor, from_state: ProjectState, to_state: ProjectState
+    ) -> None: ...
+    def extension_exists(self, schema_editor: BaseDatabaseSchemaEditor, extension: str) -> bool: ...
+    def describe(self) -> str: ...
+    @property
+    def migration_name_fragment(self) -> str: ...
```
sobolevn: Looks like this is defined in `Operation`: https://github.com/django/django/blob/67d0c4644acfd7707be4a31e8976f865509b09ac/django/db/migrations/operations/base.py#L82-L88

Can you please make sure that all parent methods stubs are not redefined?

PIG208: OK. Updated.

sobolevn (approved):
Thanks!

PR #1072, opened 2022-07-22
Fix paramters for ParallelTestSuite and RemoteTestRunner.

PIG208:
A new parameter `buffer` was added. `suites` was renamed to `subsuites`.

ParallelTestSuite:
https://github.com/django/django/blob/2fac0a18081dcc77fc860c801e5d727dc90435b3/django/test/runner.py#L468-L475

RemoteTestRunner:
https://github.com/django/django/blob/2fac0a18081dcc77fc860c801e5d727dc90435b3/django/test/runner.py#L350

`buffer` was added in https://github.com/django/django/commit/f6713cda89d323565cfdc867c977c3363edcfc3c,
`suites` was renamed in https://github.com/django/django/commit/cb6c19749d342c3dc0f97d89ff6887b220cf45b8.
Both of these changes were made in 4.1rc1.

Signed-off-by: Zixuan James Li <p359101898@gmail.com>

# I have made things!


django-stubs/test/runner.pyi
```diff
@@ -73,7 +73,12 @@ class ParallelTestSuite(TestSuite):
     subsuites: Any = ...
     processes: Any = ...
     failfast: Any = ...
```
sobolevn: Can you please also adjust these?

PIG208: Done.

adamchainz (approved):
LGTM

PR #1073, opened 2022-07-22
Fix type annotation of message_dict.

PIG208:
# I have made things!

When message_dict is accessed, it either raises an `AttributeError` or
returns a `dict`. There is no way for it to be `Optional`.

https://github.com/django/django/blob/caad462feaa84ba78ed658a9595a4a4363dad2db/django/core/exceptions.py#L170-L176


sobolevn (approved):
Thanks!

PR #1074, opened 2022-07-23
Tighten `management.get_commands` type

andersk:
It hasn’t actually been possible for `get_commands` to return `BaseCommand` items since Django 1.4. See https://github.com/django/django/pull/15876.


sobolevn (approved):
Thanks!

PR #1075, opened 2022-07-26
Improve type for @require_http_methods

adamchainz:
A list isn't strictly required - tuples and sets work too, and anything that supports `__contains__`, so use the `Container` ABC.

## Related issues

no issue


sobolevn (approved):
Thanks!

PR #1077, opened 2022-07-27
Add _nodb_cursor.

PIG208:
# I have made things!

In cases when we want to monkey patch internal functions like `destroy_test_db`, we need to have access to the cursor.

https://github.com/django/django/blob/9dff316be41847c505ebf397e4a52a0a71e0acc4/django/db/backends/base/creation.py#L26


sobolevn (commented):
Sorry, we don't include implementation details.
Use `# type[ignore]` for this.

PR #1078, opened 2022-07-28
Drop the restriction of _ValidatorCallable's return type.

PIG208:
# I have made things!

The type of the return value of a validator is neither specified nor
required in the source code and documentation.

See also: https://docs.djangoproject.com/en/4.0/ref/validators/

### DIFF
--- a/django-stubs/core/validators.pyi
+++ b/django-stubs/core/validators.pyi
@@ -7,7 +7,7 @@ from django.core.files.base import File
 EMPTY_VALUES: Any
 
 _Regex = Union[str, Pattern[str]]
-_ValidatorCallable = Callable[[Any], None]
+_ValidatorCallable = Callable[[Any], object]
 
 class RegexValidator:
     regex: _Regex = ...  # Pattern[str] on instance, but may be str on class definition


django-stubs/core/validators.pyi
```diff
@@ -7,7 +7,7 @@ from django.core.files.base import File
 EMPTY_VALUES: Any
 
 _Regex = Union[str, Pattern[str]]
-_ValidatorCallable = Callable[[Any], None]
+_ValidatorCallable = Callable[[Any], object]
```
intgr, line 10: I don't think we should relax the types so much. The fact that some type technically works and doesn't crash Django doesn't mean it's the correct type, or that it will continue working with new Django versions.

For example, this makes it valid to pass a validator function that returns boolean True/False. Users might not realize that their validator function doesn't actually do anything, because Django excepts an exception to be raised instead.

sobolevn: I agree, `None` is a part of the contract here. You either raise or not.
This can produce new errors for developers, not reduce them.

PIG208: I see, thanks!

PR #1079, opened 2022-07-28
Support Lookup generics.

PIG208:
# I have made things!

Custom `Lookup` implementation will run into
'Missing type parameters for generic type "Lookup"' without having
`Lookup` monkey-patched with django-stubs-ext.


sobolevn (approved):
Thanks! Are interested in automating this check (that types marked as `Generic` either support `__class_getitem__` or are monkey-patched)?
Because we have this problem all the time.

PIG208:
Sounds interesting. Is there a mechanism for us to find all the types that are marked as `Generic`? Starting from there, the check shouldn't be too complicated to implement.

sobolevn:
Yes, we can use `ast` module for that. For example, some different check: https://github.com/wemake-services/wemake-python-styleguide/blob/master/scripts/check_generic_visit.py

PR #1080, opened 2022-07-29
Fix OutputWrapper.style_func type

andersk:
`style_func` has a setter that converts `None` to `lambda x: x`, so the getter cannot return `None`.

https://github.com/django/django/blob/4.0.6/django/core/management/base.py#L141-L150


sobolevn (approved):
Thanks!

PR #1090, opened 2022-08-05
Improve type annotation for RunSQL

PIG208:
# I have made things!

The `sql`s to be executed do not necessarily need to be a homogeneous list
or tuple containing only lists or tuples or `str`s. It can be a mix of
everything acceptable for the cursor. `RunSQL` iterates through the list or tuple
and executes them separately.

Also, the 2-item tuple for `sql` can have a `dict` as the second item
for parameters. This behavior is the same as using
`cursor.execute` for backends except for SQLite.

A test case is added along with this change.

Relevant implementation
https://github.com/django/django/blob/5f760025004bdb02f9844011033459c30347f215/django/db/migrations/operations/special.py#L119-L133


django-stubs/db/migrations/operations/special.pyi
```diff
@@ -21,15 +21,17 @@ class SeparateDatabaseAndState(Operation):
 
 class RunSQL(Operation):
     noop: Literal[""] = ...
-    sql: Union[str, _ListOrTuple[str], _ListOrTuple[Tuple[str, Optional[_ListOrTuple[str]]]]] = ...
-    reverse_sql: Optional[Union[str, _ListOrTuple[str], _ListOrTuple[Tuple[str, Optional[_ListOrTuple[str]]]]]] = ...
+    sql: Union[str, _ListOrTuple[Union[str, Tuple[str, Union[Dict[str, Any], Optional[_ListOrTuple[str]]]]]]] = ...
```
sobolevn, line 24: Oh my...

sobolevn (approved):
Thank you!

django-stubs/db/migrations/operations/special.pyi
PIG208: Yes... I know

intgr:
This causes false positive when used with a `list[str]` variable:

```python
sqls: list[str] = ["SOME SQLS", "SOME SQLS"]
RunSQL(sql=sqls, reverse_sql=[])
```

> error: Argument "sql" to "RunSQL" has incompatible type "List[str]"; expected "Union[str, Union[List[Union[str, Tuple[str, Union[Dict[str, Any], List[str], Tuple[str, ...], Tuple[], None]]]], Tuple[Union[str, Tuple[str, Union[Dict[str, Any], List[str], Tuple[str, ...], Tuple[], None]]], ...], Tuple[]]]"  [arg-type]

I don't think there's a simple fix though. Ordinariy we would change `List[]` to `Sequence[]`, but the `RunSQL` code is actually very specific that it only accepts lists and tuples.

sobolevn:
Wild guess: maybe moving `list[str]` to the left? Like `Union[str, list[str], ...]`?
Just a guess though.

PR #1103, opened 2022-08-14
Add type and parameter fields to MediaType

yrd:
# I have made things!

This commits adds three fields that were missed in django.http.request.MediaType.

See here for how they are initialized:
https://github.com/django/django/blob/c8eb9a7c451f7935a9eaafbb195acf2aa9fa867d/django/http/request.py#L621-L624

## Related issues

None.


adamchainz (approved):
LGTM

Could you also fix up the hint for `parse_header` to mark it as returning `bytes` in the dict?

yrd:
Done 👍🏻

adamchainz (approved):

adamchainz:
Merged now that I have been added as a maintainer 😊

PR #1104, opened 2022-08-14
Improve return type for BaseCommand.execute()

adamchainz:
# I have made things!

It returns the value returned by `handle()`: https://github.com/django/django/blob/cb791a2540c289390b68a3ea9c6a79476890bab2/django/core/management/base.py#L421-L458

## Related issues

n/a


sobolevn (approved):
🏁

PR #1106, opened 2022-08-14
Improve types for DiscoverRunner

adamchainz:
# I have made things!

* `PDBDebugResult` - missing result class related to the `--pdb` flag. I noticed that `DebugSQLTextTestResult` repeats hints for things inherited from `TextTestResult` - why is that? I didn't do that for `PDBDebugResult`...
* `test_suite`, `parallel_test_suite`, `test_runner`, `test_loader`, `reorder_by` - add real types rather than `Any`
* `pdb`, `test_name_patterns`, `buffer` - args added in #909 but attribute missed
* `time_keeper` - attribute that relates to the `timing` flag
* `logger`, `shuffle`, `shuffle_seed`, `log`, `setup_shuffler` - args, attributes, and methods from Django 4.0: https://docs.djangoproject.com/en/4.0/releases/4.0/#tests
* `load_with_patterns`, `load_tests_for_label`, `_get_databases` - some missing methods

## Related issues

n/a


sobolevn (approved):

django-stubs/test/runner.pyi
```diff
@@ -1,12 +1,15 @@
 import logging
 from argparse import ArgumentParser
+from contextlib import contextmanager
 from io import StringIO
 from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, Type
```
sobolevn:
```suggestion
from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, Type, Union
```

adamchainz:
From now on, I will run at least some tests locally before making a PR 😅😅😅

PR #1107, opened 2022-08-14
Improve types for CheckMessage subclasses

adamchainz:
# I have made things!

These classes pass all but the first argument through to `CheckMessage`: https://github.com/django/django/blob/e30d6678421b7573a1995f61521f14348c9b2a17/django/core/checks/messages.py#L59-L81

For type checking purposes it's best to redeclare all the arguments so they are type checked.

## Related issues

n/a


sobolevn:
Hi! Thanks a lot for your PRs.
I've seen them, but we need to solve https://github.com/python/mypy/issues/13422 first 😞

adamchainz:
Aye... I had a crack at running the tests locally to try fix that but it felt rather unclear to me :( Might need a git bisect on mypy?

sobolevn:
I am pretty sure it is https://github.com/python/mypy/pull/11143

django-stubs/core/checks/messages.pyi
```diff
@@ -19,16 +19,16 @@ class CheckMessage:
     def is_silenced(self) -> bool: ...
 
 class Debug(CheckMessage):
-    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
+    def __init__(self, msg: str, hint: Optional[str] = ..., obj: Any = ..., id: Optional[str] = ...) -> None: ...
```
sobolevn, line 22: `level` is covered by the first call argument 👍 
https://github.com/django/django/blob/e30d6678421b7573a1995f61521f14348c9b2a17/django/core/checks/messages.py#L10

sobolevn (approved):

PR #1108, opened 2022-08-15
Fix CI

sobolevn:
Failing logs: https://github.com/typeddjango/django-stubs/runs/7834462015?check_suite_focus=true


sobolevn:
I literally have 0 ideas what is going on.
It all works on my local machine.

sobolevn:
Ok, I figured out that for some reason our CI does not see our stubs. So, everything is just `Any`.
But, why my local machine works?

At this pace I will be a cheff instead 🤦

adamchainz:
Personally my backup is goat farmer, it'd be quieter.

sobolevn:
I like the idea of a sandwich truck, something like this:
truck

Goats are great as well 😆

sobolevn:
Right now there are three things we can do:
1. Pin `mypy<0.970` in `pre-commit`
2. Wait for `0.972`, I am not sure that we can realease it any time soon
3. Find some wild way of solving `MYPYPATH` problems (even author of this feature has no idea what to do)

adamchainz:
I would vote on pinning mypy<0.970 in CI, as a temporary fix, to allow further PR's to be merged. I'm using django-stubs with 0.971 myself and not having issues - afaiu the problem is limited to the CI setup.

sobolevn:
🎉

PR #1110, opened 2022-08-16
Type `samesite` values as literals 'Lax', 'Strict' or 'None'

flaeppe:
- Adjusts supported `SESSION_COOKIE_SAMESITE` values (https://docs.djangoproject.com/en/3.2/ref/settings/#session-cookie-samesite)
- Adjusts supported `CSRF_COOKIE_SAMESITE` values (https://docs.djangoproject.com/en/3.2/ref/settings/#csrf-cookie-samesite)


django-stubs/conf/global_settings.pyi
```diff
@@ -5,7 +5,7 @@ by the DJANGO_SETTINGS_MODULE environment variable.
 
 # This is defined here as a do-nothing function because we can't import
 # django.utils.translation -- that module depends on the settings.
-from typing import Any, Dict, List, Optional, Pattern, Protocol, Sequence, Tuple, Union
+from typing import Any, Dict, List, Literal, Optional, Pattern, Protocol, Sequence, Tuple, Union
```
```suggestion
from typing import Any, Dict, List, Optional, Pattern, Protocol, Sequence, Tuple, Union

from typing_extensions import Literal
```

sobolevn (approved):
Thanks!

PR #1116, opened 2022-08-19
Updating deconstructible

karolyi:
@see https://github.com/microsoft/pyright/issues/3840

### DIFF
--- a/django-stubs/utils/deconstruct.pyi
+++ b/django-stubs/utils/deconstruct.pyi
@@ -1,3 +1,14 @@
-from typing import Any, Optional
+from typing import Any, Callable, Optional, TypeVar, overload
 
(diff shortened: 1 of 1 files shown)


django-stubs/utils/deconstruct.pyi
```diff
@@ -1,3 +1,14 @@
...
+_TCallable = TypeVar('_TCallable', bound=Callable[..., Any])
+
+
+@overload
+def deconstructible(_type: type[_T]) -> type[_T]: ...
+
+
+@overload
+def deconstructible(
+    *, path: Optional[Any] = ...
```
intgr: Can path be `Optional[str]`?

karolyi: Yep, I'll change it, hold on.

sobolevn (approved):
Thank you!

PR #1117, opened 2022-08-22
Improve hints for BaseCache

adamchainz:
# I have made things!

Add hints for some things in the source but not django-stubs: https://github.com/django/django/blob/main/django/core/cache/backends/base.py

1. Add missing `InvalidCacheKey` exception class.
2. Add some missing private attributes of BaseCache, that I found I needed for the type hints of [my cache backend for MySQL](https://django-mysql.readthedocs.io/en/latest/cache.html).
3. Add missing `validate_key` and `make_and_validate_key` methods. Rearrange `validate_key` to match source order in Django's code.
4. Add the async cache backend methods, added in Django 4.0: https://docs.djangoproject.com/en/4.0/releases/4.0/#cache . These are just copy/paste with `def` replaced with `async def a`.


sobolevn (approved):
Thank you for your contributions and patience!

adamchainz:
Thank you for maintaining this project for so long! It's taken me a while to get around to actually using it properly...

PR #1118, opened 2022-08-22
Fix CI, part 2

sobolevn:
Refs https://github.com/typeddjango/django-stubs/pull/1108

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          pip install -U pip setuptools wheel
+          pip install -U pip 'setuptools<64' wheel
           pip install -r ./requirements.txt
 
       - name: Run pre-commit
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -20,7 +20,7 @@ def find_stub_files(name: str) -> List[str]:
     readme = f.read()
 
 dependencies = [
-    "mypy>=0.930",
+    "mypy>=0.930,<0.970",  # TODO: remove after CI is fixed
```
andersk, line 23: This shouldn’t be necessary for the sake of CI. CI does `pip install -r ./requirements.txt`, and `requirements.txt` already includes `mypy==0.961` (and also `-e .[compatible-mypy]` which constrains `mypy` to `<0.970`).

sobolevn: Any ideas how to fix it?

.github/workflows/test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          pip install -U pip setuptools wheel
+          pip install -U pip 'setuptools<64' wheel
```
andersk, line 24: I think the two other copies of this need to be patched too?

andersk:
I found two ways to get it working again: #1122 or #1123.

PR #1123, opened 2022-08-26
Prevent building an editable wheel with setuptools<64

### LINKED ISSUE #1112: disregard/delete

andersk:
This seems to be needed for passing CI now.

This is an alternative to #1122; we should only merge one or the other.

Closes #1108.
Closes #1118.
Closes #1122.


sobolevn:
Thank you!

PR #1125, opened 2022-08-26
Add database Field.non_db_attrs

adamchainz:
# I have made things!

Added to Django 4.1 in https://github.com/django/django/pull/15561

It's public, mentioned in release notes with linked docs: https://docs.djangoproject.com/en/4.1/releases/4.1/#models

## Related issues

n/a


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -155,6 +155,7 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
     hidden: bool
     system_check_removed_details: Optional[Any]
     system_check_deprecated_details: Optional[Any]
+    non_db_attrs: Tuple[str]
```
```suggestion
    non_db_attrs: Tuple[str, ...]
```

?

adamchainz: right, my bad!

PR #1127, opened 2022-08-26
Add HttpRequest current_app and LANGUAGE_CODE attrs

adamchainz:
# I have made things!

Found two more optional attrs, like `user`, that Django may add to `HttpRequest`. I figured the same pragmatic approach of declaring them as always present could be taken. Also I rearranged and documented these attributes with references to where they come from.

(1) Views in `django.contrib.admin` set `request.current_app`, and the templates use it. The docs also encourage setting the attribute when writing custom views: https://docs.djangoproject.com/en/4.1/ref/contrib/admin/#adding-views-to-admin-sites

(2) `LocaleMiddleware` sets `request.LANGUAGE_CODE`, see right at the bottom of: https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#how-django-discovers-language-preference

## Related issues

n/a

### DIFF
--- a/django-stubs/http/request.pyi
+++ b/django-stubs/http/request.pyi
@@ -57,6 +57,8 @@ class HttpRequest(BytesIO):
     site: Site
     session: SessionBase
     _stream: BinaryIO
+    # django.contrib.admin uses this attribute:
+    current_app: str = ...
     # The magic. If we instantiate HttpRequest directly somewhere, it has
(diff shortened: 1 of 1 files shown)


sobolevn (approved):

adamchainz:
Found another optional attr to add, upadting PR...

sobolevn (approved):

PR #1128, opened 2022-08-26
Add db Field.flatchoices

adamchainz:
# I have made things!

Add this undocumented property that has been there since Django 1.0: https://github.com/django/django/commit/649463dd348abd6d0cab890e2372e88fc452128e

I found I had used it in a `Field` subclass.

## Related issues

n/a


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -219,6 +219,8 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
         limit_choices_to: Optional[_LimitChoicesTo] = ...,
         ordering: Sequence[str] = ...,
     ) -> _ChoicesList: ...
+    def _get_flatchoices(self) -> List[_Choice]: ...
+    flatchoices = property(_get_flatchoices)
```
sobolevn: I am not sure that this is supported by mypy: `Revealed type is "Any"`

Try using `@property` as a decorator.

Repro:

```python
class A:
    def _get_flatchoices(self) -> int: ...
    flatchoices = property(_get_flatchoices)

reveal_type(A().flatchoices)
```

adamchainz: Ah, well spotted!

sobolevn (approved):

PR #1130, opened 2022-08-27
Enhance stubs for csrf decorators.

PIG208:
`csrf_protect` and etc. are created via the `decorator_from_middleware`
helper, which doesn't modify the original signature of the wrapped view
function. Using TypeVar helps preserve the original type of the
decorated callable.

Related implementation in Django: https://github.com/django/django/blob/b92ffebb0cdc469baaf1b8f0e72dddb069eb2fb4/django/utils/decorators.py#L117-L162

## Related issues

None


sobolevn (approved):
Thanks!

PR #1131, opened 2022-08-28
Add async queryset for Django 4.1

xncbf:
# I have made things!

I added all of asynchronous queries created in django 4.1, including `contains` created in django 4.0.

## Related issues

n/a


django-stubs/db/models/manager.pyi
```diff
@@ -50,26 +50,51 @@ class BaseManager(Generic[_T]):
...
     def delete(self) -> Tuple[int, Dict[str, int]]: ...
+    async def adelete(self) -> Tuple[int, Dict[str, int]]: ...
     def update(self, **kwargs: Any) -> int: ...
+    async def aupdate(self, **kwargs: Any) -> int: ...
     def exists(self) -> bool: ...
+    async def aexists(self) -> bool: ...
     def explain(self, *, format: Optional[Any] = ..., **options: Any) -> str: ...
+    async def aexplain(self, *, format: Optional[Any] = ..., **options: Any) -> str: ...
+    def contains(self, objs: Iterable[_T]) -> bool: ...
+    async def acontains(self, objs: Iterable[_T]) -> bool: ...
```
adamchainz, line 97: this type is wrong - it doesn't take an iterable of type `_T` but a single instance of `_T`

https://docs.djangoproject.com/en/4.1/ref/models/querysets/#django.db.models.query.QuerySet.contains


```suggestion
    def contains(self, obj: _T) -> bool: ...
    async def acontains(self, obj: _T) -> bool: ...
```

django-stubs/db/models/query.pyi
```diff
@@ -57,25 +57,49 @@ class _QuerySet(Generic[_T, _Row], Collection[_Row], Reversible[_Row], Sized):
...
     def get(self, *args: Any, **kwargs: Any) -> _Row: ...
+    async def aget(self, *args: Any, **kwargs: Any) -> _Row: ...
     def create(self, **kwargs: Any) -> _T: ...
+    async def acreate(self, **kwargs: Any) -> _T: ...
     def bulk_create(
         self, objs: Iterable[_T], batch_size: Optional[int] = ..., ignore_conflicts: bool = ...
     ) -> List[_T]: ...
+    async def abulk_create(
+        self, objs: Iterable[_T], batch_size: Optional[int] = ..., ignore_conflicts: bool = ...
+    ) -> List[_T]: ...
```
adamchainz, line 70: Whilst modifying `bulk_create` let's also import the new Django 4.1 parameters.

https://docs.djangoproject.com/en/4.1/ref/models/querysets/#django.db.models.query.QuerySet.bulk_create

> The update_conflicts, update_fields, and unique_fields parameters were added to support updating fields when a row insertion fails on conflict.

```suggestion
    def bulk_create(
        self,
        objs: Iterable[_T],
        batch_size: Optional[int] = ...,
        ignore_conflicts: bool = ...,
        update_conflicts: bool = ...,
        update_fields: Optional[Collection[str]] = ...,
        unique_fields: Optional[Collection[str]] = ...,
    ) -> List[_T]: ...
    async def abulk_create(
        self,
        objs: Iterable[_T],
        batch_size: Optional[int] = ...,
        ignore_conflicts: bool = ...,
        update_conflicts: bool = ...,
        update_fields: Optional[Collection[str]] = ...,
        unique_fields: Optional[Collection[str]] = ...,
    ) -> List[_T]: ...
```

sobolevn (commented):
This is ok as the first step, but please: take in mind that we also have special handling of `.filter` and other query set  methods in out plugin.

This would need to be addressed as well.

adamchainz (commented):
damn I was just reviewing @sobolevn 😅 making a PR with these changes

django-stubs/db/models/manager.pyi
sobolevn: Thanks for catching this! I never used async django, so my knowledge is limited: I was just checking that signatures are the same as regular methods.

PR #1134, opened 2022-08-28
Improve types for QuerySet.contains and bulk_create

adamchainz:
# I have made things!

Copying in review from #1131.

## Related issues

n/a


django-stubs/db/models/query.pyi
```diff
@@ -98,8 +110,8 @@ class _QuerySet(Generic[_T, _Row], Collection[_Row], Reversible[_Row], Sized):
     async def aexists(self) -> bool: ...
     def explain(self, *, format: Optional[Any] = ..., **options: Any) -> str: ...
     async def aexplain(self, *, format: Optional[Any] = ..., **options: Any) -> str: ...
-    def contains(self, objs: Iterable[_T]) -> bool: ...
-    async def acontains(self, objs: Iterable[_T]) -> bool: ...
+    def contains(self, obj: models.Model) -> bool: ...
+    async def acontains(self, obj: models.Model) -> bool: ...
```
adamchainz, line 114: using `_T` here gave me:

```
django-stubs/db/models/query.pyi:113: error: Cannot use a covariant type variable as a parameter
django-stubs/db/models/query.pyi:114: error: Cannot use a covariant type variable as a parameter
```

But looking at the source, it seems any instance of a model class is acceptable:

https://github.com/django/django/blob/b92ffebb0cdc469baaf1b8f0e72dddb069eb2fb4/django/db/models/query.py#L1241-L1255

hence change to `models.Model`

sobolevn (approved):
Wow, `update_conflicts=True` looks like an amazing feature!

adamchainz:
Yeah!

I suppose for strict accuracy, we could use `@overload` to enforce that `ignore_conflicts` and `update_conflicts` are not passed at the same time, and that the fields are passed if `update_conflicts` is `True`, etc. But I cba to enumerate all the cases right now 😅

PR #1135, opened 2022-08-28
Fix Manager.contains and bulk_create

adamchainz:
# I have made things!

Realized that #1134 only changed the methods on `QuerySet`, not `Manager`.

Also normalized import of `Model` in `db/models/query.pyi` so that hints for methods can be copied between the two classes when required.

## Related issues

n/a


sobolevn (approved):

sobolevn:
**Offtop**

Hey, @adamchainz! A friend of mine shared this link with me today: https://github.com/trending/developers

Looks like you are a trending developer of the day! Maybe it will make your day better!
Thanks for all the hard work :)

adamchainz:
Ha thanks. You're trending at number 4! 👏

I was trending before in May too, no idea how it's computed 🤷‍♂️

Keep up the good work yourself!

PR #1143, opened 2022-09-05
Add django.db.models.functions.Collate

syre:
Collate has been part of django.db.models.functions since 3.2 afaik:

https://github.com/django/django/blob/stable/3.2.x/django/db/models/functions/comparison.py#L74
https://docs.djangoproject.com/en/3.2/ref/models/database-functions/#django.db.models.functions.Collate


django-stubs/db/models/functions/comparison.pyi
```diff
@@ -8,6 +8,7 @@ class Cast(Func):
     def __init__(self, expression: Any, output_field: Union[str, Field]) -> None: ...
 
 class Coalesce(Func): ...
+class Collate(Func): ...
```
adamchainz: can you type out its `__init__` please? it takes an expression, and a collation that's a string, it seems

syre: sure, i have updated the MR

adamchainz:
Thank you!

PR #1145, opened 2022-09-05
Add `Parser.__loaded_blocks` type hint

UnknownPlatypus:
# I have made things!
While working on a custom `BlockNode` implementation, I noticed a missing attribute to the `Parser` class named `__loaded_blocks`. This attribute is not initially on the class but attached to it by the [`do_block`](https://github.com/django/django/blob/main/django/template/loader_tags.py#L212) function. 

It is used to [track duplicate template block names](https://github.com/django/django/blob/main/django/template/loader_tags.py#L222-L231). This MR is simply adding hints for this attribute.

Thanks again for the awesome work on this project !
Cheers


sobolevn (commented):
I am not sure that this is a reasonable exception from our "almost no protected, no private names" rule.

Do you have anything to persuade me?

UnknownPlatypus:
Oh sorry for that, i was unaware of the "almost no protected, no private names" rule.

I was thinking this change was quite harmless and convenient for people working on custom `BlockNode` implementations. You have to opt in to this `__loaded_blocks` attribute else your custom `BlockNode` class might have trouble working correctly together with regular block nodes. 

But I understand custom block node might be an edge case you don't want to support. And I can probably avoid working with this private attribute.

Up to you.

sobolevn:
The problem with private attributes is that no one promises that it is going to be there in the next patch / security / minor release.

It is quite hard for us to track these. So, I am going to close your PR.
However, thanks a lot for the proposal! Please, feel free to send other improvements :)

UnknownPlatypus:
No worries. Thanks for the input !

PR #1146, opened 2022-09-08
Correct EmailMultiAlternatives.body

adamchainz:
# I have made things!

Like its superclass `EmailMessage`, `EmailMultiAlternatives` also accepts `None` for `body`, which is turned into `""`.

[source](https://github.com/django/django/blob/32536b1324e98768dd892980408a8c6b26c23fd9/django/core/mail/message.py#L240)

## Related issues

n/a


sobolevn (approved):
Thanks!

Issue #1085, opened 2022-08-04
"_MonkeyPatchedWSGIResponse" has no attribute "redirect_chain"


monosans:
# Bug report

## What's wrong

```python
from django.test import Client

c = Client()
response = c.get("/admin/", follow=True)
response.redirect_chain  # "_MonkeyPatchedWSGIResponse" has no attribute "redirect_chain"
```

## How is that should be

If the `follow=True` argument was passed, the response object must have the `redirect_chain` attribute, otherwise it must not.

Quote from [docs](https://docs.djangoproject.com/en/4.1/topics/testing/tools/#making-requests): `If you set follow to True the client will follow any redirects and a redirect_chain attribute will be set in the response object containing tuples of the intermediate urls and status codes.`

The type of this attribute is `List[Tuple[str, int]]`.

## System information

- OS: Arch Linux
- `python` version: 3.10.5
- `django` version: 4.1
- `mypy` version: 0.961 and 0.971
- `django-stubs` version: 1.12.0
- `django-stubs-ext` version: 0.5.0

adamchainz:
Fixed in #1124. Please use a Git install to pull in this type ([example](https://github.com/typeddjango/django-stubs/issues/1095#issuecomment-1241967327)) to resolve this problem until the next release.

PR #1148, opened 2022-09-09
Remove bad redirect_chain attributes

adamchainz:
# I have made things!

`redirect_chain` is monkey-patched on by the test client. I added it in #1124 . I just spotted there are bad `redirect_chain` attributes declared on the base `HttpResponse` and `TemplateResponse` classes, added in fa718b8e55e4ac74b309b0068eeced6f89f48728. These are not generally there, only the monkey-patching adds them, so we should remove these declarations.

## Related issues

Looking at #1085 made me check the history.

### DIFF
--- a/django-stubs/http/response.pyi
+++ b/django-stubs/http/response.pyi
@@ -93,7 +93,6 @@ class HttpResponseBase:
 class HttpResponse(HttpResponseBase, Iterable[bytes]):
     content = _PropertyDescriptor[object, bytes]()
     csrf_cookie_set: bool
-    redirect_chain: List[Tuple[str, int]]
     sameorigin: bool
     test_server_port: str
     test_was_secure_request: bool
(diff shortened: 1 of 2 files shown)


django-stubs/http/response.pyi
```diff
@@ -93,7 +93,6 @@ class HttpResponseBase:
 class HttpResponse(HttpResponseBase, Iterable[bytes]):
     content = _PropertyDescriptor[object, bytes]()
     csrf_cookie_set: bool
-    redirect_chain: List[Tuple[str, int]]
```
intgr, line 96: Maybe this should be added to `_MonkeyPatchedWSGIResponse`, `_MonkeyPatchedASGIResponse`?

adamchainz: it is, since #1124, as I linked in description :)

intgr: Ah thanks, sorry I missed that

sobolevn (approved):
Thank you!

PR #1150, opened 2022-09-13
move _default_manager to ModelBase

clouds56:
# I have made things!

Move _default_manager from Model (as a `@classproperty`) to ModelBase (as a `@property`)
To match what exactly happens in django
https://github.com/django/django/blob/e14d08cd894e9d91cb5d9f44ba7532c1a223f458/django/db/models/base.py#L429
(and also makes pyright/pylance happy)

## Related issues

see also #817.


django-stubs/db/models/base.pyi
```diff
@@ -16,19 +15,17 @@ class ModelState:
     adding: bool = ...
     fields_cache: ModelStateFieldsCacheDescriptor = ...
 
-class ModelBase(type): ...
+class ModelBase(type):
+    @property
+    def _default_manager(cls: Type[_Self]) -> BaseManager[_Self]: ...
```
sobolevn: I think that mypy might not like `cls` to be bound to `Type[Model]` 🤔

clouds56: Yes, mypy's right, updated.

clouds56: I have no idea how to solve this, maybe a `type: ignore` should be introduced.
Indeed, there's no guarantee that `ModelBase` would receive `cls` that bounded to `Model`.

sobolevn: I think we can fallback to the first solution + `type: ignore`

django-stubs/db/models/base.pyi
```diff
@@ -18,16 +17,17 @@ class ModelState:
 
 class ModelBase(type):
     @property
-    def _default_manager(cls: Type[_Model]) -> BaseManager[_Model]: ...
+    def objects(cls: Type[_Self]) -> BaseManager[_Self]: ... # type: ignore
```
sobolevn: Please, add a code for these ignores. `--show-error-codes`.

sobolevn (approved):
Thanks a lot!

Tests:
- https://github.com/typeddjango/django-stubs/blob/3d8d9004874bc09308cce0f1e0fd5507782c9bb8/tests/typecheck/managers/test_managers.yml#L37
- https://github.com/typeddjango/django-stubs/blob/3d8d9004874bc09308cce0f1e0fd5507782c9bb8/tests/typecheck/managers/test_managers.yml#L62

sobolevn:
Please, make sure to fix `django-stubs/db/models/base.pyi` with `black`

clouds56:
@sobolevn I believe all checks have passed, what's blocking?

PR #1152, opened 2022-09-15
Serialize `RelatedManager` attributes in cache

### LINKED ISSUE #1151: Any `RelatedManager`s does not exist on cached runs for model instances
## What's wrong

Accessing `RelatedManager`s through any `related_name=` values on model instances yields error seen below __on cached runs__

```
error: "Purchase" has no attribute "receipts"  [attr-defined]
    reveal_type(purchase.receipts)
```

Inspecting the `.mypy_cache` I can verify that there's no entry in `"names"` for such attributes. Making me highly suspect the plugin is setting `no_serialize=True` incorrectly for generated `RelatedManager`s.

## How is that should be

Cached runs should not error on attributes existing on non cached runs. i.e. Entry under `"names"` in cache should exist for attributes inheriting `RelatedManager`.

## System information

- OS:
- `python` version: 3.10
- `django` version: 4.1.1
- `mypy` version: 0.971

flaeppe:
## Related issues

Closes #1151

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -538,9 +538,7 @@ class AddRelatedManagers(ModelClassInitializer):
                     model_info=related_model_info, derived_from="_default_manager"
                 )
                 if default_reverse_manager_info:
-                    self.add_new_node_to_model_class(
-                        attname, Instance(default_reverse_manager_info, []), no_serialize=True
-                    )
+                    self.add_new_node_to_model_class(attname, Instance(default_reverse_manager_info, []))
                     continue
 
                 # The reverse manager we're looking for doesn't exist. So we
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
Can you please verify this locally with cache tuend on?

flaeppe:
Yeah, I've done that. Seeing output like:

```
note: Revealed type is "bcom.pos.models.Receipt_RelatedManager"
```

As well as finding the (previously missing) entries in cache:

```json
      "receipts": {
        ".class": "SymbolTableNode",
        "kind": "Mdef",
        "node": {
          ".class": "Var",
          "flags": [
            "is_initialized_in_class",
            "is_ready"
          ],
          "fullname": "[...].models.Purchase.receipts",
          "name": "receipts",
          "type": "[...].models.Receipt_RelatedManager"
        },
        "plugin_generated": true
      },
```

sobolevn (approved):
Caching is hard.

flaeppe:
> Caching is hard.

Indeed

For completeness I just want to add that I'm also seeing a `SymbolTableNode` for `Receipt_RelatedManager`, with a correct mro which, IIRC, is kind of the end result of the code around the changed lines here

```json
    "mro": [
      "[...].models.Receipt_RelatedManager",
      "django.db.models.manager.RelatedManager",
      "[...].models.ReceiptManager",
      "django.db.models.manager.Manager",
      "django.db.models.manager.BaseManager",
      "builtins.object"
    ],
```

PR #1156, opened 2022-09-19
Relax `fields=` arg for `refresh_from_db` to accept `Sequence`

flaeppe:
Django handles either e.g. `list` and `tuple`


adamchainz:
Thanks!

Issue #1095, opened 2022-08-10
Release schedule


aleksanb:
Can we release the newest main to PyPI?

It's possible to install main via a github url directly in pip, but that's somewhat cumbersome and i've had issues specifying the correct version of django-stubs-ext when installing that way.

PIG208:
```
https://github.com/typeddjango/django-stubs/archive/74e31c7562c7b983fbef832b13629ee8c13636ff.zip#egg=django-stubs-ext==0.6.0+git&subdirectory=django_stubs_ext
https://github.com/typeddjango/django-stubs/archive/74e31c7562c7b983fbef832b13629ee8c13636ff.zip#egg=django-stub==1.12.0+git
```
works for me. Replace the commit hash with the revision you need.

aleksanb:
> works for me. Replace the commit hash with the revision you need.

great! I was missing the subdirectory part of the url.

christianbundy:
> It's possible to install main via a github url directly in pip, but that's somewhat cumbersome and i've had issues specifying the correct version of django-stubs-ext when installing that way.

☝️

adamchainz:
You can use Pip's updated [VCS support](https://pip.pypa.io/en/stable/topics/vcs-support/) to pull with Git rather than zip files. This should be quicker than zip files now, afaiu, e.g.:

```
django-stubs @ git+https://github.com/typeddjango/django-stubs.git@354a892ecc45ef92e3200a10920e124d60f279a5
django-stubs-ext @ git+https://github.com/typeddjango/django-stubs.git@354a892ecc45ef92e3200a10920e124d60f279a5#subdirectory=django_stubs_ext
```

christianbundy:
> We haven't made a release with support yet, but until then you can install from Git (https://github.com/typeddjango/django-stubs/issues/1095#issuecomment-1241967327).

It probably goes without saying, but lots of folks use package mirrors that require a release to pypi, and can't safely install directly from this github repo. I'm happy to help push releases more regularly, but unfortunately the Git workarounds don't work for everyone.

adamchainz:
Understood. I've just started as a maintainer, I'm hoping I can help make relaeses more regular.

It's also possible to copy the stubs into your project, I [blogged about doing so](https://adamj.eu/tech/2022/08/23/python-type-hints-gradually-add-types-for-third-party-packages/) recently.

Thorbenl:
[...]

PR #1160, opened 2022-09-23
Update `django.template.base.Template.render()` argument type

### LINKED ISSUE #1147: django.template.Template.render() has incorrectly typed argument
# Bug report

## What's wrong

`django.template.Template.render()` has the following stub (see [here](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/template/base.pyi#L76)):

```python
def render(self, context: Optional[Union[Context, Dict[str, Any]]]) -> SafeString: ...
```

However, the implementation of this method below makes it impossible to pass in a `Dict[str, Any]` (see [here](https://github.com/django/django/blob/main/django/template/base.py#L169)) as it immediately attempts to access an attribute `render_context`:

```python
def render(self, context):
    "Display stage -- can be called many times"
    with context.render_context.push_state(self):
        if context.template is None:
            with context.bind_template(self):
                context.template_name = self.name
                return self._render(context)
        else:
            return self._render(context)
```

From the looks of it, the various backends define their own `Template` class, some (all?) of which have a `render()` method that accepts dictionaries, but this is not the case for the standard `django.template.Template` class.

I don't know enough of the surrounding context to say whether this is intended or not, but it seems wrong from a first glance. Apologies if this has already been answered, I did have a look through the issues on the repo to try to find similar queries.

## How is that should be

The stub for `django.template.Template.render()` should be changed to:

```python
def render(self, context: Optional[Context]) -> SafeString: ...
```

## System information

- OS: macOS 12.2.1
- `python` version: 3.10.6
- `django` version: 4.0.7
- `mypy` version: 0.971

Majsvaffla:
# I have made things!

I have corrected the type definition for the _context_ argument to `django.template.base.Template.render`. The argument is no longer typed as optional and does no longer accept a _dict_ value.

## Related issues

- Closes #1147


sobolevn (approved):
I am quite unsure what to do.

Removing this completely means that older versions of django become unsupported. 

@adamchainz what do you think?

sobolevn:
(I meant to click "Comment", not "Approve")

adamchainz:
I'd like to make some hints conditional by Django version, if that could be possible... perhaps the mypy plugin could provide the current version?

sobolevn:
As far as I know - custom builtins are not supported. https://github.com/python/mypy/issues/12172

We can do something like:

```python
_DjangoVersion4_1 = _DjangoVersion[4, 1]
```

Inside our plugin we can return `Literal[True]` for this using `get_type_analyze_hook()` from https://mypy.readthedocs.io/en/stable/extending_mypy.html

But, this is quite complex!

adamchainz:
I was thinking about typing `django.VERSION` conditionally, so we could use the same syntax as at runtime:

```py
if django.VERSION >= (4 ,1):
    ...  # new stuff
else:
    ...  # old stuff
```

It seems though that Mypy cannot type narrow on comparisons of literal ints in tuples, like below:

```py
from typing import Literal
from typing_extensions import assert_never

version: tuple[Literal[4], Literal[1]] = (4, 1)

if version >= (4, 1):
    reveal_type(version)
else:
    assert_never(version)
```

It doesn't work:

```
$ mypy example.py
example.py:7: note: Revealed type is "Tuple[Literal[4], Literal[1]]"
example.py:9: note: Revealed type is "Tuple[Literal[4], Literal[1]]"
example.py:10: error: Argument 1 to "assert_never" has incompatible type "Tuple[Literal[4], Literal[1]]"; expected "NoReturn"
```

It seems this limitation also applies to literal ints in general:

```py
from typing import Literal
from typing_extensions import assert_never

version: Literal[4] = 4

if version >= 4:
    reveal_type(version)
else:
    reveal_type(version)
    assert_never(version)
```

Perhaps the better approach is as you suggested, some alternative true/false constants. Perhaps we fix to the latest version but override that in the plugin based on the installed version of django that we find, or some config.

sobolevn:
`sys.version_info` check is hard-coded. This is why I thought about `Literal[True]` / `Literal[False]` hacks with analyzed dynamic types.

ngnpope:
Also found this related issue: https://github.com/python/typing/issues/693

Following on from Adam's example, I did note that `==` and `!=` with a literal number do work as expected:
```python
from typing import Literal
from typing_extensions import assert_never

version: Literal[4] = 4

if version == 4:
    reveal_type(version)
else:
    reveal_type(version)
    assert_never(version)

if version != 4:
    reveal_type(version)
    assert_never(version)
else:
    reveal_type(version)

if version == 4 or version == 5:
    reveal_type(version)
else:
    reveal_type(version)
    assert_never(version)
```
This yields the following:
```console
❯ mypy test.py 
test.py:7: note: Revealed type is "Literal[4]"
test.py:16: note: Revealed type is "Literal[4]"
test.py:19: note: Revealed type is "Literal[4]"
Success: no issues found in 1 source file
```
So you could possibly construct something very verbose to achieve this by exposing major and minor versions as independent variables and thrashing out all combinations.

I'm currently looking into whether we can extend `mypy` to allow for checking additional things as well as `sys.version_info` when inferring condition values. It looks like it should be relatively straightforward to piggy back off the existing logic and provide a callback... I'll have to have a go.

adamchainz:
> So you could possibly construct something very verbose to achieve this by exposing major and minor versions as independent variables and thrashing out all combinations.

We could also use a single numbers: `41` for Django 4.1, `42` for Django 4.2, etc.

> I'm currently looking into whether we can extend `mypy` to allow for checking additional things as well as `sys.version_info` when inferring condition values. It looks like it should be relatively straightforward to piggy back off the existing logic and provide a callback... I'll have to have a go.

Sick

django-stubs/template/base.pyi
```diff
@@ -73,7 +73,7 @@ class Template:
         engine: Optional[Engine] = ...,
     ) -> None: ...
     def __iter__(self) -> Iterator[Node]: ...
-    def render(self, context: Optional[Union[Context, Dict[str, Any]]]) -> SafeString: ...
+    def render(self, context: Context) -> SafeString: ...
```
intgr, line 60: Note that this has to be compatible with `django.template.backends.base._EngineTemplate` protocol defined in django-stubs.

intgr: Oh, wait. There are two `Template` classes in Django

`django.template.base.Template.render()` doesn't support dict, but `django.template.backends.django.Template.render()` is a wrapper class that does support `dict` and `None`.

`django.template.backends.django.DjangoTemplates.get_template()` wraps the first Template in the latter wrapper class. Confusing!

intgr:
> Removing this completely means that older versions of django become unsupported.

Note that this is not as bad as it actually looks.

If you get your template via the `django.template.loader.get_template()` or `get_template()` method of the template engine, you still get a Template whose `render()` supports `dict` and `None`.

As noted in my other comment:

> Oh, wait. There are two `Template` classes in Django
> 
> `django.template.base.Template.render()` doesn't support `dict`, but `django.template.backends.django.Template.render()` is a wrapper class that does support `dict` and `None`.
> 
> `django.template.backends.django.DjangoTemplates.get_template()` wraps the first Template in the latter wrapper class. Confusing!

christianbundy:
> We could also use a single numbers: 41 for Django 4.1, 42 for Django 4.2, etc.

Yep! Or a NamedTuple if we really needed this specificity:

```python3
from typing import Literal, NamedTuple
from typing_extensions import assert_never


class Version(NamedTuple):
    major: Literal[4]
    minor: Literal[1]
    
version = Version(4, 1)

if version.major == 4:
    reveal_type(version.major)
else:
    reveal_type(version.major)
    assert_never(version)

if version.major != 4:
    reveal_type(version.major)
    assert_never(version)
else:
    reveal_type(version.major)

if version.major == 4 or version.major == 5:
    reveal_type(version.major)
else:
    reveal_type(version.major)
    assert_never(version)
```

adamchainz (approved):
I think we can merge this one now, as in https://github.com/typeddjango/django-stubs/issues/1095#issuecomment-1307066177 we agreed to support just the latest Django version.

adamchainz:
I opened an issue about conditional types based on Django version: #1244

sobolevn:
Let's merge it right after `1.13.1` release

Majsvaffla:
I haven't followed along completely but if I understand correctly, I could resolve the conflicts and this is still to be merged?

sobolevn:
Yes :)

flaeppe:
Thanks for the PR and sorry for being so super slow getting it merged

PR #1162, opened 2022-09-24
Fix partially analyzed settings module

### LINKED ISSUE #312: Type from settings not correctly loaded in a model
# Bug report

I have this in settings:
```
BANK_ENTITIES = [('a', 'A Bank'), ('b', 'B Bank'),]
```

and this in my model:
```
entity = models.CharField(max_length=20, choices=settings.BANK_ENTITIES, blank=True)
```

## What's wrong

mypy fails with: 

```
models.py:247: error: Argument "choices" to "CharField" has incompatible type "tuple"; expected "Optional[Iterable[Union[Tuple[Any, Any], Tuple[str, Iterable[Tuple[Any, Any]]]]]]"
```

If I place a reveal in settings and another in models.py I get 

```
settingsbase.py:90: note: Revealed type is 'builtins.list[Tuple[builtins.str, builtins.str]]'
models.py:214: note: Revealed type is 'builtins.list'
```

## How is that should be

mypy should be able to load the full type from settings, not only that it's a list

## System information

- OS: Ubuntu 18.04
- `python` version: 3.6
- `django` version: 2.2
- `mypy` version: 0.761

PIG208:
# I have made things!

For modules `django.http` and `django.http.request`, we manually add
dependency for the module containing `AUTH_USER_MODEL` for mypy to
resolve first.

This potentially adds cyclic dependency when the models module contains
a reference to `django.conf`, and the settings depend on any modules that
rely on `django.http` or `django.http.request`. At this point, the settings module is
only partially analyzed. The symbol table is populated (so top-level settings are
accessible by the name), but the types have not been inferred.

Removing this dependency of `django.http` and `django.http.request` works. But
if the settings module also depends on `django.contrib.auth`, this still breaks.

This can potentially break the refined type for `request.user`, which does depend
on `AUTH_USER_MODEL` being available beforehand. This implicit dependency
still needs to be somehow established.

So, until these issues are addressed, this fix is not final.

## Related issues

- Closes #312 
- Refs #312

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -131,7 +131,7 @@ class NewSemanalDjangoPlugin(Plugin):
 
         # for `get_user_model()`
         if self.django_context.settings:
-            if file.fullname == "django.contrib.auth" or file.fullname in {"django.http", "django.http.request"}:
+            if file.fullname == "django.contrib.auth":
                 auth_user_model_name = self.django_context.settings.AUTH_USER_MODEL
                 try:
                     auth_user_module = self.django_context.apps_registry.get_model(auth_user_model_name).__module__
--- a/tests/typecheck/test_settings.yml
+++ b/tests/typecheck/test_settings.yml
@@ -45,3 +45,28 @@
             content: |
                 from pathlib import Path
                 MEDIA_ROOT = Path()
+
+-   case: imported_from_models
+    disable_cache: true
+    custom_settings: |
+        from other import other
+        AUTH_USER_MODEL = "my_app.MyUser"
+        INSTALLED_APPS = ["my_app"]
+        MY_LIST = [1]
+    main: |
+        from typing import TYPE_CHECKING
+        from django.conf import settings
+        if TYPE_CHECKING:
+            reveal_type(settings.MY_LIST)  # N: Revealed type is "builtins.list[builtins.int]"
+    files:
+        -   path: my_app/__init__.py
+        -   path: my_app/models.py
+            content: |
+                from django.db import models
+                import main
+                class MyUser(models.Model):
+                    pass
+        -   path: other.py
+            content: |
+                from django_stubs_ext import ValuesQuerySet
+                other = ()


andersk:
> But if the settings module also depends on `django.contrib.auth`, this still breaks.

That might not be a problem. My understanding is that a settings module is not supposed to have any dependencies on Django. That’s why `AUTH_USER_MODEL` is a string, for example.

PIG208:
Yes. Essentially the problem we want to solve here is the cyclic dependency that is not supposed to emerge from the settings.

andersk:
Hmm. Maybe my test case, and by extension the Zulip code it originated from, is just wrong for that reason.

PIG208:
It is possible that this dependency is inevitable because it is from `django_stubs_ext`, which needs to be imported into the settings module. So probably we can fix that.

andersk:
Maybe the application’s `__init__.py` is a better place for the `django_stubs_ext.monkeypatch()`? (If so, the documentation should be adjusted.)

PIG208:
I ran this fix in Zulip. The error does not go away unless this `get_user_model` dependency as well as those for `defined_model_classes` (models referred via `ForeignKey`s) get removed completely. It appears that there are more ways to trigger this error through adding depdencies.

andersk:
I got it working by breaking a bunch of import cycles in Zulip (zulip/zulip#23040), and moving `django_stubs_ext.monkeypatch()` to `zerver/__init__.py`.

To make this solution robust, ideally we’d replace this [fallback to `value.__class__`](https://github.com/typeddjango/django-stubs/blob/02999f3e868f861b0b85af7fb7c49d309396e89a/mypy_django_plugin/transformers/settings.py#L41-L49) with an error, so that we’ll see the mistake if anyone tries to reintroduce such a cycle.

- #1163

sobolevn:
Instead we merged https://github.com/typeddjango/django-stubs/pull/1163

Thanks for working on this!

PR #1167, opened 2022-09-27
Update mypy to 0.981

sobolevn:
Closes https://github.com/typeddjango/django-stubs/pull/1166


adamchainz:
Nice, that's quite a few ignore comments gone.

PR #1168, opened 2022-09-27
Allow Promise to be passed to ArrayField.verbose_name

ljodal:
This expands the interface of ArrayField to allow lazy translations to be passed to the verbose_name field. Seems like this was missed in #1139


django-stubs/contrib/postgres/fields/array.pyi
```diff
@@ -27,7 +27,7 @@ class ArrayField(CheckFieldDefaultMixin, Field[_ST, _GT]):
         base_field: Field,
         size: Optional[int] = ...,
         *,
-        verbose_name: Optional[Union[str, bytes]] = ...,
+        verbose_name: Optional[Union[_StrOrPromise, bytes]] = ...,
         name: Optional[str] = ...,
```
ljodal: Should name also accept _StrOrPromise?

sobolevn: Yes, I think so

django-stubs/contrib/postgres/fields/array.pyi
```diff
@@ -27,7 +27,7 @@ class ArrayField(CheckFieldDefaultMixin, Field[_ST, _GT]):
         base_field: Field,
         size: Optional[int] = ...,
         *,
-        verbose_name: Optional[Union[str, bytes]] = ...,
+        verbose_name: Optional[Union[_StrOrPromise, bytes]] = ...,
```
intgr: Not really in the scope of this PR, but I can't help to ask, why is `bytes` even allowed here? Makes no sense IMO. Is this for Python 2 compatibility?

The commit message adding this simply states "fixes" d43cb1fcd7961ca620aaf1d67949baf2e3c040f3 (?!)

sobolevn: I think it might be some legacy thing, I am totally fine with removing it.

intgr: OK I can open a PR for that after this is merged, if ljodal doesn't get to it first.

django-stubs/contrib/postgres/fields/array.pyi
ljodal: Aight, updated that as well

django-stubs/contrib/postgres/fields/array.pyi
ljodal: I removed the bytes while fixing `name` :)

sobolevn (approved):

sobolevn:
Thanks!

intgr:
Hrm. The `name` argument of all other fields is still typed as `Optional[str]` though.

And I think it doesn't makes sense for `Field.name` to ever be `Promise`. It determines the attribute of the `Model` instance and the database field name, it'll be embedded in migrations and database tables, etc.

ljodal:
Oh, oops, guess we should revert that bit then 😬

PR #1169, opened 2022-09-27
Reparametrize managers without explicit type parameters

### LINKED ISSUE #1022: error: Name "Sequence" is not defined

ljodal:
This extracts the reparametrization logic from #1030 in addition to removing the codepath that copied methods from querysets to managers. That code path seems to not be needed with this change.

This should properly fix #1022 as that keeps popping up from time to time when managers without type parameters are used.

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -1,5 +1,5 @@
 from collections import OrderedDict
-from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List, Optional, Set, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List, Optional, Set, Union
 
 from django.db.models.fields import Field
 from django.db.models.fields.related import RelatedField
@@ -10,11 +10,9 @@ from mypy.mro import calculate_mro
 from mypy.nodes import (
     GDEF,
     MDEF,
-    Argument,
     Block,
     ClassDef,
     Expression,
-    FuncDef,
     MemberExpr,
     MypyFile,
     NameExpr,
@@ -34,11 +32,10 @@ from mypy.plugin import (
     MethodContext,
     SemanticAnalyzerPluginInterface,
 )
-from mypy.plugins.common import add_method_to_class
 from mypy.semanal import SemanticAnalyzer
-from mypy.types import AnyType, CallableType, Instance, NoneTyp, TupleType
+from mypy.types import AnyType, Instance, NoneTyp, TupleType
 from mypy.types import Type as MypyType
-from mypy.types import TypedDictType, TypeOfAny, UnboundType, UnionType
+from mypy.types import TypedDictType, TypeOfAny, UnionType
 
 from mypy_django_plugin.lib import fullnames
 from mypy_django_plugin.lib.fullnames import WITH_ANNOTATIONS_FULLNAME
@@ -359,86 +356,6 @@ def add_new_sym_for_info(info: TypeInfo, *, name: str, sym_type: MypyType, no_se
     var.is_initialized_in_class = True
     var.is_inferred = True
     info.names[name] = SymbolTableNode(MDEF, var, plugin_generated=True, no_serialize=no_serialize)
-
-
-def build_unannotated_method_args(method_node: FuncDef) -> Tuple[List[Argument], MypyType]:
-    prepared_arguments = []
-    try:
-        arguments = method_node.arguments[1:]
-    except AttributeError:
-        arguments = []
-    for argument in arguments:
-        argument.type_annotation = AnyType(TypeOfAny.unannotated)
-        prepared_arguments.append(argument)
-    return_type = AnyType(TypeOfAny.unannotated)
-    return prepared_arguments, return_type
-
-
-def bind_or_analyze_type(t: MypyType, api: SemanticAnalyzer, module_name: Optional[str] = None) -> Optional[MypyType]:
-    """Analyze a type. If an unbound type, try to look it up in the given module name.
-
-    That should hopefully give a bound type."""
-    if isinstance(t, UnboundType) and module_name is not None:
-        node = api.lookup_fully_qualified_or_none(module_name + "." + t.name)
-        if node is not None and node.type is not None:
-            return node.type
-
-    return api.anal_type(t)
-
-
-def copy_method_to_another_class(
-    api: SemanticAnalyzer,
-    cls: ClassDef,
-    self_type: Instance,
-    new_method_name: str,
-    method_node: FuncDef,
-    return_type: Optional[MypyType] = None,
-    original_module_name: Optional[str] = None,
-) -> bool:
-    if method_node.type is None:
-        arguments, return_type = build_unannotated_method_args(method_node)
-        add_method_to_class(api, cls, new_method_name, args=arguments, return_type=return_type, self_type=self_type)
-        return True
-
-    method_type = method_node.type
-    if not isinstance(method_type, CallableType):
-        if not api.final_iteration:
-            api.defer()
-        return False
-
-    if return_type is None:
-        return_type = bind_or_analyze_type(method_type.ret_type, api, original_module_name)
-    if return_type is None:
-        return False
-
-    # We build the arguments from the method signature (`CallableType`), because if we were to
-    # use the arguments from the method node (`FuncDef.arguments`) we're not compatible with
-    # a method loaded from cache. As mypy doesn't serialize `FuncDef.arguments` when caching
-    arguments = []
-    # Note that the first argument is excluded, as that's `self`
-    for pos, (arg_type, arg_kind, arg_name) in enumerate(
-        zip(method_type.arg_types[1:], method_type.arg_kinds[1:], method_type.arg_names[1:]),
-        start=1,
-    ):
-        bound_arg_type = bind_or_analyze_type(arg_type, api, original_module_name)
-        if bound_arg_type is None:
-            return False
-        if arg_name is None and hasattr(method_node, "arguments"):
-            arg_name = method_node.arguments[pos].variable.name
-        arguments.append(
-            Argument(
-                # Positional only arguments can have name as `None`, if we can't find a name, we just invent one..
-                variable=Var(name=arg_name if arg_name is not None else str(pos), type=arg_type),
-                type_annotation=bound_arg_type,
-                initializer=None,
-                kind=arg_kind,
-                pos_only=arg_name is None,
-            )
-        )
-
-    add_method_to_class(api, cls, new_method_name, args=arguments, return_type=return_type, self_type=self_type)
-
-    return True
 
 
 def add_new_manager_base(api: SemanticAnalyzerPluginInterface, fullname: str) -> None:
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -508,19 +508,12 @@ def reparametrize_any_manager_hook(ctx: ClassDefContext) -> None:
     if preserve_typevars:
         return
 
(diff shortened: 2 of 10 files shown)


ljodal:
In addition to new type errors being discovered in our codebase, I've seen one new side-effect of this, which I'm unsure if we should consider a problem or not. Say you have a manager defined like this:

```python
from typing import TypeVar
from django.db import models

T = TypeVar("T", bound="Model")

class MyManager(models.Manager[T]):
    pass

class MyModel(models.Model):
    pass
```

If you subclass this without specifying generics, like this:

```python
class SecondManager(MyManager):
    pass
```

you'll get an error saying `Type argument "T" of "MyManager" must be a subtype of "MyModel"  [type-var]`.

But as I'm writing this I realize I might be able to make the logic smart enough to pick the "last" parameter in the MRO stack, if that sounds reasonable? Then `SecondManager` should be interpreted as `SecondManager[T: MyModel]` rather than `SecondManager[T: Model]`. I'll take a stab and see :)

tests/typecheck/managers/test_managers.yml
```diff
@@ -386,9 +386,11 @@
 -   case: override_manager_create1
     main: |
         from myapp.models import MyModel
-        MyModel.objects.create()
     installed_apps:
         - myapp
+    out: |
+        myapp/models:4: error: Return type "MyModel" of "create" incompatible with return type "_T" in supertype "BaseManager"
+        myapp/models:5: error: Incompatible return value type (got "_T", expected "MyModel")
```
ljodal, line 393: This is the failing test I mentioned in the PR description. I decided to fix it like this, basically saying this is an error. I think it should be fine to raise an error if you've restricted the type of an overridden method but the type vars of to the manager class

ljodal:
> [...]
> But as I'm writing this I realize I might be able to make the logic smart enough to pick the "last" parameter in the MRO stack, if that sounds reasonable? Then `SecondManager` should be interpreted as `SecondManager[T: MyModel]` rather than `SecondManager[T: Model]`. I'll take a stab and see :)

I've implemented this fix now, so I think this should be good to go. It does not cause any new issues that I've seen in our codebase other than surfacing more type errors that were previously hidden behind implicit `Any`s

mypy_django_plugin/transformers/managers.py
```diff
@@ -466,3 +466,55 @@ def create_new_manager_class_from_as_manager_method(ctx: DynamicClassDefContext)
         # Note that the generated manager type is always inserted at module level
         SymbolTableNode(GDEF, new_manager_info, plugin_generated=True),
     )
+
+
+def reparametrize_any_manager_hook(ctx: ClassDefContext) -> None:
+    """
+    Add explicit generics to manager classes that are defined without generic.
```
```suggestion
    Add implicit generics to manager classes that are defined without generic.
```

mypy_django_plugin/transformers/managers.py
```diff
@@ -466,3 +466,55 @@ def create_new_manager_class_from_as_manager_method(ctx: DynamicClassDefContext)
...
+
+    Eg.
+
+        class MyManager(models.Manager): ...
+
+    is interpreted as::
```
```suggestion
    is interpreted as::

```

sobolevn (commented):
Can you please rebase your branch? 
I've just merged a similar PR, I want to be sure that it is fine.

sobolevn (approved):
Oh, you just did :)

sobolevn:
@flaeppe any comments?

mypy_django_plugin/transformers/managers.py
```diff
@@ -466,3 +466,55 @@ def create_new_manager_class_from_as_manager_method(ctx: DynamicClassDefContext)
...
+        (base for base in manager.node.bases if base.type.has_base(fullnames.BASE_MANAGER_CLASS_FULLNAME)),
+        None,
+    )
+    if parent_manager is None:
+        return
+
+    is_missing_params = (
+        len(parent_manager.args) == 1
+        and isinstance(parent_manager.args[0], AnyType)
+        and parent_manager.args[0].type_of_any is not TypeOfAny.explicit
```
flaeppe: There's also a `TypeOfAny.from_omitted_generics`. Not sure what we should use here, but we might perhaps be a bit more narrow if we do `TypeOfAny.from_omitted_generics`?

If it even works, might be cases I'm forgetting..

flaeppe (commented):
I only have one thought, I'm a bit unsure about it though

mypy_django_plugin/transformers/managers.py
flaeppe: Otherwise we could overwrite something like `TypeOfAny.from_error` etc etc

ljodal: Aight, I'll take a look and see what type of any we actually get in these cases and make it specific to that if possible :)

mypy_django_plugin/lib/helpers.py
```diff
@@ -361,86 +358,6 @@ def add_new_sym_for_info(info: TypeInfo, *, name: str, sym_type: MypyType, no_se
...
-    That should hopefully give a bound type."""
-    if isinstance(t, UnboundType) and module_name is not None:
-        node = api.lookup_fully_qualified_or_none(module_name + "." + t.name)
-        if node is not None and node.type is not None:
-            return node.type
-
-    return api.anal_type(t)
-
-
-def copy_method_to_another_class(
```
flaeppe, line 389: How great, we finally get rid of this one!

mypy_django_plugin/transformers/managers.py
sobolevn: This will trigger an error if `disallow_any_generics = True` is set.
I am not sure we want this.

ljodal: I'm not following @sobolevn, you mean as the code is written now or if using `TypeOfAny.from_omitted_generics`?

ljodal: I've updated this to only run if the any type is `TypeOfAny.from_omitted_generics`, which seems to work great 👍

flaeppe: Correct me if I'm wrong, but I think what @sobolevn was trying to say is that even if we implicitly "repair" the omitted generic, mypy will still emit an error. Given you run with `--disallow-any-generics`

So one might argue that there's no need for us to fix these. Since one can find and make them explicit via mypy signaling where you've missed populating the generic by using the flag.

There's still a use case here though, and that's for our dynamically generated managers. In case they don't have any model arg populated. Since a user can't declare those explicitly

ljodal: Ah, okay. I'm following now, so basically we should remove this entire branch and accept that if you don't specify the generics of the class you'll get `Any`? I think one argument for keeping this is third party libraries that are not typed. With this you'll at least get correct types on the queryset in your code.

I guess we could also put this behind a flag?

That said, I don't think we're really changing the behaviour of the mypy plugin here, it's always been doing this stuff, it's just been doing it using a different approach (creating a new class)

flaeppe: It's important we do this for implicit managers from Django though (e.g. related managers)

Unless that's done someplace else

ljodal: I don't think that hits the new hook here, as that's (as far as I know) only called for `class MyClass:` statements in code. Related managers are handled elsewhere I think?

sobolevn: Sorry for causing so much confusion :)

I was thinking about cases where our users have `disallow_any_generics = True` set, current implementation looks like it can still cause this error. Let's add a test case with `disallow_any_generics = True` set. I am pretty sure that we won't be able to fix / suppress this error, because it happens too early: https://github.com/python/mypy/blob/ddd917767aa8135f3b1aeef47b0bb0616a4b63fb/mypy/typeanal.py#L1462

But, we can at least be sure: what happens and what to expect.
Maybe even mention this in the docs 🤔

ljodal: Ah, okay :) I'll add a test case!

I don't think we should try to suppress the error message if possible though, if the user has enabled that flag I'd expect them to want a warning. At least in our codebase the implicit generics we add are less specific than what you'd manually add, so I consider this a sort of in-between solution until all of the project is fully typed (we're working in a fairly large codebase, gradually adding types)

tests/typecheck/managers/test_managers.yml
```diff
@@ -386,9 +386,11 @@
 -   case: override_manager_create1
     main: |
         from myapp.models import MyModel
-        MyModel.objects.create()
     installed_apps:
         - myapp
+    out: |
+        myapp/models:4: error: Return type "MyModel" of "create" incompatible with return type "_T" in supertype "BaseManager"
```
sobolevn, line 392: This looks like a problem: `BaseManager` defines `create() -> [T: Model]`, and our `ModelManger(BaseManager)` defines it as `create() -> MyModel`. Why is that an error?

Refs https://github.com/typeddjango/django-stubs/pull/1169#issuecomment-1259676102

mypy_django_plugin/transformers/managers.py
ljodal: Hmm, it seems like this interacts a bit weirdly with `disallow_any_generics = True`. You'll get the expected warning from mypy and it seems like whatever we do in the hook is overwritten and the return types are changed back to `Any` 🤔  I added a test that demonstrates this behaviour

ljodal: Ah, I dug a bit deeper and when that option is enabled the `Any` type is changed to `TypeOfAny.from_error`, so the hook doesn't do anything. I think that's fair enough if you've enabled that option?

tests/typecheck/managers/test_managers.yml
ljodal: Hmm, yeah, I can dig a bit deeper into this

ljodal: The error here is exactly the same as you'd get from mypy if you did this:

```python
from typing import TypeVar
from django.db import models

class MyManager(models.Manager[T]):
    def create(self, **kwargs) -> "MyModel":
        ...

class MyModel(models.Model):
    ...
```

So it's expected based on what the plugin is doing, but I guess it might be surprising?

sobolevn: Yes, I now understand. The error is correct, but is very confusing.

Last question: can we somehow make it more clear? Or complete remove it (eventhough it iw correct)?

ljodal: Hmm, I can see what I can do, but I'm not aware of any methods we have available to change mypy's warnings?

At the very least I can add a note about this in the readme. I _might_ also be able to detect when this has happened and omit an additional error maybe, but I'm not sure if that helps much?

sobolevn: We can add a `note` saying that it probably should be `class MyManager(models.Manager[MyModel]):`, not `class MyManager(models.Manager[T]):`

How does that sound?

sobolevn: Or we can just ignore it for now :)
And wait for some complains first.

ljodal: I don't think it's easy to detect these cases, given how many methods on `Manager` might be affected by this. I added a new section to the FAQ about this issue, so hopefully that should be enough.

sobolevn (approved):
Thank you so much! Awesome feature.

PR #1170, opened 2022-09-27
Disallow passing bytes for Field verbose_name

intgr:
As discussed in #1168 Field `verbose_name` argument should not accept bytes.

Also fixed an omission in `GenericIPAddressField` that had `verbose_name` type `Any`.


sobolevn (approved):
Thanks!

PR #1173, opened 2022-09-29
Fix types for sessions

ljodal:
The session key can be None, which type stubs no longer allow after #1129, and the session_dict can be `dict[str, Any]`


django-stubs/contrib/sessions/backends/base.pyi
```diff
@@ -24,14 +24,14 @@ class SessionBase(Dict[str, Any]):
     def items(self): ...
     def clear(self) -> None: ...
     def is_empty(self) -> bool: ...
-    def _get_session_key(self) -> str: ...
+    def _get_session_key(self) -> Optional[str]: ...
     def _set_session_key(self, value: str) -> None: ...
```
ljodal:
```suggestion
    def _set_session_key(self, value: Optional[str]) -> None: ...
```

ljodal:
I found some problems with `session_dict` as well in `BaseSession`, so I fixed those as well

adamchainz (approved):
LGTM!

sobolevn (commented):
Thank you! Can you please run `black` to fix https://github.com/typeddjango/django-stubs/actions/runs/3385096206/jobs/5622845846 ?

sobolevn:
I think I can enable `pre-commit.ci` to fix this automatically 🤔

adamchainz:
Yes pre-commit.ci ftw!

adamchainz:
I guess my local pre-commit didn't get applied because this was a rebase commit, and git doesn't fire hooks during rebase 😢

PR #1176, opened 2022-09-29
Fix type of <fieldname>_id when using ForeignKey(to_field=)

intgr:
Previously mypy_django_plugin would always use the field type of target model's primary key, but `to_field` can refer to a different field type.

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -144,16 +144,16 @@ class DjangoContext:
         return helpers.get_private_descriptor_type(field_info, "_pyi_lookup_exact_type", is_nullable=field.null)
 
     def get_related_target_field(
-        self, related_model_cls: Type[Model], field: "ForeignKey[Any]"
+        self, related_model_cls: Type[Model], field: "ForeignKey[Any, Any]"
     ) -> "Optional[Field[Any, Any]]":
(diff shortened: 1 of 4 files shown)


mypy_django_plugin/django/context.py
```diff
@@ -143,6 +143,20 @@ def get_field_lookup_exact_type(
...
+    def get_related_target_field(
+        self, related_model_cls: Type[Model], field: "ForeignKey[Any, Any]"
+    ) -> "Optional[Field[Any, Any]]":
+        # ForeginKey only supports one `to_fields` item (ForeignObject supports many)
+        assert len(field.to_fields) == 1
+        to_field_name = field.to_fields[0]
+        if to_field_name:
+            rel_field = related_model_cls._meta.get_field(to_field_name)
+            if not isinstance(rel_field, Field):
+                return None  # Not supported
```
intgr, line 155: If it's referring to a non-field, Django system checks would probably throw an error. I could also use `assert` here.

sobolevn (approved):
Great stuff! Thank you!

intgr:
@sobolevn Thanks for your continued maintenance.

There is already tons of good stuff merged in django-stubs, any plans for a release?

sobolevn:
Yes! Soon :)

Issue #997, opened 2022-06-16
Incorrect typing for ManyToMany add()


markedwards:
# Bug report

## What's wrong

The typing of a ManyToMany relation's `add()` method incorrectly assumes that model ids are int, but they can be other types, such as UUID.  Attempting to call `add(model_id)` causes a mypy error if the id is not an int:

```
Argument 1 has incompatible type "UUID"; expected "Union[FooModel, int]"
```

## How is that should be

The `add()` method should accept the related model, and whatever the type of the related model's id is.

superlevure:
I have the same issue with `set()` method, any workaround?

superlevure:
Looks like a duplicate of https://github.com/typeddjango/django-stubs/issues/997 too

PedroPerpetua:
This bug is still present as of `1.14.0`.

palfrey:
Still present in 5.0.0

PR #1179, opened 2022-10-02
Fix type of <fieldname>_id in query lookups when using ForeignKey(to_field=)

intgr:
This is similar to #1176, but applies to query lookups instead of model attribute access.

This should be merged after #1178.

### DIFF
--- a/django-stubs/db/models/fields/__init__.pyi
+++ b/django-stubs/db/models/fields/__init__.pyi
@@ -516,6 +516,7 @@ class DateTimeField(DateField[_ST, _GT]):
 class UUIDField(Field[_ST, _GT]):
     _pyi_private_set_type: Union[str, uuid.UUID]
     _pyi_private_get_type: uuid.UUID
+    _pyi_lookup_exact_type: Union[uuid.UUID, str]
     def __init__(
         self,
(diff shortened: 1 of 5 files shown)


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -516,6 +516,7 @@ class DateTimeField(DateField[_ST, _GT]):
 class UUIDField(Field[_ST, _GT]):
     _pyi_private_set_type: Union[str, uuid.UUID]
     _pyi_private_get_type: uuid.UUID
+    _pyi_lookup_exact_type: Union[uuid.UUID, str]
```
sobolevn, line 519: Unrelated, merged in a different PR.

sobolevn (changes requested):
LGTM! Thanks! Please, fix merge conflict.

palfrey:
What needs to be done to progress this one? I've just hit https://github.com/typeddjango/django-stubs/issues/997 and would love to be able to resolve that, which I _think_ this does.

intgr:
Sorry for dropping the ball here. Ugh, the rebase will be messy. :)

PR #1182, opened 2022-10-06
Add missing re-export for JSONObject.

PIG208:
# I have made things!

`JSONObject` should be re-exported from `django.db.models.functions`.

Source code: https://github.com/django/django/blob/c58a8acd413ccc992dd30afd98ed900897e1f719/django/db/models/functions/__init__.py


adamchainz (approved):

adamchainz:
Thanks!

PR #1191, opened 2022-10-18
Fix `RelatedManager` not callable

UnknownPlatypus:
# I have made things!

mypy raise an exception when we try to call a `RelatedManager`. However, [django allows that](https://github.com/django/django/blob/main/django/db/models/fields/related_descriptors.py#L648-L651).

Here is a minimal reproducible exemple:

```python
from django.db import models 

class CustomUser(User):
    staffs = models.Manager()
    
class SalesMan(models.Model):
    client = models.ManyToManyField(CustomUser)


sales_man = SalesMan.objects.first()

reveal_type(sales_man.client) # Revealed type is "django.db.models.manager.RelatedManager[test.t.CustomUser]"
sales_man.agent(manager="staffs") # error: "RelatedManager[CustomUser]" not callable
```

I added the missing stubs and a test validating it was failing before and passing after this patch.

## Related issues
None


sobolevn (approved):
Thank you! 👍

PR #1193, opened 2022-10-19
Update cache timeout parameters to floats

adamchainz:
# I have made things!

[The docs](https://docs.djangoproject.com/en/4.1/topics/cache/#django.core.caches.cache.set) only say that cache timeouts should be “numbers”. But Django’s base cache tests, applied to all builtin backends, deliberately checks floats:

https://github.com/django/django/blob/d62563cbb194c420f242bfced52b37d6638e67c6/tests/cache/tests.py#L658-L661

I also discovered this when adding type hints to Django-MySQL, which has a [cache backend](https://django-mysql.readthedocs.io/en/latest/cache.html), whose tests use float timeouts.

Using `float` as the type is still compatible with `int`, thanks to Mypy’s [duck type compatibility](https://mypy.readthedocs.io/en/stable/duck_type_compatibility.html?highlight=float#duck-type-compatibility).

## Related issues

n/a


sobolevn (approved):

sobolevn:
Thank you!

PR #1198, opened 2022-10-20
Generic sitemap

### LINKED ISSUE #1000: The type for location in the class Sitemap is wrong.
https://github.com/typeddjango/django-stubs/blob/8d8b8cd1fc8b54eac799905dbbffc4e21708b6da/django-stubs/contrib/sitemaps/__init__.pyi#L24

The item shouldn't just be Model. As you can see it can also just be a string too if creating a sitemap of pages that aren't in models:

````python
class StaticViewSitemap(Sitemap):
    priority = 1
    changefreq = "always"

    def items(self) -> List[str]:
        return ["home", "about", "contact", "recommendations", "privacy-policy", "blog"]

    def location(self, item: str) -> str:
        return reverse(item)
````

### LINKED ISSUE #1192: Sitemap type anotation is too restrictive
# Bug report

## What's wrong

I'm having a similar issue to #1000 when type annotating django sitemaps.

The following is not passing:

```python
class OfferSitemap(Sitemap):
    priority = 1
    changefreq = "always"

    def items(self) -> QuerySet[Offer]:
        return Offer.objects.all()

    def location(self, item: Offer) -> str:
        return reverse(
            "myapp:detail-offer",
            kwargs={
                "provider_name": item.provider.url_name,
                "offer_name": item.url_name,
            },
        )

# error: Argument 1 of "location" is incompatible with supertype "Sitemap"; supertype defines the argument type as "Model"
# note: This violates the Liskov substitution principle
```

## How is that should be

From the docs: 

> [items()](https://docs.djangoproject.com/en/4.1/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.items) is a method that returns a [sequence](https://docs.python.org/3/glossary.html#term-sequence) or QuerySet of objects. The objects returned will get passed to any callable methods corresponding to a sitemap property ([location](https://docs.djangoproject.com/en/4.1/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.location), [lastmod](https://docs.djangoproject.com/en/4.1/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.lastmod), [changefreq](https://docs.djangoproject.com/en/4.1/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.changefreq), and [priority](https://docs.djangoproject.com/en/4.1/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.priority)).

I think the type could be more generic because the `items` method is the one that determines the type of the `item` variable every other methods are using. Maybe taking a similar approach to #800 with an `_Item` Typevar ?

Let me know if you think this is a correct approach and I'll be happy to submit a PR to address that.

Cheers

## System information

- OS: **ubuntu20.04**
- `python` version: **3.8**
- `django` version: **3.2.15**
- `mypy` version: **0.950**
- `django-stubs` version: **1.11.0**
- `django-stubs-ext` version: **0.5.0**

UnknownPlatypus:
# I have made things!

Implement generic type for the `Sitemap` class and also add the [get_latest_lastmod](https://docs.djangoproject.com/en/4.1/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.get_latest_lastmod) method stub (added in django 4.1)

## Related issues
- Closes #1000 
- Closes #1192


tests/typecheck/contrib/sitemaps/test_generic_sitemap.yml
```diff
@@ -0,0 +1,49 @@
+-   case: test_items_custom_model
+    main: |
+        from django.contrib.sitemaps import Sitemap
+        from django.db.models import QuerySet
+        from django.urls import reverse
+        from myapp.models import Offer
+
+        class OfferSitemap(Sitemap):
```
sobolevn: Do you mean 
```suggestion
        class OfferSitemap(Sitemap[Offer]):
```
?

UnknownPlatypus: Yeah you're right. I just updated this and the other test.

tests/typecheck/contrib/sitemaps/test_generic_sitemap.yml
```diff
@@ -0,0 +1,49 @@
+-   case: test_items_custom_model
+    main: |
+        from django.contrib.sitemaps import Sitemap
+        from django.db.models import QuerySet
+        from django.urls import reverse
+        from myapp.models import Offer
+
+        class OfferSitemap(Sitemap[Offer]):
```
sobolevn, line 9: We also need a case with

```python
class WrongOfferSitemap(Sitemap[Offer]):
     def items() -> BadReturn: ...
     def locations(arg: BadArg): ...
```

UnknownPlatypus: Indeed, I've added this

tests/typecheck/contrib/sitemaps/test_generic_sitemap.yml
```diff
@@ -0,0 +1,62 @@
...
+            def location(self, item: Offer) -> str:
+                return reverse(
+                    "myapp:detail-offer",
+                    kwargs={
+                        "provider_name": item.provider,
+                        "offer_name": item.trc, # E: "Offer" has no attribute "trc"
+                    },
+                )
+
+        class WrongOfferSitemap(Sitemap[Offer]):
```
sobolevn, line 25: One last question: what about `GenericSiteMap`? Should we test it?

UnknownPlatypus: Yeah I was unsure about this one. I don't know how people are generally using it. 

Maybe I should add a case close to [the documentation exemple](https://docs.djangoproject.com/en/4.1/ref/contrib/sitemaps/#django.contrib.sitemaps.GenericSitemap) to make sure everything pass ?

sobolevn: 👍

tests/typecheck/contrib/sitemaps/test_generic_sitemap.yml
```diff
@@ -26,12 +27,23 @@
...
             def location(self, item: str) -> int:
                 return 1
+
+        info_dict = {
+          "queryset": Offer.objects.all(),
+        }
+        urlpatterns = [
+            path(
+                'sitemap.xml', sitemap,
+                {'sitemaps': {'offers': GenericSitemap(info_dict, priority=0.6)}},
```
sobolevn: and one more for an incorrect version

sobolevn (approved):
Thank you!

UnknownPlatypus:
Thanks for the review :raised_hands: !

PR #1199, opened 2022-10-21
Introduce `QuerySetAny` type for QuerySet isinstance checks

### LINKED ISSUE #704: [Question] What is an alternative to isinstance(obj, QuerySet)?

PIG208:
This also re-export QuerySetAlias for external access to generic _QuerySet. We use this in our test cases as well. The approach taken here is making `QuerySet` a subclass of `_QuerySet[_T, _T]`, and make `_QuerySetAlias` a type alias of `_QuerySet[_T, _T]`.

`_QuerySetAlias[_T]` should be used internally whenever the type parameter is needed, replacing `QuerySet[_T]` entirely within the stubs; both `QuerySet` and `_QuerySetAlias` are accepted in use cases where the type parameter is not needed.

# I have made things!

## Related issues

- Closes #704
- Refs #794

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,32 @@ func(MyModel.objects.annotate(foo=Value("")).get(id=1))  # OK
 func(MyModel.objects.annotate(bar=Value("")).get(id=1))  # Error
 ```
 
+### How do I check if something is an instance of QuerySet in runtime?
+
+A limitation of making `QuerySet` generic is that you can not use
+it for `isinstance` checks.
+
+```python
+from django.db.models.query import QuerySet
+
+def foo(obj: object) -> None:
+    if isinstance(obj, QuerySet): # Error: Parameterized generics cannot be used with class or instance checks
+        ...
+```
+
+To get around with this issue without making `QuerySet` non-generic,
+Django-stubs provides `django_stubs_ext.QuerySetAny`, a non-generic
+variant of `QuerySet` suitable for runtime type checking:
+
+```python
+from django_stubs_ext import QuerySetAny
+
+def foo(obj: object) -> None:
+    if isinstance(obj, QuerySetAny):  # OK
+        ...
+```
+
+
 ## Related projects
 
 - [`awesome-python-typing`](https://github.com/typeddjango/awesome-python-typing) - Awesome list of all typing-related things in Python.
--- a/django-stubs/db/models/query.pyi
+++ b/django-stubs/db/models/query.pyi
@@ -209,6 +209,8 @@ class RawQuerySet(Iterable[_T], Sized):
     def resolve_model_init_order(self) -> Tuple[List[str], List[int], List[Tuple[str, int]]]: ...
     def using(self, alias: Optional[str]) -> RawQuerySet[_T]: ...
 
+_QuerySetAny = _QuerySet
+
 QuerySet = _QuerySet[_T, _T]
 
 class Prefetch:
--- a/django_stubs_ext/django_stubs_ext/__init__.py
+++ b/django_stubs_ext/django_stubs_ext/__init__.py
@@ -1,3 +1,4 @@
+from .aliases import QuerySetAny as QuerySetAny
 from .aliases import StrOrPromise, StrPromise
 from .aliases import ValuesQuerySet as ValuesQuerySet
 from .annotations import Annotations as Annotations
@@ -7,6 +8,7 @@ from .types import AnyAttrAllowed as AnyAttrAllowed
 
 __all__ = [
     "monkeypatch",
+    "QuerySetAny",
     "ValuesQuerySet",
     "WithAnnotations",
     "Annotations",
--- a/django_stubs_ext/django_stubs_ext/aliases.py
+++ b/django_stubs_ext/django_stubs_ext/aliases.py
@@ -1,16 +1,18 @@
 import typing
 
 if typing.TYPE_CHECKING:
-    from django.db.models.query import _T, _QuerySet, _Row
+    from django.db.models.query import _T, _QuerySet, _QuerySetAny, _Row
     from django.utils.functional import _StrOrPromise as StrOrPromise
     from django.utils.functional import _StrPromise as StrPromise
 
+    QuerySetAny = _QuerySetAny
     ValuesQuerySet = _QuerySet[_T, _Row]
 else:
     from django.db.models.query import QuerySet
     from django.utils.functional import Promise as StrPromise
 
+    QuerySetAny = QuerySet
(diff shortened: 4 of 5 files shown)


PIG208:
One concern is that we are forcing end-users to use `QuerySetAlias` in type annotations.

PIG208:
TODO: Add test cases for `isinstance` checks with `QuerySet`

intgr:
If this gets merged, all downstream users using `QuerySet[MyModel]` need to rewrite them as `QuerySetAlias[MyModel]`? That's a pretty high cost, in my codebase there are 230 instances of `QuerySet[]` hints.

I understand this is to fix `isinstance(x, QuerySet)`, which I have not needed anywhere so far.

I agree the latter issue deserves to be fixed, but before inflicting such a large migration on users, we should be quite sure there are no better alternatives. Maybe there's something that can be done with mypy plugins to avoid this?

sobolevn:
Yeah, I am also quite worried about this. This seems very high-risk.

adamchainz:
I am also worried about making such a change. We should maintain `QuerySet[MyModel]` as a priority. Django was even changed to support it: https://github.com/django/django/commit/578c03b276e . It wouldn't be possible to remove that without a whole deprecation cycle anyway.

PIG208:
I think we can also do this the other way around, by making `QuerySetAlias` non-generic, and continuing using `QuerySet[_T]`.

PIG208:
The PR is updated.

The approach taken here is making `_QuerySetAlias` a subclass of `_QuerySet[_T, _T]` dedicated for `isinstance` checks, and leave `QuerySet` unchanged as the type alias of `_QuerySet[_T, _T]`.

A slight disadvantage of this approach is that `QuerySet` is not a subclass of `QuerySetAlias`, and thus using `QuerySetAlias` in type annotation is generally incorrect. But the fix should suffice for the `isinstance` check use case. This is noted in the added documentation.

README.md
```diff
@@ -270,6 +270,33 @@ func(MyModel.objects.annotate(foo=Value("")).get(id=1))  # OK
...
+def foo(obj: object) -> None:
+    isinstance(obj, QuerySet) # Error: Parameterized generics cannot be used with class or instance checks
+```
+
+To get around with this issue without making `QuerySet` non-generic,
+Django-stubs provides `django_stubs_ext.QuerySetAlias`, a non-generic
+variant of `QuerySet` dedicated for runtime type checking:
+
+```python
+from django_stubs_ext import QuerySetAlias
```
sobolevn: Do others have any name ideas for this?

I fear that `Alias` is too generic. It does not show the purpose of this thing.

Maybe `QuerySetInstance`?

adamchainz: `QuerySetInstance` 👍

django-stubs/db/models/query.pyi
```diff
@@ -209,6 +209,8 @@ class RawQuerySet(Iterable[_T], Sized):
     def resolve_model_init_order(self) -> Tuple[List[str], List[int], List[Tuple[str, int]]]: ...
     def using(self, alias: Optional[str]) -> RawQuerySet[_T]: ...
 
+class _QuerySetInstance(_QuerySet[_T, _T]): ...
```
sobolevn: Maybe `QuerySetInstance = _QuerySet`?

sobolevn (commented):
This way we can remove a new class. It might affect some cases like invariant lists. 

(Please, also add a test for `list[QuerySet].append(query_set_instance_obj)` case).

README.md
intgr: Alternatively `QuerySetAny` as in "QuerySet of any model"?

PIG208: `QuerySetAny` sounds better, since this type is essentially `_QuerySet[Any, Any]`.

django-stubs/db/models/query.pyi
PIG208: Updated.

sobolevn (approved):
Thank you! Looks great now!

I will let others to provide their feedback before merging.

PIG208:
Renamed the test case to `test_querysetany`.

README.md
```diff
@@ -270,6 +270,30 @@ func(MyModel.objects.annotate(foo=Value("")).get(id=1))  # OK
...
+
+To get around with this issue without making `QuerySet` non-generic,
+Django-stubs provides `django_stubs_ext.QuerySetAny`, a non-generic
+variant of `QuerySet` dedicated for runtime type checking:
+
+```python
+from django_stubs_ext import QuerySetAny
+
+def foo(obj: object) -> None:
+    isinstance(obj, QuerySetAny)  # OK
```
adamchainz: Can we make the examples a little bit more realistic?

```suggestion
def foo(obj: object) -> None:
    if isinstance(obj, QuerySet): # Error: Parameterized generics cannot be used with class or instance checks
        ...
```

To get around with this issue without making `QuerySet` non-generic,
Django-stubs provides `django_stubs_ext.QuerySetAny`, a non-generic
variant of `QuerySet` dedicated for runtime type checking:

```python
from django_stubs_ext import QuerySetAny

def foo(obj: object) -> None:
    if isinstance(obj, QuerySetAny):  # OK
        ...
```

adamchainz (approved):
LGTM, I think the new name is suitable.

README.md
PIG208: Good point! Fixed.

README.md
```diff
@@ -270,6 +270,32 @@ func(MyModel.objects.annotate(foo=Value("")).get(id=1))  # OK
...
+    if isinstance(obj, QuerySet): # Error: Parameterized generics cannot be used with class or instance checks
+        ...
+```
+
+To get around with this issue without making `QuerySet` non-generic,
+Django-stubs provides `django_stubs_ext.QuerySetAny`, a non-generic
+variant of `QuerySet` suitable for runtime type checking:
+
+```python
+from django_stubs_ext import QuerySetAny
```
sobolevn, line 291: Question for other PRs: do we have other types to take care of?

PIG208:
I think this is ready to merge.

sobolevn (approved):
Ok, `QuerySetAny` also seems like a good name!

intgr:
`QuerySetAny` is now deprecated and can be replaced with a simple `isinstance(x, QuerySet)` check. More details here:

* https://github.com/typeddjango/django-stubs/discussions/2194

PR #1200, opened 2022-10-21
Add hints for instance variables in console.EmailBackend

leamingrad:
Prior to this change, the console-based email backend inherited directly
from the base email backend, so it did not declare the extra variables
that are set on the instance ([here](https://github.com/django/django/blob/main/django/core/mail/backends/console.py#L12-L13)).

This change adds hints for the `stream` and `_lock` variables that these
instances have.

I didn't add any tests as I couldn't see one that would be obviously useful,
but I can add one if needed.

## Related issues

N/A (I can raise one if needed)

### DIFF
--- a/django-stubs/core/mail/backends/console.pyi
+++ b/django-stubs/core/mail/backends/console.pyi
@@ -1,3 +1,8 @@
+import threading
+from typing import TextIO
+
 from django.core.mail.backends.base import BaseEmailBackend
 
-class EmailBackend(BaseEmailBackend): ...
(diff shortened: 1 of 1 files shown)


adamchainz (approved):
LGTM, thx!

adamchainz:
> I didn't add any tests as I couldn't see one that would be obviously useful,
> but I can add one if needed.

Yeah we don't need tests for basic hints like this. Tests are more useful for complex types, like `@overload` or generics, or where the plugin is involved.

PR #1201, opened 2022-10-24
Allow `assertQuerysetEqual` to compare with a collection of values

flaeppe:
It should be possible to do e.g.

```python
self.assertQuerysetEqual(
    MyModel.objects.all(),
    [(1, timezone.now(), Decimal(1337), 1337), ...],
    transform=attrgetter("pk", "created_at", "some_decimal", "some_int", ...),
)
```

(Not just compare with pk's, which I think the old hints suggested?)


sobolevn (approved):
Thanks!

PR #1202, opened 2022-10-25
Improve type of `TestCase.captureOnCommitCallbacks`

flaeppe:
Also decorate it with `@classmethod` and `@contextmanager`


django-stubs/test/testcases.pyi
```diff
@@ -216,7 +218,11 @@ class TransactionTestCase(SimpleTestCase):
 class TestCase(TransactionTestCase):
     @classmethod
     def setUpTestData(cls) -> None: ...
-    def captureOnCommitCallbacks(cls, *, using: Optional[str] = ..., execute: bool = ...): ...
+    @classmethod
+    @contextmanager
+    def captureOnCommitCallbacks(
+        cls, *, using: Optional[str] = ..., execute: bool = ...
```
adamchainz: `using` should never have been marked `Optional`, it's always a string, it just has a default (indicated by `= ...`)	

```suggestion
        cls, *, using: str = ..., execute: bool = ...
```

django-stubs/test/testcases.pyi
```diff
@@ -216,7 +218,11 @@ class TransactionTestCase(SimpleTestCase):
...
+    ) -> Generator[Sequence[Callable[[], Any]], None, None]: ...
```
adamchainz: We should be specific and use `List` as the return type. This way tests that use list-specific operations still type check fine.

```suggestion
    ) -> Generator[List[Callable[[], Any]], None, None]: ...
```

adamchainz (approved):

django-stubs/test/testcases.pyi
flaeppe: Sure, we can do that. I went with `Sequence` to avoid highlighting that it's mutable, since it felt strange to expose that (e.g. `callbacks.append()` within the context could perhaps mess around with the implementation) even though it's a list.

adamchainz: Yeah, it would be nice to lock down weird operations with type hints, but to avoid type hints introducing unnecessary churn to projects, I think it's most pragmatic to document Django as-is.

sobolevn: Yes, +1

This is the same we do in typeshed: documenting things as-is.

PR #1204, opened 2022-10-26
Add lots of missing argument & return type hints

intgr:
Discovered by setting mypy options `disallow_untyped_defs`, `disallow_incomplete_defs`.


intgr:
I didn't fix all of the places -- there are tons more missing hints in `django.contrib.gis` and `django.db.models.fields`, so these settings can't be added to `mypy.ini` yet.

sobolevn (commented):
I love and hate PRs like this :)

I love them because you did a great work annotating so many files and functions!

I hate them because I am not able to review them properly :(

I will read thought it, but I won't check every param / annotation with the source code. Please, forgive that!

intgr:
> I hate them because I am not able to review them properly :(

I agree, there's too much stuff all over the place. But the idea is that eventually we can enable `disallow_untyped_defs`, `disallow_incomplete_defs` checks in CI so incomplete definitions could no longer be added.

django-stubs/contrib/admin/checks.pyi
```diff
@@ -16,6 +16,6 @@ class ModelAdminChecks(BaseModelAdminChecks):
 class InlineModelAdminChecks(BaseModelAdminChecks):
     def check(self, inline_obj: BaseModelAdmin, **kwargs: Any) -> Sequence[CheckMessage]: ...  # type: ignore
 
-def must_be(type: Any, option: Any, obj: Any, id: Any): ...
-def must_inherit_from(parent: Any, option: Any, obj: Any, id: Any): ...
-def refer_to_missing_field(field: Any, option: Any, obj: Any, id: Any): ...
+def must_be(type: Any, option: Any, obj: Any, id: Any) -> Sequence[CheckMessage]: ...
```
sobolevn: I am pretty sure it is not raw `Sequence`. Is it `list` or `tuple`?

django-stubs/contrib/contenttypes/forms.pyi
```diff
@@ -1,8 +1,12 @@
...
 
-from django.forms.models import BaseModelFormSet
+from django.db.models import Model
+from django.forms.models import BaseModelFormSet, ModelForm
 
-class BaseGenericInlineFormSet(BaseModelFormSet):
+_M = TypeVar("_M", bound=Model)
+_ModelFormT = TypeVar("_ModelFormT", bound=ModelForm)
+
+class BaseGenericInlineFormSet(BaseModelFormSet[_M, _ModelFormT]):
```
sobolevn, line 9: Should we add it to our monkeypatching? Since it is generic now.

sobolevn:
CC @adamchainz I really need a second pair of eyes 🙏

django-stubs/contrib/admin/checks.pyi
```diff
@@ -16,6 +16,6 @@ class ModelAdminChecks(BaseModelAdminChecks):
...
+def must_inherit_from(parent: Any, option: Any, obj: Any, id: Any) -> Sequence[CheckMessage]: ...
+def refer_to_missing_field(field: Any, option: Any, obj: Any, id: Any) -> Sequence[CheckMessage]: ...
```
adamchainz: let's use the concrete types

```suggestion
def must_be(type: Any, option: Any, obj: Any, id: Any) -> List[CheckMessage]: ...
def must_inherit_from(parent: Any, option: Any, obj: Any, id: Any) -> List[CheckMessage]: ...
def refer_to_missing_field(field: Any, option: Any, obj: Any, id: Any) -> List[CheckMessage]: ...
```

django-stubs/contrib/admin/helpers.pyi
```diff
@@ -170,7 +171,7 @@ class InlineAdminForm(AdminForm):
     def pk_field(self) -> AdminField: ...
     def fk_field(self) -> AdminField: ...
     def deletion_field(self) -> AdminField: ...
-    def ordering_field(self): ...
+    def ordering_field(self) -> AdminField: ...
```
adamchainz: This is removed in Django upstream since code never used it, should be fine to remove it from stubs prematurely

https://github.com/django/django/commit/f3a06b06b59b3d24d5c2c14a34a8051879203c45

```suggestion
```

django-stubs/contrib/admin/utils.pyi
```diff
@@ -45,7 +45,7 @@ class NestedObjects(Collector):
     def nested(self, format_callback: Callable = ...) -> List[Any]: ...
     def can_fast_delete(self, *args: Any, **kwargs: Any) -> bool: ...
 
-def model_format_dict(obj: Union[Model, Type[Model], QuerySet, Options[Model]]): ...
+def model_format_dict(obj: Union[Model, Type[Model], QuerySet, Options[Model]]) -> Dict[str, str]: ...
```
adamchainz: docstring:

> Return a `dict` with keys 'verbose_name' and 'verbose_name_plural',
>    typically for use with string formatting.

could use a `TypedDict` if you feel like it

django-stubs/contrib/auth/backends.pyi
```diff
@@ -33,7 +34,7 @@ class ModelBackend(BaseBackend):
         is_active: bool = ...,
         include_superusers: bool = ...,
         obj: Optional[Model] = ...,
-    ): ...
+    ) -> QuerySet[AbstractBaseUser]: ...
```
adamchainz, line 37: shouldn't this be with `_AnyUser`?

django-stubs/contrib/auth/models.pyi
```diff
@@ -51,7 +52,7 @@ class UserManager(BaseUserManager[_T]):
         include_superusers: bool = ...,
         backend: Optional[str] = ...,
         obj: Optional[Model] = ...,
-    ): ...
+    ) -> QuerySet[AbstractBaseUser]: ...
```
adamchainz: shouldn't this be the user class type param?

```suggestion
    ) -> QuerySet[_T]: ...
```

django-stubs/core/management/commands/sqlmigrate.pyi
```diff
@@ -10,4 +10,4 @@ from django.db.migrations.loader import MigrationLoader as MigrationLoader
 
 class Command(BaseCommand):
     output_transaction: bool = ...
-    def execute(self, *args: Any, **options: Any): ...
+    def execute(self, *args: Any, **options: Any) -> Optional[str]: ...
```
adamchainz, line 13: did you check all management commands' `execute` / `handle` methods?

django-stubs/db/models/expressions.pyi
```diff
@@ -118,8 +118,21 @@ class F(Combinable):
...
+        nulls_first: bool = ...,
+        nulls_last: bool = ...,
+    ) -> OrderBy: ...
+    def desc(
+        self,
+        *,
+        expression: Union[Expression, F, Subquery],
+        nulls_first: bool = ...,
+        nulls_last: bool = ...,
+    ) -> OrderBy: ...
```
adamchainz, line 135: nice

django-stubs/utils/asyncio.pyi
```diff
@@ -1,3 +1,8 @@
-from typing import Any
+from typing import Any, Callable, TypeVar, overload
 
-def async_unsafe(message: Any): ...
+_C = TypeVar("_C", bound=Callable)
+
+@overload
+def async_unsafe(message: str) -> Callable[[_C], _C]: ...
+@overload
+def async_unsafe(message: _C) -> _C: ...
```
adamchainz, line 8: nice

django-stubs/utils/dateformat.pyi
```diff
@@ -37,20 +37,20 @@ class DateFormat(TimeFormat):
     timezone: Optional[_TzInfoT]
     year_days: Any = ...
     def __init__(self, obj: Union[builtin_datetime, builtin_time, date]) -> None: ...
-    def b(self): ...
+    def b(self) -> str: ...
```
adamchainz, line 40: interesting how just some of these were missed

django-stubs/utils/functional.pyi
```diff
@@ -24,7 +24,7 @@ class Promise:
     def __mod__(self, rhs: Any) -> Any: ...
     def __add__(self, other: Any) -> Any: ...
     def __radd__(self, other: Any) -> Any: ...
-    def __deepcopy__(self, memo: Any): ...
+    def __deepcopy__(self: _T, memo: Any) -> _T: ...
```
adamchainz, line 27: @sobolevn  should we use `typing_extensions.Self` ?

adamchainz (commented):
192 additions / 133 deletions is not too bad... had a quick look.

Is there any way to set the options [in per-module settings](https://mypy.readthedocs.io/en/stable/config_file.html#per-module-and-global-options) in `mypy.ini`, for at least some modules?

django-stubs/utils/functional.pyi
sobolevn: Not yet, bound type var is still the way to go.

django-stubs/contrib/admin/checks.pyi
intgr: It's actually `list`, but all other returns in this file are typed as `Sequence`. I'll change all of them to `List` then.

django-stubs/contrib/contenttypes/forms.pyi
intgr: `BaseModelFormSet` doesn't seem to be patched either.

django-stubs/contrib/admin/checks.pyi
intgr: Changed other returns in this file too.

django-stubs/core/management/commands/sqlmigrate.pyi
intgr: Why "all management commands"? This is just one command: sqlmigrate

I used the same return as `BaseCommand`.

django-stubs/contrib/auth/backends.pyi
intgr: Won't work

> django-stubs/contrib/auth/backends.pyi:37: error: Type argument "Union[AbstractBaseUser, AnonymousUser]" of "_QuerySet" must be a subtype of "Model"

intgr:
> Is there any way to set the options [in per-module settings](https://mypy.readthedocs.io/en/stable/config_file.html#per-module-and-global-options) in mypy.ini, for at least some modules?

I tried, but couldn't get it to work. I suspect it might not work because the directory is called `django-stubs`, but module is actually `django`.

In any case I'll do a follow-up shortly that enables these checks globally.

django-stubs/core/management/commands/sqlmigrate.pyi
intgr: I did check all functions that are missing return type annotations.

I didn't check all `Command` classes systematically in case they have differing return types, I think that's out of scope for this PR.

django-stubs/contrib/admin/utils.pyi
```diff
@@ -45,7 +45,11 @@ class NestedObjects(Collector):
     def nested(self, format_callback: Callable = ...) -> List[Any]: ...
     def can_fast_delete(self, *args: Any, **kwargs: Any) -> bool: ...
 
-def model_format_dict(obj: Union[Model, Type[Model], QuerySet, Options[Model]]): ...
+class ModelFormatDict(TypedDict):
```
sobolevn, line 48: I think that this does not exist in django, so it is `_ModelFormatDict`

django-stubs/contrib/auth/backends.pyi
intgr: This method does a DB query, so it only returns user models -- it can't return `AnonymousUser`.

Similar to how `authenticate()` returns `AbstractBaseUser`.

sobolevn (approved):
I think we should just merge it. We can fix any further problems later.

sobolevn:
Thanks everyone! Major update!

django-stubs/contrib/admin/utils.pyi
intgr: Ooops, sorry! I pushed this fix to the wrong branch. Will be fixed in #1206.

sobolevn: Ok, no problem!

PR #1206, opened 2022-10-27
Fill remaining missing hints with Any & disallow partial hints

intgr:
This one just adds `Any` to all the remaining missing hints (most of those files are unhinted anyway) and disallows partial hints in mypy.ini
```
disallow_untyped_defs = true
```


sobolevn:
@intgr can you please solve merge conflicts? And I can then start reviewing it :)

django-stubs/contrib/gis/gdal/layer.pyi
```diff
@@ -5,28 +5,28 @@ from django.contrib.gis.gdal.base import GDALBase as GDALBase
 class Layer(GDALBase):
     ptr: Any = ...
     def __init__(self, layer_ptr: Any, ds: Any) -> None: ...
-    def __getitem__(self, index: Any): ...
+    def __getitem__(self, index: Any) -> Any: ...
     def __iter__(self) -> Any: ...
-    def __len__(self): ...
+    def __len__(self) -> Any: ...
```
sobolevn, line 10: Technically it is always `int`

django-stubs/contrib/gis/gdal/raster/band.pyi
```diff
@@ -6,40 +6,40 @@ class GDALBand(GDALRasterBase):
...
-    ): ...
+    ) -> Any: ...
 
 class BandList(list):
     source: Any = ...
     def __init__(self, source: Any) -> None: ...
     def __iter__(self) -> Any: ...
-    def __len__(self): ...
-    def __getitem__(self, index: Any): ...
+    def __len__(self) -> Any: ...
```
sobolevn, line 44: And here

django-stubs/contrib/gis/geos/coordseq.pyi
```diff
@@ -6,27 +6,27 @@ class GEOSCoordSeq(GEOSBase):
     ptr_type: Any = ...
     def __init__(self, ptr: Any, z: bool = ...) -> None: ...
     def __iter__(self) -> Any: ...
-    def __len__(self): ...
-    def __getitem__(self, index: Any): ...
+    def __len__(self) -> Any: ...
```
sobolevn, line 9: `-> int`

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -181,16 +181,16 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
...
-    def __get__(self: _T, instance: None, owner) -> _T: ...
+    def __get__(self: _T, instance: None, owner: Any) -> _T: ...
     # Model instance access
     @overload
-    def __get__(self, instance: Model, owner) -> _GT: ...
+    def __get__(self, instance: Model, owner: Any) -> _GT: ...
     # non-Model instances
     @overload
-    def __get__(self: _T, instance, owner) -> _T: ...
+    def __get__(self: _T, instance: Any, owner: Any) -> _T: ...
```
intgr, line 193: I'm unsure what these `owner` fields are.

django-stubs/contrib/gis/geos/geometry.pyi
```diff
@@ -12,127 +12,127 @@ class GEOSGeometryBase(GEOSBase):
     has_cs: bool = ...
     __class__: Any = ...
     def __init__(self, ptr: Any, cls: Any) -> None: ...
-    def __copy__(self): ...
-    def __deepcopy__(self, memodict: Any): ...
+    def __copy__(self) -> Any: ...
```
sobolevn, line 15: Can be turned into `(self: _T) -> _T` in this PR or later (probably later, since we have more `__copy__` methods to fix)

And the same for `__deepcopy__`

django-stubs/contrib/gis/geos/point.pyi
```diff
@@ -8,21 +8,21 @@ class Point(GEOSGeometry):
         self, x: Optional[Any] = ..., y: Optional[Any] = ..., z: Optional[Any] = ..., srid: Optional[Any] = ...
     ) -> None: ...
     def __iter__(self) -> Any: ...
-    def __len__(self): ...
+    def __len__(self) -> Any: ...
```
sobolevn, line 11: `-> int`

django-stubs/contrib/gis/geos/polygon.pyi
```diff
@@ -5,15 +5,15 @@ from django.contrib.gis.geos.geometry import GEOSGeometry as GEOSGeometry
 class Polygon(GEOSGeometry):
     def __init__(self, *args: Any, **kwargs: Any) -> None: ...
     def __iter__(self) -> Any: ...
-    def __len__(self): ...
+    def __len__(self) -> Any: ...
```
sobolevn, line 8: And here

django-stubs/contrib/gis/measure.pyi
```diff
@@ -7,36 +7,36 @@ class MeasureBase:
...
+    def __add__(self, other: Any) -> Any: ...
+    def __iadd__(self, other: Any) -> Any: ...
+    def __sub__(self, other: Any) -> Any: ...
+    def __isub__(self, other: Any) -> Any: ...
+    def __mul__(self, other: Any) -> Any: ...
+    def __imul__(self, other: Any) -> Any: ...
+    def __rmul__(self, other: Any) -> Any: ...
+    def __truediv__(self, other: Any) -> Any: ...
+    def __itruediv__(self, other: Any) -> Any: ...
+    def __bool__(self) -> Any: ...
```
sobolevn, line 22: `-> bool`

django-stubs/contrib/gis/measure.pyi
```diff
@@ -7,36 +7,36 @@ class MeasureBase:
...
-    def __mul__(self, other: Any): ...
+    def __mul__(self, other: Any) -> Any: ...
 
 class Area(MeasureBase):
     STANDARD_UNIT: Any = ...
     UNITS: Any = ...
     ALIAS: Any = ...
     LALIAS: Any = ...
-    def __truediv__(self, other: Any): ...
+    def __truediv__(self, other: Any) -> Any: ...
```
sobolevn, line 39: Can be removed

mypy.ini
```diff
@@ -12,6 +12,8 @@ warn_unused_ignores = True
 warn_redundant_casts = True
 warn_unused_configs = True
 warn_unreachable = True
+disallow_untyped_defs = true
+disallow_incomplete_defs = true
```
intgr, line 16: This makes lots of tests fail because test code in `*.yml` is missing annotations

```
-   case: transaction_atomic_decorator
    main: |
            ...
            @transaction.atomic(savepoint=True)
            def atomic_method2(self):
```

I can add missing hints to test functions too.

intgr:
>     def __len__(self) -> Any: ...
> Technically it is always int

The idea here was to fill in all missing hints so that we could increase the strictness in `mypy.ini`, to make sure people don't forget to add hints to new arguments or return values in the future.

I'm not making anything worse because `Any` was implied already before for all these unhinted items.

I suppose I could do manual tweaks here and there, but it's increasing the scope of a change that's already enormous. Is that what we want?

sobolevn:
> I'm not making anything worse because Any 

Ok!

mypy.ini
sobolevn: How many functions are there? If you have time for this - it is very useful!
If you don't - we can for now use different settings for tests / code.

django-stubs/db/models/fields/__init__.pyi
flaeppe: I think it's `type[M] | None = None` (if not in pyi)

Where

```
M = TypeVar("M", bound=Model)
```

And I think it's not none when instance is not none.

Making this an `@overload` function

flaeppe: So in this specific overload both any's should be `: None` I think

adamchainz:
I'm okay with this, but after merging, can you open an issue or PR for `-> int` on all `__len__` methods? [autotyping](https://github.com/JelleZijlstra/autotyping) can also help with adding some type hints automatically.

mypy.ini
intgr: There aren't too many, but I'll submit a separate PR for clarity.

sobolevn:
is it ready to be reviewed? or should we merge tests first?

intgr:
It contains all changes from the other PR also. But other than that it should be ready.

sobolevn:
I've merged test changes, there are conflicts now 👀

intgr:
Rebased.

sobolevn (approved):
Thanks again! Maybe someone else will have some extra input.

django-stubs/db/models/fields/__init__.pyi
intgr: @flaeppe You're probably right but I'm wary of touching these since I don't totally understand what is happening here. Are you up for opening a follow-up PR?

flaeppe: I think it should be fine keeping it as `Any` since what's exposed as an API is `_T`. The `instance` and `owner` types would only be relevant for the implementation body, but this is in stubs so we won't be affected here.

The only time they'd come in to play and have an impact would be while/if subclassing `Field` and overriding it's descriptor methods. If we keep `Any`, we'll allow a child class to declare them as whatever type they deem fit, while doing e.g. `type[M] | None` we'll restrain it.

Though, in general I don't see any obvious use cases for overriding the descriptor for `Field`, but I might be wrong here. The normal Django use case would be to implement a separate field descriptor class (https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.Field.descriptor_class) to decide what goes in and out of a model field

adamchainz (approved):
LGTM as a step towards stricter Mypy config.

PR #1209, opened 2022-10-27
Add django.db.models.base.model_unpickle

### LINKED ISSUE #1207: django.db.models.base.model_unpickle missing
apologies for cutting out the template -- most of the fields weren't relevant to my issue

I'm not sure what the correct typing of this is (I'm not all that familiar with django) https://github.com/django/django/blob/de2c2127b66e77a034c01c81753c5c08e651a5b4/django/db/models/base.py#L2464-L2471

but any access of this is giving an `[attr-defined]` error in mypy:

```python
django.db.models.base.model_unpickle
```

```
error: Module has no attribute "model_unpickle"  [attr-defined]
```

adamchainz:
# I have made things!

This is an internal Django function for pickling, but easy to add the types for.

To doublecheck the arg types, I added a breakpoint and ran Django’s test suite:

```
diff --git a/django/db/models/base.py b/django/db/models/base.py
index 584db319da..7450e2fe7d 100644
--- a/django/db/models/base.py
+++ b/django/db/models/base.py
@@ -2468,6 +2468,7 @@ def model_unpickle(model_id):
     else:
         # Backwards compat - the model was cached directly in earlier versions.
         model = model_id
+    breakpoint()
     return model.__new__(model)
```

## Related issues

Fixes #1207.

### DIFF
--- a/django-stubs/db/models/base.pyi
+++ b/django-stubs/db/models/base.pyi
@@ -1,4 +1,4 @@
-from typing import Any, Collection, Dict, Iterable, List, Optional, Sequence, Set, Tuple, Type, TypeVar
+from typing import Any, Collection, Dict, Iterable, List, Optional, Sequence, Set, Tuple, Type, TypeVar, Union
 
 from django.core.checks.messages import CheckMessage
 from django.core.exceptions import MultipleObjectsReturned as BaseMultipleObjectsReturned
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
I trust you on this one, thank you!

PR #1211, opened 2022-10-27
Add BaseDetailView.object

### LINKED ISSUE #605: DetailView has no attribute "object"
Hi all. I have my DetailView with overrided get_context_data method. All work, but mypy return `categories\views\subcategory_detail_view.py:14: error: "SubcategoryDetailView" has no attribute "object"`.

If i change `self.object.contests.get_open_contests()` to `self.get_object().contests.get_open_contests()`, then mypy return `error: "Model" has no attribute "contests"`

How i can fix it ?
Also i have two question:

1. Why django-stubs doesn't see self.object in view
2. Why django-stubs doesn't see models relationsheep

My View:
```
from typing import Dict, Any, Type

from django.views.generic import DetailView

from categories.models import Subcategory


class SubcategoryDetailView(DetailView):
    model: Type[Subcategory] = Subcategory
    template_name: str = 'categories/subcategory_detail_view.html'

    def get_context_data(self, **kwargs: Any) -> Dict[str, Any]:
        context: Dict[str, Any] = super().get_context_data(**kwargs)
        context['open_contests'] = self.object.contests.get_open_contests()
        return context
```

My Model:
```
from typing import Union

from django.db import models
from django.db.models import ForeignKey
from django.db.models.expressions import Combinable

from categories.models import AbstractCategory, Category


class Subcategory(AbstractCategory):
    parent: ForeignKey[Union[Category, Combinable], Category] = models.ForeignKey(Category, on_delete=models.CASCADE, related_name='subcategories')

    class Meta(AbstractCategory.Meta):
        verbose_name_plural: str = 'Subcategories'

    def __str__(self) -> str:
        return f"{self.parent.name} -> {self.name}"

    @property
    def contests_count(self) -> int:
        return self.contests.filter(status='open').count()
```

adamchainz:
# I have made things!

Django source: https://github.com/django/django/blob/84206607d6bfd61e7f7a88b51163ffd4153e3b5a/django/views/generic/detail.py#L108

A bit weird it's not in `SingleObjectMixin`, which references it, but whatever, that needs fixing in Django.

## Related issues

Fixes #605.

### DIFF
--- a/django-stubs/views/generic/detail.pyi
+++ b/django-stubs/views/generic/detail.pyi
@@ -20,6 +20,7 @@ class SingleObjectMixin(Generic[_M], ContextMixin):
     def get_context_object_name(self, obj: _M) -> Optional[str]: ...
 
 class BaseDetailView(SingleObjectMixin[_M], View):
+    object: _M
     def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse: ...
 
 class SingleObjectTemplateResponseMixin(TemplateResponseMixin):


sobolevn (approved):
👍

PR #1213, opened 2022-10-28
Fixed `F`'s `asc` and `desc` args

g-as:
# I have made things!

Fixed `asc` and `desc` signature by removing `expression`.

I also updated `BaseExpression`'s `asc` and `desc` (they are the same).

```python

class OrderBy(Expression):
    ...

    def __init__(self, expression, descending=False, nulls_first=False, nulls_last=False):
        ...

    ...

@deconstructible
class F(Combinable):
    """An object capable of resolving references to existing query objects."""

    ...

    def asc(self, **kwargs):
        return OrderBy(self, **kwargs)

    def desc(self, **kwargs):
        return OrderBy(self, descending=True, **kwargs)
        
    ...
```


django-stubs/db/models/expressions.pyi
```diff
@@ -121,15 +132,13 @@ class F(Combinable):
     def asc(
```
sobolevn, line 132: Do we still need these method? They are duplicating the `BaseExpression` ones.
Or is it django-version specific?

g-as: `F` doesn't inherit from `BaseExpression` but from `class Combinable:`

sobolevn: Oh, I am sorry :)

g-as: No worries.

sobolevn: Source: https://github.com/django/django/blob/09397f5cfade93c2e79a391ded7ab6b01e51730e/django/db/models/expressions.py#L818-L819

sobolevn (approved):

sobolevn:
Thank you! 👍

g-as:
Cheers.

PR #1215, opened 2022-10-28
BaseDeleteView now extends FormMixin

### LINKED ISSUE #1214: DeleteView doesn't expose get_form_kwargs because its missing grandparent type FormMixin
# Bug report
## What's wrong
When extending Django's generic DeleteView, mypy doesn't realize `get_form_kwargs()` is defined by DeleteView -> BaseDeleteView -> FormMixin.

```python
from typing import Any
from django.views.generic.edit import DeleteView, FormMixin
from django.db import models

class MyDeleteView(DeleteView[models.Model]):

    def get_form_kwargs(self) -> dict[str, Any]:
        # Doesn't realize DeleteView -> BaseDeleteView -> FormMixin
        # error: Returning Any from function declared to return "Dict[str, Any]"
        return super().get_form_kwargs()
        # return super(FormMixin, self).get_form_kwargs()
        # Fixes highlighting but now we get:
        # error: Argument 2 for "super" not an instance of argument 1
```

## How is that should be

Django's source code shows that:
*  DeleteView inherits BaseDeleteView
* BaseDeleteView has the FormMixin which includes `get_form_kwargs()`

```python
# django/views/generic/edit.py
...
class BaseDeleteView(DeletionMixin, FormMixin, BaseDetailView):
...
```
I would guess the stubs would need to be changed to reflect this but not sure if this is what we are looking for:
```python
class BaseDeleteView(DeletionMixin[_M], FormView[_ModelFormT], BaseDetailView[_M]):
    object: _M
```
## System information

- OS: MacOS 10.15.7
- `python` version: Python 3.10.7
- `django` version: 4.1.2
- `mypy` version: 0.982

adamchainz:
# I have made things!

Fixed conflict version of #849.

Changed in Django 4.0: https://github.com/django/django/commit/3a45fea0832c5910acee6e0d29f230f347a50462

https://docs.djangoproject.com/en/4.0/releases/4.0/#generic-views

## Related issues

Fixes #1214.

### DIFF
--- a/django-stubs/views/generic/edit.pyi
+++ b/django-stubs/views/generic/edit.pyi
@@ -66,9 +66,9 @@ class DeletionMixin(Generic[_M]):
     def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse: ...
     def get_success_url(self) -> str: ...
 
-class BaseDeleteView(DeletionMixin[_M], BaseDetailView[_M]):
+class BaseDeleteView(DeletionMixin[_M], FormMixin[_ModelFormT], BaseDetailView[_M]):
     object: _M
 
-class DeleteView(SingleObjectTemplateResponseMixin, BaseDeleteView[_M]):
+class DeleteView(SingleObjectTemplateResponseMixin, BaseDeleteView[_M, _ModelFormT]):
     object: _M
     template_name_suffix: str = ...
--- a/django-stubs/views/generic/edit.pyi
+++ b/django-stubs/views/generic/edit.pyi
@@ -66,9 +66,9 @@ class DeletionMixin(Generic[_M]):
     def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse: ...
     def get_success_url(self) -> str: ...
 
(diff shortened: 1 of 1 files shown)


django-stubs/views/generic/edit.pyi
```diff
@@ -66,9 +66,9 @@ class DeletionMixin(Generic[_M]):
     def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse: ...
     def get_success_url(self) -> str: ...
 
-class BaseDeleteView(DeletionMixin[_M], BaseDetailView[_M]):
+class BaseDeleteView(DeletionMixin[_M], FormMixin[_ModelFormT], BaseDetailView[_M]):
```
sobolevn: It is better to explicitly add `Generic[_M, _ModelFormT]` when two vars are used this way.

Otherwise, it is easy to mix things up.

adamchainz: So, like the commit I just pushed?

sobolevn (approved):
Thank you!

django-stubs/views/generic/edit.pyi
sobolevn: Yes :)

PR #1219, opened 2022-10-31
Allow expressions in CheckConstraint.check

### LINKED ISSUE #867: check in CheckConstraint should support expression
# Bug report

## What's wrong

As of Django 3.1 ([#30484](https://code.djangoproject.com/ticket/30484)), the CheckConstraint supports `Q` and boolean `Expression`, but only `Q` is allowed in current type stubs.

By following it is the code snippets from `tests/constraints/models.py` that will throw typing error:
```python
            models.CheckConstraint(
                check=models.expressions.RawSQL(
                    "price < %s", (1000,), output_field=models.BooleanField()
                ),
                name="%(app_label)s_price_lt_1000_raw",
            ),
            models.CheckConstraint(
                check=models.expressions.ExpressionWrapper(
                    models.Q(price__gt=500) | models.Q(price__lt=500),
                    output_field=models.BooleanField(),
                ),
                name="%(app_label)s_price_neq_500_wrap",
            ),
```

## How is that should be

Maybe like #806, make `check` accepts `Union[Q, Combinable]`.

## System information

- OS: Debian 11
- `python` version: 3.9
- `django` version: 3.2
- `mypy` version: 0.931

adamchainz:
# I have made things!

Add support for `check` being any boolean expression. Not *all* `BaseExpression`s are allowed - only those where `conditional` is `True`: https://github.com/django/django/blob/c01e76c95caa1bd2995f1b6c186cbe8726c92ef9/django/db/models/constraints.py#L65-L68 . But we don’t have generic type hint support for that, so we can allow all `BaseExpression`s.

## Related issues

Fixes #867. @littlebtc would you like to review?


sobolevn:
I will leave some time for the mentioned people to review :)

intgr (approved):
In an effort to work through the backlog of unmerged PRs, I have converted this to new syntax. We should merge this.

adamchainz:
LGTM thanks for updating.

PR #1220, opened 2022-10-31
Re-allow expressions in UniqueConstraint

adamchainz:
# I have made things!

Re-do of #806 which was unfortunately reverted in the mega PR #909. I’m not sure why it was undone there.

Using `@overload` because `expressions` and `fields` are mutually exclusive: https://github.com/django/django/blob/c01e76c95caa1bd2995f1b6c186cbe8726c92ef9/django/db/models/constraints.py#L147-L150 . Added tests for that behaviour.

## Related issues

@knyghty would you like to check? And @linusg, since you spotted the reversion?

### DIFF
--- a/tests/typecheck/db/models/test_constraints.yml
+++ b/tests/typecheck/db/models/test_constraints.yml
@@ -1,0 +1,37 @@
+-   case: unique_constraint_expressions
+    main: |
+        from django.db.models import Q, UniqueConstraint
+        from django.db.models.functions import Lower
+
+        UniqueConstraint(
+            Lower('name').desc(),
+            'category',
+            name='unique_lower_name_category',
+        )
+
+-   case: unique_constraint_fields
+    main: |
+        from django.db.models import Q, UniqueConstraint
+        from django.db.models.functions import Lower
+
+        UniqueConstraint(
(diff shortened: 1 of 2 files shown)


tests/typecheck/db/models/test_constraints.yml
```diff
@@ -0,0 +1,37 @@
...
+        UniqueConstraint(
+            Lower('name'),
+            fields=['name'],
+            name='unique_mess',
+        )
+    regex: true
+    out: |
+        main:4: error: No overload variant of "UniqueConstraint" matches argument types "Lower", "List\[str\]", "str"
+        main:4: note: Possible overload variants:
+        main:4: note:     .*
```
sobolevn, line 36: Nice usage of regex! 👍

sobolevn (approved):
Thanks!

tests/typecheck/db/models/test_constraints.yml
adamchainz: Nice plugin, writing all the variants would have been unnecessarily verbose!

linusg (approved):
Thanks (as well :smile:)!

PR #1222, opened 2022-11-03
Enable `pre-commit.ci`

sobolevn:
Testing `pre-commit.ci`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,3 +33,11 @@ repos:
     rev: 3.9.2
     hooks:
       - id: flake8
+
+ci:
+    autofix_commit_msg: '[pre-commit.ci] auto fixes from pre-commit.com hooks'
+    autofix_prs: true
+    autoupdate_commit_msg: '[pre-commit.ci] pre-commit autoupdate'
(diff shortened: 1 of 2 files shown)


sobolevn:
This commit https://github.com/typeddjango/django-stubs/pull/1222/commits/f03bff8487905aa049ba4b024820b08def9ea6e6 contains `black` violation.

We should get a new commit with the fix.

sobolevn:
And here we go: https://github.com/typeddjango/django-stubs/pull/1222/commits/2306d98611764a322da82bc1ec6930fc4649c61e 🎉

adamchainz:
Sick!

adamchainz:
Why skip flake8?

sobolevn:
I think `flake8` has no autofixes. 
And since we check `flake8` in `lint` job, there's no need to duplicate it.

adamchainz:
pre-commit.ci isn’t only for tools with autofixes. It presents the output of linters like flake8 in its UI. I think we don’t need to run pre-commit twice, we can just use pre-commit.ci for everything. That’s what I do on my projects. Opened #1250 to do so.

PR #1223, opened 2022-11-03
Run checks against Django 4.1 and Python 3.11

michael-k:
- Run the `mypy-self-check` and `typecheck` jobs on GitHub Actions against Python 3.11
- Run the `typecheck` job on GitHub Actions against Django 4.1
- Fix compatibility of enums with Python 3.11
- Add stubs for `PY310` and `PY311`
- Add trove classifiers (django-stubs already has one for Django 4.1 from typeddjango/django-stubs#1167)

## Related issues

n/a

### DIFF
--- a/django-stubs/db/models/enums.pyi
+++ b/django-stubs/db/models/enums.pyi
@@ -1,5 +1,11 @@
 import enum
+import sys
 from typing import Any, List, Tuple
+
+if sys.version_info >= (3, 11):
+    enum_property = enum.property
+else:
+    enum_property = property
 
 class ChoicesMeta(enum.EnumMeta):
     names: List[str] = ...
@@ -12,7 +18,7 @@ class Choices(enum.Enum, metaclass=ChoicesMeta):
     def __str__(self) -> str: ...
     @property
     def label(self) -> str: ...
-    @property
+    @enum_property
     def value(self) -> Any: ...
 
 # fake
@@ -23,7 +29,7 @@ class _IntegerChoicesMeta(ChoicesMeta):
     values: List[int] = ...
 
 class IntegerChoices(int, Choices, metaclass=_IntegerChoicesMeta):
-    @property
+    @enum_property
     def value(self) -> int: ...
 
 # fake
@@ -34,5 +40,5 @@ class _TextChoicesMeta(ChoicesMeta):
     values: List[str] = ...
 
 class TextChoices(str, Choices, metaclass=_TextChoicesMeta):
-    @property
+    @enum_property
     def value(self) -> str: ...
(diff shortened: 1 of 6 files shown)


scripts/typecheck_tests.py
```diff
@@ -15,6 +15,7 @@
     "2.2": "44e7cca62382f2535ed0f5d2842b433f0bd23a57",
     "3.2": "0153a63a674937e4a56d9d5e4ca2d629b011fbde",
     "4.0": "67d0c4644acfd7707be4a31e8976f865509b09ac",
+    "4.1": "7dfd29b84e5a27d61ae62cb5ed9122d5a99ee1be",
```
michael-k, line 18: 4.1.3 because it's the first release that supports Python 3.11.  I can point that to 4.0 if required.
https://github.com/django/django/commit/7dfd29b84e5a27d61ae62cb5ed9122d5a99ee1be

sobolevn:
Technically 3.11 is not fully supported by the current mypy.
But, I think we are not going to use `except*` anytime soon.

django-stubs/db/models/enums.pyi
```diff
@@ -34,5 +40,5 @@ class _TextChoicesMeta(ChoicesMeta):
     values: List[str] = ...
 
 class TextChoices(str, Choices, metaclass=_TextChoicesMeta):
-    @property
+    @enum_property
```
sobolevn, line 43: This won't be supported by `mypy`.

michael-k: Without it the `mypy-self-check` job failed with

```
> Run mypy --cache-dir=/dev/null --no-incremental django-stubs
django-stubs/db/models/enums.pyi:16: error: Signature of "value" incompatible with supertype "Enum"
django-stubs/db/models/enums.pyi:27: error: Signature of "value" incompatible with supertype "Enum"
django-stubs/db/models/enums.pyi:38: error: Signature of "value" incompatible with supertype "Enum"
Found 3 errors in 1 file (checked 675 source files)
Error: Process completed with exit code 1.
```

See https://github.com/michael-k/django-stubs/actions/runs/3385444913/jobs/5623591875

intgr: Yes I think this is necessary. The same errors also appear in https://github.com/typeddjango/django-stubs/pull/1223

Maybe add a test with `reveal_type()` on the `value` property just in case?

adamchainz: > The same errors also appear in #1223

That is this PR, did you mean to link to a different one?

intgr: Ah yes, I meant https://github.com/typeddjango/djangorestframework-stubs/pull/284

sobolevn (approved):
Thanks, let's try it!

PR #1226, opened 2022-11-03
Cancel old pipelines

sobolevn:
Closes #1225


sobolevn:
Yes, old build are canceled.

adamchainz:
Good work.

PR #1228, opened 2022-11-03
Add LANGUAGE_COOKIE_* settings

andersk:
https://docs.djangoproject.com/en/4.1/ref/settings/#language-cookie-httponly
https://docs.djangoproject.com/en/4.1/ref/settings/#language-cookie-samesite


django-stubs/conf/global_settings.pyi
```diff
@@ -61,6 +61,8 @@ LANGUAGE_COOKIE_NAME: str = ...
 LANGUAGE_COOKIE_AGE: Optional[int] = ...
 LANGUAGE_COOKIE_DOMAIN: Optional[str] = ...
 LANGUAGE_COOKIE_PATH: str = ...
+LANGUAGE_COOKIE_HTTPONLY: bool = ...
+LANGUAGE_COOKIE_SAMESITE: Optional[Literal["Lax", "Strict", "None", False]] = ...
```
adamchainz: Why make the samesite settings `Optional`?

`None` is not documented as an allowed value:

https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-SESSION_COOKIE_SAMESITE

andersk: `None` is the default for `LANGUAGE_COOKIE_SAMESITE`. The rest is for consistency. Django itself gives the same meaning to `False` and `None` (not to be confused with `"None"`), as far as I can tell.

flaeppe: The previous change I made for some of these was to remove `Optional`, to align with the documentation: https://github.com/typeddjango/django-stubs/pull/1110

I'm completely guessing right now, but might `None` be due to backwards compatibility?

It feels like what is documented should take precedence here. At least I would like to get an error running Mypy if I'm using `None` and instead should use e.g. `False` to disable, which the docs state.

adamchainz: Yeah I also think we should go with what is documented, as that’s all Django promises to support. Yes `None` happens to work, and that’s unlikely to change, but `False` is all that is guaranteed.

django-stubs/conf/global_settings.pyi
```diff
@@ -61,6 +61,8 @@ LANGUAGE_COOKIE_NAME: str = ...
 LANGUAGE_COOKIE_AGE: Optional[int] = ...
 LANGUAGE_COOKIE_DOMAIN: Optional[str] = ...
 LANGUAGE_COOKIE_PATH: str = ...
+LANGUAGE_COOKIE_HTTPONLY: bool = ...
+LANGUAGE_COOKIE_SAMESITE: Literal["Lax", "Strict", "None", False] = ...
```
andersk, line 65: This is definitely wrong—the default value *is* `None` ([documentation](https://docs.djangoproject.com/en/4.1/ref/settings/#language-cookie-samesite), [code](https://github.com/django/django/blob/4.1.3/django/conf/global_settings.py#L171)), so you *can* read `settings.LANGUAGE_COOKIE_SAMESITE` and get `None`, so this annotation breaks type safety.

andersk: - #1246

sshishov:
`LANGUAGE_COOKIE_SECURE` is missing, guys

sobolevn:
PR is welcome!

PR #1229, opened 2022-11-03
Transform user_passes_test signature

### LINKED ISSUE #1058: `user_passes_test` should take in `AUTH_USER_MODEL`
Starting with #1038 `user_passes_test` is typed to accept `AbstractBaseUser | AnonymousUser`. While technically correct it would be great if the `mypy` plugin could update this to be the actual model defined in settings. For example we have code like this that now complains:

```python
@user_passes_test(lambda u: u.is_superuser)
def my_view():
    ...
```

ljodal:
Add a transformer that updates the signature of `user_passes_test` to reflect `settings.AUTH_USER_MODEL`.

## Related issues

Fixes #1058

### DIFF
--- a/mypy_django_plugin/transformers/auth.py
+++ b/mypy_django_plugin/transformers/auth.py
@@ -1,0 +1,40 @@
+from mypy.plugin import FunctionSigContext
+from mypy.types import CallableType, FunctionLike, Instance, UnionType
+
+from mypy_django_plugin.django.context import DjangoContext
+from mypy_django_plugin.lib import helpers
+
+
+def transform_user_passes_test(ctx: FunctionSigContext, django_context: DjangoContext) -> FunctionLike:
+    """
+    Update the signature of user_passes_test to reflect settings.AUTH_USER_MODEL
+    """
+
+    auth_user_model = django_context.settings.AUTH_USER_MODEL
+    try:
+        model_cls = django_context.apps_registry.get_model(auth_user_model)
+    except LookupError:
+        return ctx.default_signature
+    model_cls_fullname = helpers.get_class_fullname(model_cls)
+    user_model_info = helpers.lookup_fully_qualified_typeinfo(helpers.get_typechecker_api(ctx), model_cls_fullname)
+    if user_model_info is None:
+        return ctx.default_signature
+
+    if not ctx.default_signature.arg_types or not isinstance(ctx.default_signature.arg_types[0], CallableType):
+        return ctx.default_signature
+
+    test_func_type = ctx.default_signature.arg_types[0]
+
+    if not test_func_type.arg_types or not isinstance(test_func_type.arg_types[0], UnionType):
+        return ctx.default_signature
+    union = test_func_type.arg_types[0]
+
+    new_union = UnionType([Instance(user_model_info, [])] + union.items[1:])
+
+    new_test_func_type = test_func_type.copy_modified(
+        arg_types=[new_union] + test_func_type.arg_types[1:],
+    )
+
+    return ctx.default_signature.copy_modified(
+        arg_types=[new_test_func_type] + ctx.default_signature.arg_types[1:],
+    )
--- a/tests/typecheck/contrib/auth/test_decorators.yml
+++ b/tests/typecheck/contrib/auth/test_decorators.yml
@@ -39,12 +39,18 @@
     main: |
         from typing import Union
         from django.contrib.auth.decorators import user_passes_test
-        from django.contrib.auth.models import AnonymousUser
+        from django.contrib.auth.models import AnonymousUser, AbstractBaseUser
         from django.http import HttpRequest, HttpResponse
         from users.models import User
-        def check_user(user: Union[User, AnonymousUser]) -> bool: ...
-        @user_passes_test(check_user)
-        def view_func(request: HttpRequest) -> HttpResponse: ...
+
+        def check_user1(user: Union[User, AnonymousUser]) -> bool: ...
+        user_passes_test(check_user1)
+        def check_user2(user: User) -> bool: ...
+        user_passes_test(check_user2) # E: Argument 1 to "user_passes_test" has incompatible type "Callable[[User], bool]"; expected "Callable[[Union[User, AnonymousUser]], bool]"
+        def check_user3(user: Union[AnonymousUser, User]) -> bool: ...
+        user_passes_test(check_user3)
+        def check_user4(user: Union[AnonymousUser, AbstractBaseUser]) -> bool: ...
+        user_passes_test(check_user4)
     custom_settings: |
         AUTH_USER_MODEL = "users.User"
         INSTALLED_APPS = ("django.contrib.auth", "django.contrib.contenttypes", "users")
--- a/tests/typecheck/contrib/auth/test_decorators.yml
+++ b/tests/typecheck/contrib/auth/test_decorators.yml
@@ -35,6 +35,28 @@
         @user_passes_test(lambda u: u.get_username().startswith('super'))
         def view_func(request: HttpRequest) -> HttpResponse: ...
         reveal_type(view_func)  # N: Revealed type is "def (request: django.http.request.HttpRequest) -> django.http.response.HttpResponse"
+-   case: user_passes_test_custom_user_model
+    main: |
+        from typing import Union
+        from django.contrib.auth.decorators import user_passes_test
+        from django.contrib.auth.models import AnonymousUser
+        from django.http import HttpRequest, HttpResponse
+        from users.models import User
+        def check_user(user: Union[User, AnonymousUser]) -> bool: ...
+        @user_passes_test(check_user)
+        def view_func(request: HttpRequest) -> HttpResponse: ...
+    custom_settings: |
+        AUTH_USER_MODEL = "users.User"
+        INSTALLED_APPS = ("django.contrib.auth", "django.contrib.contenttypes", "users")
+    files:
+        -   path: users/__init__.py
+        -   path: users/models.py
+            content: |
+                from django.contrib.auth.models import AbstractBaseUser
+                from django.db import models
+                class User(AbstractBaseUser):
+                    email = models.EmailField(unique=True)
+                    USERNAME_FIELD = "email"
 -   case: user_passes_test_bare_is_error
     main: |
         from django.http import HttpRequest, HttpResponse
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -29,6 +29,7 @@ from mypy.plugin import (
     ClassDefContext,
     DynamicClassDefContext,
     FunctionContext,
+    FunctionSigContext,
     MethodContext,
     SemanticAnalyzerPluginInterface,
 )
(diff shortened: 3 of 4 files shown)


tests/typecheck/contrib/auth/test_decorators.yml
```diff
@@ -35,6 +35,28 @@
...
+-   case: user_passes_test_custom_user_model
+    main: |
+        from typing import Union
+        from django.contrib.auth.decorators import user_passes_test
+        from django.contrib.auth.models import AnonymousUser
+        from django.http import HttpRequest, HttpResponse
+        from users.models import User
+        def check_user(user: Union[User, AnonymousUser]) -> bool: ...
+        @user_passes_test(check_user)
+        def view_func(request: HttpRequest) -> HttpResponse: ...
```
sobolevn: You need to assert types here, current case is not enough.

What do you expect to happen?

ljodal: Without the changes in the PR mypy will produce a type error on line 46 (because `check_user` doesn't match `Callable[[AbstratBaseUser|AnonymousUser], bool]`), so the test fails. With this PR the signature of `user_passes_test` is changed so there's no type errors and thus the test passes.

tests/typecheck/contrib/auth/test_decorators.yml
```diff
@@ -35,6 +35,28 @@
...
         def view_func(request: HttpRequest) -> HttpResponse: ...
         reveal_type(view_func)  # N: Revealed type is "def (request: django.http.request.HttpRequest) -> django.http.response.HttpResponse"
+-   case: user_passes_test_custom_user_model
+    main: |
+        from typing import Union
+        from django.contrib.auth.decorators import user_passes_test
+        from django.contrib.auth.models import AnonymousUser
+        from django.http import HttpRequest, HttpResponse
+        from users.models import User
+        def check_user(user: Union[User, AnonymousUser]) -> bool: ...
```
sobolevn: What about `def check_user2(user: User) -> bool: ...` ?

sobolevn: And `def check_user(user: Union[ AnonymousUser, User]) -> bool: ...`

ljodal: The former would be an error, because it doesn't accept `AnonymousUser`, while the latter is perfectly valid. Note that this change only changes the signature of `user_passes_test`, all the type checking is done by mypy. So basically the signature is changed from:

```python
def user_passes_test(test_func: Callable[[AbstractBaseUser, AnonymousUser], bool]], ...):
```

to:
```python
def user_passes_test(test_func: Callable[[settings.AUTH_USER_MODEL, AnonymousUser], bool]], ...):
```

So anything valid matching the first signature would still match the second one, but the second one allows writing more specific types (using your configured user model).

I added a few more tests cases to demonstrate this.

intgr:
This seems like significant complexity to dynamically update the signature of just one function.

There are lots more places in Django that depend on `AUTH_USER_MODEL`. Are you planning to extend this and add update the other places as well, and can this be implemented generically?

`request.user` seems like an important one for example.

ljodal:
> This seems like significant complexity to dynamically update the signature of just one function.
> 
> There are lots more places in Django that depend on `AUTH_USER_MODEL`. Are you planning to extend this and add update the other places as well, and can this be implemented generically?
> 
> `request.user` seems like an important one for example.

This is already done for `request.user`: https://github.com/typeddjango/django-stubs/blob/e88f942f861809d2b824ecc12b908c8c90bbf6d2/mypy_django_plugin/transformers/request.py#L10

Same with `get_user_model()`: https://github.com/typeddjango/django-stubs/blob/e88f942f861809d2b824ecc12b908c8c90bbf6d2/mypy_django_plugin/transformers/settings.py#L11

I'm not sure if there's an easy way to do this generically (I'm also unsure how many places this is needed). I've just fixed the only place where this was causing new type issues in our codebase

sobolevn:
I think instead we can make `AUTH_USER_MODEL` an actual type and use it everywhere 🤔

ljodal:
I'm not exactly sure if I'm following how you'd envision that working? Altering the type of `settings.AUTH_USER_MODEL` to be a type alias? Wouldn't that cause issues when directly accessing that setting and expecting a string?

I'm happy to do some cleanup here and add a helper to extract the 'current' user type (to reduce duplication), but I don't see how this could be done in another way.

sobolevn:
I think we can at least experiment to do something like this:
1. Import `settings.AUTH_USER_MODEL` model from runtime
2. Pass it through our regular plugin machinery
3. Redefine `settings.AUTH_USER_MODEL` to be an alias for that type

So, ideally we can change all (most) `User` references in stubs with `AUTH_USER_MODEL` type. Like this:

```python
def user_passes_test(func: Callable[[AUTH_USER_MODEL | AnonymousUser], bool]) -> ...: ...`
```

However, I don't say that:
1. It is a good idea, I am not completely sure how it can be implemented
2. That it will work at all

ljodal:
One immediate problem I see with that is that `settings.USER_AUTH_MODEL` should stay defined as a string, as someone might rely on that at runtime as well. I guess `get_user_model()` is the API to use to get the actual model. I don't think that would be accepted as a type though. Maybe we could add something like this somewhere, but I'm not sure where? And I'm not sure if it would even work?

```python
User = get_user_model()
```

sobolevn:
It can work because of `get_dynamic_class_hook()`, but I am not sure this will be accepted as a type.

Ok, we can try to populate `_USER_AUTH_MODEL` magic thing 🤔

ljodal:
Hmm, wouldn't that break usage of the stubs without the plugin? For this to work properly I'd think it has to be defined somewhere in the stubs at least?

mypy_django_plugin/transformers/auth.py
```diff
@@ -0,0 +1,40 @@
...
+        return ctx.default_signature
+
+    if not ctx.default_signature.arg_types or not isinstance(ctx.default_signature.arg_types[0], CallableType):
+        return ctx.default_signature
+
+    test_func_type = ctx.default_signature.arg_types[0]
+
+    if not test_func_type.arg_types or not isinstance(test_func_type.arg_types[0], UnionType):
+        return ctx.default_signature
+    union = test_func_type.arg_types[0]
```
flaeppe, line 30: It's not obvious why we're extracting a Union here. We should add a comment displaying why (context of us expecting `AbstractBaseUser | AnonymousUser` isn't anywhere near here)

```suggestion
    # Extract the Union argument for user out of the test function:
    # user: AbstractBaseUser | AnonymousUser
    if not test_func_type.arg_types or not isinstance(test_func_type.arg_types[0], UnionType):
        return ctx.default_signature
    union = test_func_type.arg_types[0]
```

mypy_django_plugin/transformers/auth.py
```diff
@@ -0,0 +1,40 @@
...
+
+    new_union = UnionType([Instance(user_model_info, [])] + union.items[1:])
```
flaeppe, line 32:
```suggestion
    # Replace AbstractBaseUser with the model type AUTH_USER_MODEL points to
    # user: <AUTH_USER_MODEL> | AnonymousUser
    new_union = UnionType([Instance(user_model_info, [])] + union.items[1:])
```

mypy_django_plugin/transformers/auth.py
```diff
@@ -0,0 +1,40 @@
...
+
+    new_test_func_type = test_func_type.copy_modified(
+        arg_types=[new_union] + test_func_type.arg_types[1:],
+    )
```
flaeppe, line 36:
```suggestion
    # Update the test function with the adjusted user annotation
    new_test_func_type = test_func_type.copy_modified(
        arg_types=[new_union] + test_func_type.arg_types[1:],
    )
```

mypy_django_plugin/transformers/auth.py
```diff
@@ -0,0 +1,40 @@
...
+
+    return ctx.default_signature.copy_modified(
+        arg_types=[new_test_func_type] + ctx.default_signature.arg_types[1:],
+    )
```
flaeppe, line 40:
```suggestion
    # Update user_passes_test function signature to expect the new test function signature as first argument
    return ctx.default_signature.copy_modified(
        arg_types=[new_test_func_type] + ctx.default_signature.arg_types[1:],
    )
```

mypy_django_plugin/transformers/auth.py
```diff
@@ -0,0 +1,40 @@
+from mypy.plugin import FunctionSigContext
+from mypy.types import CallableType, FunctionLike, Instance, UnionType
+
+from mypy_django_plugin.django.context import DjangoContext
+from mypy_django_plugin.lib import helpers
+
+
+def transform_user_passes_test(ctx: FunctionSigContext, django_context: DjangoContext) -> FunctionLike:
```
flaeppe, line 8: Can't we name this function a bit more specifically?

e.g.

```suggestion
def attach_AUTH_USER_MODEL_to_user_passes_test(ctx: FunctionSigContext, django_context: DjangoContext) -> FunctionLike:
```

flaeppe (commented):
I added a few comment suggestions to `transform_user_passes_test`.

Feel free to adjust them as you find most appropriate. All I'm finding important is to have some comments there to make intention and what is happening a bit more obvious.

flaeppe:
> It can work because of `get_dynamic_class_hook()`, but I am not sure this will be accepted as a type.
> 
> Ok, we can try to populate `_USER_AUTH_MODEL` magic thing 🤔

I got a spin on this idea:

`settings.AUTH_USER_MODEL` is Django utilising its lazy referencing (`"<app_label>.<object_name>"`), and we also know that it looks for a registered model under the hood. Which means that (probably currently) we've typed or can assume `type[Model]` wherever `settings.AUTH_USER_MODEL` has been resolved.

So, what if the stubs add some variable: `_AUTH_USER_MODEL = NewType("_AUTH_USER_MODEL", Type[Model])`? That might perhaps trigger `get_dynamic_class_hook` and give us a statically declared type which is "bound" to be correct but the plugin is able to refine it into a even more specific type?

Then the stubs replaces all places where we expect `settings.AUTH_USER_MODEL` to have been resolved e.g.

```diff
- -> Type[Model]:
+ -> _AUTH_USER_MODEL:
```

Which (hopefully) would align all types.

sobolevn:
I like this idea (I think that adding a hook is a must, however - otherwise, we will have too many errors in users' code).

ljodal:
Hmm, interesting. Sounds like it's worth a try! Maybe best to try in a separate PR though?

flaeppe:
Yeah a separate PR would be best

ljodal:
Are we sure we can't just use `get_auth_user()` together with `get_dynamic_class_hook()` though? Something like this?

```python
# stubs.pyi

User = get_auth_user()

class HttpRequest:
    user: User | AnonymousUser
```

sobolevn:
I think we should, because it is a perfect use-case for it.

flaeppe:
I don't think that works. Isn't it only TypeAlias that can be annotated with?

And that function call can't return a type alias, I've tried make the plugin spoof that kind of aliasing as well, in #1699

flaeppe:
Mypy makes `User` a variable(`Var`) and that can't be annotated with

ljodal:
But can't we change that using `get_dynamic_class_hook`?

flaeppe:
> But can't we change that using `get_dynamic_class_hook`?

Probably, I've tried that in #1699 but it isn't trivial

ljodal:
I think I managed to make it work: #1730

Main trick was replacing the symbol with a placeholder when deferring, as otherwise the variable error popped up (that seems like a bug in mypy)

PR #1233, opened 2022-11-06
Support more use cases of `gettext_lazy`.

### LINKED ISSUE #1231: Error when setting `gettext_lazy` as class attribute
# Bug report

## What's wrong

We received an update in cookiecutter-django to update to the latest django-stubs, but the type checks fail:

https://github.com/cookiecutter/cookiecutter-django/actions/runs/3385553002/jobs/5623835120#step:5:2492

 The reason is because the `success_message` attribute for `SuccessMessageMixin` should accept lazy string.

https://github.com/typeddjango/django-stubs/blob/8f97bf880d3022772581ea4cf8b5bf5297a27bad/django-stubs/contrib/messages/views.pyi#L6-L9

References:
- https://github.com/typeddjango/django-stubs/pull/689
- https://github.com/typeddjango/django-stubs/issues/1137

I assume there are a other mixin with the same problem. 

## Related error:

Note that the CI run above also have a similar error regarding the model admin `fieldsets` attribute (defined [here](https://github.com/cookiecutter/cookiecutter-django/blob/ca82618526d8f2424c721a4bfaeb8abc54b5b030/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/users/admin.py#L16-L32)) - I'm happy to open a separate issue about it, but maybe we should consider a overall fix?

## How is that should be

The type annotation for success_message should accept lazy strings too:

```python
class SuccessMessageMixin: 
     success_message: _StrOrPromise = ... 
```

## System information

- OS:
- `python` version: 3.10
- `django` version: 4.0
- `mypy` version: 0.982
- `django-stubs-ext` version: n/a (I don't think this package is installed?)

PIG208:
## Related issues

- Closes #1231


sobolevn (approved):
LGTM! Thank you

PR #1234, opened 2022-11-06
Prefetch attributes

g-as:
# I have made things!

I added `Prefetch` public attributes, based on \_\_init\_\_:
```
class Prefetch:
    def __init__(self, lookup, queryset=None, to_attr=None):
        # `prefetch_through` is the path we traverse to perform the prefetch.
        self.prefetch_through = lookup
        # `prefetch_to` is the path to the attribute that stores the result.
        self.prefetch_to = lookup
        if queryset is not None and (
            isinstance(queryset, RawQuerySet)
            or (
                hasattr(queryset, "_iterable_class")
                and not issubclass(queryset._iterable_class, ModelIterable)
            )
        ):
            raise ValueError(
                "Prefetch querysets cannot use raw(), values(), and values_list()."
            )
        if to_attr:
            self.prefetch_to = LOOKUP_SEP.join(
                lookup.split(LOOKUP_SEP)[:-1] + [to_attr]
            )

        self.queryset = queryset
        self.to_attr = to_attr

    ...
```
https://github.com/django/django/blob/stable/4.0.x/django/db/models/query.py#L1786-L1808

### DIFF
--- a/django-stubs/db/models/query.pyi
+++ b/django-stubs/db/models/query.pyi
@@ -214,6 +214,10 @@ _QuerySetAny = _QuerySet
 QuerySet = _QuerySet[_T, _T]
 
 class Prefetch:
+    prefetch_through: str
+    prefetch_to: str
+    queryset: Optional[QuerySet]
+    to_attr: Optional[str]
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks!

PR #1242, opened 2022-11-08
Add django.db.models.query.BaseIterable and subclasses

### LINKED ISSUE #1237: Module "django.db.models.query" has no attribute "ModelIterable"
# Bug report
## What's wrong
The following types are not defined in `django.db.models.query`:
* BaseIterable
* ModelIterable
* RawModelIterable
* ValuesIterable
* ValuesListIterable
* NamedValuesListIterable
* FlatValuesListIterable

## System information

- OS:
- `python` version: 3.11.0
- `django` version: 4.0.4
- `mypy` version: .982

CircuitCipher:
# I have made things!
This PR addresses #1237 and adds the following classes to `django.db.models.query`:

- BaseIterable
- ModelIterable
- RawModelIterable
- ValuesIterable
- ValuesListIterable
- NamedValuesListIterable
- FlatValuesListIterable

## Related issues
- Closes #1237

### DIFF
--- a/django-stubs/db/models/query.pyi
+++ b/django-stubs/db/models/query.pyi
@@ -31,7 +31,7 @@ from django.db.models.sql.query import Query, RawQuery
 _T = TypeVar("_T", bound=Model, covariant=True)
 _Row = TypeVar("_Row", covariant=True)
 _QS = TypeVar("_QS", bound="_QuerySet")
-_TupleT = TypeVar("_TupleT", bound=tuple[Any], covariant=True)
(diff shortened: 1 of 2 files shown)


adamchainz (commented):
Maybe we could add `QuerySet._iterable_class` at the same time? It seems to me that if you need to touch these classes, you probably want that attr defined.

Otherwise, LGTM

CircuitCipher:
> Maybe we could add `QuerySet._iterable_class` at the same time? It seems to me that if you need to touch these classes, you probably want that attr defined.

I'll see what I can do about it today

adamchainz (approved):

PR #1243, opened 2022-11-09
Improve some DB expression types

adamchainz:
# I have made things!

* Remove several things from `GeoAggregate` and `GeoFuncMixin` where the inherited types are correct
* Add missing kwargs `**extra_context` to `Func.as_sql`
* Remove `Agggregate.as_sql` since it’s a pass-through to `Func.as_sql`
* Reorganize `db/models/expressions.pyi` to follow the same order as upstream
* Add missing classes `DurationExpression`, `TemporalSubtraction`, `Star`, and `OrderByList`

## Related issues

n/a

Spotted the problem with `Aggregate` when adding Mypy to Django-MySQL, which has a custom aggregate class.

### DIFF
--- a/django-stubs/contrib/gis/db/models/aggregates.pyi
+++ b/django-stubs/contrib/gis/db/models/aggregates.pyi
@@ -1,7 +1,6 @@
 from typing import Any
 
 from django.db.models import Aggregate
-from django.db.models.sql.compiler import _AsSqlType
 
 class GeoAggregate(Aggregate):
     is_extent: bool
--- a/django-stubs/contrib/gis/db/models/aggregates.pyi
+++ b/django-stubs/contrib/gis/db/models/aggregates.pyi
@@ -4,22 +4,8 @@ from django.db.models import Aggregate
 from django.db.models.sql.compiler import _AsSqlType
 
 class GeoAggregate(Aggregate):
-    function: Any
     is_extent: bool
-    @property
-    def output_field(self) -> Any: ...
-    def as_sql(
-        self, compiler: Any, connection: Any, function: Any | None = ..., **extra_context: Any
-    ) -> _AsSqlType: ...
     def as_oracle(self, compiler: Any, connection: Any, **extra_context: Any) -> Any: ...
-    def resolve_expression(
-        self,
-        query: Any | None = ...,
-        allow_joins: bool = ...,
-        reuse: Any | None = ...,
-        summarize: bool = ...,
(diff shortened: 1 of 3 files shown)


django-stubs/db/models/expressions.pyi
```diff
@@ -211,21 +221,30 @@ class Case(Expression):
...
-class Col(Expression):
-    target: Field
-    alias: str
-    contains_column_references: Literal[True] = ...
-    possibly_multivalued: Literal[False] = ...
-    def __init__(self, alias: str, target: Field, output_field: Optional[Field] = ...) -> None: ...
+class Subquery(BaseExpression, Combinable):
+    template: str = ...
+    query: Query = ...
+    extra: Dict[Any, Any] = ...
```
intgr: Since it comes from kwargs, keys are always strings

```suggestion
    extra: Dict[str, Any] = ...
```

adamchainz: Well spotted, I was just moving this around but worth improving whilst we touch it.

intgr:
Sorry, lots of conflicts here due to #1253

I promise I will review this if you resolve the conflicts. :)

adamchainz:
Rebased! Tests failed due to an apt issue, fix in #1313.

django-stubs/contrib/gis/db/models/aggregates.pyi
```diff
@@ -4,22 +4,8 @@ from django.db.models import Aggregate
 from django.db.models.sql.compiler import _AsSqlType
```
intgr:
```suggestion
```

django-stubs/db/models/expressions.pyi
```diff
@@ -148,30 +154,24 @@ class OuterRef(F):
...
-    template: str
-    nulls_first: bool
-    nulls_last: bool
-    descending: bool
-    expression: Expression | F | Subquery
-    def __init__(
+    def __init__(self, *expressions: Any, output_field: Field | None = ..., **extra: Any) -> None: ...
+    def as_sql(
         self,
-        expression: Expression | F | Subquery,
```
intgr, line 170: Reordering the classes here made it **very** difficult to understand what actually changed here. I had to order them back to get a readable diff.

In the future, if you want to reorder items, please do it as a separate PR without any additional changes.

intgr (approved):
Looks good, one import can be cleaned up.

PR #1246, opened 2022-11-09
Add Optional to LANGUAGE_COOKIE_SAMESITE

andersk:
Reverse an incorrect edit that was squashed into my change in #1228. The default value of `LANGUAGE_COOKIE_SAMESITE` is `None` ([documentation](https://docs.djangoproject.com/en/4.1/ref/settings/#language-cookie-samesite), [code](https://github.com/django/django/blob/4.1.3/django/conf/global_settings.py#L171)), so an annotation that excludes `None` is not type-safe.


adamchainz (approved):
Yes, thanks, I spotted your comment after merge and have been meaning to come back to that. You're right that, for whatever reason, Django has `None` here, so we should match it.

PR #1247, opened 2022-11-10
CI: Enable testing with Python 3.11

intgr:
Also updated pre-commit & pytest to use Python 3.10.


intgr:
Oops, duplicate of #1223.

PR #1249, opened 2022-11-11
Improve system check registry hints

### LINKED ISSUE #1098: Using the django.core.checks.register makes function untyped
I'm experimenting with type annotations and chose a fairly simple project to add type annotations to (just a couple of additional Django checks). However I ran into the following:

## What's wrong

```python
from django.core.checks import register

@register("foo")
def check_foo_is_bar(app_configs: Apps, **kwargs: Any) -> List[Warning]:
    ...
```

Results in mypy complaining:

`Untyped decorator makes function "check_foo_is_bar" untyped`

## How it should be

The register decorator is currently annotated as returning `Any`. If somehow it can be annotated as being essentially the same as `CheckRegistry.register` all would be well (I think). Is that possible?

## System information

- `python` version: Python 3.8.12
- `django` version: 4.1
- `mypy` version: 0.961

## Additional information

It's possible to work around this by using the decorator as a function instead:

```python
def check_foo_is_bar(app_configs: Apps, **kwargs: Any) -> List[Warning]:
    ...

register(check_foo_is_bar, "foo")
```

But it would be nice if the decorator worked as well.

intgr:
The `@register` decorator from `django.core.checks` was previously untyped.

Fixes #1098.

### DIFF
--- a/django-stubs/core/checks/registry.pyi
+++ b/django-stubs/core/checks/registry.pyi
@@ -1,4 +1,5 @@
-from typing import Any, Callable, Iterable, Protocol, Sequence, TypeVar, overload
+from collections.abc import Callable, Iterable, Sequence
+from typing import Any, Protocol, TypeVar, overload
 
 from django.apps.config import AppConfig
 from django.apps.registry import Apps
@@ -48,7 +49,7 @@ class CheckRegistry:
     def tags_available(self, deployment_checks: bool = ...) -> set[str]: ...
     def get_checks(self, include_deployment_checks: bool = ...) -> list[_ProcessedCheckCallable]: ...
 
-registry: CheckRegistry
+registry: CheckRegistry = ...
 register = registry.register
 run_checks = registry.run_checks
 tag_exists = registry.tag_exists
--- a/django-stubs/core/checks/registry.pyi
+++ b/django-stubs/core/checks/registry.pyi
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Protocol, Sequence, TypeVar, overload
+from typing import Any, Callable, Iterable, Protocol, Sequence, TypeVar, overload
 
 from django.apps.config import AppConfig
 from django.apps.registry import Apps
@@ -19,7 +19,7 @@ class Tags:
     urls: str
 
 class _CheckCallable(Protocol):
-    def __call__(self, app_configs: Apps, **kwargs: Any) -> Sequence[CheckMessage]: ...
+    def __call__(self, *, app_configs: Apps, **kwargs: Any) -> Iterable[CheckMessage]: ...
 
 _C = TypeVar("_C", bound=_CheckCallable)
 
--- a/django-stubs/core/checks/registry.pyi
+++ b/django-stubs/core/checks/registry.pyi
@@ -19,7 +19,9 @@ class Tags:
     urls: str
 
 class _CheckCallable(Protocol):
-    def __call__(self, *, app_configs: Apps, **kwargs: Any) -> Iterable[CheckMessage]: ...
+    def __call__(
+        self, *, app_configs: Apps, databases: Sequence[str] | None, **kwargs: Any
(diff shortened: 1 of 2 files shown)


django-stubs/core/checks/registry.pyi
```diff
@@ -18,7 +19,9 @@ class Tags:
     translation: str = ...
     urls: str = ...
 
-_CheckCallable = Callable[..., Sequence[CheckMessage]]
+class _CheckCallable(Protocol):
+    def __call__(self, app_configs: Apps, **kwargs: Any) -> Sequence[CheckMessage]: ...
```
adamchainz: Perhaps this [should return an `Iterable`](https://github.com/django/django/blob/bccfaa23864a3960ec5562d2f95d1238f1348ee6/django/core/checks/registry.py#L89)?

And `app_configs` is kwarg-only, and we could add `databases`... (and test a `database`-using check)

```suggestion
    def __call__(self, *, app_configs: Apps, databases: Optional[Iterable[str]] **kwargs: Any) -> Iterable[CheckMessage]: ...
```

intgr: Sure.

As for `databases`, I think Sequence is better than Iterable. It has more methods in common with `list` and the documentation also specifies this to be a list.

intgr:
@sobolevn Any ideas why only "pre-commit.ci" checks are running, but not the remaining workflows?

sobolevn:
No :(

intgr:
After making a new push, the workflows are now running. I guess it was a temporary GitHub glitch.

intgr:
This should be ready for review. @adamchainz do you have time to look at this again?

intgr:
Ping, this has been waiting for 3 weeks. @adamchainz @sobolevn?

django-stubs/core/checks/registry.pyi
```diff
@@ -19,7 +19,10 @@ class Tags:
     translation: str
     urls: str
 
-_CheckCallable: TypeAlias = Callable[..., Sequence[CheckMessage]]
+class _CheckCallable(Protocol):
+    def __call__(
+        self, *, app_configs: Apps, databases: Sequence[str] | None, **kwargs: Any
```
adamchainz: `app_configs` has the wrong type - it's a list of `AppConfig`, or `None`.

See https://docs.djangoproject.com/en/4.1/topics/checks/#writing-your-own-checks

> The check function must accept an app_configs argument; this argument is the list of applications that should be inspected. If None, the check must be run on all installed apps in the project.

And an example check function in one of my packages: https://github.com/adamchainz/django-linear-migrations/blob/main/src/django_linear_migrations/apps.py#L129

tests/typecheck/core/test_checks.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    from django.core.checks import register, Warning, CheckMessage
+    from django.apps.registry import Apps
+
+
+    @register("foo", deploy=True)
+    def check_foo(app_configs: Apps, databases: Optional[Sequence[str]], **kwargs: Any) -> list[Warning]:
+        if databases and 'databass' in databases:
+            return [Warning("Naughty list")]
+        return []
```
adamchainz: most checks don't take a `databases` argument, so I'd like a test that checks that situation too

adamchainz (commented):
Some changes - writing and pushing a fix.

adamchainz:
Tests failing until #1313...

adamchainz (approved):
I think this is ready to go, but I'm not going to merge as I was the last to touch it.

sobolevn (approved):

sobolevn:
Thanks everyone!

jaap3:
Lovely stuff, thanks everyone!

django-stubs/core/checks/registry.pyi
intgr: Doh that was sloppy. Thanks for the fixes!

PR #1250, opened 2022-11-11
Run pre-commit only on pre-commit.ci

adamchainz:
# I have made things!

Follow-up to #1222. Only run pre-commit on pre-commit.ci, and run flake8 there too.

## Related issues

n/a


sobolevn (approved):
Yes, this looks more optimal. Thanks! TIL

PR #1251, opened 2022-11-12
fix: Use supported version in typecheck_tests command in Contributing guide

adamantike:
The `typecheck_tests.py` script only supports the Django versions included in its `DJANGO_COMMIT_REFS` map. Version `3.1` is no longer present, so the command from the Contributing guide failed.

For consistency, the Django version used for the `stubgen-django.py` commands has been also updated.


intgr (approved):
Sorry that this somehow was ignored for months!

PR #1252, opened 2022-11-12
only allow `redirect_chain` attribute in response when client sets the follow flag

hoefling:
Signed-off-by: Oleg Hoefling <oleg.hoefling@ionos.com>

# I have made things!

The `redirect_chain` attribute was added in #1124. At runtime, the response only has this attribute set when the request was made with `follow=True`:

```py
In [1]: from django.test import Client

In [2]: c = Client()

In [3]: c.get('/spam', follow=True).redirect_chain
Not Found: /spam
2022-11-12 15:42:28 [WARNING ] django.request :: Not Found: /spam
Out[3]: []

In [4]: c.get('/spam', follow=False).redirect_chain
Not Found: /spam
2022-11-12 15:42:33 [WARNING ] django.request :: Not Found: /spam
Input In [4], in <cell line: 1>()
----> 1 c.get('/spam', follow=False).redirect_chain

AttributeError: 'HttpResponseNotFound' object has no attribute 'redirect_chain'
```
This PR replicates this behaviour.

## Related issues

No issue


django-stubs/test/client.pyi
```diff
@@ -31,6 +33,11 @@ from django.template.base import Template
 from django.test.utils import ContextList
 from django.urls import ResolverMatch
 
+if sys.version_info >= (3, 8):
```
sobolevn: You can just use `typing_extensions`. It is the same thing.

django-stubs/test/client.pyi
```diff
@@ -159,36 +168,110 @@ class Client(ClientMixin, _RequestFactory[_MonkeyPatchedWSGIResponse]):
...
+    def get(
+        self, path: str, data: Any = ..., follow: Literal[True] = ..., secure: bool = ..., **extra: Any
+    ) -> _MonkeyPatchedWSGIResponseRedirect: ...
+    @overload  # type: ignore
+    def post(
+        self,
+        path: str,
+        data: Any = ...,
+        content_type: str = ...,
+        follow: Literal[False] = ...,
```
sobolevn: Can you please test this case: `x: bool; r.post(follow=x)`?

tests/typecheck/test/test_client.yml
```diff
@@ -9,8 +9,10 @@
       reveal_type(response.client)  # N: Revealed type is "django.test.client.Client"
       reveal_type(response.context)  # N: Revealed type is "Union[django.test.utils.ContextList, builtins.dict[builtins.str, Any]]"
       reveal_type(response.content)  # N: Revealed type is "builtins.bytes"
-      reveal_type(response.redirect_chain)  # N: Revealed type is "builtins.list[Tuple[builtins.str, builtins.int]]"
+      response.redirect_chain  # E: "_MonkeyPatchedWSGIResponse" has no attribute "redirect_chain"
       response.json()
+      redirected_response = client.get('foo', follow=True)
```
sobolevn: We also need a test for `follow=False`.

sobolevn (commented):
Thanks, looks like a good idea!

django-stubs/test/client.pyi
hoefling: @sobolevn good catch - I completely forgot about the non-literal case 👍

tests/typecheck/test/test_client.yml
hoefling: @sobolevn let me move the `follow` variants in a separate test case to not to clutter the `client_methods` too much. I can also test all combinations of different methods there.

PR #1253, opened 2022-11-13
Introduce `flake8-pyi` in `pre-commit` checks

### LINKED ISSUE #128: Setup flake8-pyi to lint stubs

hoefling:
# I have made things!

This PR adds `pre-commit` check using `flake8-pyi`. Most of the warnings are fixed alongside, warnings that conflict with the current `mypy` state (e.g. https://github.com/python/mypy/issues/11098, https://github.com/python/mypy/issues/12393 or https://github.com/python/mypy/issues/13437) are disabled with an explicit comment linking to the related `mypy` issue.

Warnings that were disabled globally:

* `Y021` - looks like docstrings are used here and there, esp. the one for `Field` stub is useful
* `Y024` - switching from `namedtuple` to `NamedTuple` can be done in an automated fashion, but fields will be typed with `Any`. Not sure if this will be of any help
* `Y043` - IMO renaming the aliases is more harm than good

## Related issues

Closes #128


.pre-commit-config.yaml
```diff
@@ -33,6 +33,10 @@ repos:
     rev: 5.0.4
     hooks:
       - id: flake8
+        additional_dependencies:
+        - flake8-pyi
+        types: []
+        files: ^.*.pyi?$
```
adamchainz, line 39: This has stopped flake8 from running on the repo’s Python files, so we need two configurations. Also, please pin the version of flake8-pyi so there are no surprise failures when a new version is released.

```suggestion
      - id: flake8
      - id: flake8
        additional_dependencies:
        - flake8-pyi==22.10.0
        types: []
        files: ^.*.pyi?$
```

adamchainz (commented):
Thank you for working on setting this up.

The PR is *way* too big to review or merge as-is. I would prefer to split this up to activate one or a two rules at a time.

Also I hope you didn’t manually rewrite for changes like PEP 585, [pyupgrade can do those](https://github.com/asottile/pyupgrade/#pep-585-typing-rewrites) automatically. That’s another tool we could add to pre-commit.

sobolevn (approved):
This looks great! I will merge this asap, so no merge-conflicts on this side will emerge.

We can later work on any further improvements.

Thank you so much, this is amazing! 🎉 👍 😊

sobolevn:
> The PR is way too big to review or merge as-is. I would prefer to split this up to activate one or a two rules at a time.

Sorry, @adamchainz I've missed this comment :(

sobolevn:
I personally like to merge big style changes as-is, because otherwise it is really hard to work with merge conflicts and other PRs.

This becomes a kind of a blocker to the whole process.

.pre-commit-config.yaml
hoefling: @adamchainz the glob includes `.py` files (note the trailing question mark), so the `flake8` hook will run on both `.py` and `.pyi` files. I can introduce a second hook if you want tho 😎

hoefling:
@adamchainz I'm not against splitting the PR in separate ones, but even fixing a single rule per PR will end in a huge diff (esp. the PEP 604/PEP 585 stuff). Do you want me to go with enabling rules per modules in flake8 config? This is surely doable, but will take time, as there's nothing I can automate on a short sight, will have to redo the scripts first.

Edit: too late, @sobolevn merged already 😎

tests/typecheck/test/test_client.yml
```diff
@@ -35,3 +34,76 @@
       async_factory = AsyncRequestFactory()
       async_request = async_factory.get('foo')
       reveal_type(async_request)  # N: Revealed type is "django.core.handlers.asgi.ASGIRequest"
+-   case: client_follow_flag
+    main: |
+      from django.test.client import Client
+      client = Client()
```
intgr, line 40: What's up with these test changes? Did you base this on some other branch that isn't `master`?

intgr:
Thanks! I think the pipe syntax is so much easier on the eyes than `Union[]` and `Optional[]`.

tests/typecheck/test/test_client.yml
hoefling: @intgr whoops you are right, my bad 🤦  this also closed #1252. Will prepare a PR that reverts the changes shortly

.pre-commit-config.yaml
adamchainz: Good point, missed that! No need for a second hook, I think, unless flake8-pyi rules will be problematic for plain python files.

intgr:
I notice you simplified `float | int` to `float` in many places.

Are we affected by this now? https://github.com/python/typeshed/pull/9130

> Since mypy 0.990 type promotions was limited.
> This means that complex is not longer promoted to int/float, therefore we should adapt the types to list all possible types

sobolevn:
Yes, we are. `float | int` and `bytes | bytearray | memoryview`

intgr:
Should we revert this part of the PR?

Maybe hoeflint can re-run the script on an earlier commit, with this particular change disabled. And then submit a PR with the diffs.

sobolevn:
Yes, we should 👍

hoefling:
Yep, can revert that. IIRC there were only `float | int` and `complex | int` cases here and there.

django-stubs/utils/functional.pyi
```diff
@@ -83,17 +85,16 @@ class SimpleLazyObject(LazyObject):
...
 
 class classproperty(Generic[_Get]):
-    fget: Optional[Callable[[_Self], _Get]] = ...
-    def __init__(self, method: Optional[Callable[[_Self], _Get]] = ...) -> None: ...
-    def __get__(self, instance: Optional[_Self], cls: Type[_Self] = ...) -> _Get: ...
-    def getter(self, method: Callable[[_Self], _Get]) -> classproperty[_Get]: ...
+    fget: Callable[[Self], _Get] | None
+    def __init__(self: Self, method: Callable[[Self], _Get] | None = ...) -> None: ...
+    def __get__(self: Self, instance: Self | None, cls: type[Self] = ...) -> _Get: ...
+    def getter(self: Self, method: Callable[[Self], _Get]) -> classproperty[_Get]: ...
```
intgr, line 97: These `self: Self` annotations seem to cause regression #1285, I have submitted a PR to fix it in #1287.

PR #1254, opened 2022-11-13
Pin flake8-pyi to 22.10.0

hoefling:
# I have made things!

Addressing the suggestion from @adamchainz made [here](https://github.com/typeddjango/django-stubs/pull/1253#discussion_r1020937212).

## Related issues

Refs #128


adamchainz (approved):

adamchainz:
Thanks for following up!

PR #1257, opened 2022-11-14
only allow `redirect_chain` attribute in response when client sets the follow flag

hoefling:
Signed-off-by: Oleg Hoefling <oleg.hoefling@ionos.com>

# I have made things!

This is #1252, reintroduced. Description copied for the sake of completeness:

The `redirect_chain` attribute was added in #1124. At runtime, the response only has this attribute set when the request was made with `follow=True`:

```py
In [1]: from django.test import Client

In [2]: c = Client()

In [3]: c.get('/spam', follow=True).redirect_chain
Not Found: /spam
2022-11-12 15:42:28 [WARNING ] django.request :: Not Found: /spam
Out[3]: []

In [4]: c.get('/spam', follow=False).redirect_chain
Not Found: /spam
2022-11-12 15:42:33 [WARNING ] django.request :: Not Found: /spam
Input In [4], in <cell line: 1>()
----> 1 c.get('/spam', follow=False).redirect_chain

AttributeError: 'HttpResponseNotFound' object has no attribute 'redirect_chain'
```
This PR replicates this behaviour.

## Related issues

Refs #1252

### DIFF
--- a/django-stubs/test/client.pyi
+++ b/django-stubs/test/client.pyi
@@ -131,7 +131,7 @@ class Client(ClientMixin, _RequestFactory[_MonkeyPatchedWSGIResponse]):
     ) -> None: ...
     # Silence type warnings, since this class overrides arguments and return types in an unsafe manner.
     def request(self, **request: Any) -> _MonkeyPatchedWSGIResponse: ...
-    @overload  # type: ignore
+    @overload  # type: ignore[override]
     def get(
         self, path: str, data: Any = ..., follow: Literal[False] = ..., secure: bool = ..., **extra: Any
     ) -> _MonkeyPatchedWSGIResponse: ...
@@ -143,7 +143,7 @@ class Client(ClientMixin, _RequestFactory[_MonkeyPatchedWSGIResponse]):
     def get(
         self, path: str, data: Any = ..., follow: bool = ..., secure: bool = ..., **extra: Any
     ) -> _MonkeyPatchedWSGIResponse: ...
-    @overload  # type: ignore
+    @overload  # type: ignore[override]
     def post(
         self,
         path: str,
@@ -167,7 +167,7 @@ class Client(ClientMixin, _RequestFactory[_MonkeyPatchedWSGIResponse]):
     def post(
         self, path: str, data: Any = ..., content_type: str = ..., follow: bool = ..., secure: bool = ..., **extra: Any
     ) -> _MonkeyPatchedWSGIResponse: ...
-    @overload  # type: ignore
+    @overload  # type: ignore[override]
     def head(
         self, path: str, data: Any = ..., follow: Literal[False] = ..., secure: bool = ..., **extra: Any
     ) -> _MonkeyPatchedWSGIResponse: ...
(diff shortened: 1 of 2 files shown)


django-stubs/test/client.pyi
```diff
@@ -128,25 +131,136 @@ class Client(ClientMixin, _RequestFactory[_MonkeyPatchedWSGIResponse]):
     ) -> None: ...
     # Silence type warnings, since this class overrides arguments and return types in an unsafe manner.
     def request(self, **request: Any) -> _MonkeyPatchedWSGIResponse: ...
-    def get(  # type: ignore
+    @overload  # type: ignore
```
adamchainz: whilst you're modifying this, can use the specific code?

django-stubs/test/client.pyi
```diff
@@ -128,25 +131,136 @@ class Client(ClientMixin, _RequestFactory[_MonkeyPatchedWSGIResponse]):
...
+    def get(
```
sobolevn, line 135: I think we can remove this overload, it is the same as `bool`

sobolevn (commented):
The same is for all other overloads.

Do you agree?

django-stubs/test/client.pyi
hoefling: @sobolevn this will cause a false positive tho for the default value, e.g. `Client().get('foo').response_chain` will be inferred as existing since the `follow: Literal[True]` case matches and I can't put it behind the broader `follow: bool` case. The `client_follow_flag` test I added should also fail when the first override is deleted.

django-stubs/test/client.pyi
hoefling: @adamchainz done, thank you for pointing this out!

adamchainz: Thanks, I’d like us to require specific codes throughout at some point.

tests/typecheck/test/test_client.yml
```diff
@@ -35,3 +34,76 @@
...
+      response.redirect_chain  # E: "_MonkeyPatchedWSGIResponse" has no attribute "redirect_chain"
+
+      response = client.delete('foo')
+      response.redirect_chain  # E: "_MonkeyPatchedWSGIResponse" has no attribute "redirect_chain"
+      response = client.delete('foo', follow=False)
+      response.redirect_chain  # E: "_MonkeyPatchedWSGIResponse" has no attribute "redirect_chain"
+      response = client.delete('foo', follow=True)
+      reveal_type(response.redirect_chain)  # N: Revealed type is "builtins.list[Tuple[builtins.str, builtins.int]]"
+      x: bool
+      response = client.delete('foo', follow=x)
```
sobolevn, line 108: Maybe we should allow `bool`? I think the main usecase for this is in unit-tests. They will be checked in all cases, but we can just make things harder for users by producing errors (and possibly false-positives).

hoefling: @sobolevn sorry, I cannot follow you. Or do you mean "not allow `bool`" instead? Only boolean literals?

sobolevn: Sorry :(

I mean that using `bool` should not produce this error. Because users will get a unittest failure if it is not `True` anyway. And I don't want to have false-positives in test code.

Is it better now? 😊

django-stubs/test/client.pyi
intgr: Hmm. Ordinarily you would remove the default value (`= ...`) from the argument in one of the overloads, like

```python
@overload
def get(self, path: str, data: Any = ..., follow: Literal[True], secure: bool = ..., **extra: Any) -> _MonkeyPatchedWSGIResponseRedirect: ...
#                                         ^^^^^^^^^^^^^^^^^^^^^
@overload
def get(self, path: str, data: Any = ..., follow: bool = ..., secure: bool = ..., **extra: Any) -> _MonkeyPatchedWSGIResponse: ...
```

But here it's not allowed because it follows another argument with default `data`.

I'm not aware of any tricks to make it work. May indeed have to use 3 overloads.

PR #1260, opened 2022-11-15
Update to mypy 0.991 for compatible-mypy & CI

intgr:
Made a few tweaks that were needed to get our CI passing with mypy 0.991.

~~There is still a new regression with `mypy_django_plugin` that I haven't investigated, which happens with Django 3.2 only. Created an issue for it: #1261  Disabled testing with Django 2.2 and 3.2 in CI for now.~~ Thanks to @flaeppe for the fix!

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,13 +7,14 @@ ignore_missing_imports = True
 incremental = True
 strict_optional = True
 show_traceback = True
-warn_no_return = False
 warn_unused_ignores = True
 warn_redundant_casts = True
 warn_unused_configs = True
 warn_unreachable = True
 disallow_untyped_defs = true
 disallow_incomplete_defs = true
+show_error_codes = False
+disable_error_code = empty-body
 
 plugins =
     mypy_django_plugin.main
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -249,6 +249,8 @@ class NewSemanalDjangoPlugin(Plugin):
             and sym.node.has_base(fullnames.BASE_MANAGER_CLASS_FULLNAME)
         ):
             return reparametrize_any_manager_hook
+        else:
+            return None
 
     def get_base_class_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
         # Base class is a Model class definition
@@ -309,6 +311,8 @@ class NewSemanalDjangoPlugin(Plugin):
             "django_stubs_ext.annotations.WithAnnotations",
         ):
             return partial(handle_annotated_type, django_context=self.django_context)
+        else:
+            return None
 
     def get_dynamic_class_hook(self, fullname: str) -> Optional[Callable[[DynamicClassDefContext], None]]:
         # Create a new manager class definition when a manager's '.from_queryset' classmethod is called
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ dependencies = [
 ]
 
 extras_require = {
-    "compatible-mypy": ["mypy>=0.980,<0.990"],
+    "compatible-mypy": ["mypy>=0.991,<0.1000"],
 }
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ dependencies = [
 ]
 
 extras_require = {
-    "compatible-mypy": ["mypy>=0.991,<0.1000"],
+    "compatible-mypy": ["mypy>=0.991,<1.0"],
 }
 
 setup(
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,4 +9,4 @@ psycopg2-binary
 -e .[compatible-mypy]
 
 # Overrides:
-mypy==0.982
+mypy==0.991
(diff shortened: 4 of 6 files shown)


setup.py
```diff
@@ -31,7 +31,7 @@ def find_stub_files(name: str) -> List[str]:
 ]
 
 extras_require = {
-    "compatible-mypy": ["mypy>=0.980,<0.990"],
+    "compatible-mypy": ["mypy>=0.991,<0.1000"],
```
sobolevn:
```suggestion
    "compatible-mypy": ["mypy>=0.991,<1.0"],
```

.github/workflows/test.yml
```diff
@@ -67,10 +67,8 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.8', '3.9', '3.10']
-        django-version: ['3.2', '4.0']
+        django-version: ['4.0']
         include:
-          - python-version: '3.7'
-            django-version: '2.2'
```
intgr: @sobolevn Here's one way to "solve" #1261, just disable testing with Django 2.2 and 3.2. 🤦

intgr: Looks like #1329 has a better fix, reverting this.

intgr:
Restored CI testing with Django 2.2 and 3.2.

This is now ready for review. Hoping to get this included in the next release (#1334). Not sure who has time to look at this, maybe @flaeppe or @ljodal?

mypy.ini
```diff
@@ -7,13 +7,14 @@ ignore_missing_imports = True
...
 strict_optional = True
 show_traceback = True
-warn_no_return = False
 warn_unused_ignores = True
 warn_redundant_casts = True
 warn_unused_configs = True
 warn_unreachable = True
 disallow_untyped_defs = true
 disallow_incomplete_defs = true
+show_error_codes = False
```
intgr, line 16: `show_error_codes` default changed in mypy 0.990.

I'm reverting to the old behavior, because so many `.yml` test files have mypy error messages without the error code.

mypy.ini
```diff
@@ -7,13 +7,14 @@ ignore_missing_imports = True
...
+disable_error_code = empty-body
```
intgr, line 17: Replaced `warn_no_return = False` with `disable_error_code = empty-body`, the latter is less dangerous as it now only applies to empty functions.

mypy_django_plugin/main.py
```diff
@@ -309,6 +311,8 @@ def get_type_analyze_hook(self, fullname: str) -> Optional[Callable[[AnalyzeType
             "django_stubs_ext.annotations.WithAnnotations",
         ):
             return partial(handle_annotated_type, django_context=self.django_context)
+        else:
+            return None
```
intgr, line 315: This was added because I removed `warn_no_return = False` mypy setting.

intgr:
I really want this for the next release, but it has been difficult to get anyone to review this. Maybe @ngnpope?

mypy.ini
flaeppe: Should we try to work towards `show_error_codes = True`? It feels reasonable that tests are granular enough to fail if a specific error code changes.

It could be an option to try to resolve this over time. Best thing would probably be to have a linter that disallows `type: ignore` i.e. without code in the test suite. I suppose that could be tracked in an issue.

flaeppe (approved):
I think this looks good, just had 1 comment regarding `show_error_codes`

intgr:
Awesome, thanks!

> Should we try to work towards `show_error_codes = True`? It feels reasonable that tests are granular enough to fail if a specific error code changes.

For adding `type: ignore` comments it would be helpful.

But in the `.yml` test suite, we already compare the full error messages, which is specific enough. I doubt mypy will ever have two different error codes with the same message.

It would make the already long `# E: ...` lines even longer. It's also a bit finicky because the error code is separated by two spaces (might not be obvious that a diff is between one and two spaces).

flaeppe:
> Awesome, thanks!
> 
> > Should we try to work towards `show_error_codes = True`? It feels reasonable that tests are granular enough to fail if a specific error code changes.
> 
> For adding `type: ignore` comments it would be helpful.
> 
> But in the `.yml` test suite, we already compare the full error messages, which is specific enough. I doubt mypy will ever have two different error codes with the same message.
> 
> It would make the already long `# E: ...` lines even longer. It's also a bit finicky because the error code is separated by two spaces (might not be obvious that a diff is between one and two spaces).

Ah, right, didn't think about that it was the reverse case that was the issue. I suppose if we'd prefer matching on code instead of a complete message one could go with `# ER: .*msg.*[some-code]` (regex match). But I agree, it's good enough matching the message, like the suite does now.

ngnpope:
> I really want this for the next release, but it has been difficult to get anyone to review this. Maybe @ngnpope?

Sorry, I do want to help out, but I have to squeeze it in around everything else... 🙂 Me right now: 🤹🏻 

> I suppose if we'd prefer matching on code instead of a complete message one could go with # ER: .*msg.*[some-code] (regex match).

I don't think that would work so well in most cases because the messages are often formatted with the type values which we then would not detect changes for...

intgr:
> Sorry, I do want to help out, but I have to squeeze it in around everything else...

No worries :)

PR #1265, opened 2022-11-16
GeoDjango spatial reference fixes

mjakob:
# I have made things!

- Make `django.contrib.gis.gdal.srs.SpatialReference()` accept an EPSG code as an `int`, not just a string.
- Make `django.contrib.gis.gdal.srs.CoordTransform()` only accept `SpatialReference`s as arguments.
- Narrow `django.contrib.gis.geos.GEOSGeometryBase.transform()` and `django.contrib.gis.gdal.OGRGeometry.transform()` to accept `CoordTransform`, `SpatialReference`, `int`, or `str` instead of `Any`.

All changes should be according to Django's documentation. A test case is included.

## Related issues

None that I am aware of.


tests/typecheck/contrib/gis/test_gdal.yml
```diff
@@ -0,0 +1,20 @@
+-   case: test_transform
```
sobolevn: Thanks, but I don't think this is worth adding a new test case.

We add tests for plugin related things / complex overloads.

sobolevn (commented):
I can say honestly, that I don't have any clue how any of this works :)

Maybe some other maintainers have experience with geo module?

tests/typecheck/contrib/gis/test_gdal.yml
mjakob: Removed!

I hesitated myself after I wrote it, but when I saw *gis_tests* [in the ignored test modules](https://github.com/typeddjango/django-stubs/blob/master/scripts/enabled_test_modules.py#L8) i decided to keep it in.

mjakob:
If it's of any help, the Django code is pretty clear with what types it expects:
- [CoordTransform()](https://github.com/django/django/blob/2848e5d0ce5cf3c31fe87525536093b21d570f69/django/contrib/gis/gdal/srs.py#L349)
- [OGRGeometry.transform()](https://github.com/django/django/blob/2848e5d0ce5cf3c31fe87525536093b21d570f69/django/contrib/gis/gdal/geometries.py#L400)
- [GEOSGeometryBase.transform()](https://github.com/django/django/blob/2848e5d0ce5cf3c31fe87525536093b21d570f69/django/contrib/gis/geos/geometry.py#L472) actually creates an OGRGeometry from the GEOSGeometry and calls that instance's transform() method.

Finally, the Django documentation has [an example of creating a SpatialReference from an integer](https://docs.djangoproject.com/en/4.1/ref/contrib/gis/gdal/#django.contrib.gis.gdal.SpatialReference). This is what triggered this PR since it causes an error in mypy.

intgr (approved):
Sorry that this got delayed for so long. None of the regular TypedDjango reviewers are familiar with GeoDango. But this looks good to me.

PR #1270, opened 2022-11-23
WIP: Implement Managers with generic QuerySets

### LINKED ISSUE #863: Making QuerySet classes part of the type of Manager classes
This is not a bug report, but a suggestion for a feature. This is a breaking change and requires some hackery. It's something I'm doing in my project, and something to consider for other projects. At the very least this will explain to others how to do what I'm doing.

I use this library with [Pyright](https://github.com/microsoft/pyright) and without the `mypy` plugin. I have found a rudimentary way of getting custom `QuerySet` class methods to appear for my managers by doing something close to the following:

```py

QS = TypeVar('QS', bound='QuerySet')  # type: ignore

class TypedManager(Generic[M, QS], Manager[M]):
   def all(self) -> QS: ...

    # ... Many other methods are re-declared here to use `QS` as the QuerySet type  ...
```

I can then declare my QuerySet and Manager classes like so:

```py
# `cast` just to avoid runtime errors, as I don't monkey patch the class.
# In my codebase I have a wrapper type for this.
class FooQuerySet(cast('Type[QuerySet[Foo]]', QuerySet)):
    def custom_method(self) -> 'FooQuerySet':
        ...

class FooManager(TypedManager['Foo', FooQuerySet]):
    ...
```

Then the type of `Foo.objects.all().custom_method()` will be known. One of the caveats is that you cannot use QuerySet methods in the Manager class with `from_queryset`, as there's no way to transform types like that in Python typing at the moment, especially considering the `queryset_only` flag. As such, you have to use `.all()` before calling custom QuerySet methods.

I suggest changing the `Manager` class type arguments to include the `QuerySet` type, especially if these generic parameters are ever going to end up in the Django codebase itself. It's the only way to accurately express the types when doing type checking without a plugin system.

One thing to note is that you have to disable type errors for the `bound`, but the rest of the code will work in Pyright. This might not work in all type checkers for Python. Ideally, Python typing should be updated to make it possible to express the following instead:

```py
QS = TypeVar('QS', bound='QuerySet[M]')
```

This is just a feature missing in Python typing at the moment. It is probably worth pushing for Python type checkers to be able to do this before implementing something like this.

w0rp:
# I have made things!

I'm creating this pull request now primarily because I need some help implementing this. I am updating the Manager class to hold the QuerySet class for the manager as a generic type parameter, [as per the issue](https://github.com/typeddjango/django-stubs/issues/863).

I have gotten some parts to work, and the types should work for type checkers without plugin support. Figuring out how to get the mypy plugin to work however is _absolutely maddening_. There are several hooks where a manager is changed in one place, and then again in a totally different place, and then maybe again in another, and I am having difficulty figuring out exactly how to change the Manager types so they have the right type parameters in the plugin code.

If I can fix the plugin and test cases, this patch will improve type information for mypy and other type checkers that don't have plugin support like Pyright.

## Related issues

Closes #863


flaeppe:
This is late, I know, but I have one question here, while I'm aware that this is something that isn't really enforced today either; If a manager is generic over a model and a `QuerySet`, while a `QuerySet` is generic over a model, how do we then get insurance that it's the same __model__ for both `Manager` and `QuerySet`?

To some extent I do agree that a manager should/could be generic over a model and queryset though.

And the plugin yes, it's quite the spaghetti while it also have to do a lot of work puzzling together managers with querysets in different and multiple stages of mypy's processing.

sobolevn (commented):
This is a breaking change. I feel like we should wait a few releases with it, because the last one was quite complex for our users.

Plus, I want to be sure that we have enough benefits in doing this.

Viicos:
> how do we then get insurance that it's the same **model** for both `Manager` and `QuerySet`?

afaik you can't. Actually the current `_QS` type variable is invalid (@w0rp currently added a type ignore). This is a use case that probably requires https://github.com/python/typing/issues/548. I'll explore the path of this PR a bit more and see if I can find something suitable.

finlayacourt:
Now that we have default types would it be possible to impliment this change in a non-breaking way?

```python
_T = TypeVar("_T", bound=Model, covariant=True)
_QS = TypeVar("_QS", bound=QuerySet[Any], covariant=True, default=QuerySet[_T])

class BaseManager(Generic[_T, _QS]):
    def get_queryset(self) -> _QS: ...
```

The default also ensures that the queryset is bound by the correct model unless the type param is explicitly overridden

PR #1272, opened 2022-11-26
Type hint improvements for string promises, manager, query set

benspaulding:
I originally dove in to address issues where type hints only accepted `str` but should accept `_StrOrPromise | None`, so there are two commits about that.

One commit changes the hints in the signatures of `django.db.models.query._QuerySet.as_manager()` and `django.db.models.manager.BaseManager.from_queryset()` to use generics/typevars instead of `Any`.

Finally, I found two places with what I think are unnecessary `# type: ignore` comments.

- If you want each of those in a different PR, please let me know!
- If you only want some of the commits feel free to cherry-pick them yourself, if that is easier.

Let me know if you have any feedback or questions. Thanks!


sobolevn:
> One commit changes the hints in the signatures of django.db.models.query._QuerySet.as_manager() and django.db.models.manager.BaseManager.from_queryset() to use generics/typevars instead of Any.

This is handled by our plugin, but in case it will help users without the plugin to have better types (and won't break anything): I am ok with it.

sobolevn (approved):

benspaulding:
@sobolevn [wrote](https://github.com/typeddjango/django-stubs/pull/1272#issuecomment-1328019584):
> > …
>
> This is handled by our plugin, but in case it will help users without the plugin to have better types (and won't break anything): I am ok with it.

Ah, good to know! I didn’t realize it handled that. I was looking through the stubs to see why I wasn’t getting what I expected, but that was in my editor, not with mypy directly. So if it works for y’all it would be handy to see this accepted.

Thanks!

sobolevn (approved):
Thank you!

PR #1278, opened 2022-12-02
Add Django 4.0 trigram word classes

### LINKED ISSUE #1277: Missing TrigramWordSimilarity, TrigramWordDistance
# Bug report

Hi there, thanks for the great library!

## What's wrong

django-stubs is missing type hints for some of the postgres full text search classes that were added in Django 4.0

https://github.com/django/django/blob/8d160f154f0240a423e83ffe0690e472f837373c/django/contrib/postgres/search.py

- `TrigramWordBase`
- `TrigramWordDistance`
- `TrigramStrictWordDistance`
- `TrigramWordSimilarity`
- `TrigramStrictWordSimilarity`

django-stubs has existing type hints for `TrigramBase`, `TrigramDistance`, etc so it should be pretty straightforward to copy those examples.

## How is that should be

I should be able to import these classes without raising a type error.

## System information

- OS:
- `python` version:
- `django` version:
- `mypy` version:
- `django-stubs-ext` version:

michael-lazar:
# I have made things!

Documentation for these classes:

 https://docs.djangoproject.com/en/4.1/ref/contrib/postgres/search/#trigramwordsimilarity

## Related issues

- Closes [#1277](https://github.com/typeddjango/django-stubs/issues/1277)

### DIFF
--- a/django-stubs/contrib/postgres/search.pyi
+++ b/django-stubs/contrib/postgres/search.pyi
@@ -105,26 +105,11 @@ class TrigramBase(Func):
     def __init__(self, expression: _Expression, string: str, **extra: Any) -> None: ...
 
 class TrigramWordBase(Func):
-    def __init__(self, expression: _Expression, string: str, **extra: Any) -> None: ...
+    def __init__(self, string: str, expression: _Expression, **extra: Any) -> None: ...
 
-class TrigramSimilarity(TrigramBase):
-    function: str
-
-class TrigramDistance(TrigramBase):
-    function: str
-    arg_joiner: str
-
-class TrigramWordDistance(TrigramWordBase):
-    function: str
-    arg_joiner: str
-
-class TrigramStrictWordDistance(TrigramWordBase):
-    function: str
-    arg_joiner: str
-
-class TrigramWordSimilarity(TrigramWordBase):
-    function: str
-
-class TrigramStrictWordSimilarity(TrigramWordBase):
-    function: str
-
+class TrigramSimilarity(TrigramBase): ...
+class TrigramDistance(TrigramBase): ...
+class TrigramWordDistance(TrigramWordBase): ...
+class TrigramStrictWordDistance(TrigramWordBase): ...
+class TrigramWordSimilarity(TrigramWordBase): ...
+class TrigramStrictWordSimilarity(TrigramWordBase): ...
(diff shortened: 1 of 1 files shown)


django-stubs/contrib/postgres/search.pyi
```diff
@@ -104,9 +104,27 @@ class SearchHeadline(Func):
 class TrigramBase(Func):
     def __init__(self, expression: _Expression, string: str, **extra: Any) -> None: ...
 
+class TrigramWordBase(Func):
+    def __init__(self, expression: _Expression, string: str, **extra: Any) -> None: ...
```
intgr: Hmm Django documentation has the arguments in the other order, for `TrigramWordSimilarity` https://docs.djangoproject.com/en/4.1/ref/contrib/postgres/search/#django.contrib.postgres.search.TrigramWordSimilarity

```
TrigramWordSimilarity(string, expression, **extra)
```

Or is the documentation wrong?

django-stubs/contrib/postgres/search.pyi
```diff
@@ -104,9 +104,27 @@ class SearchHeadline(Func):
...
+
+class TrigramStrictWordDistance(TrigramWordBase):
+    function: str
+    arg_joiner: str
+
+class TrigramWordSimilarity(TrigramWordBase):
+    function: str
+
+class TrigramStrictWordSimilarity(TrigramWordBase):
+    function: str
```
intgr: The `Func` base class already defines `function: str`, no need to repeat attributes already defined higher in class hierarchy.

intgr (changes requested):
Thanks for the PR

django-stubs/contrib/postgres/search.pyi
michael-lazar: The documentation is correct and I need to fix this, thanks for double checking!

IMO it's super confusing that the arguments are swapped with the other trigram functions, but the rationale is given in the Django PR: https://github.com/django/django/pull/14833#discussion_r702749182

django-stubs/contrib/postgres/search.pyi
michael-lazar: Sure, I can do the same thing for `arg_joiner` too.

intgr (approved):
Looks good 👍

intgr:
CI will be fixed in #1279

sobolevn (approved):

sobolevn:
Thanks both of you!

Issue #1282, opened 2022-12-06
Item "AbstractBaseUser" of "Union[AbstractBaseUser, AnonymousUser]" has no attribute "is_superuser"


Archmonger:
# Bug report

## What's wrong

`AbstractBaseUser` does not appear to be properly type hinted

```python
# This causes the following MyPy error...
# Item "AbstractBaseUser" of "Union[AbstractBaseUser, AnonymousUser]" has no attribute "is_superuser"
@method_decorator(user_passes_test(lambda u: u.is_superuser), name="dispatch")
class GeneralSettingsView(UpdateView):
    ...
```

## How is that should be

All expected attributes off [`AbstractBaseUser`](https://docs.djangoproject.com/en/4.1/topics/auth/customizing/#django.contrib.auth.models.AbstractBaseUser) should be exposed to MyPy

## System information

- OS:
- `python` version: 3.10
- `django` version: 4.1
- `mypy` version: 0.982
- `django-stubs` version: 1.13
- `django-stubs-ext` version: 0.7.0

sobolevn:
pr is welcome!

Archmonger:
Some additional digging tells me that `is_superuser` is not a valid field of `AbstractBaseUser`. So the issue I'm experiencing appears to be #1058.

However, I did notice that the `clean` field is currently not defined within the `AbstractBaseUser` type hints.

Is it worth PRing this if the definition is only `def clean(self) -> None: ...`?

Technically unneeded, but depends on if this repo prioritizes completeness or not.

Archmonger:
Closing in favor of #1058

PR #1283, opened 2022-12-06
Add `clean()` to `AbstractBaseUser`

Archmonger:
Discovered a missing definition while investigating #1282.

This technically might not be needed, but can be merged depending on if this repo prioritizes completeness.

## Related issues

- Refs #1282


sobolevn (commented):
I think it is defined in a super type: https://github.com/typeddjango/django-stubs/blob/b8a1dae68bd2cc2eb3d86b92173895058f384a29/django-stubs/db/models/base.pyi#L42

Archmonger:
Technically that super type would be an override of the base definition, in the case of a `UserModel`.

sobolevn:
I mean that all `Model` instances have `clear` method. So, any `User` would have `clearn` method from it.

PR #1287, opened 2022-12-08
Fix @classproperty decorator

### LINKED ISSUE #1285: Model.object and djangos @classproperties not working since 1.13.1
# Bug report

Hello :-) first of all thank you very much for releasing the new version 1.13.1 so quickly after I asked for it!
But now I am facing two new problems, which might be related so I'm posting them together.

## Steps to reproduce

See my separately crafted repo with minimal steps to reproduce:
https://github.com/djbrown/mypy-django-bugs

But basically it's:

* `pip install django mypy django-stubs==1.13.0`
* generate simple django project and include above bug snippets
* execute mypy check -> ✔
* `pip install django-stubs==1.13.1`
* execute mypy check -> ❌
* `pip install django-stubs==1.13.0`
* execute mypy check -> ✔

So it's definitely the new version..
I can also add a failing test or workflow so you don't have to reproduce manually.

<details>
<summary>[Collapsed: This part of the report was resolved]</summary>

## no attribute Model.objects
```python
def invalid(model: Model):
    model.objects
```
Results in `error: "Model" has no attribute "objects"  [attr-defined]`
I already tried to fix by enabling monkey-patching in settings.py or module via [django_stubs_ext](https://github.com/typeddjango/django-stubs/tree/master/django_stubs_ext) or use ModelBase or TypeVar instead of Model.
Because in https://github.com/typeddjango/django-stubs/pull/1150 the object property was moved from Model to ModelBase - but that doesn't work for me.

Maybe also related issues:
- https://github.com/typeddjango/django-stubs/issues/83
- https://github.com/typeddjango/django-stubs/issues/1068
  - https://github.com/typeddjango/django-stubs/issues/16

</details>

## cannot access classproperties
```python
from django.test import LiveServerTestCase
class ServerTestCase(LiveServerTestCase):
    def invalid(self, arg: str):
        self.live_server_url
```
Results in:
```
error: Argument 1 to "__get__" of "classproperty" has incompatible type "ServerTestCase"; expected "Optional[classproperty[Any]]"  [arg-type]
error: Argument 2 to "__get__" of "classproperty" has incompatible type "Type[ServerTestCase]"; expected "Type[classproperty[Any]]"  [arg-type]
```
I already tried to fix by enabling monkey-patching in settings.py or module via [django_stubs_ext](https://github.com/typeddjango/django-stubs/tree/master/django_stubs_ext), or access classproperty via class or super class.
Accessing normal class attributes works fine though.

## System information

- OS: ubuntu
- `python` version: 3.10.4
- `django` version: 4.1.4
- `mypy` version: 0.991

intgr:
Removed unnecessary `Self` hints on classproperty methods.

## Related issues

The removal of `self: Self` fixes #1285. Regressed in #1253.


django-stubs/utils/functional.pyi
```diff
@@ -92,9 +92,9 @@ _Get = TypeVar("_Get", covariant=True)
 
 class classproperty(Generic[_Get]):
     fget: Callable[[Self], _Get] | None
-    def __init__(self: Self, method: Callable[[Self], _Get] | None = ...) -> None: ...
-    def __get__(self: Self, instance: Self | None, cls: type[Self] = ...) -> _Get: ...
-    def getter(self: Self, method: Callable[[Self], _Get]) -> classproperty[_Get]: ...
+    def __init__(self, method: Callable[[Self], _Get] | None = ...) -> None: ...
```
sobolevn: Now `Self` seems unbound. It can be `Any`

djbrown:
LGTM and the fix works locally for me :)

django-stubs/utils/functional.pyi
intgr: I don't actually understand how this mechanism works 😆 I just reverted the unwanted changes and verified via the test that it works.

If I add `self: Self` here, it no longer works.

sobolevn: mypy binds variables in two cases:
1. If they are in `Generic[]` definition, like we have here with `_Get`
2. If it is used more than once in a function: `def some(a: A) -> A: return a`

Here `Self` is unbound and should be changed to `Any` with a comment, I guess.

djbrown: maybe it could be done like in [python/typeshed stdlib/builtins.pyi classmethod](https://github.com/python/typeshed/blob/307dd08a6e09f800cb239d37f0eb475345c7c67c/stdlib/builtins.pyi#L128-L139)
```python
_R_co = TypeVar("_R_co", covariant=True)
...
class classmethod(Generic[_R_co]):
    @property
    def __func__(self) -> Callable[..., _R_co]: ...
    @property
    def __isabstractmethod__(self) -> bool: ...
    def __init__(self: classmethod[_R_co], __f: Callable[..., _R_co]) -> None: ...
    def __get__(self, __obj: _T, __type: type[_T] | None = ...) -> Callable[..., _R_co]: ...
    if sys.version_info >= (3, 10):
        __name__: str
        __qualname__: str
        @property
        def __wrapped__(self) -> Callable[..., _R_co]: ...
```

[PEP 673 - Self Type](https://peps.python.org/pep-0673/) describes `Self` as shorthand for: `TClassName = TypeVar("TClassName", bound="ClassName")`. So maybe covariant TypeVar would be the fix here?
Also PEP 673 was only just introducted in python 3.11. Maybe that could be a problem too?

djbrown: But then again djangos [classproperty](https://github.com/django/django/blob/0dd29209091280ccf34e07c9468746c396b7778e/django/utils/functional.py#L61-L75) is implemented a bit differently than CPythons [classmethod](https://github.com/python/cpython/blob/main/Objects/funcobject.c#L888)

djbrown: I just saw that `_Get` already is the covariant TypeVar, so that should be fine.
Looking at the [typehints for @property](https://github.com/python/typeshed/blob/307dd08a6e09f800cb239d37f0eb475345c7c67c/stdlib/builtins.pyi#L1168-L1185) I think I now understand what @sobolevn means (simply replace the remaining `Self` with `Any`):
```python
    fget: Callable[[Any], _Get] | None
    def __init__(self, method: Callable[[Any], _Get] | None = ...) -> None: ...
    def __get__(self, instance: Any | None, cls: type[Any] = ...) -> _Get: ...
    def getter(self, method: Callable[[Any], _Get]) -> classproperty[_Get]: ...
```

intgr: Indeed works just as well with `Any`. I've pushed this change.

sobolevn (approved):

PR #1292, opened 2022-12-13
Improve ModelAdmin and decorator hints and attributes, add tests

intgr:
Some follow-up additions on top of #1267:

* Added tests for admin decorators.
* The `@display` decorator now supports usage for `ModelAdmin` methods as well.
* Changed `_StrPromise` to `_StrOrPromise` where plain `str` should be supported as well.
* `ModelAdmin.display` didn't previously allow functions returning booleans.
* `ModelAdmin.actions` didn't previously allow having accurate type hints of the passed function (only allowed base class `ModelAdmin`)

### DIFF
--- a/django-stubs/contrib/admin/decorators.pyi
+++ b/django-stubs/contrib/admin/decorators.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Callable, Sequence
-from typing import Any, TypeAlias, TypeVar, overload
+from typing import Any, TypeVar, Union, overload
 
 from django.contrib.admin import ModelAdmin
 from django.contrib.admin.sites import AdminSite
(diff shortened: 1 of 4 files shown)


django-stubs/contrib/admin/decorators.pyi
```diff
@@ -1,50 +1,53 @@
...
+from django.utils.functional import _StrOrPromise
+from typing_extensions import TypeAlias
 
-_T = TypeVar("_T")
 _Model = TypeVar("_Model", bound=Model)
 _ModelAdmin = TypeVar("_ModelAdmin", bound=ModelAdmin)
 _Request = TypeVar("_Request", bound=HttpRequest)
 _QuerySet = TypeVar("_QuerySet", bound=QuerySet)
+# This is deliberately different from _DisplayT defined in contrib.admin.options
+_DisplayCallable: TypeAlias = Union[Callable[[_ModelAdmin, _Model], Any], Callable[[_Model], Any]]  # noqa: Y037
```
intgr, line 18: Need to suppress Y037. Can't use the pipe syntax here for some reason, otherwise [mypy-self-check (3.9)](https://github.com/typeddjango/django-stubs/actions/runs/3687912473/jobs/6242079013#logs) fails with:

> django-stubs/contrib/admin/decorators.pyi:18: error: Unsupported left operand type for | ("object")

django-stubs/contrib/admin/options.pyi
```diff
@@ -154,7 +156,7 @@ class ModelAdmin(BaseModelAdmin[_ModelT]):
     delete_selected_confirmation_template: _TemplateForResponseT | None
     object_history_template: _TemplateForResponseT | None
     popup_response_template: _TemplateForResponseT | None
-    actions: Sequence[Callable[[ModelAdmin, HttpRequest, QuerySet], None] | str] | None
+    actions: Sequence[_ActionCallable[Any, _ModelT] | str] | None
```
intgr, line 159: It s unfortunate that we have replace`ModelAdmin` with `Any` here. But hopefully in the future it's possible to use `Self`.

sobolevn (approved):
Looks good, thanks!

Issue #1295, opened 2022-12-20
Error when using `gettext_lazy` with AdminSite and ModelAdmin


rclement:
# Bug report

## What's wrong

When using `gettext_lazy` (or the `_` shortcut) with attributes of `AdminSite` and `ModelAdmin`, some errors are still thrown by `mypy`.

### `AdminSite`

The following snippet:

```python
from django.contrib import admin
from django.utils.translation import gettext_lazy as _


class CustomAdminSite(admin.AdminSite):
    site_title = _("Custom Admin Site")
    site_header = _("Admin Site")
    index_title = _("A customized admin site")
```

results in the following `mypy` errors:

```bash
app/admin.py:6: error: Incompatible types in assignment (expression has type "_StrPromise", base class "AdminSite" defined the type
as "str")  [assignment]
        site_title = _("Custom Admin Site")
                     ^~~~~~~~~~~~~~~~~~~~~~
app/admin.py:7: error: Incompatible types in assignment (expression has type "_StrPromise", base class "AdminSite" defined the type
as "str")  [assignment]
        site_header = _("Admin Site")
                      ^~~~~~~~~~~~~~~~
app/admin.py:8: error: Incompatible types in assignment (expression has type "_StrPromise", base class "AdminSite" defined the type
as "str")  [assignment]
        index_title = _("A customized admin site")
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### `ModelAdmin`

The following snippet:

```python
from django.contrib import admin, messages

class CustomAdmin(admin.ModelAdmin[Model]):
    @admin.display(description=_("Custom Attribute"))
    def custom_attribute(self, obj: models.Model) -> int:
        return 12

    def custom_method(self, request: HttpRequest) -> None:
        self.message_user(request, _("Custom error message"), level=messages.ERROR)
```

results in the following `mypy` errors:

```bash
app/admin.py:4: error: Argument "description" to "display" has incompatible type "_StrPromise"; expected "Optional[str]" 
[arg-type]
        @admin.display(description=_("Custom Attribute"))
                                   ^~~~~~~~~~~~~~~~~
app/admin.py:9: error: Argument 2 to "message_user" of "ModelAdmin" has incompatible type "_StrPromise"; expected "str" 
[arg-type]
                    self.message_user(request, _("Custom error message"), level=messages.ERROR)
                                               ^~~~~~~
```

## How is that should be

[...]

PR #1296, opened 2022-12-20
Enhance support for gettext_lazy in AdminSite and ModelAdmin

rclement:
## Changes

- Added support of `gettext_lazy` for selected attributes in `contrib.admin.AdminSite`
- Added support of `gettext_lazy` for the `message` argument of `contrib.admin.ModelAdmin.message_user`
- Updated related tests

## Related issues

Refs #1295


sobolevn (approved):
Thank you!

I've removed an extra test, because we generally don't test simple annotations :)

PR #1297, opened 2022-12-21
Add typing for data parameter in get from RequestFactory, test Client and urlencode

nils-van-zuijlen:
… and urlencode

Following on https://github.com/typeddjango/djangorestframework-stubs/pull/267 by adding the same typing information to the corresponding django stubs.

The data parameter in both `get` methods is fed to `urlencode`,  and that function takes either a mapping or a list of tuples of URL parameters.


django-stubs/test/client.pyi
```diff
@@ -1,9 +1,9 @@
-from collections.abc import Awaitable, Callable, Iterable, Iterator
+from collections.abc import Awaitable, Callable, Iterable, Iterator, Mapping
 from io import BytesIO
 from json import JSONEncoder
 from re import Pattern
 from types import TracebackType
-from typing import Any, Generic, NoReturn, TypeVar
+from typing import Any, Generic, NoReturn, TypeAlias, TypeVar
```
intgr: > django-stubs/test/client.pyi:6: error: Module "typing" has no attribute "TypeAlias"

Import it from `typing_extensions` instead.

intgr (approved):
I added the `from typing_extensions import TypeAlias` myself and this now passes tests.

In the related `djangorestframework-stubs` PR (https://github.com/typeddjango/djangorestframework-stubs/pull/267#issuecomment-1361485446) we had the dilemma about allowing `int`. I don't feel strongly about that, it can be added later if someone finds it useful.

intgr:
Thanks! 🥳

PR #1299, opened 2022-12-25
Improve GeoDjango type-hints

ciscorn:
# I have made things!

- Provide more specific type-hints for GeoDjango modules.
- Fixed some missing exports in `contrib/gis/**/__init__.py`
- Removed unintentional re-exports in many modules.

## Related issues

No


sobolevn (commented):
Oh wow! Thanks a lot! This is amazing!

But, I literally have 0 experience with `gis`.
Maybe @intgr or @adamchainz have more knowledge?

intgr:
Nope, I have no experience with Django GIS.

intgr (approved):
Looks good on a cursory read. As noted above, the TypedDjango team doesn't use GeoDjango. I'm inclined to just merge this as is, it's filling out types that used to be just `Any`.

Changes like this are rarely perfect and we shouldn't expect them to be. If it causes regressions for users, they will be reported and can be improved.

intgr:
There was some overlap between this PR and #1265 (which just got merged to master).

I have pushed a merge with the latest master.

django-stubs/contrib/gis/gdal/datasource.pyi
```diff
@@ -1,16 +1,18 @@
 from typing import Any
 
 from django.contrib.gis.gdal.base import GDALBase as GDALBase
+from django.contrib.gis.gdal.driver import Driver as Driver
+from django.contrib.gis.gdal.layer import Layer as Layer
```
intgr: These probably shouldn't be re-exported. The `GDALBase` above (which you didn't touch, to be fair) looks wrong too.

```suggestion
from django.contrib.gis.gdal.base import GDALBase
from django.contrib.gis.gdal.driver import Driver
from django.contrib.gis.gdal.layer import Layer
```

django-stubs/contrib/gis/gdal/feature.pyi
```diff
@@ -1,27 +1,30 @@
 from typing import Any
 
 from django.contrib.gis.gdal.base import GDALBase as GDALBase
+from django.contrib.gis.gdal.field import Field as Field
+from django.contrib.gis.gdal.geometries import OGRGeometry as OGRGeometry
+from django.contrib.gis.gdal.geomtype import OGRGeomType as OGRGeomType
```
intgr: Looks like ordinary imports, shouldn't be re-exported.

```suggestion
from django.contrib.gis.gdal.base import GDALBase
from django.contrib.gis.gdal.field import Field
from django.contrib.gis.gdal.geometries import OGRGeometry
from django.contrib.gis.gdal.geomtype import OGRGeomType
```

django-stubs/contrib/gis/gdal/geometries.pyi
```diff
@@ -1,136 +1,142 @@
-from typing import Any
+from typing import Any, AnyStr
 
 from django.contrib.gis.gdal import CoordTransform, SpatialReference
 from django.contrib.gis.gdal.base import GDALBase as GDALBase
+from django.contrib.gis.gdal.envelope import Envelope as Envelope
+from django.contrib.gis.gdal.geomtype import OGRGeomType as OGRGeomType
```
intgr: Looks like ordinary imports, shouldn't be re-exported.

```suggestion
from django.contrib.gis.gdal.base import GDALBase
from django.contrib.gis.gdal.envelope import Envelope
from django.contrib.gis.gdal.geomtype import OGRGeomType
```

intgr (changes requested):
After looking into this some more, there are lots of re-exports that look like they shouldn't be re-exports.

Usually Django only has re-exports in `__init__.py` files. Also items in explicit module `__all__ = [...]` should be re-exported. The remaining typically should not be.

intgr:
PS: I don't expect you to fix all the pre-existing incorrect re-exports. Just fix the new ones you have added in this PR.

django-stubs/contrib/gis/utils/layermapping.pyi
```diff
@@ -2,8 +2,8 @@ from collections.abc import Mapping
 from pathlib import Path
 from typing import Any, Protocol
 
+import django.contrib.gis.gdal.field
 from django.contrib.gis.gdal import DataSource, OGRGeomType
-from django.contrib.gis.gdal.field import Field as OGRField
```
intgr: `from django.contrib.gis.gdal.field import Field as OGRField` is preferrable here.

It doesn't count as a re-export; re-exports are only those that have the same name (e.g. `import Field as Field`)

intgr: Documented here: https://typing.readthedocs.io/en/latest/source/stubs.html#imports

intgr (approved):
Thanks! Waiting for the CI to finish.

intgr:
Maybe @mjakob can try this out and share some feedback.

django-stubs/contrib/gis/utils/layermapping.pyi
ciscorn: Thank you for your review!

PR #1302, opened 2022-12-27
Added types for template.backends.django.Template

smithdc1:
# I have made things!

This is an alternative to #994. The intention is to address the comments made there (hopefully tests pass now), and uses the `__EngineTemplate` Protocol.

### DIFF
--- a/django-stubs/template/backends/django.pyi
+++ b/django-stubs/template/backends/django.pyi
@@ -1,10 +1,12 @@
 from collections.abc import Iterator
 from typing import Any, NoReturn
 
+from django.template.base import Origin as _Origin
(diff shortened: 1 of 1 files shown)


django-stubs/template/backends/django.pyi
```diff
@@ -1,10 +1,12 @@
 from collections.abc import Iterator
 from typing import Any, NoReturn
 
+from django.template.base import Origin as _Origin
```
sobolevn: No need to use `_`, it is only useful when we have name conflicts.

sobolevn (approved):
Looks like a reasonable solution! Thanks!

PR #1306, opened 2022-12-31
change render_to_string context type

mschoettle:
# I have made things!

Change `context` argument of `render_to_string` to `dict[str, Any] | None`

The documentation states for the `context` argument

> A [dict](https://docs.python.org/3/library/stdtypes.html#dict) to be used as the template’s context for rendering.

See: https://docs.djangoproject.com/en/dev/topics/templates/#django.template.loader.render_to_string
See: https://docs.djangoproject.com/en/dev/topics/templates/#django.template.engine.render_to_string

If I am misunderstanding anything with the change to `Mapping[_ContextKeys, ...]` let me know please.

## Related issues

Refs #1298


sobolevn (commented):
Since we don't have typeddict support anyway, we can use `dict`

sobolevn:
Thanks!

PR #1308, opened 2023-01-05
Remove unintentional GeoDjango re-exports

intgr:
Follow-up from discussion on #1299.

Stub files with `from ... import SomeSymbol as SomeSymbol` are explicit re-exports (https://typing.readthedocs.io/en/latest/source/stubs.html#imports)

Django does employ intentional re-exports to make imports shorter and more convenient (usually in `__init__.py` files). But many of these re-exports were under **longer** module paths. And these unintentional re-exports are likely to be less stable over Django releases. E.g.

```python
# Instead of:
from django.contrib.gis.geos.geometry import hex_regex
from django.contrib.gis.sitemaps.kml import Sitemap
# Use:
from django.contrib.gis.geometry import hex_regex
from django.contrib.sitemaps import Sitemap
```

Now mypy will warn users when using these re-exports.

### DIFF
--- a/django-stubs/contrib/gis/admin/widgets.pyi
+++ b/django-stubs/contrib/gis/admin/widgets.pyi
@@ -1,7 +1,7 @@
 from logging import Logger
 from typing import Any
 
-from django.forms.widgets import Textarea as Textarea
+from django.forms.widgets import Textarea
 
 geo_context: Any
 logger: Logger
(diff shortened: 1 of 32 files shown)


intgr:
@ciscorn would be great if you could take a look at this PR.

ciscorn:
@intgr Looks good to me. We shouldn't rely on these re-exports.

christianbundy (approved):
this is great, thank you!

PR #1310, opened 2023-01-05
Type cache delete functions to return bool

flaeppe:
Ref: https://docs.djangoproject.com/en/4.1/topics/cache/#django.core.caches.cache.delete


intgr:
The boolean return was added in Django 3.1, so all supported versions support this. :+1:

intgr (approved):

PR #1311, opened 2023-01-05
Revert "Used _ContextKeys for context like dicts"

intgr:
This reverts commit cb2f2339ff31851ff29811c27288f821e59a2b8e, PR #1298.

That change causes false positives with common template rendering patterns, and reasons for the change were not well explained.

See my 3 comments at https://github.com/typeddjango/django-stubs/pull/1298#issuecomment-1372050848

## Related issues

Refs #1298, #1306.


mschoettle:
Sorry, this might not be the feedback you are hoping for. 

As the contributor of #1306 I am in favour of reverting this. I agree that context should be `dict[str, Any]`. I think the Django documentation supports this as well: https://docs.djangoproject.com/en/4.1/ref/templates/api/#rendering-a-context

intgr:
> Sorry, this might not be the feedback you are hoping for.

Contrary, feedback from community members and prior contributors is very much appreciated.

There's also tacit endorsement from ljodal in https://github.com/typeddjango/django-stubs/issues/1316#issuecomment-1397727993 

> For the specific PR in question here I'm also unsure what the goal of the initial PR was, but I predict a lot of errors in our codebases as well

I think I'm within my rights to consider PR this approved by community members and merge it.

Of course this reversion doesn't have to be final, if there's a good reason and a way to fix the issues with the original PR.

mschoettle:
Thank you for your work and help in advancing this project!

PR #1314, opened 2023-01-09
Add support for `django.core.cache.backends.memcached.PyMemcacheCache`

### LINKED ISSUE #1284: Missing class PyMemcacheCache in core/cache/backends/memcached.pyi
# Bug report

## What's wrong

In Django 3.2+ source code in core/cache/backends/memcached.py there's:
```
class PyMemcacheCache(BaseMemcachedCache):
    """An implementation of a cache binding using pymemcache."""
    def __init__(self, server, params):
        import pymemcache.serde
        super().__init__(server, params, library=pymemcache, value_not_found_exception=KeyError)
        self._class = self._lib.HashClient
        self._options = {
            'allow_unicode_keys': True,
            'default_noreply': False,
            'serde': pymemcache.serde.pickle_serde,
            **self._options,
        }
```
but in stubs in core/cache/backends/memcached.pyi there is no stub for PyMemcacheCache.
## How is that should be
 I believe the missing part in core/cache/backends/memcached.pyi is:
```
class PyMemcacheCache(BaseMemcachedCache):
    def __init__(self, server: Union[str, Sequence[str]], params: Dict[str, Any]) -> None: ...
```

## System information

- OS:
- `python` version: 3.8.13
- `django` version: 3.2.16
- `mypy` version: 0.991

ngnpope:
Fixes #1284.

### DIFF
--- a/django-stubs/core/cache/backends/memcached.pyi
+++ b/django-stubs/core/cache/backends/memcached.pyi
@@ -18,3 +18,6 @@ class MemcachedCache(BaseMemcachedCache):
 
 class PyLibMCCache(BaseMemcachedCache):
     def __init__(self, server: str | Sequence[str], params: dict[str, Any]) -> None: ...
+
+class PyMemcacheCache(BaseMemcachedCache):
+    def __init__(self, server: str | Sequence[str], params: dict[str, Any]) -> None: ...


intgr (approved):
Thanks!

PR #1317, opened 2023-01-10
Avoid index errors that cause Mypy to raise Internal Error

jimmylai:
# Avoid index errors that cause Mypy to raise Internal Error

Mypy failed by the following internal error in our codebase
```
error: INTERNAL ERROR -- Please try using mypy master on GitHub:
If this issue continues with mypy master, please report a bug at https://github.com/python/mypy/issues
version: 0.991
```
After investigation, it's caused by an IndexError in mypy_django_plugin.transformers.orm_lookups
This fix adds boundary checks on arg_types and arg_names to avoid the IndexError.
With this patch, Mypy runs successfully without raising internal errors.


sobolevn (approved):

sobolevn:
Thank you!

PR #1319, opened 2023-01-12
Support lazy strings for form errors and ValidationError

mschoettle:
# I have made things!

Add support for using lazy strings in form `add_error` and `ValidationError`.

## Related issues

Refs #689 

Side comment: Reading #689 and the example in https://github.com/typeddjango/django-stubs/pull/1139#issuecomment-1232167698 finally made me understand why this change was necessary. Would it make sense to add an entry to the FAQ in case there are still any occurrences missing?


sobolevn (approved):
> Would it make sense to add an entry to the FAQ in case there are still any occurrences missing?

Yes, please

PR #1320, opened 2023-01-12
Add FAQ entry regarding StrPromise to README

mschoettle:
# I have made things!

Follow-up from my question in #1319, this PR adds a FAQ entry to the README with a small code snippet.

The first commit addresses some markdownlint violations (still 1 left).

## Related issues

Refs #1319, #689

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ This happens because these Django classes do not support [`__class_getitem__`](h
 
 ### How can I create a HttpRequest that's guaranteed to have an authenticated user?
 
-Django's built in [`HttpRequest`](https://docs.djangoproject.com/en/4.0/ref/request-response/#django.http.HttpRequest) has the attribute `user` that resolves to the type
+Django's built in [`HttpRequest`](https://docs.djangoproject.com/en/4.1/ref/request-response/#django.http.HttpRequest) has the attribute `user` that resolves to the type
 
 ```python
 Union[User, AnonymousUser]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -296,6 +291,22 @@ def foo(obj: object) -> None:
...
+```python
+from django.utils.translation import gettext_lazy as _
+
+from django_stubs_ext import StrOrPromise
+
+errors: list[StrOrPromise] = []
+errors.add(_('Error message'))
+```
+
+If this is reported on Django code, please report an issue or open a pull request.
```
intgr: Sometimes also the correct solution is to call `str(value)` on the `StrPromise`, for example when passing to a third party library that expects a `str` type value. Although when done incorrectly, this may result in the translation being evaluated too early.

mschoettle: Thanks, I can add this. Are you aware of any documentation on when it might be sufficient to evaluate translations (too) early?

intgr: This looks like the perfect piece of documentation to link to: https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#working-with-lazy-translation-objects

intgr: In particular:

> Using gettext_lazy() and ngettext_lazy() to mark strings in models and utility functions is a common operation. When you’re working with these objects elsewhere in your code, you should ensure that you don’t accidentally convert them to strings, because they should be converted as late as possible (so that the correct locale is in effect). This necessitates the use of the helper function described next.

mschoettle: Thanks, that's great. I revised the text and removed the code snippet.

intgr:
Thanks! I suggest you add a direct link to https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#working-with-lazy-translation-objects as well. Maybe instead of one of the GitHub links.

mschoettle:
> Thanks! I suggest you add a direct link to https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#working-with-lazy-translation-objects as well. Maybe instead of one of the GitHub links.

Ah, apologies! I forgot to add the link to the text. I had it in the second paragraph (fixed). If you prefer that I replace one of the GitHub links as well I can do that (I would change the link to "cannot be used interchangeably" then).

intgr (approved):
This looks good to me.

intgr:
> If you prefer that I replace one of the GitHub links as well

Na it was just a thought.

PR #1322, opened 2023-01-13
fix: Fix override_settings decorator typing

martinlehoux:
- My solution is quite simple, maybe some TestContextDecorators do change the class/function


intgr (approved):
Looks good to me, thanks!

PR #1241, opened 2022-11-08
Version 1.13.1 release

sobolevn:
Let's hope that it will work!

PR #1326, opened 2023-01-21
Version 1.13.2 release

### LINKED ISSUE #1316: Next release planning (1.13.2)
I'd like to make a new django-stubs release soonish.

I think this is a blocker: 
* #1311 

~~I don't know how to unlock mypy 0.991 support other than to disable testing with Django 3.2 and 2.2 (due to #1261). That's not ideal, but it is a step towards the new versioning policy we discussed in https://github.com/typeddjango/django-stubs/issues/1095#issuecomment-1307066177 to only support the latest Django version.~~
* #1260

Nice to have PRs:
* #1309
* #1308
* #1315

Would this be version 1.13.2, 1.14.0 or go with https://github.com/typeddjango/django-stubs/issues/1095#issuecomment-1307066177 and call it 4.1.0?

@sobolevn @adamchainz feedback welcome (and anyone else)

intgr:
Closes #1316.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ extras_require = {
 
 setup(
     name="django-stubs",
-    version="1.13.1",
+    version="1.13.2",
     description="Mypy stubs for Django",
     long_description=readme,
     long_description_content_type="text/markdown",


intgr:
This is routine just like #1241, I don't think there's anything here to review.

Issue #1261, opened 2022-11-15
Plugin regression with mypy 0.99x: FieldDoesNotExist


intgr:
`mypy_django_plugin` causes a mypy internal error when running `typecheck_tests.py` on Django 3.2 source with mypy 0.991. Does not occur with Django 4.0.

See `typecheck` failures in PR #1260.

I have not yet had the time to investigate this. Any help would be welcome.

```
  File "/home/runner/work/django-stubs/django-stubs/mypy_django_plugin/transformers/fields.py", line 117, in set_descriptor_types_for_field_callback
  File "/home/runner/work/django-stubs/django-stubs/mypy_django_plugin/transformers/fields.py", line 215, in transform_into_proper_return_type
  File "/home/runner/work/django-stubs/django-stubs/mypy_django_plugin/transformers/fields.py", line 41, in _get_current_field_from_assignment
  File "/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/site-packages/django/db/models/options.py", line 610, in get_field
  File "mypy/checkexpr.py", line 1004, in apply_function_plugin
  File "mypy/checkexpr.py", line 1181, in check_call_expr_with_callee_type
  File "mypy/checkexpr.py", line 1264, in check_call
  File "mypy/checkexpr.py", line 1466, in check_callable_call
  File "mypy/checkexpr.py", line 410, in visit_call_expr
  File "mypy/checkexpr.py", line 4665, in accept
  File "mypy/checkexpr.py", line 530, in visit_call_expr_inner
  File "mypy/nodes.py", line 1841, in accept
/home/runner/work/django-stubs/django-stubs/django-source/tests/model_fields/models.py:176: error: INTERNAL ERROR -- Please try using mypy master on GitHub:
FieldDoesNotExist: RenamedField has no field named 'modelname'
```

intgr:
Please let me know if this issue affects any real codebase as well. Currently it has only been observed in our CI.

Possible work-around: downgrade to mypy 0.982.

flaeppe:
From my digging:

`RenamedField` is [declared in Django's test suite](https://github.com/django/django/blob/4c85beca9d8bf24573d741e26ea36295c69af5b2/tests/model_fields/models.py#L175-L176) like:

```python
class RenamedField(models.Model):
    modelname = models.IntegerField(name="fieldname", choices=((1, 'One'),))
```

[...]

PR #1328, opened 2023-01-22
Fix crash when filtering on implicit `pk` field on abstract model

### LINKED ISSUE #1189: Plugin crash on abstract model filtering with custom Manager
python 3.10
django 4.1
mypy 0.982
django-stubs HEAD (a78c592)
django-stubs-ext HEAD (a78c592)

Crash in plugin when using an abstract model with a custom manager:

```
./myproj/app/models.py:15: error: INTERNAL ERROR -- Please try using mypy master on GitHub:
version: 0.982
  File "mypy/checkexpr.py", line 4619, in accept
  File "mypy/nodes.py", line 1822, in accept
  File "mypy/checkexpr.py", line 385, in visit_call_expr
  File "mypy/checkexpr.py", line 505, in visit_call_expr_inner
  File "mypy/checkexpr.py", line 1155, in check_call_expr_with_callee_type
  File "mypy/checkexpr.py", line 1238, in check_call
  File "mypy/checkexpr.py", line 1438, in check_callable_call
  File "mypy/checkexpr.py", line 995, in apply_function_plugin
  File "django-stubs/mypy_django_plugin/transformers/orm_lookups.py", line 40, in typecheck_queryset_filter
    lookup_type = django_context.resolve_lookup_expected_type(ctx, model_cls, lookup_kwarg)
  File "django-stubs/mypy_django_plugin/django/context.py", line 388, in resolve_lookup_expected_type
    lookup_parts, field_parts, is_expression = query.solve_lookup_type(lookup)
  File "/code/myproj/.venv/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1217, in solve_lookup_type
    _, field, _, lookup_parts = self.names_to_path(lookup_splitted, self.get_meta())
  File "/code/myproj/.venv/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1656, in names_to_path
    name = opts.pk.name
AttributeError: 'NoneType' object has no attribute 'name'
./myproj/app/models.py:15: : note: use --pdb to drop into pdb
```

Minimal reproduction example (with or without django_stubs_ext.monkeypatch):

NB: This needs to be in a models file in an installed app.
```
from django.db import models


class BaseManager(models.Manager):
    pass


    objects = BaseManager()

    class Meta:
        abstract = True

    def lock(self) -> None:
        type(self).objects.filter(pk=self.pk).select_for_update().values_list("pk")


class MyModel(BaseModel):
    pass

```

flaeppe:
# I have made things!

Avoid crashing when queryset filters on an implicit `pk` field for a model, when model is abstract.

I tried to figure out #1261 and while type checking Django's test suite I came across the same error as in #1189. But reproduced when filtering on `django.contrib.auth.models.AbstractUser`.

## Related issues

Closes #1189 

I wasn't able to reproduce with the example found in the linked issue though.

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -384,7 +384,7 @@ class DjangoContext:
 
     def resolve_lookup_expected_type(self, ctx: MethodContext, model_cls: Type[Model], lookup: str) -> MypyType:
         query = Query(model_cls)
-        if lookup.startswith("pk") and query.get_meta().pk is None:
+        if lookup == "pk" or lookup.startswith("pk__") and query.get_meta().pk is None:
             # Primary key lookup when no primary key field is found, model is presumably
             # abstract and we can't say anything about 'pk'.
             return AnyType(TypeOfAny.implementation_artifact)
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -384,6 +384,10 @@ class DjangoContext:
 
     def resolve_lookup_expected_type(self, ctx: MethodContext, model_cls: Type[Model], lookup: str) -> MypyType:
         query = Query(model_cls)
+        if lookup.startswith("pk") and query.get_meta().pk is None:
+            # Primary key lookup when no primary key field is found, model is presumably
+            # abstract and we can't say anything about 'pk'.
+            return AnyType(TypeOfAny.implementation_artifact)
         try:
             lookup_parts, field_parts, is_expression = query.solve_lookup_type(lookup)
             if is_expression:
--- a/tests/typecheck/models/test_abstract.yml
+++ b/tests/typecheck/models/test_abstract.yml
@@ -1,0 +1,7 @@
+- case: test_filter_on_abstract_user_pk
+  main: |
+    from django.contrib.auth.models import AbstractUser
+    AbstractUser.objects.get(pk=1)
+    reveal_type(AbstractUser().pk)  # N: Revealed type is "Any"
+  installed_apps:
+    - django.contrib.auth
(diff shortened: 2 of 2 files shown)


mypy_django_plugin/django/context.py
```diff
@@ -384,6 +384,10 @@ def resolve_lookup_into_field(
 
     def resolve_lookup_expected_type(self, ctx: MethodContext, model_cls: Type[Model], lookup: str) -> MypyType:
         query = Query(model_cls)
+        if lookup.startswith("pk") and query.get_meta().pk is None:
```
intgr: `lookup.startswith("pk")` can match other column names, e.g. someone may abbreviate `private_key` as `pkey` which starts with `pk`

Does `solve_lookup_type()` fail in this situation? If not, the `lookup_parts` variable seems like better way to detect this.

flaeppe: Yeah, true. While quite unlikely to also match on `.pk is None` we should still avoid it.

`solve_lookup_type()` is what's crashing. We could probably change the statement to:

```python
if lookup == "pk" or lookup.startswith("pk__") and query.get_meta().pk is None:
    ...
```

flaeppe: While there's also the option to mimic Django

```python
lookup_splitted, _ = lookup.split(LOOKUP_SEP, 1)
if lookup_splitted == "pk" and query.get_meta().pk is None:
    ...
```

intgr: Either of these look fine to me. I'll leave it up to you.

flaeppe: I went with the first comment to avoid an import of`LOOKUP_SEP`

intgr (approved):
Looks good to me, waiting for CI.

simonpercivall:
Actually, #1189 wasn't closed by this. My example was faulty, it was a `.values_list("pk")` that causes the issue.

PR #1331, opened 2023-01-22
Allow returning Response from admin @action callbacks

### LINKED ISSUE #1330: Recent change to admin action decorators bans return values
# Bug report

Example code:

```python3
@admin.action(description="get some JSON")
def get_json(
    self,
    request: HttpRequest,
    queryset: QuerySet[SomeModel],
) -> HttpResponse:       
    return HttpResponse(
        json.dumps({"some": "json"}),
        "application/json",
        status.HTTP_200_OK,
    )
```

## What's wrong

In older versions of django-stubs, this code was fine, because `@admin.action` was typed as:

```python3
def action(
    function: Callable[[ModelAdmin, HttpRequest, QuerySet], None] | None = ...,
    *,
    permissions: Sequence[str] | None = ...,
    description: str | None = ...,
) -> Callable: ...
```

But in https://github.com/typeddjango/django-stubs/pull/1267, this was changed to hard-code the return value as `None`.

## How is that should be

I'm not 100% sure what return values Django allows, but I think this should at least allow HTTP responses.

Link: https://docs.djangoproject.com/en/4.1/ref/contrib/admin/actions/#actions-that-provide-intermediate-pages

## System information

- OS: Linux
- `python` version: 3.10.8
- `django` version: 4.0.8
- `mypy` version: 0.982

christianbundy:
Problem: The Django docs say that you can "return an [HttpResponse](https://docs.djangoproject.com/en/4.1/ref/request-response/#django.http.HttpResponse) (or subclass) from your action", but the stubs currently say that actions can only return `None`.

Solution: Update stubs to return `HttpResponse | None` instead.

## Related issues

Closes #1330 

Documentation: https://docs.djangoproject.com/en/4.1/ref/contrib/admin/actions/#actions-that-provide-intermediate-pages


intgr:
Pinging @flaeppe for visibility.

intgr:
Thanks for the PR!

Please add a test case to `tests/typecheck/contrib/admin/test_decorators.yml` as well. Preferrably into the existing `test_admin_decorators_action` testcase, as more tests slow down the test suite.

django-stubs/contrib/admin/decorators.pyi
```diff
@@ -17,20 +17,22 @@ _QuerySet = TypeVar("_QuerySet", bound=QuerySet)
 # This is deliberately different from _DisplayT defined in contrib.admin.options
 _DisplayCallable: TypeAlias = Union[Callable[[_ModelAdmin, _Model], Any], Callable[[_Model], Any]]  # noqa: Y037
 _DisplayCallableT = TypeVar("_DisplayCallableT", bound=_DisplayCallable)
+_ActionReturn = TypeVar("_ActionReturn", bound=Optional[HttpResponseBase])
```
intgr: We prefer `HttpResponseBase | None` instead of `Optional[]` in `.pyi` files.

intgr:
The failures from test suite aren't really issues, you just need to update the expected messages in the test cases. https://github.com/typeddjango/django-stubs/actions/runs/3983617421/jobs/6832674566

btw you can run individual tests locally with e.g. `pytest -k errors_for_invalid_action_signature`

django-stubs/contrib/admin/decorators.pyi
```diff
@@ -17,20 +17,22 @@ _QuerySet = TypeVar("_QuerySet", bound=QuerySet)
...
 
 @overload
 def action(
     function: Callable[[_ModelAdmin, _Request, _QuerySet], None],
     permissions: Sequence[str] | None = ...,
     description: _StrOrPromise | None = ...,
-) -> Callable[[_ModelAdmin, _Request, _QuerySet], None]: ...
+) -> Callable[[_ModelAdmin, _Request, _QuerySet], Optional[HttpResponseBase]]: ...
```
ngnpope: Why still `Optional[HttpResponseBase]` here and not `_ActionReturn` if you've introduced that below?

django-stubs/contrib/admin/decorators.pyi
```diff
@@ -17,20 +17,22 @@ _QuerySet = TypeVar("_QuerySet", bound=QuerySet)
...
 @overload
 def action(
     *,
     permissions: Sequence[str] | None = ...,
     description: _StrOrPromise | None = ...,
 ) -> Callable[
-    [Callable[[_ModelAdmin, _Request, _QuerySet], None]], Callable[[_ModelAdmin, _Request, _QuerySet], None]
+    [Callable[[_ModelAdmin, _Request, _QuerySet], _ActionReturn]],
+    Callable[[_ModelAdmin, _Request, _QuerySet], _ActionReturn],
```
ngnpope, line 35: Can we simplify this to avoid repeating `Callable[[_ModelAdmin, _Request, _QuerySet], _ActionReturn]` and define `ActionCallable` and `ActionCallableT` as was done for the `@display` decorator? These decorators, in general, work the same way, so a similar typing implementation seems to make sense.

ngnpope (changes requested):
It would be good to have some tests to prevent future regressions.

christianbundy:
Thanks y'all! I appreciate the quick reviews.

Just pushed some changes to resolve these issues, please let me know what you think.

tests/typecheck/contrib/admin/test_decorators.yml
```diff
@@ -73,6 +73,11 @@
     @admin.action(description="Some text here", permissions=["test"])
     def freestanding_action_fancy(modeladmin: "MyModelAdmin", request: HttpRequest, queryset: QuerySet[MyModel]) -> None: ...
 
+    @admin.action
+    def freestanding_action_http_response(modeladmin: "MyModelAdmin", request: HttpRequest, queryset: QuerySet[MyModel]) -> HttpResponse: ...
+
+    @admin.action
+    def freestanding_action_file_response(modeladmin: "MyModelAdmin", request: HttpRequest, queryset: QuerySet[MyModel]) -> FileResponse: ...
```
intgr, line 80: Also add the new functions to the `actions = [...]` list below to ensure they are accepted there.

django-stubs/contrib/admin/decorators.pyi
```diff
@@ -17,21 +17,21 @@ _QuerySet = TypeVar("_QuerySet", bound=QuerySet)
...
+) -> _ActionCallableT: ...
 @overload
 def action(
     *,
     permissions: Sequence[str] | None = ...,
     description: _StrOrPromise | None = ...,
-) -> Callable[
-    [Callable[[_ModelAdmin, _Request, _QuerySet], None]], Callable[[_ModelAdmin, _Request, _QuerySet], None]
-]: ...
+) -> Callable[[_ActionCallableT], _ActionCallableT,]: ...
```
intgr: This comma looks redundant?

```suggestion
) -> Callable[[_ActionCallableT], _ActionCallableT]: ...
```

django-stubs/contrib/admin/options.pyi
```diff
@@ -131,7 +131,7 @@ class BaseModelAdmin(Generic[_ModelT]):
 
 _DisplayT: TypeAlias = _ListOrTuple[str | Callable[[_ModelT], str | bool]]
 _ModelAdmin = TypeVar("_ModelAdmin", bound=ModelAdmin)
-_ActionCallable: TypeAlias = Callable[[_ModelAdmin, HttpRequest, QuerySet[_ModelT]], None]
+_ActionCallable: TypeAlias = Callable[[_ModelAdmin, HttpRequest, QuerySet[_ModelT]], Optional[HttpResponseBase]]
```
intgr:
```suggestion
_ActionCallable: TypeAlias = Callable[[_ModelAdmin, HttpRequest, QuerySet[_ModelT]], HttpResponseBase | None]
```

django-stubs/contrib/admin/decorators.pyi
```diff
@@ -17,21 +17,21 @@ _QuerySet = TypeVar("_QuerySet", bound=QuerySet)
 # This is deliberately different from _DisplayT defined in contrib.admin.options
 _DisplayCallable: TypeAlias = Union[Callable[[_ModelAdmin, _Model], Any], Callable[[_Model], Any]]  # noqa: Y037
 _DisplayCallableT = TypeVar("_DisplayCallableT", bound=_DisplayCallable)
+_ActionCallable: TypeAlias = Callable[[_ModelAdmin, _Request, _QuerySet], HttpResponseBase | None]  # noqa: Y037
+_ActionCallableT = TypeVar("_ActionCallableT", bound=_ActionCallable)
```
intgr: I think there was a good reason why these `Callable[...]` were duplicated before at each overload.

I'm not sure why exactly -- maybe mypy doesn't support TypeVars inside TypeVars -- but after your changes, this no longer raises an error with mypy at the location of the decorator:

```
    @admin.action(description="Some text here", permissions=["test"])
    def freestanding_action_fancy(modeladmin: int, request: int, queryset: int) -> None: ...
```

I suppose this logic could also use a few _negative_ test cases -- checking that mypy correctly flags incorrect usage of the parameters or return types.

flaeppe: Yes, `TypeVar` can't be bound to a `TypeAlias`, it'll resolve to `Any` before it's matched with generic arguments of a class or a callable.

A few non-happy test cases should make that obvious.

christianbundy: Sounds good, I've reverted the previous change and added more tests to highlight this issue. I've left `_DisplayCallableT` alone, but I'd imagine it probably has a similar problem.

ngnpope: I see… Yes, if it's a problem, then the equivalent for display is also surely broken.

christianbundy: I wonder whether it'd be worthwhile to enable `disallow_any_generics` for this repo -- I put together a quick repro, and it looks like if we enable the rule we'll get warnings for this: https://mypy-play.net/?mypy=latest&python=3.11&flags=disallow-any-generics&gist=bf198a38181d36c2105ce3fef468d702

ngnpope: Yeah, maybe, but best looked into separately later.

intgr (approved):
Thanks, this looks great! I'll merge this later unless someone else has additional feedback.

django-stubs/contrib/admin/decorators.pyi
intgr: I'm afraid the fallout from `disallow_any_generics` will be enormous. But I may be wrong :)

christianbundy: > Found 630 errors in 155 files (checked 675 source files)

You aren't wrong. 😬 Here's a quick stab at it though: https://github.com/typeddjango/django-stubs/pull/1339

PR #1332, opened 2023-01-23
Add PASSWORD_RESET_TIMEOUT_DAYS

christianbundy:
Problem: `PASSWORD_RESET_TIMEOUT` was added in 3.1, `PASSWORD_RESET_TIMEOUT_DAYS` was removed in 4.0

Solution: Add the new one, remove the old one.

Links:

- https://docs.djangoproject.com/en/4.0/ref/settings/#password-reset-timeout
- https://docs.djangoproject.com/en/dev/internals/deprecation/#deprecation-removed-in-4-0

### DIFF
--- a/django-stubs/conf/global_settings.pyi
+++ b/django-stubs/conf/global_settings.pyi
@@ -383,8 +383,8 @@ LOGIN_REDIRECT_URL: str
 
 LOGOUT_REDIRECT_URL: str | None
 
-# The number of days a password reset link is valid for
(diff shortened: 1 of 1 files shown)


intgr (approved):
Thanks! Looks good to me.

For people still using older Django who need `PASSWORD_RESET_TIMEOUT_DAYS`, you could use `# type: ignore`. But if that's problematic for some reason, I'm not opposed to restoring the old one also for compatibility.

christianbundy:
Thank you!

PR #1335, opened 2023-01-23
Fix some errors in GeoDjango stubs

### LINKED ISSUE #1333: "OGRGeometry" has no attribute "__iter__" (not iterable)
# Bug report

## What's wrong

Iterating on the return type of `Layer.get_geoms` throws:

`"OGRGeometry" has no attribute "__iter__" (not iterable)`

I think these lines are wrong:

https://github.com/typeddjango/django-stubs/blob/2e504dff6042a792c2a4f6f1b85e77a02c9029d6/django-stubs/contrib/gis/gdal/layer.pyi#L43-L46

## How is that should be

```py 
>>> type(layer.get_geoms())
<class 'list'>
>>> type(layer.get_geoms()[0])
<class 'django.contrib.gis.gdal.geometries.Polygon'>
>>> type(layer.get_geoms(geos=True)[0])
<class 'django.contrib.gis.geos.polygon.Polygon'>
```

I think the type definitions should be as follows (mind also the switched `geos` parameter):

```py
    @overload
    def get_geoms(self, geos: Literal[False] = ...) -> list[OGRGeometry]: ...
    @overload
    def get_geoms(self, geos: Literal[True]) -> list[GEOSGeometry]: ...
```
What do you think, @ciscorn? 

## System information

- OS: Docker image `python:3.9.16-bullseye`
- `python` version: 3.9.16
- `django` version: 4.1.4
- `mypy` version: 0-991

### LINKED ISSUE #1336: error: "bytes" has no attribute "tobytes"
# Bug report

## What's wrong

mypy: `error: "bytes" has no attribute "tobytes" `

Contrary to `shapely.LineString` the `wkb` property of `django.contrib.gis.geos.LineString` is not of type `bytes`, but of type `memoryview`.

```py
>>> type(geometry.boundary)
<class 'django.contrib.gis.geos.collections.MultiLineString'>
>>> type(geometry.boundary.wkb)
<class 'memoryview'>
>>> type(geometry.boundary.wkb.tobytes())
<class 'bytes'>
```

https://github.com/typeddjango/django-stubs/blob/2e504dff6042a792c2a4f6f1b85e77a02c9029d6/django-stubs/contrib/gis/gdal/geometries.pyi#L62-L63

## How is that should be

Probably:

```py
 @property 
 def wkb(self) -> memoryview: ... 
```
cc @ciscorn 

Background information:
- Commit updating Django docs: https://github.com/django/django/commit/7c884afe5acc00d1df83390848227f42918ec0d4
- Commit giving details about `bytes` vs `memoryview`: https://github.com/python/typing/commit/f6c17dd8ccca955374e465172de44e04b0a6c037

## System information

- OS: Docker image `python:3.9.16-bullseye`
- `python` version: 3.9.16
- `django` version: 4.1.4
- `mypy` version: 0-991

ciscorn:
# I have made things!

Fix #1333, Fix #1336


django-stubs/contrib/gis/gdal/layer.pyi
```diff
@@ -41,7 +41,7 @@ class Layer(GDALBase):
     spatial_filter: OGRGeometry | None
     def get_fields(self, field_name: str) -> list[Any]: ...
     @overload
-    def get_geoms(self, geos: Literal[True] = ...) -> OGRGeometry: ...
+    def get_geoms(self, geos: Literal[True] = ...) -> GEOSGeometry: ...
     @overload
-    def get_geoms(self, geos: Literal[False]) -> GEOSGeometry: ...
```
StefanBrand, line 46: I think these should be `list[*Geometry]`.

ciscorn: Thank you!

django-stubs/contrib/gis/gdal/layer.pyi
```diff
@@ -41,7 +41,7 @@ class Layer(GDALBase):
     spatial_filter: OGRGeometry | None
     def get_fields(self, field_name: str) -> list[Any]: ...
     @overload
-    def get_geoms(self, geos: Literal[True] = ...) -> OGRGeometry: ...
+    def get_geoms(self, geos: Literal[True] = ...) -> list[GEOSGeometry]: ...
     @overload
-    def get_geoms(self, geos: Literal[False]) -> GEOSGeometry: ...
+    def get_geoms(self, geos: Literal[False]) -> list[OGRGeometry]: ...
```
intgr: Looks like the default for `geos=` is False, so the param default `= ...` should be moved to the second overload?

Doc: https://docs.djangoproject.com/en/4.1/ref/contrib/gis/gdal/#django.contrib.gis.gdal.Layer.get_geoms

ciscorn: You are right. Thanks!

django-stubs/contrib/gis/geos/geometry.pyi
```diff
@@ -82,7 +82,7 @@ class GEOSGeometryBase(GEOSBase):
     def json(self) -> str: ...
     geojson: str
     @property
-    def wkb(self) -> bytes: ...
+    def wkb(self) -> memoryview: ...
     @property
     def ewkb(self) -> bytes: ...
```
intgr: Looks like `ewkb` in Django has similar code and returns `memoryview` as well?

Have a look at other `bytes` properties in these classes: `hexewkb`, `hex`?

ciscorn: It seems ewkb needs to be fixed ...

```python
>>> ds = gdal.DataSource("sample.shp")

# GEOSGeometry
>>> type(ds[0][0].geom.geos.ewkb)
memoryview
>>> type(ds[0][0].geom.geos.hexewkb)
bytes
>>> type(ds[0][0].geom.geos.hex)
bytes

# OGRGeometry
>>> type(ds[0][0].geom.wkb)
memoryview
>>> type(ds[0][0].geom.hex)
bytes
```

intgr (approved):
Looks good to me, waiting for CI

PR #1339, opened 2023-01-24
Disallow `Any` generics

### LINKED ISSUE #64: Make stubs compatible with disallow_any_generics=True

christianbundy:
Problem: After [this thread](https://github.com/typeddjango/django-stubs/pull/1331#discussion_r1084587812) I realized that I've been taking `disallow_any_generics` for granted in my other projects, and realized that without it enabled in django-stubs it's very easy to make silly mistakes and much harder to notice when accidentally passing `Any`.

Solution: I tried enabling `disallow_any_generics` and then just slowly worked through each type error. 

## Related issues

Maybe closes https://github.com/typeddjango/django-stubs/issues/64? I'm not sure.


django-stubs/contrib/admin/decorators.pyi
```diff
@@ -6,31 +6,33 @@ from django.contrib.admin.sites import AdminSite
...
-_ModelAdmin = TypeVar("_ModelAdmin", bound=ModelAdmin)
+_ModelAdmin = TypeVar("_ModelAdmin", bound=ModelAdmin[Model])
 _Request = TypeVar("_Request", bound=HttpRequest)
-_QuerySet = TypeVar("_QuerySet", bound=QuerySet)
+_QuerySet = TypeVar("_QuerySet", bound=QuerySet[Model])
 # This is deliberately different from _DisplayT defined in contrib.admin.options
 _DisplayCallable: TypeAlias = Union[Callable[[_ModelAdmin, _Model], Any], Callable[[_Model], Any]]  # noqa: Y037
-_DisplayCallableT = TypeVar("_DisplayCallableT", bound=_DisplayCallable)
+_DisplayCallableT = TypeVar("_DisplayCallableT", bound=_DisplayCallable[ModelAdmin[Model], Model])
+_ActionReturn = TypeVar("_ActionReturn", bound=HttpResponseBase | None)
```
christianbundy, line 20: Note: this currently builds off of my other branch, https://github.com/typeddjango/django-stubs/pull/1331.

mypy-test.ini
```diff
@@ -0,0 +1,28 @@
...
+strict_optional = True
+show_traceback = True
+warn_no_return = False
+warn_unused_ignores = True
+warn_redundant_casts = True
+warn_unused_configs = True
+warn_unreachable = True
+disallow_untyped_defs = true
+disallow_incomplete_defs = true
+disallow_any_generics = false
```
christianbundy, line 18: What's the right way to do this? I want `disallow_any_generics = true` in our `mypy.ini`, but I'd like to disable it in tests since I don't want to touch every test. Solutions that came to mind:

- Rewrite all the tests (😭)
- Add a `mypy_config` declaration to all failing tests
- Disable for the `mypy-myapp.*` module in `mypy.ini` -- I couldn't get this working.
- Worst possible solution: copy and paste `mypy.ini` to `mypy-test.ini` for this proof of concept.

django-stubs/middleware/csrf.pyi
```diff
@@ -24,7 +24,11 @@ def rotate_token(request: HttpRequest) -> None: ...
 class CsrfViewMiddleware(MiddlewareMixin):
     def process_request(self, request: HttpRequest) -> None: ...
     def process_view(
-        self, request: HttpRequest, callback: Callable | None, callback_args: tuple, callback_kwargs: dict[str, Any]
+        self,
+        request: HttpRequest,
+        callback: Callable[..., Any] | None,
```
christianbundy, line 29: My goal was to match the existing type unless there was an obvious improvement. For example, when replacing `Callable` I'd use `reveal_type()` to ensure I'm matching it 1:1. I tried very hard to avoid adding `Any` unless that's truly what was previously being inferred.

tests/typecheck/contrib/admin/test_options.yml
```diff
@@ -12,7 +12,7 @@
       from django.db.models.query import QuerySet
       from django.utils.translation import gettext_lazy as _
 
-      def an_action(modeladmin: "A", request: HttpRequest, queryset: QuerySet) -> None:
+      def an_action(modeladmin: "A", request: HttpRequest, queryset: QuerySet[Model]) -> None:
```
intgr, line 15: I recommend using some *real* model in tests, like `contrib.auth.models.User`, to be more similar to actual usage in the field.

django-stubs/views/generic/list.pyi
```diff
@@ -15,14 +15,14 @@ class MultipleObjectMixin(Generic[_M], ContextMixin):
...
-    paginator_class: type[Paginator]
+    paginator_class: type[Paginator[_M]]
     page_kwarg: str
     ordering: str | Sequence[str] | None
     def get_queryset(self) -> _SupportsPagination[_M]: ...
     def get_ordering(self) -> str | Sequence[str] | None: ...
     def paginate_queryset(
         self, queryset: _SupportsPagination[_M], page_size: int
-    ) -> tuple[Paginator, Page, _SupportsPagination[_M], bool]: ...
+    ) -> tuple[Paginator[_M], Page[_M], _SupportsPagination[_M], bool]: ...
```
christianbundy, line 25: In a few cases, I took a leap of faith and decided to `Paginator[_M]` rather than `Paginator[Model]`.

tests/typecheck/contrib/admin/test_options.yml
christianbundy: Oops, I don't think I'd meant to commit this. IIRC there's a test failure in both this file and `test_decorators`, at least partially because of the `display()` fixes, but I appreciate the suggestion and I'll do that when resolving the failing tests.

django-stubs/views/generic/list.pyi
intgr: Yep these really improve our typing accuracy, thanks!

intgr:
Note that I just merged https://github.com/typeddjango/django-stubs/pull/1260, which had changes in mypy.ini. Sorry for the conflicts.

christianbundy:
No problem at all, thanks for letting me know so I can resolve the conflicts real quick.

intgr:
I think this change is desirable and will improve the quality of our stubs.

There are two new conflicts in `.pyi` and some tests also need updating. To avoid this treadmill of conflicts, I can delay merging any further stubs changes for a few days.

intgr:
I tried this change with my real codebase and it causes significant false positives.

Not sure why this one appears with `filter(somecolumn__in=...)`

> error: Incompatible type for lookup 'somecolumn__in': (got "List[Any]", expected "str")  [misc]

These seem to be from changing `Something[Any]` to `Something[Model]`:

> error: Argument 1 to "Subquery" has incompatible type "_QuerySet[WithAnnotations[******], Dict[str, Any]]"; expected "Union[Query, _QuerySet[Model, Model]]"

> error: Value of type variable "_DisplayCallableT" of function cannot be "Callable[[MyModelAdmin, MyModel], str]"  [type-var]

Maybe you should revert back to `Something[Any]` in places you're not sure about?

Issue #1276, opened 2022-12-02
Dynamic settings fail as of 1.13.0


markedwards:
# Bug report

## What's wrong

We build Django settings dynamically, instead of using a `settings.py` file with static properties.  This worked fine with django-stubs up until 1.13.0, but all of our settings fail with 1.13.0:

```
'Settings' object has no attribute 'FOO'
```

This seems to be caused by #1163.

## How is that should be

I understand the reasoning given in #1163, but we are pinned at django-stubs 1.12.0 and would have to completely reimplement our installation in order to unblock.  Can dynamically-constructed settings be supported with django-stubs, or is that door permanently closed?  Maybe it could be a non-default option we could enable, accepting the potential downsides?

## System information

- OS:
- `python` version: 3.10.8
- `django` version: 4.1.3
- `mypy` version: 0.991
- `django-stubs` version: 1.13.0
- `django-stubs-ext` version: 0.7.0

virtualxdriver:
This notably breaks projects that make use of `django-configurations`.

BranislavBajuzik:
This also breaks projects that use [`django-split-settings`](https://github.com/wemake-services/django-split-settings).

intgr:
@sobolevn What do you think a proper solution for this issue would look like?

Being able to toggle disable the `django.conf.settings` integration in the mypy plugin would be one solution. Are there any alternatives that you can think of?

I know you have spoken against making the plugin more configurable. But IME the django-stubs plugin code is frequently fragile and causes mypy crashes ([backlog of mypy-plugin issues](https://github.com/typeddjango/django-stubs/labels/mypy-plugin)). Being able to toggle off and on certain parts of the plugin would seem helpful and allow users to pick which parts they like even if other parts don't work for them. It might also help in debugging/localizing which part of plugin code is at fault.

PR #1341, opened 2023-01-24
Skip executing `apps.ready` for installed apps

### LINKED ISSUE #1312: mypy can’t build migration graph
# Bug report

## What's wrong
i have a django project, is runs and works correctly. But when I try to launch mypy, it shows some errors:
`mypy src --install-types --namespace-packages --implicit-reexport --config-file ./pyproject.toml

```
  File "/Users/viktorzadabin/HipHub/backend/venv/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/mypy/__main__.py", line 15, in console_entry
    main()
  File "mypy/main.py", line 95, in main
  File "mypy/main.py", line 174, in run_build
  File "mypy/build.py", line 193, in build
  File "mypy/build.py", line 248, in _build
  File "mypy/build.py", line 507, in load_plugins
  File "mypy/build.py", line 488, in load_plugins_from_config
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/mypy_django_plugin/main.py", line 73, in __init__
    self.django_context = DjangoContext(self.plugin_config.django_settings_module)
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/mypy_django_plugin/django/context.py", line 83, in __init__
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/mypy_django_plugin/django/context.py", line 67, in initialize_django
    apps.populate(settings.INSTALLED_APPS)
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/django/apps/registry.py", line 124, in populate
    app_config.ready()
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/django_prometheus/apps.py", line 23, in ready
    ExportMigrations()
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/django_prometheus/migrations.py", line 52, in ExportMigrations
    executor = MigrationExecutor(connections[alias])
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/django/db/migrations/executor.py", line 18, in __init__
    self.loader = MigrationLoader(self.connection)
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/django/db/migrations/loader.py", line 58, in __init__
    self.build_graph()
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/django/db/migrations/loader.py", line 276, in build_graph
    self.graph.validate_consistency()
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/django/db/migrations/graph.py", line 198, in validate_consistency
    [n.raise_error() for n in self.node_map.values() if isinstance(n, DummyNode)]
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/django/db/migrations/graph.py", line 198, in <listcomp>
    [n.raise_error() for n in self.node_map.values() if isinstance(n, DummyNode)]
  File "/Users/viktorzadabin/HipHub/backend/venv/lib/python3.10/site-packages/django/db/migrations/graph.py", line 60, in raise_error
    raise NodeNotFoundError(self.error_message, self.key, origin=self.origin)
django.db.migrations.exceptions.NodeNotFoundError: Migration hub.0002_initial dependencies reference nonexistent parent node ('profile', '0001_initial')
```

but django server  runs correctly with all applied migrations

## How is that should be

mypy should launch successfully

## System information

- OS: MacOS Ventura 13.1
- `python` version: 3.10.9
- `django` version: 4.1.4
- `mypy` version: 0.991

flaeppe:
# I have made things!

I got an idea from reading https://github.com/typeddjango/django-stubs/issues/1337#issuecomment-1401668445, that it'd be possible and also preferable to avoid triggering `AppConfig.ready` for `INSTALLED_APPS`. For multiple reasons:

- It'll avoid reporting of issues of `django-stubs` crashing for reasons unrelated to `django-stubs` (https://github.com/typeddjango/django-stubs/issues?q=is%3Aissue+%22in+ready%22).
- I don't think there should ever be a mypy plugin requirement existing in an `AppConfig.ready`. It should work without it.
- It should (in the general case) make startup time of `django-stubs` shorter

The approach taken here is quite invasive(i.e. monkey patching with `unittest.mock`), it could be done in multiple ways. Consider it a PoC.

## Related issues

Closes #1312

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -1,9 +1,8 @@
 import os
 import sys
 from collections import defaultdict
-from collections.abc import Generator
 from contextlib import contextmanager
-from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, MutableMapping, Optional, Set, Tuple, Type, Union
+from typing import TYPE_CHECKING, Any, Dict, Generator, Iterable, Iterator, Optional, Set, Tuple, Type, Union
 from unittest import mock
 
 from django.core.exceptions import FieldError
@@ -51,14 +50,14 @@ def temp_environ() -> Iterator[None]:
         os.environ.update(environ)
 
 
-class AppConfigs(dict, MutableMapping[str, "AppConfig"]):  # type: ignore[type-arg]
+class AppConfigs(Dict[str, "AppConfig"]):
     """
     A mapping for 'AppConfig' that monkey patches 'ready' method on insert
     """
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
-        self.patches: dict[str, mock._patch[mock.MagicMock]] = {}
+        self.patches: Dict[str, mock._patch[mock.MagicMock]] = {}
 
     def __setitem__(self, key: str, value: "AppConfig") -> None:
         self.patches[key] = mock.patch.object(value, "ready", autospec=True)
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -1,8 +1,10 @@
 import os
 import sys
 from collections import defaultdict
+from collections.abc import Generator
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, Optional, Set, Tuple, Type, Union
+from unittest import mock
 
 from django.core.exceptions import FieldError
 from django.db import models
@@ -32,6 +34,7 @@ except ImportError:
 
 
 if TYPE_CHECKING:
+    from django.apps.config import AppConfig
     from django.apps.registry import Apps  # noqa: F401
     from django.conf import LazySettings  # noqa: F401
     from django.contrib.contenttypes.fields import GenericForeignKey
@@ -46,6 +49,42 @@ def temp_environ() -> Iterator[None]:
     finally:
         os.environ.clear()
         os.environ.update(environ)
+
+
+class AppConfigs(dict[str, "AppConfig"]):
+    """
+    A mapping for 'AppConfig' that monkey patches 'ready' method on insert
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.patches: dict[str, mock._patch[mock.MagicMock]] = {}
+
+    def __setitem__(self, key: str, value: "AppConfig") -> None:
+        self.patches[key] = mock.patch.object(value, "ready", autospec=True)
+        super().__setitem__(key, value)
+        self.patches[key].start()
+
+
+@contextmanager
+def skip_apps_ready(apps: "Apps") -> Generator[None, None, None]:
+    """
+    Context manager that monkey patches the apps registry's container for app configs so
+    that we avoid executing custom 'ready' methods. This both saves time but more
+    importantly avoids executing custom runtime code for arbitrary apps. There should be
+    no need for django-stubs plugin to have triggered 'ready' in order to type check
+    correctly.
+    """
+    if apps.ready:
+        # Don't overwrite a readied apps instance. Calling '.ready' will be a noop.
+        yield None
+    else:
+        apps.app_configs = AppConfigs()
+        try:
+            yield None
+        finally:
+            for patch in apps.app_configs.patches.values():
+                patch.stop()
 
 
 def initialize_django(settings_module: str) -> Tuple["Apps", "LazySettings"]:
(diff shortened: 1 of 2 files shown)


flaeppe:
@intgr @sobolevn do you have any opinion on this?

Not really an argument, just additional info; I ran this on one of our projects with a 6 second speedup. Although we do a lot of stuff during startup there.

intgr:
Ah sorry. Seems like a good idea, although not a fan of the monkeypatching.

What gave me pause is that I imagine some people might be using the `ready()` callbacks to impement some hacks or monkeypatching, e.g. changing model or field options, which mypy then would fail to pick up. So I was wondering whether it should be configurable. There could be some other use cases that I can't think of. But there's no precedent currently for making plugin features toggleable.

intgr:
> I ran this on one of our projects with a 6 second speedup

If this is purely from the `ready()` callbacks, it sounds like a rather poor developer experience, since every `./manage.py` invocation incurs this delay as well.

If the ready callbacks are only necessary for e.g. `runserver` to function, this would be better solved by Django passing some information to `ready()` about what it's supposed to be preparing for. I wonder if antying like this has been discussed upstream for Django.

> It'll avoid reporting of issues of django-stubs crashing for reasons unrelated to django-stubs

IMO this is a major design issue with the current plugin design -- importing checked code into the mypy runtime is problematic for many reasons (also mentioned in https://github.com/typeddjango/django-stubs/issues/1337#issuecomment-1401668445). This PR only fixes a symptom of that, but the list of issues reported with this symptom is impressive for sure.

Ideally we should also think about what a better design would look like. Although such a solution would probably take a lot longer.

flaeppe:
> What gave me pause is that I imagine some people might be using the `ready()` callbacks to impement some hacks or monkeypatching, e.g. changing model or field options, which mypy then would fail to pick up. ...

Just as the real goal of the plugin should be to not have to boot django's runtime, I don't think there should be any reason the plugin should support some arbitrary runtime code to be executed. I wouldn't claim that as a reason not to monkey patch ourselves, if we'd benefit from it.

> If this is purely from the ready() callbacks, it sounds like a rather poor developer experience, since every ./manage.py invocation incurs this delay as well.

The timing isn't important and it has nothing to do with django or django-stubs. It was specifics for that project is all.

> If the ready callbacks are only necessary for e.g. `runserver` to function, this would be better solved by Django passing some information to `ready()` about what it's supposed to be preparing for. I wonder if antying like this has been discussed upstream for Django.

I think `ready()` is what you like or need it to be. Django just gives you a hook to do stuff when it's ready, but hasn't started yet.

> IMO this is a major design issue with the current plugin design -- importing checked code into the mypy runtime is problematic for many reasons (also mentioned in https://github.com/typeddjango/django-stubs/issues/1337#issuecomment-1401668445). This PR only fixes a symptom of that, but the list of issues reported with this symptom is impressive for sure.
>
> Ideally we should also think about what a better design would look like. Although such a solution would probably take a lot longer.

While all that might be true, I think it'd help out skipping `ready()` as I think that'd be the main entrypoint for the plugin triggering runtime code of projects it type checks. Leading to arbitrary runtime errors during type checking.

What I'm trying to say is; that I'm thinking patching `ready()` will reduce the surface of runtime code touched by the plugin, thus minimising errors unrelated to our plugin.

intgr:
Are you opposed to making this configurable?

You've demonstrated the issues solved by this change well. But it could open some other can of worms that causes issues for a different set of users.

> I think ready() is what you like or need it to be. Django just gives you a hook to do stuff when it's ready, but hasn't started yet.

Yep, that's the thing. Django itself always unconditionally calls these hooks and we have no idea what users may be doing in those. We _do_ initialize Django within the mypy process; this change would make a Django project in mypy context behave differently from a usual Django runtime context. Then we expect to be able to introspect models and settings in this different runtime context.

We could hypothesize that no user does anything in these ready() hooks that affects the introspection behavior, but I'm afraid Hyrum's Law will prove us wrong.

If some poor soul does hit this issue, it would be better if we can reply "OK just disable this option" than "welp, just keep using the old django-stubs version then". The latter has been our de facto stance in #1276 for example.

flaeppe:
> Are you opposed to making this configurable?

Not really, no.

But if so I'd prefer enabling the monkey patch of `ready()` as default, thus opting out of skipping `ready()`. Not too strong opinions about that either though.

intgr:
Yes agreed about defaulting to enabled.

intgr:
@flaeppe Are you planning to revive this? I hope I haven't discouraged you. :)

flaeppe:
> @flaeppe Are you planning to revive this? I hope I haven't discouraged you. :)

Sure, I just haven't gotten around to it.

Issue #1023, opened 2022-06-27
Could not resolve manager type for  "core.models.FooModel.objects" [django-manager-missing]


MartinThoma:
# Bug report

I've updated `django-stubs-ext` from 0.4.0 to 0.5.0 and got a lof of error messages from mypy. Before, mypy didn't complain

There might actually be nothing wrong with django-stubs, but maybe I need to add a type annotation. The problem is that I have no idea what I need to change.

The code has many unit tests. It's very unlikely that all of those locations mypy now complains about actually have issues that would prevent proper code execution.

## What's wrong

mypy shows several `django-manager-missing` errors.

## How is that should be

No mypy errors shown

## System information

- OS: Ubuntu
- `python` version: 3.8
- `django` version: 3.2
- `mypy` version: 0.942
- `django-stubs` version: 1.12.0
- `django-stubs-ext` version: 0.5.0

jose-reveni:
I have been getting the same error and I think it might be related to #1022

flaeppe:
I wouldn't directly say it's related to #1022, unless you're seeing that error too.

The error you're seeing was intentionally introduced in #999.

I'd start with investigating the types for your model managers. I think manager names are outputted in the errors you're seeing?

Could you display how you set up one of those managers generating the error?

MartinThoma:
I didn't setup any manager explicitly for those models.

I create those models like this:

```python
from django.db import models
import uuid

class Base(models.Model):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)

class FooModel(Base):
    foo = models.CharField(max_length=100)
    bar = models.BooleanField()
```

And I use it like this: `FooModel.objects.all()`.

There is, of course, way more going on, but that is the gist of it. I can try to make a minimal example with pinned dependencies some time this week if that helps.

flaeppe:
Interesting, I'm pretty certain we have test cases covering non-explicit managers.

But I'll double check that.

A minimal example would be helpful too!

aleksanb:
UUID might hack itself onto the base manager of models that use it, perhaps?
Similar to the way django money was found to be behaving.

MartinThoma:
I actually also use Django money. I thought it was not related :thinking:

flaeppe:
Hm, one important lead would be what the type the manager is during runtime.

@MartinThoma since it's an implicit manager. Could you investigate what type(FooModel.objects) outputs?

flaeppe:
[...]

PR #1342, opened 2023-01-24
Fix crash when FK references unknown 'app_label'

flaeppe:
# I have made things!

Plugin previously crashed running the provided test case, now it yields an intended error message instead.

## Related issues

Ref: #1312

### DIFF
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -928,3 +928,18 @@
 
                 class SalesMan(models.Model):
                     client = models.ManyToManyField(CustomUser)
+
+-   case: test_fails_if_app_label_is_unknown_in_relation_field
+    main: |
+        from installed.models import InstalledModel
+    installed_apps:
+      - installed
+    files:
+        -   path: installed/__init__.py
+        -   path: installed/models.py
+            content: |
+                from django.db import models
+                class InstalledModel(models.Model):
+                    non_installed = models.ForeignKey(  # E: Cannot find model 'not_installed.NonInstalledModel' referenced in field 'non_installed'
+                        "not_installed.NonInstalledModel", on_delete=models.CASCADE
+                    )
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
LGTM

tests/typecheck/fields/test_related.yml
```diff
@@ -928,3 +928,18 @@
...
+        from installed.models import InstalledModel
+    installed_apps:
+      - installed
+    files:
+        -   path: installed/__init__.py
+        -   path: installed/models.py
+            content: |
+                from django.db import models
+                class InstalledModel(models.Model):
+                    non_installed = models.ForeignKey(  # E: Cannot find model 'not_installed.NonInstalledModel' referenced in field 'non_installed'
```
intgr, line 943: Is it also possible to ignore this error using a `# type: ignore` comment?

intgr (approved):
Cool, this makes sense.

tests/typecheck/fields/test_related.yml
intgr: I tested it and indeed it works.

flaeppe: I didn't think mypy supported suppressing any errors produced by a plugin (yet?)

Refs:
- https://github.com/typeddjango/django-stubs/issues/1023#issuecomment-1178853194
- https://github.com/python/mypy/issues/12987

intgr: I thought the same. Yet the test passes successfully with the following diff:

```patch
diff --git a/tests/typecheck/fields/test_related.yml b/tests/typecheck/fields/test_related.yml
index 7e9758ae..cc863373 100644
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -940,6 +940,6 @@
             content: |
                 from django.db import models
                 class InstalledModel(models.Model):
-                    non_installed = models.ForeignKey(  # E: Cannot find model 'not_installed.NonInstalledModel' referenced in field 'non_installed'
+                    non_installed = models.ForeignKey(  # type: ignore
                         "not_installed.NonInstalledModel", on_delete=models.CASCADE
                     )
```

I'm not complaining 😆

flaeppe: How interesting, have to compare how this error is emitted versus the one that couldn't be ignored

PR #1345, opened 2023-01-25
Allow int, bool for data argument in test.client.Client.get()

### LINKED ISSUE #1327: Recent changes to test.client.Client.get() do not accept valid data types
# Bug report

## What's wrong

`test.client.Client.get()` expects the `data` arg to be `Mapping[str, str | bytes | Iterable[str | bytes]] | Iterable[tuple[str, str | bytes | Iterable[str | bytes]]] | None` but the method works with other options as well such as when the value is a bool. ie. `self.client.get(url, data={'somevar': True}`
## How is that should be

get() data with a bool value should be accepted.

Here is an example from the codebase in question

```
class QueryArgs(TypedDict):
            query_type: str
            only_selected: bool
            ingredients: list[str]

class TestParams(TypedDict):
            query_args: QueryArgs  # Feels like I should be able to just define the TypedDict inline here
            # query_args: Union[Mapping[str, Union[str, bytes, Iterable[Union[str, bytes]]]], Iterable[Tuple[str, Union[str, bytes, Iterable[Union[str, bytes]]]]], None]
            expected: QuerySet[Recipe]

test_matrix: list[TestParams] = [
            {
                "query_args": {
                    "query_type": "any",
                    "only_selected": False,
                    "ingredients": [self.rum.slug, self.orange_curacao.slug],
                }
                ,
                "expected": Recipe.objects.all(),
            },
]

# there are actually more tests in the matrix
for t in test_matrix:
            with self.subTest(t["query_args"]):
                query_args = t["query_args"]
                response = self.client.get(self.url, data=query_args)
                self.assertEqual(200, response.status_code)
                self.assertQuerysetEqual(t["expected"], response.context["object_list"])
```

This works with django-stubs 1.13.1, but fails with 1.13.1 and 1.13.2 even though self.client.get() works as expected and translates the python True and False correctly.

## System information

- OS:
- `python` version: 3.11.1
- `django` version: 4.1.5
- `mypy` version: 0.991

nils-van-zuijlen:
## Related issues

Closes #1327


intgr:
Looks like you forgot the changes for `_GetDataType` for the `Client.get` method :)

btw is `bool` allowed now as well implicitly?

intgr:
Ping @nils-van-zuijlen

intgr (approved):
Thanks, looks good. Waiting for CI.

intgr:
I verified that `bool` is indeed implicitly accepted, since it's a subtype of `int` in Python (for awkward historical reasons).

PR #1360, opened 2023-02-06
Upgrade to Mypy 1.0

christianbundy:
# I have made things!

Upgrade to Mypy 1.0

## Related issues

Refs https://github.com/python/mypy/issues/13685

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ dependencies = [
 ]
 
 extras_require = {
-    "compatible-mypy": ["mypy>=0.991,<1.0"],
+    "compatible-mypy": ["mypy>=1.0,<1.1"],
 }
 
 setup(
--- a/django-stubs/db/models/fields/files.pyi
+++ b/django-stubs/db/models/fields/files.pyi
@@ -70,7 +70,7 @@ class FileField(Field):
         error_messages: _ErrorMessagesT | None = ...,
     ) -> None: ...
     # class access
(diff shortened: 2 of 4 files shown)


intgr:
Awesome. Is there a reason this is still in draft?

christianbundy:
Nope! Just didn't notice that CI passed. ✅

setup.py
```diff
@@ -31,7 +31,7 @@ def find_stub_files(name: str) -> List[str]:
 ]
 
 extras_require = {
-    "compatible-mypy": ["mypy>=0.991,<1.0"],
+    "compatible-mypy": ["mypy>=1.0,<1.1"],
```
ljodal, line 34: Do we need to bump the minimum, or is that the policy? Seems like there's no braking changes here?

intgr: This only affects those that use `django-stubs[compatible-mypy]`, yes we bump lower bound in that case.

ljodal: Ah right, nvm me, I'm blind 😇

intgr (approved):
Looks fine to me.

sobolevn (approved):
LGTM! Thanks a lot!

adamchainz:
Nice one!

noamcohen97:
Nice! Is there a new release planned?

intgr:
@noamcohen97 See #1363

PR #1364, opened 2023-02-08
Update Model subclass DoesNotExist type

mateenkasim:
# I have made things!

- Updates stub for `DoesNotExist` error on model subclass
- In VSCode, Pylance considers all DoesNotExist errors to be the same error, when the reality is that each model has its own independent DoesNotExist. This is discussed in the referenced issue on Pylance's GH, where they attribute it to the stub. This PR updates the stub using their suggestion.
```python
try:
    # some django code
except ModelA.DoesNotExist:
    # handle ModelA error
except ModelB.DoesNotExist:
    # pylance thinks this code is unreachable before this PR
```

## Related issues

- Refs: https://github.com/microsoft/pylance-release/issues/2770

## Notes
- In all honesty, I don't truly understand stubs – I just copied the suggestion from the folks at Pylance + formatting described in the contribution guide
- Formatting and testing went through smoothly, except for `tests/typecheck/models/test_ineritance.yml/django_contrib_gis_base_model_mixin_inheritance`, which errored because I don't have something called GDAL installed? Let me know if I truly caused this error, or if it's just a machine-specific thing.

### DIFF
--- a/django-stubs/db/models/base.pyi
+++ b/django-stubs/db/models/base.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Collection, Iterable, Sequence
-from typing import Any, Final, TypeVar
+from typing import Any, TypeVar
 
 from _typeshed import Self
 from django.core.checks.messages import CheckMessage
@@ -7,6 +7,7 @@ from django.core.exceptions import MultipleObjectsReturned as BaseMultipleObject
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.db.models.manager import BaseManager
 from django.db.models.options import Options
+from typing_extensions import Final
 
 _Self = TypeVar("_Self", bound=Model)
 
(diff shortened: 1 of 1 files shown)


mschoettle:
Cool! Does it make sense to make it `Final` at the same time? I don't think it is meant to be overridable, is it?

mateenkasim:
I think that makes sense – a `DoesNotExist` error has one clear purpose, as opposed to something like a `save` function which is often overridden.

I left the `Final` out because it's not used much in this repo, but I'm happy to make that change!

mschoettle (approved):
Thanks, LGTM!

django-stubs/db/models/base.pyi
```diff
@@ -1,5 +1,5 @@
 from collections.abc import Collection, Iterable, Sequence
-from typing import Any, TypeVar
+from typing import Any, Final, TypeVar
```
intgr: Looks like you need to import from `typing_extensions` for Python 3.7.

intgr (approved):
Thanks!

django-stubs/db/models/base.pyi
mateenkasim: Good catch - thanks!

PR #1365, opened 2023-02-09
Add new names for assert functions from Django 4.2

adamchainz:
# I have made things!

Under [misc release notes](https://docs.djangoproject.com/en/4.2/releases/4.2/#id1):

> `SimpleTestCase.assertFormsetError()` is deprecated in favor of `assertFormSetError()`.
> `TransactionTestCase.assertQuerysetEqual()` is deprecated in favor of `assertQuerySetEqual()`.

So we need the new capitalizations for people testing on Django 4.2 (👋).

## Related issues

n/a

### DIFF
--- a/django-stubs/test/testcases.pyi
+++ b/django-stubs/test/testcases.pyi
@@ -109,7 +109,7 @@ class SimpleTestCase(unittest.TestCase):
         msg_prefix: str = ...,
     ) -> None: ...
     @overload
-    def assertFormError(
+    def assertFormError(  # old signature, deprecated in Django 4.1
         self,
         response: HttpResponseBase,
         form: str,
@@ -117,6 +117,7 @@ class SimpleTestCase(unittest.TestCase):
         errors: list[str] | str,
         msg_prefix: str = ...,
     ) -> None: ...
+    # assertFormsetError (lowercase "set") deprecated in Django 4.2
     @overload
     def assertFormsetError(
         self,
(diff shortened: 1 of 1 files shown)


intgr (approved):
Seems reasonable to me.

Maybe add a comment like "deprecated in Django 4.2+" above the old methods?

adamchainz:
I did initially, but then I noticed the `@overloads` for `assertFormSetError` that take `response` are there, which are deprecated from Django 4.1. I can comment both...

adamchainz:
Pushed with comments, I'll let you merge

django-stubs/test/testcases.pyi
```diff
@@ -109,14 +109,15 @@ class SimpleTestCase(unittest.TestCase):
...
-    def assertFormError(
+    def assertFormError(  # old signature, deprecated in Django 4.1
         self,
         response: HttpResponseBase,
         form: str,
         field: str | None,
         errors: list[str] | str,
         msg_prefix: str = ...,
     ) -> None: ...
+    # assertFormsetError (lowercase “set”) deprecated in Django 4.2
```
intgr: Seems weird to have curly quotes in source code IMO. Let's avoid Unicode where it's not necessary.

```suggestion
    # assertFormsetError (lowercase "set") deprecated in Django 4.2
```

django-stubs/test/testcases.pyi
```diff
@@ -185,6 +205,7 @@ class TransactionTestCase(SimpleTestCase):
     fixtures: Any
     multi_db: bool
     serialized_rollback: bool
+    # assertQuerysetEqual (lowercase “set”) deprecated in Django 4.2
```
intgr:
```suggestion
    # assertQuerysetEqual (lowercase "set") deprecated in Django 4.2
```

intgr (approved):
Thanks! Sorry to nitpick. :)

django-stubs/test/testcases.pyi
adamchainz: Sure, force of habit from writing!

intgr:
Thanks

PR #1366, opened 2023-02-16
Fixed Collector.data and NestedObjects.data types

### LINKED ISSUE #1271: `Collector.data` and `NestedObjects.data` have different types, and are incorrect
`NestedObjects` currently has:
https://github.com/typeddjango/django-stubs/blob/fa972f1632fb69861c913f2bf359fcf4d2284d7a/django-stubs/contrib/admin/utils.pyi#L33

`Collector` currently has:
https://github.com/typeddjango/django-stubs/blob/fa972f1632fb69861c913f2bf359fcf4d2284d7a/django-stubs/db/models/deletion.pyi#L60

_AFAIK_ `NestedObjects.data` is closer to correct, but I think is wrong, insofar as it's actually something like:
```
Dict[Type[Model], Set[Model] | List[Model]]
```

That is:
- the keys are the model **classes**,
- the values are _initially_ a `set` of model **instances**, but certain methods (notably, `sort` and `delete`) will mutate it into a `list` instead.

[See in Django main](
https://github.com/django/django/blob/855f5a36e7c8e7a8ce3f62d6ef8c9ae3e073ae3d/django/db/models/deletion.py#L101-L102), but this looks the same at least as far back as `3.2` (and possibly ... forever?):
```
# Initially, {model: {instances}}, later values become lists.
self.data = defaultdict(set)
```

Again AFAIK, `NestedObjects` doesn't actually modify the concept of the `data` attribute in any way, so should be the same as `Collector`

Closest vaguely related ticket I could find was:
- #842 which doesn't seem like a good overlap 🤷

RyanWalker277:
# Fixed Collector.data and NestedObjects.data types

## Related issues

Fixes #1271 

### Code changes

Specified `dict[type[Model], set[Model] | list[Model]]` type for `Collecor.data` attribute

### DIFF
--- a/django-stubs/db/models/deletion.pyi
+++ b/django-stubs/db/models/deletion.pyi
@@ -57,7 +57,7 @@ class RestrictedError(IntegrityError):
 
 class Collector:
     using: str
-    data: dict[type[Model], Collection[Any]]
+    data: dict[type[Model], set[Model] | list[Model]]
     field_updates: Any
     restricted_objects: Any
     fast_deletes: Any
--- a/django-stubs/db/models/deletion.pyi
+++ b/django-stubs/db/models/deletion.pyi
@@ -57,7 +57,7 @@ class RestrictedError(IntegrityError):
 
 class Collector:
     using: str
-    data: dict[str, Any]
+    data: dict[type[Model], Collection[Any]]
     field_updates: Any
     restricted_objects: Any
     fast_deletes: Any
--- a/django-stubs/db/models/deletion.pyi
+++ b/django-stubs/db/models/deletion.pyi
@@ -57,7 +57,7 @@ class RestrictedError(IntegrityError):
 
 class Collector:
     using: str
-    data: Any
+    data: dict[str, Any]
     field_updates: Any
     restricted_objects: Any
     fast_deletes: Any
--- a/django-stubs/contrib/admin/utils.pyi
+++ b/django-stubs/contrib/admin/utils.pyi
@@ -30,7 +30,7 @@ def get_deleted_objects(
 ) -> tuple[list[Model], dict[str, int], set[str], list[str]]: ...
 
 class NestedObjects(Collector):
-    data: dict[str, Any]
+    data: dict[type[Model], set[Model] | list[Model]]
     dependencies: dict[Any, Any]
     fast_deletes: list[Any]
     field_updates: dict[Any, Any]
--- a/django-stubs/contrib/contenttypes/management/commands/remove_stale_contenttypes.pyi
+++ b/django-stubs/contrib/contenttypes/management/commands/remove_stale_contenttypes.pyi
@@ -1,12 +1,13 @@
 from typing import Any
 
 from django.core.management import BaseCommand
+from django.db.models.base import Model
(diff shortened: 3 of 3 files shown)


django-stubs/db/models/deletion.pyi
```diff
@@ -57,7 +57,7 @@ class RestrictedError(IntegrityError):
 
 class Collector:
     using: str
-    data: Any
+    data: dict[str, Any]
```
intgr: Looking at Django code, looks like the key is model class, not string?

```python
    def __init__(self, using, origin=None):
        ...
        # Initially, {model: {instances}}, later values become lists.
        self.data = defaultdict(set)
```

```python
    def add(self, objs, source=None, nullable=False, reverse_dependency=False):
        ...
        model = objs[0].__class__
        instances = self.data[model]
```

Also, incomplete types are fine, but looks like value be more precisely typed as `Collection[Any]` (common base class between list and set).

```suggestion
    data: dict[type[Model], Collection[Any]]
```

RyanWalker277: Thanks for the review! The keys are indeed the model classes not strings. I have made the necessary changes. Just wanted to ask whether `Collection[Any]` is just there for increased type precision or it serves some other purpose as well? I was a bit curious about this part but couldn't figure this out.

intgr: Collection is just for improved type precision.

intgr (approved):
Looks good to me.

intgr:
Take a look at the CI failures

django-stubs/db/models/deletion.pyi
kezabelle: Perhaps there's some nuance to this that I'm missing, but why would you want to end up typing the values as `Any`? Everything in the dictionary's values is _supposed_ to be a homogenous bundle of `Model`, non?

To summarise what I put in the original ticket, AFAIK the _exact_ type match ought to be something like`Dict[Type[Model], Set[Model] | List[Model]]`, which fair enough, you might wish to relax down to `Collection`, but what necessitates throwing the `Any` type in?

RyanWalker277: Hey! Thanks for reviewing my changes. I appreciate your feedback. To address your concern, the reason why "Any" was used instead of specifying the exact type of the values in the dictionary is because some parts of the codebase may use different data structures that inherit from the "Model" class. However, I see your point that it may be better to specify the type as "Set[Model] | List[Model]" for clarity. I will do the necessary changes. Thanks for bringing that up!

RyanWalker277:
As suggested by kezabelle, I have made the changes to the code and used the type match `Dict[Type[Model], Set[Model] | List[Model]]` instead of typing the values as Any in the dictionary. Would appreciate it if you could review the changes and let me know if there is anything else that needs to be done.

intgr (approved):
Thanks.

FWIW the downside of using `Model` instead of `Any` is that when you want to access the fields and methods of a particular model subclass (e.g. `auth.models.User`), mypy won't let you do that without an explicit `cast()` or `isinstance()` check. But either is fine by me.

PR #1370, opened 2023-02-21
Fixed types for `vary_on_headers`

### LINKED ISSUE #1340: Correctly type `vary_on_headers` decorator
## What's wrong

Noticed this while reviewing #1339. The `vary_on_headers` decorator doesn't preserve the type (arguments and return) when used on a function.

RyanWalker277:
# Fixed types for `vary_on_headers`

## Related issues
Fixes #1340


django-stubs/views/decorators/vary.pyi
```diff
@@ -1,7 +1,10 @@
 from collections.abc import Callable
 from typing import Any, TypeVar
 
+from typing_extensions import ParamSpec
+
+_P = ParamSpec("_P")
 _F = TypeVar("_F", bound=Callable[..., Any])
 
-def vary_on_headers(*headers: Any) -> Callable: ...
+def vary_on_headers(*headers: Callable[_P, _F]) -> Callable[_P, _F]: ...
```
intgr: This is not correct, the arguments to `headers` are presumably header names as strings?

And the `Callable[]` signature does not preserve the arguments of the wrapped function.

RyanWalker277: Yes you are right! My bad, sorry for that. I dug a bit deeper and found that `functools.wraps` will preserve the types while using the `vary_on_headers` ( as I mentioned in #1340 ). 

```
import functools
from typing import Any, TypeVar, cast

_F = TypeVar("_F", bound=Callable[..., Any])

def vary_on_headers(*headers: Any) -> Callable[[_F], _F]:
    def decorator(func: _F) -> _F:
        @functools.wraps(func)
        def wrapper(*args: Any, **kwargs: Any) -> Any:
            return func(*args, **kwargs)
        return cast(_F, wrapper)
    return decorator

def vary_on_cookie(func: _F) -> _F: ...
```
This is a sample implementation, it can be achieved through `Paramspec` as well, without the unsafe `cast()`. But in any case we would need to impliment `vary_on_headers` which fails the flake8 pre-commit hook. 

```
django-stubs/views/decorators/vary.pyi:11:13: Y010 Function body must contain only "..."
django-stubs/views/decorators/vary.pyi:12:9: Y048 Function body should contain exactly one statement
django-stubs/views/decorators/vary.pyi:13:5: Y048 Function body should contain exactly one statement
```
IMO it needs to be done on the implementation side when the function is actually implemented. LMK your thoughts on this.

django-stubs/views/decorators/vary.pyi
```diff
@@ -1,7 +1,17 @@
+import functools
 from collections.abc import Callable
-from typing import Any, TypeVar
+from typing import Any, TypeVar, cast
 
 _F = TypeVar("_F", bound=Callable[..., Any])
 
-def vary_on_headers(*headers: Any) -> Callable: ...
+def vary_on_headers(func: _F) -> _F: ...
```
intgr: Like I said in #1340, this can be hinted just like  `require_http_methods`

```suggestion
def vary_on_headers(*headers: str) -> Callable[[_F], _F]: ...
```

RyanWalker277: Got it now! Thanks a ton for the help.

django-stubs/views/decorators/vary.pyi
```diff
@@ -1,7 +1,8 @@
+import functools
 from collections.abc import Callable
-from typing import Any, TypeVar
+from typing import Any, TypeVar, cast
```
intgr: These imports are no longer needed

RyanWalker277: Sorry, my bad!

RyanWalker277: Sorry, my bad!

django-stubs/views/decorators/vary.pyi
```diff
@@ -1,7 +1,8 @@
+import functools
```
intgr: Hehe this one isn't needed either

```suggestion
```

intgr (approved):
Applied the import cleanup. Looks good now. Will merge later.

django-stubs/views/decorators/vary.pyi
RyanWalker277: Extremely sorry! That one just slipped through. Thanks a ton for the help.

PR #1373, opened 2023-02-22
Fix ListFilter's expected_parameters return type

pheki:
Currently `ListFilter`'s expected_parameters' return type is defined as `list[str] | None`, which is incorrect in my opinion as:
-  If a `FieldListFilter` subclass returns `None` it will raise [here](https://github.com/django/django/blob/ad18a0102cc2968914232814c6554763f15abbe3/django/contrib/admin/filters.py#L140) as `None` is not iterable
- The default implementation of `SimpleListFilter` [returns](https://github.com/django/django/blob/ad18a0102cc2968914232814c6554763f15abbe3/django/contrib/admin/filters.py#L111) `[None]`, which is not allowed by the current return type.

This PR changes the return type to `list[str | None]`.

## Related issues

- None, but I can open one if required


intgr (approved):
Thanks!

PR #1377, opened 2023-02-23
Annotate BaseCreateView.object attribute as Model | None

### LINKED ISSUE #986: Incorrect `BaseCreateView.object` annotation
# Bug report

## What's wrong

An example of a `CreateView` looks like this (from the Django docs)

```python
from django.views.generic.edit import CreateView
from myapp.models import Author

class AuthorCreateView(CreateView):
    model = Author
    fields = ['name']
```

but the model attribute on the `BaseCreateView` that `CreateView` inherits from is currently defined as `None` [here](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/views/generic/edit.pyi#L52).

## How is that should be

I think object should be annotated something like this

```python
class BaseCreateView(ModelFormMixin[_M, _ModelFormT], ProcessFormView):
    object: Optional[_M]
    def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse: ...
    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse: ...
```

RyanWalker277:
## Related issues

Fixes #986 

# Code Changes

Specified a `_M | None` type for `object` attribute of `BaseCreateView` class


intgr (approved):
Makes sense, thanks

PR #1378, opened 2023-02-23
Release script: Also clean build directories before a build

intgr:
To avoid future incidents like https://github.com/typeddjango/djangorestframework-stubs/issues/362

See https://github.com/typeddjango/djangorestframework-stubs/issues/362#issuecomment-1441790233 for the full background about what happened in djangorestframework-stubs 1.9.0.


sobolevn (approved):
LGTM

PR #1379, opened 2023-02-24
Fixed type of AdminSite.site_url to allow None

LincolnPuzey:
This is allowed to be None as per the docs
https://docs.djangoproject.com/en/dev/ref/contrib/admin#django.contrib.admin.AdminSite.site_url


sobolevn (approved):
Thank you!

PR #1392, opened 2023-03-10
Add support for nested manager and queryset class definitions

federicobond:


intgr:
Thanks for the PR. @flaeppe do you want to review this?

mypy_django_plugin/lib/helpers.py
```diff
@@ -69,12 +69,28 @@ def lookup_fully_qualified_sym(fullname: str, all_modules: Dict[str, MypyFile])
...
+    module_file = None
+    while True:
+        module_file = all_modules.get(module)
+        if module_file:
+            break
+        else:
+            if "." not in module:
+                return None
+            module, parent_cls = module.rsplit(".", 1)
+            cls_name = ".".join([parent_cls, cls_name])
```
flaeppe: I can't recall exactly where but I think mypy has utility that tells you what scope is currently being processed. Could we investigate if that could be useful/a possibility here? Instead of splitting and retrying on the fullname, potentially reimplementing what mypy has for us.

With that said, perhaps it could be up to the caller to be aware of class level scope and rsplit the fullname before passing it to `lookup_fully_qualified_sym`?

What would/could such an approach look like?

tests/typecheck/managers/querysets/test_as_manager.yml
```diff
@@ -212,3 +212,21 @@
...
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+
+                class MyModel(models.Model):
+                    class MyQuerySet(models.QuerySet):
```
flaeppe, line 228: Not really related to this PR, but I'm finding it odd that the plugin dishonors/repairs `disallow_any_generics`.

To be correct under that config, it should look like below here

```suggestion
                    class MyQuerySet(models.QuerySet["MyModel"]):
```

@intgr I know removing the auto population of missing generics would be quite breaking. But do you think it's worth entertaining the idea?

(I think the case is similar for the manager test in this change set)

mypy_django_plugin/lib/helpers.py
federicobond: I'll investigate this, makes more sense. It ended up being more hackish than I would have liked.

federicobond: I found [this](https://github.com/python/mypy/blob/486a51b6e49fff63c05b139185a560c85db7f7be/mypy/lookup.py#L13-L61) in the mypy codebase, which apparently could be a full replacement for the `lookup_fully_qualified_sym` function in this plugin, but I am having a bit of trouble integrating everything because I am to unable place a pdb breakpoint in the plugin code while running the tests. It just hangs there when it finds a `breakpoint()` or `pdb.set_trace()` call.

Is there any way I can setup my django-stubs installation to be able to debug tests?

This is the command I am using:

```
pytest -x --mypy-same-process -s tests/typecheck/managers/querysets/test_annotate.yml::annotate_using_with_annotations_typeddict
```

tests/typecheck/managers/querysets/test_as_manager.yml
intgr: I think the *real* reason is that we don't set `disallow_any_generics` in django-stubs and it's [disabled by default](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_any_generics) in mypy. 😆

flaeppe: But, I think, until Django supports all `__class_getitem__` our monkey patching expects. It's impossible to have a successful run in or with `django-stubs` under `disallow_any_generics` (?)

mypy_django_plugin/lib/helpers.py
flaeppe: The `lookup_fully_qualified` you've found looks promising, but I'm wondering why it wasn't used in the first place. I would suspect, since they're named so similar, someone was aware of it when implementing `lookup_fully_qualified_sym`.

Although, I'd say our test suite should back up if `lookup_fully_qualified` is a valid replacement. So if the suite is green after a swap, everything is fine IMO.

IIRC I've not been able to use breakpoints either. Been resorting to the good ol' print debugging 🙂

federicobond: Ok, so I went deeper into the rabbit hole and rewrote the pull request. The version I mentioned from mypy is quite basic and does not support all of mypy_django_plugin use cases. This rewritten version should be clearer, easier to maintain than my previous one and now passes all tests. It's also probably a bit more performant, as it avoids unnecessary string joining and splitting.

I also found a way to use breakpoints with pytest_mypy_plugins. I will submit a pull request shortly to address that.

mypy_django_plugin/lib/helpers.py
```diff
@@ -69,10 +69,27 @@ def lookup_fully_qualified_sym(fullname: str, all_modules: Dict[str, MypyFile])
...
+        module_file = all_modules.get(module)
+        if module_file:
+            break
+        if "." not in module:
+            return None
+        module, parent_cls = module.rsplit(".", 1)
+        parent_classes.insert(0, parent_cls)
+
+    scope: Union[MypyFile, TypeInfo] = module_file
+    for parent_cls in parent_classes:
```
flaeppe, line 83: `.append` and then iterating in reverse works just as good here right? Perhaps a micro optimization but should at least theoretically be faster, as average for insert is `O(n)`

Ref: https://wiki.python.org/moin/TimeComplexity


```suggestion
        parent_classes.append(parent_cls)

    scope: Union[MypyFile, TypeInfo] = module_file
    for parent_cls in reversed(parent_classes):
```

federicobond: Yes, though n is so small in the vast majority of cases as to make this mostly a matter of readability.

flaeppe: Yes it probably won't be noticeable.

flaeppe (approved):
LGTM

mypy_django_plugin/lib/helpers.py
flaeppe: > ... I also found a way to use breakpoints with pytest_mypy_plugins. I will submit a pull request shortly to address that.

Cool, sounds great

federicobond: This is the pull request: https://github.com/typeddjango/pytest-mypy-plugins/pull/108

PR #1404, opened 2023-03-19
Added missing stub for 'Query.join()' 

### LINKED ISSUE #1402: `Query.join()` is missing from the stubs
# Bug report

## What's wrong

```
model_utils/managers.py:459: error: "Query" has no attribute "join"  [attr-defined]
            new_qs.query.join(conn, reuse=None)
            ^~~~~~~~~~~~~~~~~
```

I checked `db/models/query.pyi` and indeed the `Query` class has no `join()` method, while it does in Django itself.

This seems to be a similar situation to #1399, but I didn't want to stretch the scope of that issue.

## System information

- OS: Ubuntu Linux 22.04
- `python` version: 3.10.6
- `django` version: 4.1.7
- `mypy` version: 1.1.1

RyanWalker277:
## Related issues
 Fixes #1402 
 ### Code changes

Added a stub for the join method with the proper type annotations. The join method takes in a `string` for the `table` parameter and a `tuple` of two `strings` for the on parameter, which represents the columns from each table that should be joined. Hope it clears things up!


django-stubs/db/models/sql/query.pyi
```diff
@@ -197,6 +197,7 @@ class Query(BaseExpression):
     def check_filterable(self, expression: Any) -> None: ...
     def build_lookup(self, lookups: Sequence[str], lhs: Expression | Query, rhs: Any) -> Lookup: ...
     def try_transform(self, lhs: Expression | Query, name: str) -> Transform: ...
+    def join(self, table: str, on: tuple[str, str] = ..., **kwargs: Any) -> None: ...
```
sobolevn: I found a different signature: https://github.com/django/django/blob/e10c1688f96e3b2d202fe401472b7b25f6105969/django/db/models/sql/query.py#L993

django-stubs/db/models/sql/query.pyi
```diff
@@ -197,6 +197,7 @@ class Query(BaseExpression):
     def check_filterable(self, expression: Any) -> None: ...
     def build_lookup(self, lookups: Sequence[str], lhs: Expression | Query, rhs: Any) -> Lookup: ...
     def try_transform(self, lhs: Expression | Query, name: str) -> Transform: ...
+    def join(self, join: str, reuse: str | None = ..., reuse_with_filtered_relation: bool = ...) -> str: ...
```
sobolevn: `join` argument is clearly not a `str`: https://github.com/django/django/blob/e10c1688f96e3b2d202fe401472b7b25f6105969/django/db/models/sql/query.py#L1021

sobolevn: It is described here: https://github.com/django/django/blob/e10c1688f96e3b2d202fe401472b7b25f6105969/django/db/models/sql/query.py#L995-L997

RyanWalker277: Sorry for the confusion. I'll update it accordingly. Thanks for pointing it out.

RyanWalker277: > It is described here: https://github.com/django/django/blob/e10c1688f96e3b2d202fe401472b7b25f6105969/django/db/models/sql/query.py#L995-L997

According to the description, 'join' is either a `base_table_class` or `join_class`. I think `Type[BaseTable | Join]` would be the correct type argument in this case. Let me know your thoughts on this one. I am making a commit to test the change on the CI as pytest is giving some unexpected results at my end.

sobolevn (approved):
Thank you

PR #1407, opened 2023-03-20
Added stub for where method

### LINKED ISSUE #1399: "Query" has no attribute "where"
# Bug report

## What's wrong

Good morning!

I'm adapting this django code snippet into my codebase and I encountered a type error: https://gist.github.com/noviluni/d86adfa24843c7b8ed10c183a9df2afe

The relevant piece looks like this

```python
object_list: models.QuerySet

...

if not object_list.query.where:
```

and mypy raises the following type error

```
error: "Query" has no attribute "where"  [attr-defined]
```

This appears to be a missing declaration in django-stubs. We can see in the django codebase that `Query.where` is a public variable that's defined in the `__init__` method ([source](https://github.com/django/django/blob/d2b688b966f5d30414899549412d370e1317ddb8/django/db/models/sql/query.py#L250
)).

```
self.where = WhereNode()
```

but it's missing from the corresponding django-stubs class.

https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/sql/query.pyi

## How is that should be

Accessing `Query.where` should not raise a type error

## System information

- OS:
- `python` version: 3.10
- `django` version: 4.1.7
- `mypy` version: 1.0.0

RyanWalker277:
# Added a stub for `Query.where`

Fixes #1399 

### Description

Added a new `where` attribute to the `Query` class definition. Hope it fixes this!

### DIFF
--- a/django-stubs/db/models/sql/query.pyi
+++ b/django-stubs/db/models/sql/query.pyi
@@ -50,6 +50,7 @@ class Query(BaseExpression):
     default_ordering: bool
     standard_ordering: bool
     used_aliases: set[str]
+    where: type[WhereNode]
     filter_is_sticky: bool
     subquery: bool
     group_by: None | Sequence[Combinable] | Sequence[str] | Literal[True]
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Confirmed: https://github.com/django/django/blob/f9f9215d3e5500d9a8d0bec0936480ba34d42e88/django/db/models/sql/query.py#L250

Thanks!

PR #1411, opened 2023-03-22
Allow setting SECRET_KEY to be str or bytes, add SECRET_KEY_FALLBACKS, update PasswordResetTokenGenerator

christianbundy:
Problem: `SECRET_KEY` could be either `str` or `bytes`, and it doesn't look like we have any support for `SECRET_KEY_FALLBACKS`.

Solution: Fix type annotation and add `SECRET_KEY_FALLBACKS`. Also update usages in `PasswordResetTokenGenerator`.

See-also: https://github.com/django/django/pull/15198/files

### DIFF
--- a/django-stubs/conf/global_settings.pyi
+++ b/django-stubs/conf/global_settings.pyi
@@ -168,7 +168,11 @@ IGNORABLE_404_URLS: list[Pattern[str]]
 # A secret key for this particular Django installation. Used in secret-key
 # hashing algorithms. Set this in your settings, or Django will complain
 # loudly.
-SECRET_KEY: str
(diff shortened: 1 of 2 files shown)


intgr (approved):
Indeed https://docs.djangoproject.com/en/4.2/ref/settings/#secret-key states

> Uses of the key shouldn’t assume that it’s text or bytes. Every use should go through [force_str()](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.encoding.force_str) or [force_bytes()](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.encoding.force_bytes) to convert it to the desired type.

PR #1418, opened 2023-03-30
Allow lazy translation strings for constraint 'violation_error_message' arguments, attributes

realsuayip:
# I have made things!

## Related issues

- Improvement on https://github.com/typeddjango/django-stubs/pull/1263

PR #1419, opened 2023-03-31
Added BaseConstraint.default_violation_error_message attribute hint

RyanWalker277:
# Refs
#1418 

# Changes

Added `default_violation_error_message` in `BaseConstraint` stubs


django-stubs/db/models/constraints.pyi
```diff
@@ -16,6 +16,7 @@ class Deferrable(Enum):
 class BaseConstraint:
     name: str
     violation_error_message: _StrOrPromise | None
+    default_violation_error_message: str = ...
```
intgr: This won't allow lazy translations. Lazy translations are somewhat tricky.

Note code in Django:

```
...
class BaseConstraint:
    default_violation_error_message = _("Constraint “%(name)s” is violated.")
```

The `_(...)` or `gettext_lazy(...)` here returns a lazy-translated string, it's not actually an instance of `str`, we call it `_StrPromise`.

Typing it as `str` would prevent this assignment.

Most contexts that accept `_StrPromise` also accept `str`, so we use the `_StrOrPromise` type there (it's just an alias for `str | _StrPromise`).

This here should use `_StrOrPromise` as well.

intgr: We don't usually add `= ...` to class attributes -- that's only for function/method arguments.

```suggestion
    default_violation_error_message: str
```

intgr (changes requested):
Thanks for picking this up

django-stubs/db/models/constraints.pyi
RyanWalker277: Will keep this in mind in future changes.Thanks for explaining it so well!

RyanWalker277: Will keep this in mind, thanks for the review =)

intgr (approved):

PR #1420, opened 2023-04-01
Allow 'BaseExpression' in 'UniqueConstraint.expressions'

realsuayip:
# I have made things!

Currently, `UniqueConstraint` does not allow entries such as `Lower()` (it should), I changed the typing so that it is compatible `Index`, which has similar structure.


sobolevn (approved):
Thank you!

PR #1423, opened 2023-04-04
Don't require tomli on Python>=3.11

### LINKED ISSUE #1408: Don't require `tomli` on Python 3.11
# Bug report

## What's wrong

Hi!

I'm in the process of upgrading my app to use Python 3.11 and noticed that the only library that requires `tomli` is `django-stubs`, as all other have it marked as an optional dependency required only if Python version is lower than 3.11.

Screenshot 2023-03-20 at 16 57 32

## How is that should be

`django-stubs` should only install `tomli` on Python <3.11  and use `tomllib` on Python 3.11. As far as I understand their interfaces are compatible, which is why all the libraries listed above just use `tomllib` on Python 3.11 and `tomli` on older versions

Example from `black`:

https://github.com/psf/black/blob/ef6e079901d53a42dfae4ab10b081ce7a73a47b5/pyproject.toml#L71

https://github.com/psf/black/blob/ef6e079901d53a42dfae4ab10b081ce7a73a47b5/src/black/files.py#L26-L34

As far as I see, the replacement should be pretty straightforward and I'd be happy to provide a PR.

## System information

- OS: macOS 12.6.3
- `python` version: 3.11.2
- `django` version: 4.1.7
- `mypy` version: 1.0.1
- `django-stubs` version: 1.14.0

monosans:
Closes #1408

### DIFF
--- a/mypy_django_plugin/config.py
+++ b/mypy_django_plugin/config.py
@@ -5,7 +5,7 @@ from functools import partial
 from pathlib import Path
 from typing import Any, Callable, Dict, NoReturn, Optional
 
-if sys.version_info >= (3, 11):
+if sys.version_info[:2] >= (3, 11):
     import tomllib
 else:
     import tomli as tomllib
--- a/mypy_django_plugin/config.py
+++ b/mypy_django_plugin/config.py
@@ -1,10 +1,14 @@
 import configparser
+import sys
 import textwrap
 from functools import partial
 from pathlib import Path
 from typing import Any, Callable, Dict, NoReturn, Optional
 
-import tomli
+if sys.version_info >= (3, 11):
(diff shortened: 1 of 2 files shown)


mypy_django_plugin/config.py
```diff
@@ -1,10 +1,14 @@
 import configparser
+import sys
 import textwrap
 from functools import partial
 from pathlib import Path
 from typing import Any, Callable, Dict, NoReturn, Optional
 
-import tomli
+if sys.version_info >= (3, 11):
```
sobolevn:
```suggestion
if sys.version_info[:2] >= (3, 11):
```

sobolevn (approved):
Thanks!

PR #1424, opened 2023-04-04
Add BaseUserCreationForm

monosans:
Added in Django 4.2: <https://docs.djangoproject.com/en/4.2/topics/auth/default/#django.contrib.auth.forms.BaseUserCreationForm>


sobolevn:
Everytime something new is added to Django / changed, some part of me is very scared :)

So, let's start supporting 4.2 then! It is a LTS release, so yeah!

sobolevn (approved):
Thanks! Friendly ping to @intgr about 4.2

PR #1425, opened 2023-04-04
Added decorators.AsyncGetResponseCallable attribute hint

### LINKED ISSUE #1269: Add GetResponseAsyncCallable? / typing of MiddlewareMixin
Given the following middleware, type checking fails:

> error: Incompatible types in "await" (actual type "HttpResponse", expected type "Awaitable[Any]")  [misc]
> error: Returning Any from function declared to return "HttpResponse"  [no-any-return]

```python
from django.utils.decorators import async_only_middleware
from django.utils.deprecation import GetResponseCallable


@async_only_middleware
def my_middleware(get_response: "GetResponseCallable"):
    async def middleware(request: "HttpRequest") -> "HttpResponse":
        response = await get_response(request)
        # …
        return response
    return middleware
```

Adding and using `AsyncGetResponseCallable` fixes this:
```diff
diff --git i/django-stubs/utils/deprecation.pyi w/django-stubs/utils/deprecation.pyi
index e15e0ae0..fa5a8d5d 100644
--- i/django-stubs/utils/deprecation.pyi
+++ w/django-stubs/utils/deprecation.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Callable
-from typing import Any, Protocol
+from typing import Any, Awaitable, Protocol

 from django.http.request import HttpRequest
 from django.http.response import HttpResponse
@@ -32,6 +32,9 @@ class DeprecationInstanceCheck(type):
 class GetResponseCallable(Protocol):
     def __call__(self, __request: HttpRequest) -> HttpResponse: ...

+class AsyncGetResponseCallable(Protocol):
+    def __call__(self, __request: HttpRequest) -> Awaitable[HttpResponse]: ...
+
 class MiddlewareMixin:
     get_response: GetResponseCallable
     def __init__(self, get_response: GetResponseCallable = ...) -> None: ...
```

This could also be handled in `GetResponseCallable` directly (via union), but would be less explicit then.

Maybe it is also a good idea to add support to it to MiddlewareMixin, where it would be typed based on the input of `get_response` passed to `__init__` then.

RyanWalker277:
# Refs

Fixes #1269 

# Changes

Added `AsyncGetResponseCallable` attribute hint in `utils.decorators`


sobolevn (approved):
Thanks!

PR #1427, opened 2023-04-06
Improve Form, FormSet, BoundField, ErrorDict, ErrorList rendering types. Add IntegerField.step_size

monosans:
Added in Django 4.1: <https://docs.djangoproject.com/en/4.2/ref/forms/api/#as-div>


intgr (commented):
Looking at Django code, these methods actually come from `RenderableFormMixin`, which is a base class for both `BaseFormSet`, `BaseForm`. It's better if we fix the inheritance hierarchy here.

But if you don't feel like implementing that right now, this is already an improvement, I can also merge this as is.

monosans:
> It's better if we fix the inheritance hierarchy here.

Done. Let me know if this needs to be broken down into multiple PRs.

intgr (approved):
Awesome work, thanks!

PR #1430, opened 2023-04-07
Add type hint for JSONObject.__init__

realsuayip:
# I have made things!

https://docs.djangoproject.com/en/4.2/ref/models/database-functions/#jsonobject

https://github.com/django/django/blob/2eb1f37260f0e0b71ef3a77eb5522d2bb68d6489/django/db/models/functions/comparison.py#L144


sobolevn (approved):
Thank you!

PR #1431, opened 2023-04-09
fix: don't expose Manager.delete

### LINKED ISSUE #1426: Don't expose delete() on Manager
# Bug report

## What's wrong

Given a model `Foo`, calling `Foo.objects.delete()` on it passes type checking but fails at runtime:

> AttributeError: 'Manager' object has no attribute 'delete'

## How is that should be

Can we remove `delete()` from `Manager`?

Django docs: https://docs.djangoproject.com/en/dev/topics/db/queries/#deleting-objects

> Note that [delete()](https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.query.QuerySet.delete) is the only [QuerySet](https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.query.QuerySet) method that is not exposed on a [Manager](https://docs.djangoproject.com/en/dev/topics/db/managers/#django.db.models.Manager) itself. This is a safety mechanism to prevent you from accidentally requesting Entry.objects.delete(), and deleting all the entries.

## System information

- OS:
- `python` version: 3.10.10
- `django` version: 4.1.7
- `mypy` version: 1.1.1
- `django-stubs-ext` version:

mschoettle:
# I have made things!

Remove `Manager.delete` since it is the only method not exposed from the `QuerySet` by the `Manager`.

Let me know if I should add a test for this.

## Related issues

Closes #1426


sobolevn (approved):
They are defined here on QuerySet: https://github.com/typeddjango/django-stubs/blob/217c58105a69f7ffdb9239d8e02ff3f5e37a9a93/django-stubs/db/models/query.pyi#L111-L112

Thanks!

PR #1436, opened 2023-04-13
Fix ForeignKey queryset filters for cases where the to argument is a …

Kircheneer:
…string like 'app_label.model_name'.

# I have made things!

## Related issues

Refs #626 (probably closes, but my use case is only similar, not exactly the same.

This fixes lookups for related_cls where it was assumed the `to` field on a foreign key would always be a class. If it is a string however (using `app_label.model_name` notation), this breaks some things.

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, Optional, Sequence, Set, Tuple, Type, Union
 
-from django.core.exceptions import FieldError
+from django.core.exceptions import FieldError, FieldDoesNotExist
 from django.db import models
 from django.db.models.base import Model
 from django.db.models.expressions import Expression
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
Thanks, but this would need a test case.

Kircheneer:
> Thanks, but this would need a test case.

Sure! Apparently it has also broken a bunch of things, I will take a look into that.

Kircheneer:
I have switched approaches here a little because I discovered this only applies when abstract base classes are involved. From the docs:
> Relationships defined this way on [abstract models](https://docs.djangoproject.com/en/dev/topics/db/models/#abstract-base-classes) are resolved when the model is subclassed as a concrete model and are not relative to the abstract model’s app_label:

I have changed the logic to accommodate for this and added a test.

mypy_django_plugin/django/context.py
```diff
@@ -385,7 +385,25 @@ def solve_lookup_type(
...
+            pass
+        query_parts = lookup.split("__")
+        try:
+            field = query.get_meta().get_field(query_parts[0])
+            if len(query_parts) == 1:
+                return [], [query_parts[0]], False
+            sub_query = Query(field.related_model).solve_lookup_type("__")
+            entire_query_parts = [query_parts[0], *sub_query[1]]
+            return sub_query[0], entire_query_parts, sub_query[2]
+        except FieldDoesNotExist:
```
sobolevn, line 400: Can we please only catch this for `field = query.get_meta().get_field(query_parts[0])`?

Other line should not be in this `try`

sobolevn (commented):
Thank you! Just one nit left :)

mypy_django_plugin/django/context.py
Kircheneer: Thank you, good point.

sobolevn (approved):
🎉

PR #1438, opened 2023-04-13
Add the RenameIndex migration operation introduced in Django 4.1

### LINKED ISSUE #1437: Missing stub for django.db.migrations.RenameIndex
# Bug report

## What's wrong

The RenameIndex migration operation introduced in Django 4.1 is missing. This results in the following error showing up on certain automatically created migrations:
```
.../0105_rename_order_shipping_date_user_orders_orde_shippin_9d8506_idx_and_more.py:32: error: Module has no attribute "RenameIndex"; maybe "RemoveIndex"?  [attr-defined]
```

## How is that should be

mypy should not report any error, the migration works normally in Django.

## System information

- OS: Ubuntu 22.04
- `python` version: 3.11
- `django` version: 4.1.7
- `mypy` version: 1.1.1

karamanolev:
## Related issues

Closes: #1437

Documentation for RenameIndex: https://docs.djangoproject.com/en/4.2/ref/migration-operations/#renameindex


django-stubs/db/migrations/operations/models.pyi
```diff
@@ -107,6 +107,19 @@ class RemoveIndex(IndexOperation):
...
+    new_name: str
+    old_name: str | None
+    old_fields: Sequence[str] | None
+    def __init__(
+        self,
+        model_name: str,
+        new_name: str,
+        old_name: str | None = ...,
+        old_fields: Sequence[str] | None = ...,
+    ) -> None: ...
```
intgr, line 121: Looks like there are also two `cached_property`ies: old_name_lower, new_name_lower.

I doubt anyone wants to use those, but still best to be complete :)

intgr (changes requested):
Thanks! By the way, for simple stub changes like this, there's no need to create an issue first. Just opening a PR is fine.

django-stubs/db/migrations/operations/models.pyi
karamanolev: Thanks for pointing those out. I completely missed them, but are now added.

intgr (approved):
🎉

karamanolev:
@intgr Sorry to bother you, is there a release coming soon that will include this change?

intgr:
See #1450. There are still several things I'd prefer to do before the release, but my free time has been limited. So possibly I may opt for a quicker but imperfect release rather than delaying it further.

PR #1439, opened 2023-04-15
Add STORAGES setting

federicobond:
STORAGES setting was [introduced](https://docs.djangoproject.com/en/4.2/releases/4.2/#id1) in Django 4.2.


sobolevn (approved):
Thank you!

PR #1441, opened 2023-04-15
Handle repeated field lookups in calls to values_list

leamingrad:
Prior to this change, if checked code was calling the values_list method
on a queryset with a repeated field lookup, for example
Model.objects.values_list("foo", "foo"), then we would fail to type the
resulting tuple correctly. The second argument would be untyped, and the
resulting tuple would have fewer fields than Django would actually give
it.

This problem only affects unnamed, non-flat tuples, as Django will error
if a repeating field lookups is attempted for these other variants.

This change ensures that when we type an unnamed tuple, we refer back to
the original field lookups when passing in the resolved values. We also
add a test to explain the issue, and prevent regressions.

## Related issues

I could not find a related issue for this, but I can raise one if that
would be helpful.

### DIFF
--- a/tests/typecheck/managers/querysets/test_values_list.yml
+++ b/tests/typecheck/managers/querysets/test_values_list.yml
@@ -26,6 +26,27 @@
                     name = models.CharField(max_length=100)
                     age = models.IntegerField()
 
+-   case: values_list_field_repetition
+    main: |
+        from myapp.models import MyUser
+
+        # Values tuples can have the same field repeated
+        values_tuple = MyUser.objects.values_list('name', 'age', 'name').get()
+        reveal_type(values_tuple)  # N: Revealed type is "Tuple[builtins.str, builtins.int, builtins.str]"
+        reveal_type(values_tuple[0])  # N: Revealed type is "builtins.str"
+        reveal_type(values_tuple[1])  # N: Revealed type is "builtins.int"
+        reveal_type(values_tuple[2])  # N: Revealed type is "builtins.str"
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+                class MyUser(models.Model):
+                    name = models.CharField(max_length=100)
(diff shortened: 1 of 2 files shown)


tests/typecheck/managers/querysets/test_values_list.yml
```diff
@@ -26,6 +26,25 @@
...
+        reveal_type(values_tuple[1])  # N: Revealed type is "builtins.str"
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+                class MyUser(models.Model):
+                    name = models.CharField(max_length=100)
```
sobolevn, line 47: Can we please have another field for this test? Something like `age = models.IntegerField`

tests/typecheck/managers/querysets/test_values_list.yml
```diff
@@ -26,6 +26,25 @@
                     name = models.CharField(max_length=100)
                     age = models.IntegerField()
 
+-   case: values_list_field_repetition
+    main: |
+        from myapp.models import MyUser
+
+        # Values tuples can have the same field repeated
+        values_tuple = MyUser.objects.values_list('name', 'name').get()
+        reveal_type(values_tuple)  # N: Revealed type is "Tuple[builtins.str, builtins.str]"
```
sobolevn: And let's use `age` here as well to be sure that this is not just a single type that repeats itself :)

sobolevn (commented):
Thanks, just one proposal :)

tests/typecheck/managers/querysets/test_values_list.yml
leamingrad: Done - good suggestion!

tests/typecheck/managers/querysets/test_values_list.yml
leamingrad: Done - good suggestion!. I've doubled-checked that without the fix the new version of the test does fail.

sobolevn (approved):
Great job! 👍

PR #1442, opened 2023-04-16
Allow lazy translation strings in email contexts

realsuayip:


sobolevn (commented):
Sorry, I cannot find any mentions of this in either:
- https://docs.djangoproject.com/en/4.2/topics/email/
- https://github.com/django/django/blob/main/django/core/mail/message.py

Can you please point to exact places in the docs / source?

realsuayip:
I could not find a reference to it either. However, I think it is a valid (and common) use case. In my experience, any user-facing string tends to be translatable in Django, and in the case they are not, some exception is raised, making it necessary to use `force_str`.

sobolevn:
@adamchainz can you please take a look?

intgr:
It has been many years since I used translations in Django, but for back when I used them, my experience matches realsuayip's description:

> In my experience, any user-facing string tends to be translatable in Django

The relevant documentation is https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#working-with-lazy-translation-objects, quote:

> The result of a gettext_lazy() call **can be used wherever you would use a string (a str object) in other Django code**, but it may not work with arbitrary Python code.

I think the Django documentation overstates it, reality is more nuanced. Obviously, there are `str` params for machine identifiers (e.g. `IntegerField(db_column=_("translate this"))` or `.prefetch_related(_("translate this"))`) that don't reasonably support lazy translations. But for user-readable message strings like email titles, the Django intends to allow lazy strings everywhere.

intgr (approved):
I looked at the Django code, but it's not straightforward. @realsuayip have you tested that these functions work with lazy strings?

As described in my previous message, I'm inclined to accept this PR, but I'll leave final decision to sobolevn.

realsuayip:
@intgr Yes, eventually all functions end up creating `EmailMessage` instances, which itself handles translatable strings.

sobolevn (approved):
Yes, looks like it works locally! Thanks everyone!

PR #1443, opened 2023-04-16
Add stubs for new async model methods from Django 4.2

bigfootjon:
# I have made things!

This adds stubs for the async methods added in Django 4.2: https://docs.djangoproject.com/en/4.2/releases/4.2/#models

## Related issues

N/A


intgr (approved):
Thanks!

PR #1445, opened 2023-04-16
Normalize __init__ 'expressions' argument for subclasses of Func

realsuayip:
In base `Func`, `expressions` is hinted as `Àny`. IMHO, the subclasses should also imitate this behavior.

Some of the subclasses expand `expressions` into named parameters, which I also typed as `Any`.

Now this is not pretty, however I'm not sure what `expressions` really accepts. I could figure out these

str (behaves like `F`)
`Expression`
`Combinable` e.g., `OuterRef` or `F` (this one is missed in many of the cases)
`None` in some cases (?)

If we can agree/conclude on possible types, it would be great. Otherwise `Any` should do fine I guess.


django-stubs/db/models/functions/text.pyi
```diff
@@ -32,7 +31,7 @@ class Concat(Func):
 
 class Left(Func):
     output_field: models.CharField
-    def __init__(self, expression: Expression | str, length: Expression | int, **extra: Any) -> None: ...
+    def __init__(self, expression: Any, length: Any, **extra: Any) -> None: ...
```
adamchainz: https://github.com/django/django/blob/0494efddc422716431b92896899284b6afebb23a/django/db/models/functions/text.py#L157-L159

```suggestion
    def __init__(self, expression: Any, length: Expression | int, **extra: Any) -> None: ...
```

django-stubs/db/models/functions/text.pyi
```diff
@@ -45,9 +44,7 @@ class Lower(Transform): ...
 
 class LPad(Func):
     output_field: models.CharField
-    def __init__(
-        self, expression: Expression | str, length: Expression | int | None, fill_text: Expression = ..., **extra: Any
-    ) -> None: ...
+    def __init__(self, expression: Any, length: Any, fill_text: Any = ..., **extra: Any) -> None: ...
```
adamchainz:
```suggestion
    def __init__(self, expression: Any, length: Expression | int | None, fill_text: Any = ..., **extra: Any) -> None: ...
```

https://github.com/django/django/blob/0494efddc422716431b92896899284b6afebb23a/django/db/models/functions/text.py#L195-L200

django-stubs/db/models/functions/text.pyi
```diff
@@ -58,11 +55,11 @@ class Ord(Transform):
 
 class Repeat(Func):
     output_field: models.CharField
-    def __init__(self, expression: Expression | str, number: Expression | int | None, **extra: Any) -> None: ...
+    def __init__(self, expression: Any, number: Any, **extra: Any) -> None: ...
```
adamchainz:
```suggestion
    def __init__(self, expression: Any, number: Expression | int | None, **extra: Any) -> None: ...
```

https://github.com/django/django/blob/0494efddc422716431b92896899284b6afebb23a/django/db/models/functions/text.py#L231-L236

django-stubs/db/models/functions/text.pyi
```diff
@@ -89,9 +86,7 @@ class StrIndex(Func):
 
 class Substr(Func):
     output_field: models.CharField
-    def __init__(
-        self, expression: Expression | str, pos: Expression | int, length: Expression | int | None = ..., **extra: Any
-    ) -> None: ...
+    def __init__(self, expression: Any, pos: Any, length: Any = ..., **extra: Any) -> None: ...
```
adamchainz:
```suggestion
    def __init__(self, expression: Any, pos: Expression | int, length: Any | None = ..., **extra: Any) -> None: ...
```
https://github.com/django/django/blob/0494efddc422716431b92896899284b6afebb23a/django/db/models/functions/text.py#L344-L346

django-stubs/db/models/functions/window.pyi
```diff
@@ -12,17 +12,17 @@ class DenseRank(Func):
 class FirstValue(Func): ...
 
 class LagLeadFunction(Func):
-    def __init__(self, expression: str | None, offset: int = ..., default: int | None = ..., **extra: Any) -> None: ...
+    def __init__(self, expression: Any, offset: Any = ..., default: Any = ..., **extra: Any) -> None: ...
```
adamchainz:
```suggestion
    def __init__(self, expression: Any, offset: int = ..., default: Any = ..., **extra: Any) -> None: ...
```
https://github.com/django/django/blob/0494efddc422716431b92896899284b6afebb23a/django/db/models/functions/window.py#L40-L53

django-stubs/db/models/functions/window.pyi
```diff
@@ -12,17 +12,17 @@ class DenseRank(Func):
...
 
 class Lag(LagLeadFunction): ...
 class LastValue(Func): ...
 class Lead(LagLeadFunction): ...
 
 class NthValue(Func):
-    def __init__(self, expression: str | None, nth: int = ..., **extra: Any) -> None: ...
+    def __init__(self, expression: Any, nth: Any = ..., **extra: Any) -> None: ...
```
adamchainz:
```suggestion
    def __init__(self, expression: Any, nth: int = ..., **extra: Any) -> None: ...
```

https://github.com/django/django/blob/0494efddc422716431b92896899284b6afebb23a/django/db/models/functions/window.py#L78-L87

django-stubs/db/models/functions/window.pyi
```diff
@@ -12,17 +12,17 @@ class DenseRank(Func):
...
 
 class Ntile(Func):
-    def __init__(self, num_buckets: int = ..., **extra: Any) -> None: ...
+    def __init__(self, num_buckets: Any = ..., **extra: Any) -> None: ...
```
adamchainz:
```suggestion
    def __init__(self, num_buckets: int = ..., **extra: Any) -> None: ...
```
https://github.com/django/django/blob/0494efddc422716431b92896899284b6afebb23a/django/db/models/functions/window.py#L99-L101

adamchainz (changes requested):
Yes `Any` is the right type for a general value in `expressions`, which are passed to `BaseExpression._parse_expressions`:

https://github.com/django/django/blob/0494efddc422716431b92896899284b6afebb23a/django/db/models/expressions.py#L911

...which wraps non-string, non-expression values in `Value()`:

https://github.com/django/django/blob/0494efddc422716431b92896899284b6afebb23a/django/db/models/expressions.py#L202-L208

`Value()` represents a value to be passed directly to the database driver, and drivers support arbitrary types, so `Any` is correct.

That said, many of the subclasses use limited types, so the existing type hints for particular parameters *were* correct. You need to inspect the code and docs to tell if Django limits types.

adamchainz:
@intgr care to review my updates?

intgr:
> Yes `Any` is the right type for a general value in expressions, which are passed to `BaseExpression._parse_expressions`

Hrm? I haven't looked into this in detail and I don't know this domain well. But changing types to `Any` seems like going backwards to me.

Surely these don't reasonably support *every* possible type, like a file, `Ellipsis`, `ModuleType` or `type[AppConfig]`?

Can't we restrict these to combinables and types that are valid as database bind variables for example? Something like `int | str | UUID | ... | Combinable | None`?

If we really can't do any reasonable type checking, we should use `object` instead of `Any`.

adamchainz:
> Can't we restrict these to combinables and types that are valid as database bind variables for example? Something like `int | str | UUID | ... | Combinable | None`?

No, because database drivers support registering extra types at runtime. This is not that uncommon, in a couple of past projects I've registered custom types with pscyopg, to work with postgres extensions.

> If we really can't do any reasonable type checking, we should use `object` instead of `Any`.

I no longer think `object` is a great type to mean "any kind of object", like I did when I wrote [this post](https://adamj.eu/tech/2021/05/07/python-type-hints-use-object-instead-of-any/). If it could be any type, it should be `Any` - it prevents unnecessary restrictions, and also helps users understand what's allowed.

Have you seen anything on why `object` could be preferred?

intgr:
Sorry for for not responding for so long.

> No, because database drivers support registering extra types at runtime. This is not that uncommon, in a couple of past projects I've registered custom types with pscyopg, to work with postgres extensions.

I've never used that myself.

Maybe this is a controversial opinion, but honestly my preference would be to use `# type: ignore` in these special cases involving nonstandard types, and get the benefit of stricter type checking in all remaining code that doesn't do anything special. Instead of sacrificing type checking everywhere just to avoid false positives in special cases.

What do you think?

> Have you seen anything on why `object` could be preferred?

OK, I don't have strong feelings about `Any` vs `object`, my understanding was similar to what you wrote in that blog.

In django-stubs, `Any` is usually a placeholder for "we haven't yet figured out what this is", but `object` is an explicit "we determined that any type is allowed".

Beyond that, there are behavior differences when used with generics (`Any` disables co-/contravariance errors) and return values (`Any` may actually be preferred). But for simple argument types like in this PR, AFAIK it doesn't matter.

adamchainz:
> Maybe this is a controversial opinion, but honestly my preference would be to use `# type: ignore` in these special cases involving nonstandard types, and get the benefit of stricter type checking in all remaining code that doesn't do anything special. Instead of sacrificing type checking everywhere just to avoid false positives in special cases.

I would prefer we got the types to "no false positives" with full coverage first, and then considered making such cases stricter. I think it will help adoption if there are no false positives.

Also, "special cases" could easily be like most queries for certain models using e.g. a custom range type. So it could cause many ignore comments...

> Beyond that, there are behavior differences when used with generics (Any disables co-/contravariance errors) and return values (Any may actually be preferred). But for simple argument types like in this PR, AFAIK it doesn't matter.

Yes my understanding is also that it won't be detectable to users in this case. That said IDE autocomplete users probably understand "Any" more than "object".

sobolevn:
In `typeshed` we ended up with multiple `Any` aliases:
- `Unused`, when any argument can be passed, because it simply does nothing
- `Incomplete`, when we use `Any` as a placeholder for "change me later"

adamchainz:
I like it. Shall we do the same here? How would we find which `Any`s need to be `Incomplete` ?

intgr:
> I would prefer we got the types to "no false positives" with full coverage first, and then considered making such cases stricter. I think it will help adoption if there are no false positives.

Right now the types are already strict. Are you proposing to relax them to `Any` now and then to make them stricter again in the future? I don't understand why we want to go through a relaxing phase, I think I'm missing something.

I think one data point about custom types is that we haven't received issues regarding this so far (none that I know of). Some such use cases may well be fixable e.g. we could add [psyopg2 custom data types](https://www.psycopg.org/docs/extras.html#range-data-types) to the allowed ones in addition to Django's native types. If we relax them to `Any`, we will not receive such feedback.

adamchainz:
Sorry, I left this til now. Yeah, you're right, let's be strict for now until there are issues and fix those more concretely.

adamchainz:
That said, I'm not quite sure how to proceed here. I couldn't find clear evidence of which types to accept for some functions, e.g. `Replace`, where the old types are somewhat wrong (`Combinable` seems incorrect), and [Django's docs](https://docs.djangoproject.com/en/4.2/ref/models/database-functions/#django.db.models.functions.Replace) are fairly minimal.

PR #1447, opened 2023-04-17
Fix ManyToManyRawIdWidget to accept ManyToManyRel argument

### LINKED ISSUE #1435: ManyToManyRawIdWidget expects to be initialized with a ManyToOneRel
# Bug report

## What's wrong

ManyToManyRawIdWidget expects `__init__` to be called with a ManyToOneRel:

```console
error: Argument 1 to "ManyToManyRawIdWidget" has incompatible type "ManyToManyRel"; expected "ManyToOneRel"  [arg-type]
```

This happens because ManyToManyRawIdWidget subclasses ForeignKeyRawIdWidget, which (correcly) takes a ManyToOneRel. 

## How is that should be

ManyToManyRawIdWidget uses a ManyToManyRel, so that should be accepted.

Maybe we could force `__init__` to take ManyToManyRel and `# type: ignore[override]` it, just like it's done for the rel attribute: https://github.com/typeddjango/django-stubs/blob/be609f1962351b99ff228309e8039ea2768bf1e0/django-stubs/contrib/admin/widgets.pyi#L52

That would violate [Liskov substitution principle](https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides), but it seems that this class already does it anyway.

## System information

- OS: linux
- `python` version: 3.8.8
- `django` version: 3.2.18
- `mypy` version: 1.1.1

RyanWalker277:
# Refs

Fixes #1435 

# Changes

This PR modifies the `ForeignKeyRawIdWidget` class to accept both `ManyToOneRel` and `ManyToManyRel`. The issue was that the ForeignKeyRawIdWidget class takes ManyToOneRel but ManyToManyRawIdWidget takes only ManyToManyRel which created problems as ManyToManyRawIdWidget inherits from ForeignKeyRawIdWidget. Modifying ForeignKeyRawIdWidget to accept both types resolves this issue.

Changes Made:

- The rel attribute is now defined as a union of ManyToOneRel and ManyToManyRel.
- The __init__ method of the ForeignKeyRawIdWidget class now accepts both ManyToOneRel and ManyToManyRel.
Impact:


django-stubs/contrib/admin/widgets.pyi
```diff
@@ -38,19 +38,23 @@ class AdminFileWidget(forms.ClearableFileInput): ...
...
 
 class ForeignKeyRawIdWidget(forms.TextInput):
-    rel: ManyToOneRel
+    rel: ManyToOneRel | ManyToManyRel
     admin_site: AdminSite
     db: str | None
     def __init__(
-        self, rel: ManyToOneRel, admin_site: AdminSite, attrs: _OptAttrs | None = ..., using: str | None = ...
+        self,
+        rel: ManyToOneRel | ManyToManyRel,
```
intgr: This seems weird to me.
 
I think `ForeignKeyRawIdWidget` should keep accepting `ManyToOneRel`, and you should add a new `__init__` method to `ManyToManyRawIdWidget` which accepts only `ManyToManyRel`.

pheki: I personally agree with @intgr on this one: `ForeignKeyRawIdWidget` doesn't work with `ManyToManyRel` and `ManyToManyRawIdWidget` doesn't work with `ManyToOneRel`

django-stubs/contrib/admin/widgets.pyi
```diff
@@ -42,15 +42,26 @@ class ForeignKeyRawIdWidget(forms.TextInput):
...
+        using: str | None = ...,
     ) -> None: ...
     def base_url_parameters(self) -> dict[str, str]: ...
     def get_context(self, name: str, value: Any, attrs: _OptAttrs | None) -> dict[str, Any]: ...
     def url_parameters(self) -> dict[str, str]: ...
     def label_and_url_for_value(self, value: Any) -> tuple[str, str]: ...
 
 class ManyToManyRawIdWidget(ForeignKeyRawIdWidget):
-    rel: ManyToManyRel  # type: ignore
+    rel: ManyToManyRel
```
intgr: Did you notice that tests are failing?

You need to restore this `type: ignore`

intgr (changes requested):
Other than the `type: ignore`, I think this is good.

django-stubs/contrib/admin/widgets.pyi
RyanWalker277: Actually I was in class at that time, thought I would look at the ci failures after reaching home 😅
Thanks for pointing this out.

pheki (approved):

intgr (approved):
Thanks

PR #1449, opened 2023-04-20
Fix aiterator

kjagiello:
This changeset fixes the following:
1. `aiterator` does not exist on `RawQuerySet` in Django, so the corresponding typedef is getting removed here
2. `aiterator` typedef was missing on `_QuerySet` even though it is supported by Django
3. `aiterator` typedef was incorrectly typed on `BaseManager`


sobolevn (approved):
Thanks!

django-stubs/db/models/manager.pyi
```diff
@@ -37,7 +37,7 @@ class BaseManager(Generic[_T]):
     # NOTE: The following methods are in common with QuerySet, but note that the use of QuerySet as a return type
     # rather than a self-type (_QS), since Manager's QuerySet-like methods return QuerySets and not Managers.
     def iterator(self, chunk_size: int = ...) -> Iterator[_T]: ...
-    async def aiterator(self, chunk_size: int = ...) -> Iterator[_T]: ...
+    def aiterator(self, chunk_size: int = ...) -> AsyncIterator[_T]: ...
```
intgr, line 40: Weird that the `async` keyword has to be removed as well. But according to this SO post https://stackoverflow.com/a/68911014/177663 that is correct.

intgr (approved):

PR #1451, opened 2023-04-21
Allow any reverse relation using ForeignObjectRel to be type checked

Photonios:
Currently, reverse relations created using just `ForeignObjectRel` aren't checked. A real-world example of this is `django-composite-foreignkey`. Reverse relations created by it aren't discovered by the mypy plugin.

If anyone has any ideas on how to write a proper test for this, let me know.


intgr:
Whoa there. You need to take a few steps back and explain what `ForeignObjectRel` does, how one would use it and what django-stubs should do with it. If possible, with vanilla Django, not an external library like `django-composite-foreignkey`.

As far as I can tell, `ForeignObjectRel` isn't documented in Django, apart from passing mentions in changelogs. Is it considered a stable API in Django? https://www.google.com/search?q=%22ForeignObjectRel%22+site%3Adocs.djangoproject.com

sobolevn:
Yes, adding a test case with pure django objects would help a lot as well.

Photonios:
> Whoa there. You need to take a few steps back and explain what `ForeignObjectRel` does, how one would use it and what django-stubs should do with it. If possible, with vanilla Django, not an external library like `django-composite-foreignkey`.

`ForeignObject` and the related `ForeignObject` are the base classes for `ForeignKey`, `ManyToOneRel` and `ManyToManyRel`.

Although `ForeignKey` does not have support for creating multi-column relationships, the underlying `ForeignObject` does support this. I've demonstrated this in the test I added in my last commit.

Django does not truly support composite/multi-column foreign keys as support for it hasn't yet been added to the schema editor and migration subsystem. `ForeignObject` is frequently used to deal databases modeled/managed outside Django. Although one cannot manage the actual database constraint with Django, the relationship _can_ be properly expressed in Django with `ForeignObjec`.

> As far as I can tell, `ForeignObjectRel` isn't documented in Django, apart from passing mentions in changelogs. Is it considered a stable API in Django? https://www.google.com/search?q=%22ForeignObjectRel%22+site%3Adocs.djangoproject.com

I don't think it's considered a truly stable/documented API, but the fact that changes to are mentioned in the changelogs indicates that the Django developers are aware there are quite a few third-party packages out there relying on it. Making it unlikely to disappear suddenly or subject to intense changes.

Some more googling reveals that `ForeignObject` and related classes are generally considered building blocks for users to build their own relationship abstractions. Examples:

* https://code.djangoproject.com/ticket/19385
* https://devblog.kogan.com/blog/custom-relationships-in-django
* https://schinckel.net/2021/07/14/django-implied-relationship/

On top of that, the class has existed since Django 1.6, introduced here: https://github.com/django/django/commit/266de5f9ae9e9f2fbfaec3b7e4b5fb9941967801

All-in; I think it's worth supporting. It would allow everyone using `ForeignObject` directly or one of the third-party packages to support multi-column foreign keys to have their code type-checked correctly.

---

The test I added demonstrates how to use `ForeignObject` directly to create a virtual multi-column foreign key. The test fails against master.

tests/typecheck/models/test_related_fields.yml
```diff
@@ -87,3 +87,38 @@
               b = models.ForeignKey(Model2, related_name="test2", on_delete=models.CASCADE)
 
               objects = Model4Manager()
+
+-   case: test_related_name_foreign_object_multi_column
+    main: |
+      from app1.models import Model1, Model2
+
+      reveal_type(Model2.model_1) # N: Revealed type is "django.db.models.fields.related.ForeignObject[Any, Any]"
```
intgr: If you also add a check for `Model2().model_1`, does it correctly show up as `Model1` type?

intgr (commented):
Thanks for the explanation, makes lots of sense to me.

tests/typecheck/models/test_related_fields.yml
Photonios: No, it gets typed as `Any`. I will have a look to see if I can fix this.

Photonios: I've pushed a fix and updated the tests. It type-checks correctly now.

mypy_django_plugin/lib/fullnames.py
```diff
@@ -30,7 +31,12 @@
     BASE_MANAGER_CLASS_FULLNAME,
 }
 
-RELATED_FIELDS_CLASSES = {FOREIGN_KEY_FULLNAME, ONETOONE_FIELD_FULLNAME, MANYTOMANY_FIELD_FULLNAME}
+RELATED_FIELDS_CLASSES = {
+    FOREIGN_OBJECT_FULLNAME,
+    FOREIGN_KEY_FULLNAME,
+    ONETOONE_FIELD_FULLNAME,
+    MANYTOMANY_FIELD_FULLNAME,
+}
```
sobolevn:
```suggestion
RELATED_FIELDS_CLASSES = frozenset((
    FOREIGN_OBJECT_FULLNAME,
    FOREIGN_KEY_FULLNAME,
    ONETOONE_FIELD_FULLNAME,
    MANYTOMANY_FIELD_FULLNAME,
))
```

sobolevn (approved):
LGTM! Thank you!

mypy_django_plugin/lib/fullnames.py
Photonios: Thanks! I've applied the suggestion.

mypy_django_plugin/lib/fullnames.py
```diff
@@ -30,7 +31,14 @@
     BASE_MANAGER_CLASS_FULLNAME,
 }
 
-RELATED_FIELDS_CLASSES = {FOREIGN_KEY_FULLNAME, ONETOONE_FIELD_FULLNAME, MANYTOMANY_FIELD_FULLNAME}
+RELATED_FIELDS_CLASSES = frozenset(
+    (
+        FOREIGN_OBJECT_FULLNAME,
+        FOREIGN_KEY_FULLNAME,
+        ONETOONE_FIELD_FULLNAME,
```
intgr, line 38: Both `ForiegnKey` and `OneToOneField` are subclasses of `ForeignObject`, so these could be removed now.

`ManyToManyField` has a separate inheritance hierarchy though.

sobolevn: Good catch!

intgr: Also this will additionally apply to `GenericRelation`, which is another subclass of `ForeignObject`. Seems like a good thing, but ideally we should add a test to make sure it works as expected.

Issue #1450, opened 2023-04-20
Next release planning (4.2.0)


intgr:
The next django-stubs release is way overdue. Since our last release, mypy 1.2.0 and Django 4.2 have been released. `compatible-mypy` version is already upgraded (#1429), and we already have some improvements for Django 4.2 support -- although it's far from complete.

~~This also brings a scope increase to `django-stubs-ext`, it now offers a typehinted base class for `Model` `Meta` (#1375).~~

### New version numbering scheme

As discussed previously in https://github.com/typeddjango/django-stubs/issues/1095#issuecomment-1307066177 and few other places, we want to adopt a new version numbering scheme, similar to [typeshed](https://github.com/python/typeshed) `types-*` packages, where first two digits indicate latest Django version, and the third digit version is just a sequential counter of django-stubs releases. So first release after Django 4.2 will be django-stubs 4.2.0, next 4.2.1 etc.

The downsides of this are:
* It could be considered incompatible with semver. In particular, while we try to minimize the impact, `django-stubs-ext` changes *could* have adverse effects on production code.
* We will lose the ability to communicate major or minor changes via the version number. But AFAIK we haven't really used that in the history of TypedDjango.

**If someone has arguments against this version scheme change, now is the last opportunity to voice them.**

This new scheme would apply to both `django-stubs` and `django-stubs-ext` (refs #1409). To avoid confusion, we should keep these version numbers in sync in the future (either we always release both, or if django-stubs-ext release is skipped, the next django-stubs-ext release can skip some numbers to catch up to django-stubs).

I want to re-iterate that this does not in any way change our policy for older Django versions. Type stubs have always tracked the latest Django version, but Django itself tends to be fairly backwards compatible. And we will continue running CI tests with older Django versions and make sure we don't break them in a big way.

### Action items

* [x] We should add CI testing with Django 4.2.
* [x] Check if [djangorestframework-stubs](https://github.com/typeddjango/djangorestframework-stubs) CI needs changing.
* [x] Adding 4.2 to `DJANGO_COMMIT_REFS` in `typecheck_tests.py`. I'm afraid updating the ignore lists will be a significant amount of work. Perhaps it's not a release blocker.
* [x] There are a few recent unreviewed pull requests, we should work through those.
* [x] Document how to use `django-stubs-ext` `TypedModelMeta`.

Please let me know if I missed something.

intgr:
[...]

PR #1452, opened 2023-04-26
Run CI with Django 4.2 and declare support in package metadata

intgr:
* An addition to `requirements.txt` is enough to use Django version 4.2 in CI.
* Added `Framework :: Django :: 4.2` package classifier

Refs #1450.


intgr:
Looks like Django 4.2 is not compatible with Python 3.7. I think it's time to drop testing with Python 3.7 anyway, it's EOL in 2 months. And it will unblock pre-commit 3.x update.

sobolevn (approved):

PR #1453, opened 2023-04-26
CI: Stop testing with Python 3.7

intgr:
Python 3.7 will be officially EOL on 27 June 2023, that's in 2 months.

This simplifies updating to Django 4.2 (#1452), which no longer supports Python 3.7, and allows upgrading pre-commit (#1422).

I won't yet **require** Python 3.8, because the package is still compatible with 3.7, will update requirement after the 4.2.0 release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks!

adamchainz:
🔥 bye bye 3.7 🔥

PR #1455, opened 2023-04-26
CI: Add Django 4.2 to typecheck and update other Django versions

intgr:
Wow, this was a lot easier than I expected. There was no need to update any ignores!

* Updated the default Django version in typecheck_tests to 4.2.
* Made the `typecheck_tests.py` script executable and added a shebang line.

Refs #1450.


intgr:
Merging this myself as trivial change.

PR #1456, opened 2023-04-26
Rename BaseModelMeta to TypedModelMeta and document it

intgr:
Changes and documents the feature added in #1375.

@sobolevn I've actually changed my mind and I now prefer the name `TypedModelMeta`, what do you think?

I think this way it's more explicit what it means -- it's not actually a "base class" (it's only an alias of `object` at runtime) -- and this one more clearly describes its function.

Refs #1450.


sobolevn (approved):
LGTM 👍

PR #1458, opened 2023-04-27
Version 4.2.0 release (django-stubs, django-stubs-ext)

### LINKED ISSUE #1095: Release schedule

### LINKED ISSUE #1409: django-stubs-ext version schema
# Bug report

## What's wrong

I want to automatically update to newer versions of libraries, and so I use a version specifier like `~=1.0` to upgrade to versions  `1.x.x`, hoping that dependencies follow semver. When the version starts with `0`, I need to use `~=0.1.0` to upgrade to all `0.1.x` versions because semver defines bumping the minor version (i.e. `0.2.0`) as a breaking change.

I think django-stubs follows this, at least in spirit, but django-stubs-ext only bumps the minor version: https://pypi.org/project/django-stubs-ext/#history

## How is that should be

I would love for django-stubs-ext to either:

- Bump the patch release for mostly-backward-compatible changes (similar to how django-stubs bumps the minor version), or
- My preference: bump the major version to `1` to match django-stubs and continue bumping the minor version the way we have been

## System information

N/A

### LINKED ISSUE #1450: Next release planning (4.2.0)

intgr:
## Related issues

* Closes #1450
* Closes #1095
* Closes #1409


README.md
```diff
@@ -49,25 +49,12 @@ This fully working [typed boilerplate](https://github.com/wemake-services/wemake
...
-| 1.6.0        | 0.780 | 2.2.x \|\| 3.x | ^3.6
-| 1.5.0        | 0.770 | 2.2.x \|\| 3.x | ^3.6
-| 1.4.0        | 0.760 | 2.2.x \|\| 3.x | ^3.6
-| 1.3.0        | 0.750 | 2.2.x \|\| 3.x | ^3.6
-| 1.2.0        | 0.730 | 2.2.x | ^3.6
-| 1.1.0        | 0.720 | 2.2.x | ^3.6
-| 0.12.x       | old semantic analyzer (<0.711), dmypy support | 2.1.x | ^3.6
+| django-stubs | mypy version | django version                   | python version
+|--------------|--------------|----------------------------------| ---- |
+| 4.2.0        | 1.2.x        | 3.2.x or 4.0.x or 4.1.x or 4.2.x | ^3.7
```
sobolevn: Maybe like this?

```suggestion
| 4.2.0        | 1.2.x        | 4.2.x (other versions are partly supported) | ^3.8
```

Note that we don't support 3.7 anymore.

django_stubs_ext/setup.py
```diff
@@ -10,9 +10,11 @@
     "typing-extensions",
 ]
 
+# NB! Keep version in sync with django-stubs.
```
sobolevn:
```suggestion
# NB! Keep major.minor version in sync with django-stubs.
```

sobolevn (commented):
Thank you!

README.md
intgr: > Note that we don't support 3.7 anymore.

I just removed testing with Python 3.7, but we're still fully compatible Python 3.7.

I want to write in release notes that this is the last django-stubs that works with Python 3.7, and drop it officially after that.

django_stubs_ext/setup.py
intgr: I think we should keep all `major.minor.patch` in sync. Otherwise it's just awkward when django-stubs 4.2.10 uses django-stubs-ext 4.2.3.

That's what I described in #1450

> This new scheme would apply to both django-stubs and django-stubs-ext (refs https://github.com/typeddjango/django-stubs/issues/1409). To avoid confusion, we should keep these version numbers in sync in the future (either we always release both, or if django-stubs-ext release is skipped, the next django-stubs-ext release can skip some numbers to catch up to django-stubs).

But it's a good idea to make this comment more precise.

README.md
sobolevn: But, `django` says:

> Django 4.2 supports Python 3.8, 3.9, 3.10, and 3.11. We highly recommend and only officially support the latest release of each series.

In my opinion, it is better to raise this contraint. Otherwise, people might think that 3.7 + 4.2 is supported.

django_stubs_ext/setup.py
sobolevn: Ok, I don't see a problem with that! Releasing empty releases is not a big problem. Getting out of sync is.

README.md
intgr: Yes, but django-stubs doesn't require Django 4.2. As you note, older Django versions are "partly supported".

People may still be using django-stubs with Django 3.2 and Python 3.7. And continuing to support this for one release doesn't require us to do any extra work.

intgr: As for "4.2.x (other versions are partly supported)", maybe we should still list Django versions that our CI is testing with? So maybe add a new column "Django partial support" that lists "4.1, 4.0, 3.2"?

intgr: Because if we ever need to do a hard break in Django compatibility, this would be a good way to communicate that.

sobolevn: Yes, good idea! 👍

sobolevn: I am going to insist on removing 3.7 though:
1. It is 2 month before sunsetting
2. It is not tested anymore (so any commit might break this support)
3. Some django versions do not support it
4. It is not very popular among our users

Снимок экрана 2023-04-27 в 14 04 54

intgr: > It is 2 month before sunsetting

So it's still supported by upstream.

> It is not tested anymore (so any commit might break this support)

Yes, but I want to make this release right away when we get this over with. I'm not merging any new commits between now and the release.

> It is not very popular among our users

It's over 5%, seems significant to me.

sobolevn: Ok :)

README.md
```diff
@@ -49,25 +49,12 @@ This fully working [typed boilerplate](https://github.com/wemake-services/wemake
...
-| 1.3.0        | 0.750 | 2.2.x \|\| 3.x | ^3.6
-| 1.2.0        | 0.730 | 2.2.x | ^3.6
-| 1.1.0        | 0.720 | 2.2.x | ^3.6
-| 0.12.x       | old semantic analyzer (<0.711), dmypy support | 2.1.x | ^3.6
+| django-stubs | Mypy version | Django version | Django partial support | Python version |
+|--------------|--------------|----------------|------------------------|----------------|
+| 4.2.0        | 1.2.x        | 4.2            | 4.1, 4.0, 3.2          | 3.7 - 3.11     |
+| 1.16.0       | 1.1.x        | 4.1            | 4.0, 3.2               | 3.7 - 3.11     |
+| 1.15.0       | 1.0.x        | 4.1            | 4.0, 3.2               | 3.7 - 3.11     |
+| 1.14.0       | 0.990+       | 4.1            | 4.0, 3.2               | 3.7 - 3.11     |
```
intgr, line 57: Does this look good to you now? Should I go ahead with the release?

sobolevn: Yes, thank you! 🎉 🥳

django_stubs_ext/setup.py
intgr: > I think we should keep all major.minor.patch in sync.

Oh I remember now what my main reason for doing this was: if we keep them in sync, don't need separate `git tag`s for django-stubs and django-stubs-ext, and thus also don't need seperate relase notes.

MichaelVoelkel:
Sorry, one question, what do you mean with "partial support" even? We need to stick with Django 3.2 for now, so does it now make sense to go down to a version that does not say "partial support" or is it still best to use the latest version?

intgr:
The changes between major Django versions aren't that big. I think you're better off upgrading, because `django-stubs` gets more accurate with every version, most improvements apply to all relevant Django versions.

When you experience mypy errors due to differences between Django versions, I think `# type: ignore` comments are a better way to deal with it than staying stuck with old django-stubs. But YMMV.

PR #1459, opened 2023-04-27
Completely drop Python 3.7 compatibility

intgr:
* Django 4.2 no longer supports Python 3.7
* Our CI dependency `pre-commit` no longer supports Python 3.7 (#1422)
* Python 3.7 support ends in 2 months (https://endoflife.date/python)
* We can clean up a few type stubs

## Related issues

* Related: #1453


django-stubs/core/mail/message.pyi
```diff
@@ -66,12 +63,8 @@ class SafeMIMEMultipart(MIMEMixin, MIMEMultipart):  # type: ignore
     def __setitem__(self, name: str, val: str) -> None: ...
 
 _AttachmentContent: TypeAlias = bytes | EmailMessage | Message | SafeMIMEText | str
-# switch to tuple once https://github.com/python/mypy/issues/11098 is fixed
-# remove Optional once python 3.7 is dropped (Tuple[str | None, ...] works with mypy on py3.10)
 _AttachmentTuple: TypeAlias = (
-    Tuple[str, _AttachmentContent]
-    | Tuple[Optional[str], _AttachmentContent, str]
-    | Tuple[str, _AttachmentContent, None]
+    tuple[str, _AttachmentContent] | tuple[str | None, _AttachmentContent, str] | tuple[str, _AttachmentContent, None]
```
intgr, line 67: Hrm? These comments were referencing Python 3.7, but I'm not sure whether this has anything to do with Python 3.7.

`str | None` or `tuple[str]` don't work on Python 3.8 runtime either.

But AFAIK it doesn't matter since `.pyi` files are only evaluated by mypy, not Python itself. Maybe I'm missing something.

sobolevn: These files are parsed by python, but not executed. I don't remember the origin of this comment, but I am happy that it is removed now :)

intgr: > These files are parsed by python, but not executed.

Genuinely interested, when does Python parse `.pyi` files?

Is this somehow related to the `ast` module (which used to be `typed_ast` but got merged into Python 3.8's `ast`)?

sobolevn: Yes, exactly. `mypy` uses `ast.parse` to get the Python's AST and then `mypy` builds its own AST around it.

intgr:
Anyone who wants to review this?

sobolevn (commented):
LGTM, except this change: https://github.com/typeddjango/django-stubs/blob/master/.pre-commit-config.yaml#L19

intgr:
Ah thanks! You already mentioned it in https://github.com/typeddjango/djangorestframework-stubs/pull/399 but I forgot.

sobolevn (approved):
Thank you!

PR #1460, opened 2023-04-27
Fix missing `__init__.py` in django_stubs_ext.db

monosans:
The django_stubs_ext 4.2.0 release does not have a `db` package because it is missing `__init__.py`.


adamchainz:
@monosans what issues are you seeing? After PEP420, folders without `__init__.py` should still function as packages.

@intgr if this is an issue, let's add https://github.com/adamchainz/flake8-no-pep420 to flake8 configuration.

intgr:
This PR references the issue described here: https://github.com/typeddjango/django-stubs/issues/1450#issuecomment-1525663652

~~I tested `python setup.py sdist` from this branch and it still produced an archive without `db/models.py`. Unless I messed something up.~~

~~I think something needs to be changed in `django_stubs_ext/setup.py`, but I haven't yet looked into it in depth.~~

intgr:
My bad, this PR indeed fixes the issue, thanks!

```
├── django-stubs-ext-4.2.0
│   ├── PKG-INFO
│   ├── README.md
│   ├── django_stubs_ext
│   │   ├── __init__.py
│   │   ├── aliases.py
│   │   ├── annotations.py
│   │   ├── db
│   │   │   ├── __init__.py
│   │   │   └── models.py
```

intgr (approved):
Thanks for the help!

PR #1462, opened 2023-04-27
Fix _make_token_with_timestamp() argument secret

christianbundy:
Problem: the `_make_token_with_timestamp` stub says that it takes a third argument, `legacy: bool`, but in reality it's `secret: str | bool`.

Solution: Update the stub.

Link: https://github.com/django/django/blob/720abed34304ec410c04a2e7a7dec7be4dae0d61/django/contrib/auth/tokens.py#L84


django-stubs/contrib/auth/tokens.pyi
```diff
@@ -10,7 +10,7 @@ class PasswordResetTokenGenerator:
     algorithm: str
     def make_token(self, user: AbstractBaseUser) -> str: ...
     def check_token(self, user: AbstractBaseUser | None, token: str | None) -> bool: ...
-    def _make_token_with_timestamp(self, user: AbstractBaseUser, timestamp: int, legacy: bool = ...) -> str: ...
+    def _make_token_with_timestamp(self, user: AbstractBaseUser, timestamp: int, secret: str | bool = ...) -> str: ...
```
intgr: Does `bool` really make sense?

This gets passed to `salted_hmac()`, which calls `force_bytes(secret)`. Should it be `str | bytes` instead?

intgr:
@christianbundy Ping?

django-stubs/contrib/auth/tokens.pyi
christianbundy: I think you're right, I can't think of why I would've done `bool` here since it's the same type as https://github.com/typeddjango/django-stubs/pull/1411

christianbundy:
```suggestion
    def _make_token_with_timestamp(self, user: AbstractBaseUser, timestamp: int, secret: str | bytes = ...) -> str: ...
```

intgr (approved):
Thanks!

PR #1463, opened 2023-04-27
Tweak package license metadata and rename files

intgr:
* Fixes `SetuptoolsDeprecationWarning: The license_file parameter is deprecated, use license_files instead.`
* Turns out that `django-stubs-ext` package was missing LICENSE.txt because the `license_file=` was pointing to `../` parent directory
* Moved this key from `setup.cfg` to `setup.py`
* Made `setup.py` files executable and added shebang lines, since I'm used to running `./setup.py`
* Renamed `LICENSE.txt` to `LICENSE.md`


django_stubs_ext/setup.py
```diff
@@ -19,6 +20,7 @@
     long_description=readme,
     long_description_content_type="text/markdown",
     license="MIT",
+    license_files=["LICENSE.txt"],
```
sobolevn: I think that it is better to use `.md` file instead. Let's just run `mv LICENCE.txt LICENCE.md`

sobolevn (approved):

PR #1465, opened 2023-04-27
Fix decorator return types

### LINKED ISSUE #1464: `sensitive_variables` decorator causes incorrect return type
# Bug report

## What's wrong

A function decorated with `sensitive_variables` appears to return `Any` instead of it's actual return type. For example,
```python
from django.views.decorators.debug import sensitive_variables

@sensitive_variables("num")
def my_fn(num: int) -> int:
    return num

def other_fn() -> int:
    return my_fn(1)  # <<< mypy error: Returning Any from function declared to return "int"  [no-any-return]
```

## How is that should be

The above code snipped should not produce an error. Mypy should detect that `my_fn(1)` returns an `int`.

## System information

- OS:
- `python` version: 3.11.3
- `django` version: 4.1.8
- `mypy` version: 1.1.1

Kostia-K:
Some decorators do not preserve the type of the decorated function. Code sample in the related issue.

## Related issues

- Fixes https://github.com/typeddjango/django-stubs/issues/1464


django-stubs/views/decorators/http.pyi
```diff
@@ -11,6 +11,6 @@ require_GET: Callable[[_F], _F]
 require_POST: Callable[[_F], _F]
 require_safe: Callable[[_F], _F]
 
-def condition(etag_func: Callable | None = ..., last_modified_func: Callable | None = ...) -> Callable: ...
+def condition(etag_func: Callable | None = ..., last_modified_func: Callable | None = ...) -> Callable[[_F], _F]: ...
```
sobolevn: Can you please also type `etag_func` and `last_modified_func` while we are at it? https://docs.djangoproject.com/en/4.2/topics/conditional-view-processing/#the-condition-decorator

sobolevn (commented):
Thank you! Good catch

django-stubs/views/decorators/http.pyi
Kostia-K: Yes, good idea. I've updated both according to the docs.
Also, in the same commit, changed sensitive variables and post parameters from `Any` to `str` because that's what Django expects.

sobolevn (approved):
Thank you! 👍

PR #1468, opened 2023-04-29
Drop Django 4.0 compatibility

intgr:
Django 4.0 has been EOL for 3 weeks, no reason to keep running CI for it (https://endoflife.date/django)

But since we still support Django LTS 3.2, the 4.0 support is likely to continue being functional anyway.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,8 +74,6 @@ jobs:
           - python-version: '3.8'
             django-version: '2.2'
           - python-version: '3.11'
-            django-version: '4.0'
-          - python-version: '3.11'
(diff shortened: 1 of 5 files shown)


intgr:
Also, should we drop testing with Django 2.2? It has been EOL for 13 months. We don't really advertise 2.2 support anywhere, but CI is still testing with it.

OTOH it hasn't really been a maintenance burden to keep the CI running with 2.2. I don't have any strong feelings either way.

sobolevn:
I think we should drop 2.2
Because in theory 4.2 and older versions can conflict in definitions.

sobolevn (approved):

PR #1469, opened 2023-04-30
Add `robust` parameter to `transaction.on_commit`

realsuayip:
## Related issues

https://github.com/django/django/pull/15766


django-stubs/db/transaction.pyi
```diff
@@ -20,7 +20,7 @@ def get_rollback(using: str | None = ...) -> bool: ...
 def set_rollback(rollback: bool, using: str | None = ...) -> None: ...
 @contextmanager
 def mark_for_rollback_on_error(using: str | None = ...) -> Iterator[None]: ...
-def on_commit(func: Callable, using: str | None = ...) -> None: ...
+def on_commit(func: Callable, using: str | None = ..., robust: bool = ...) -> None: ...
```
sobolevn: Can you please also add better `Callable` type arguments for `func` param?

django-stubs/db/transaction.pyi
```diff
@@ -20,7 +20,7 @@ def get_rollback(using: str | None = ...) -> bool: ...
 def set_rollback(rollback: bool, using: str | None = ...) -> None: ...
 @contextmanager
 def mark_for_rollback_on_error(using: str | None = ...) -> Iterator[None]: ...
-def on_commit(func: Callable, using: str | None = ...) -> None: ...
+def on_commit(func: Callable[[], Any], using: str | None = ..., robust: bool = ...) -> None: ...
```
sobolevn: I think that return value of `func` is just ignored, isn't it? In this case it is `Callable[[], object]`

Please, update it here as well https://github.com/typeddjango/django-stubs/pull/1469/files#diff-b7fe04c8e35ac07ecc22682f8b72c3610500ec8f0d60a30bca745f1caca6d315R118

sobolevn (approved):
Thanks!

PR #1484, opened 2023-05-10
Allow using async StreamingHttpResponse

### LINKED ISSUE #1483: StreamingHttpResponse should now support async iterators
In Django 4.2, support for async streaming responses was added (https://docs.djangoproject.com/en/4.2/releases/4.2/#requests-and-responses)

I believe we should update the declaration here:
https://github.com/typeddjango/django-stubs/blob/8b1e20c35a2917f6ea220b4c89201d7e2e3c8692/django-stubs/http/response.pyi#L107

to become something like:
```python
class StreamingHttpResponse(HttpResponseBase, Iterable[bytes]):
    streaming_content = _PropertyDescriptor[Iterable[object] | AsyncIterator[object], Iterator[bytes] | AsyncIterator[bytes]]()
    def __init__(self, streaming_content: Iterable[object] | AsyncIterator[object]= ..., *args: Any, **kwargs: Any) -> None: ...
    def __iter__(self) -> Iterator[bytes]: ...
    async def __aiter__(self) -> AsyncIterator[bytes]: ...
    def getvalue(self) -> bytes: ... 
```

Alexerson:
# I have made things!
This is a first attempt at implementing types for async StreamingHttpResponse.

## Related issues
- Closes #1483 

Hello, this is a draft for https://github.com/typeddjango/django-stubs/issues/1483.
I don’t have more time to look at it today, and I don’t know how to properly implement the following:
- if the setter of the property, or the value passed to `__init__` is an `Iterable`, then the getter should be an `Iterator`, but if the setter or the value is an `AsyncIterable`, then the getter should be an `AsyncIterator`. 

I feel I want an overload, but those won’t work with `_PropertyDescriptor`, or a "linked TypeVar", but that doesn’t make sense.  If anyone knows how to do this, feel free to help me. Otherwise I’ll do more research a bit later this week.

### DIFF
--- a/django-stubs/http/response.pyi
+++ b/django-stubs/http/response.pyi
@@ -105,6 +105,7 @@ class HttpResponse(HttpResponseBase, Iterable[bytes]):
     def getvalue(self) -> bytes: ...
 
 class StreamingHttpResponse(HttpResponseBase, Iterable[bytes], AsyncIterable[bytes]):
+    is_async: bool
     streaming_content = _PropertyDescriptor[
         Iterable[object] | AsyncIterable[object], Iterator[bytes] | AsyncIterator[bytes]
     ]()
@@ -114,7 +115,6 @@ class StreamingHttpResponse(HttpResponseBase, Iterable[bytes], AsyncIterable[byt
     def __iter__(self) -> Iterator[bytes]: ...
     def __aiter__(self) -> AsyncIterator[bytes]: ...
     def getvalue(self) -> bytes: ...
-    is_async: bool
 
 class FileResponse(StreamingHttpResponse):
     file_to_stream: BytesIO | None
--- a/django-stubs/http/response.pyi
+++ b/django-stubs/http/response.pyi
@@ -1,8 +1,8 @@
 import datetime
-from collections.abc import Iterable, Iterator
+from collections.abc import AsyncIterable, AsyncIterator, Iterable, Iterator
 from io import BytesIO
 from json import JSONEncoder
-from typing import Any, TypeVar, overload, type_check_only
+from typing import Any, Generic, TypeVar, overload, type_check_only
 
 from django.http.cookie import SimpleCookie
 from django.utils.datastructures import CaseInsensitiveMapping, _PropertyDescriptor
@@ -104,10 +104,13 @@ class HttpResponse(HttpResponseBase, Iterable[bytes]):
     def __iter__(self) -> Iterator[bytes]: ...
     def getvalue(self) -> bytes: ...
 
-class StreamingHttpResponse(HttpResponseBase, Iterable[bytes]):
-    streaming_content = _PropertyDescriptor[Iterable[object], Iterator[bytes]]()
-    def __init__(self, streaming_content: Iterable[object] = ..., *args: Any, **kwargs: Any) -> None: ...
+_I = TypeVar("_I", Iterable[object], AsyncIterable[object])
+
+class StreamingHttpResponse(HttpResponseBase, Iterable[bytes], AsyncIterable[bytes], Generic[_I]):
+    streaming_content = _PropertyDescriptor[_I, Iterator[bytes] | AsyncIterator[bytes]]()
+    def __init__(self, streaming_content: _I = ..., *args: Any, **kwargs: Any) -> None: ...
     def __iter__(self) -> Iterator[bytes]: ...
+    def __aiter__(self) -> AsyncIterator[bytes]: ...
     def getvalue(self) -> bytes: ...
 
 class FileResponse(StreamingHttpResponse):
--- a/django-stubs/http/response.pyi
+++ b/django-stubs/http/response.pyi
@@ -2,7 +2,7 @@ import datetime
 from collections.abc import AsyncIterable, AsyncIterator, Iterable, Iterator
 from io import BytesIO
 from json import JSONEncoder
-from typing import Any, Generic, TypeVar, overload, type_check_only
+from typing import Any, TypeVar, overload, type_check_only
 
 from django.http.cookie import SimpleCookie
 from django.utils.datastructures import CaseInsensitiveMapping, _PropertyDescriptor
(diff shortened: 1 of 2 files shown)


django-stubs/http/response.pyi
```diff
@@ -104,10 +104,13 @@ class HttpResponse(HttpResponseBase, Iterable[bytes]):
     def __iter__(self) -> Iterator[bytes]: ...
     def getvalue(self) -> bytes: ...
 
-class StreamingHttpResponse(HttpResponseBase, Iterable[bytes]):
-    streaming_content = _PropertyDescriptor[Iterable[object], Iterator[bytes]]()
-    def __init__(self, streaming_content: Iterable[object] = ..., *args: Any, **kwargs: Any) -> None: ...
+_I = TypeVar("_I", Iterable[object], AsyncIterable[object])
```
sobolevn: Isn't it `_I = TypeVar("_I", bound=Union[Iterable[Any], AsyncIterable[Any]])`?

tests/typecheck/views/test_function_based_views.yml
```diff
@@ -1,5 +1,5 @@
--   case: http_response
-    main: |
+- case: http_response
```
sobolevn: I prefer to not have style-only changes, clean diff and correct history are more important to me :)

sobolevn (commented):
Thank you!

django-stubs/http/response.pyi
intgr: [According to Django docs](https://docs.djangoproject.com/en/4.2/ref/request-response/#id6), `streaming_content` is "An iterator of the response content, **bytestring** encoded according to HttpResponse.charset."

Why use `object` here, shouldn't it be `Iterable[bytes], AsyncIterable[bytes]`?

tests/typecheck/views/test_function_based_views.yml
Alexerson: Sorry, my editor was auto-formatting the file and I forgot to revert that bit. I'll fix it.

django-stubs/http/response.pyi
Alexerson: If you look at the code, Django will happily take anything that can be converted to bytes with the utility function: str, bytes and gettext string are all valid.

Alexerson: And most likely yes for the bound part.

intgr: > If you look at the code, Django will happily take anything that can be converted to bytes with the utility function: str, bytes and gettext string are all valid.

Sorry, I didn't realize that this loose behavior already existed before your changes, it was actually introduced in #712. So it's out of scope for this PR. Feel free to ignore my comment :)

But FWIW I don't agree with this approach. "It works" doesn't mean it's correct or intentional behavior, or it will continue to work with future Django versions. The point of type checkers is to introduce additional strictness to the otherwise fast-and-loose world of Python typing. I **want** type stubs to be stricter if that means it finds more bugs or leads to more precise code. Granted, we have to find a reasonable tradeoff between strictness and usefulness.

Alexerson: Yes, you’re right, I mostly took what was already done for sync streaming responses. 

As mentioned in #712 , in this case, the Django code base is actually explicitly testing that strings are encoded (https://github.com/django/django/blob/6e32d1fa1dafd0c9cd9f93997ecebb26cd9a1b62/tests/httpwrappers/tests.py#L672), so I believe it’s more a question of having the documentation fixed. I’ll suggest the change. https://code.djangoproject.com/ticket/34556

Alexerson:
Ok, I pushed some changes. The tests are happy, but I’m not :upside_down_face: 
In essence, I don’t think it’s possible to do conditional typing, and therefore, I don’t think we can have anything better than an union for the getter signature.
If anyone more clever than me can jump in, it would be great.

django-stubs/http/response.pyi
Alexerson: This was fixed in the doc.
https://docs.djangoproject.com/en/dev/ref/request-response/#django.http.StreamingHttpResponse

intgr: So according to updated documentation it should be `bytes | str | StrPromise`, not `object`?

Alexerson: And `| memoryview` as well then. As this would need to change for `HttpResponse` as well, I don’t know if this PR is the best place to do that, but I’ll follow your guideline.

django-stubs/http/response.pyi
```diff
@@ -104,11 +104,17 @@ class HttpResponse(HttpResponseBase, Iterable[bytes]):
     def __iter__(self) -> Iterator[bytes]: ...
     def getvalue(self) -> bytes: ...
 
-class StreamingHttpResponse(HttpResponseBase, Iterable[bytes]):
-    streaming_content = _PropertyDescriptor[Iterable[object], Iterator[bytes]]()
-    def __init__(self, streaming_content: Iterable[object] = ..., *args: Any, **kwargs: Any) -> None: ...
+class StreamingHttpResponse(HttpResponseBase, Iterable[bytes], AsyncIterable[bytes]):
+    streaming_content = _PropertyDescriptor[
```
intgr, line 109: Nit: it's common to have properties and attributes at the top.

```suggestion
    is_async: bool
    streaming_content = _PropertyDescriptor[
```

django-stubs/http/response.pyi
```diff
@@ -104,11 +104,17 @@ class HttpResponse(HttpResponseBase, Iterable[bytes]):
...
+        Iterable[object] | AsyncIterable[object], Iterator[bytes] | AsyncIterator[bytes]
+    ]()
+    def __init__(
+        self, streaming_content: Iterable[object] | AsyncIterable[object] = ..., *args: Any, **kwargs: Any
+    ) -> None: ...
     def __iter__(self) -> Iterator[bytes]: ...
+    def __aiter__(self) -> AsyncIterator[bytes]: ...
     def getvalue(self) -> bytes: ...
+    is_async: bool
```
intgr:
```suggestion
```

intgr (approved):
LGTM, thanks!

PR #1487, opened 2023-05-10
Add set/get typing for spatial Model Fields

### LINKED ISSUE #1479: contrib.gis PointField typing
Hi, I would like to help and add a stronger typing to geo fields (the one I use is PointField).
My issue today is that I have a `coordinates = models.PointField()` field, but when I use the value, it is typed as any.

I believe we could have the same approach in [PointField](https://github.com/typeddjango/django-stubs/blob/5f57a42110e04ac1bd451ac42923da89e5c3b7f0/django-stubs/contrib/gis/db/models/fields.pyi#LL115C18-L115C31) as in [CharField](https://github.com/typeddjango/django-stubs/blob/5f57a42110e04ac1bd451ac42923da89e5c3b7f0/django-stubs/db/models/fields/__init__.pyi#L273)

What do you think? Let me known if I should work on this.

martinlehoux:
## Related issues

- Closes #1479

### DIFF
--- a/django-stubs/contrib/gis/db/models/fields.pyi
+++ b/django-stubs/contrib/gis/db/models/fields.pyi
@@ -13,6 +13,7 @@ from django.contrib.gis.geos import (
     Polygon,
 )
 from django.core.validators import _ValidatorCallable
+from django.db.models.expressions import Combinable
 from django.db.models.fields import Field, _ErrorMessagesT, _FieldChoices
 from django.utils.functional import _StrOrPromise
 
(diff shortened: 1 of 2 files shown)


intgr (approved):
Thanks! LGTM

intgr:
Any reason to omit `ExtentField` and `RasterField`? Looks like `RasterField` uses type `GDALRaster`, not sure what `ExtentField` is. Feel free to submit another PR 😉

martinlehoux:
I'd say the reason was that they don't look the same xD

martinlehoux:
Weirdly, I can't seem to make this work in my project. CharField is well typed as `str`, but PointField wont get typed as `Point`. Any idea what could be different here?

PR #1488, opened 2023-05-10
Improve core.files.storages types for Django 4.2

Alexerson:
Hello :wave: ,

this PR is adding types for the new storages concepts added in 4.2:
https://docs.djangoproject.com/en/4.2/ref/files/storage/#django.core.files.storage.storages
&
https://docs.djangoproject.com/en/4.2/ref/files/storage/#the-inmemorystorage-class

I didn’t find any tests for the existing storages concepts so didn’t add any. Let me know if you’d like for me to add some.


django-stubs/core/files/storage.pyi
```diff
@@ -42,8 +43,37 @@ class FileSystemStorage(Storage):
     @property
     def directory_permissions_mode(self) -> int | None: ...
 
-class DefaultStorage(LazyObject): ...
+class InvalidStorageError(ImproperlyConfigured): ...
 
-default_storage: Any
+class StorageHandler:
+    def __init__(self) -> None: ...
```
intgr: Looks like `StorageHandler.__init__` takes an optional `backends` argument

```suggestion
    def __init__(self, backends: dict[str, Storage] | None = ...) -> None: ...
```

django-stubs/core/files/storage.pyi
```diff
@@ -42,8 +43,37 @@ class FileSystemStorage(Storage):
...
+    @property
+    def backends(self) -> dict[str, Any]: ...
```
intgr: Looks like dict values are `Storage` instances?

```suggestion
    def backends(self) -> dict[str, Storage]: ...
```

django-stubs/core/files/storage.pyi
```diff
@@ -42,8 +43,37 @@ class FileSystemStorage(Storage):
     @property
     def directory_permissions_mode(self) -> int | None: ...
 
-class DefaultStorage(LazyObject): ...
+class InvalidStorageError(ImproperlyConfigured): ...
 
-default_storage: Any
+class StorageHandler:
```
intgr: AFAIK the `.pyi` file structure should exactly mirror `.py` file structure. So `StorageHandler` this should be defined in `.../files/storage/handler.pyi` and re-exported in `.../files/storage/__init__.pyi`

This requires some refactoring though, turning `storage.pyi` file into a package.

intgr (changes requested):
Thanks! A few more tweaks :)

django-stubs/core/files/storage.pyi
Alexerson: Makes sense. I wasn't sure if a small diff was better or not in this case, but as it was pretty heavily refactored upstream, it makes sense to refactoring it here as well.
 I will look into this PR again hopefully by the end of the week. If not, mid next week.

intgr: 👍 I think this is already an improvement, I'm going to merge this along with my tweaks and I hope to see the follow-up PR later.

intgr (approved):
Let's merge this and fix the remaining things in a follow-up pull request.

Alexerson:
Ok, so it wasn’t actually so difficult to do that properly, and as this PR is not merged yet, I think we should do it properly right away.

The only part I wasn’t sure about was the typing of `mixins.StorageSettingsMixin`’s methods.

intgr:
> this PR is not merged yet

Oops, sorry, I was waiting for CI and forgot to merge afterwards.

django-stubs/core/files/storage/mixins.pyi
```diff
@@ -0,0 +1,9 @@
+from typing import Any, Literal
+
+class StorageSettingsMixin:
+    def _clear_cached_properties(
+        self,
+        setting: Literal["MEDIA_ROOT", "MEDIA_URL", "FILE_UPLOAD_PERMISSIONS", "FILE_UPLOAD_DIRECTORY_PERMISSIONS"],
+        **kwargs
```
intgr:
```suggestion
        **kwargs: Any
```

Alexerson:
No worries. The CI seems to be green again after your commit, thanks.

PR #1492, opened 2023-05-11
Adds support for db_comment option on model fields and tests.

### LINKED ISSUE #1490: Model field arg `db_comment` not supported
# Bug report

When using the field option `db_comment` introduced in Django 4.2, it triggers mypy errors such as the below.

## What's wrong

Given a model with the following structure:

```python
from django.db import models

class Task(models.Model):
    title = models.CharField(max_length="250", help_text=_("The title of the task"))
    notes = models.TextField(
        blank=True,
        null=True,
        help_text=_("Additional notes about the task; markdown formatting supported."),
        db_comment="Additional notes about the task. May contain markdown."
    )

    def __str__(self):
        return self.title
```

You get the following call-arg error for it when running `mypy`.

```shell
taskhub/models.py:75:13: error: Unexpected keyword argument "db_comment" for "TextField"  [call-arg]
        notes = models.TextField(
                ^
```
## How is that should be

It should not throw an error on that field definition. 

As a workaround for now, I'm adding `# type: ignore[call-arg]` on the fields where I'm using `db_comment`, but obviously that's less than ideal.

## System information

- OS:
- `python` version: 3.11.2
- `django` version: 4.2.1
- `mypy` version: 1.2.0

andrlik:
## Related issues

Adds support for Django 4.2's new option for model fields: `db_comment`. Adds a test for the same. 

Format is:
- Closes #1490


sobolevn (commented):
Thanks! But, we don't need tests for this:
- Tests are quite slow to run
- Change is very simple

andrlik:
Ok, I've removed the test case

sobolevn (approved):

sobolevn:
Thank you!

PR #1494, opened 2023-05-12
Fix classonlymethod decorator declaration

### LINKED ISSUE #1485: Error occure when using classonlymethod decorator
# Bug report

```
from django.utils.decorators import classonlymethod


class Foo:
    @classonlymethod
    def bar(cls):
        pass
```

Error from mypy:
```
error: Argument 1 to "classonlymethod" has
incompatible type "Callable[[Foo], Any]"; expected "Callable[[Any, VarArg(Any), KwArg(Any)], Any]"  [arg-type]
```

## What's wrong
Error occurred, but it seems to be more related to mypy itself. The problem is described [here](https://github.com/python/mypy/issues/2473)

I created this issue here, because I think the solution is very easy to fix it and what I understood from [here](https://github.com/python/mypy/issues/2473#issuecomment-272842040) there will not change how mypy treats subclasses of `classmethod`

Problem stub:
`django-stubs/utils/decorators.pyi`
```
class classonlymethod(classmethod): ...
```

## How is that should be
Changing to alias solves the error
```
classonlymethod = classmethod
```
Let me know what do you think

## System information

- OS:
- `python` version: 3.9.16
- `django` version: 3.2.18
- `mypy` version: 1.2.0

adamchainz:
# I have made things!

Switch to an alias as suggested on the issue.

## Related issues

Fixes #1485


intgr (approved):
Thanks!

RobertM15:
Thanks @adamchainz

PR #1495, opened 2023-05-12
Fix ForeignKey queryset filters on un-swapped models

### LINKED ISSUE #626: `'self'`-referencing relationship causes crash
**Traceback**

```
  File "mypy/checkexpr.py", line 3898, in accept
  File "mypy/nodes.py", line 1555, in accept
  File "mypy/checkexpr.py", line 270, in visit_call_expr
  File "mypy/checkexpr.py", line 347, in visit_call_expr_inner
  File "mypy/checkexpr.py", line 852, in check_call_expr_with_callee_type
  File "mypy/checkexpr.py", line 911, in check_call
  File "mypy/checkexpr.py", line 1023, in check_callable_call
  File "mypy/checkexpr.py", line 732, in apply_function_plugin
  File "/Users/jacob/modularhistory/.venv/lib/python3.9/site-packages/mypy_django_plugin/transformers/orm_lookups.py", line 31, in typecheck_queryset_filter
    lookup_type = django_context.resolve_lookup_expected_type(ctx, model_cls, lookup_kwarg)
  File "/Users/jacob/modularhistory/.venv/lib/python3.9/site-packages/mypy_django_plugin/django/context.py", line 335, in resolve_lookup_expected_type
    lookup_parts, field_parts, is_expression = query.solve_lookup_type(lookup)
  File "/Users/jacob/modularhistory/.venv/lib/python3.9/site-packages/django/db/models/sql/query.py", line 1052, in solve_lookup_type
    _, field, _, lookup_parts = self.names_to_path(lookup_splitted, self.get_meta())
  File "/Users/jacob/modularhistory/.venv/lib/python3.9/site-packages/django/db/models/sql/query.py", line 1484, in names_to_path
    pathinfos = field.get_path_info(filtered_relation)
  File "/Users/jacob/modularhistory/.venv/lib/python3.9/site-packages/django/db/models/fields/related.py", line 723, in get_path_info
    opts = self.remote_field.model._meta
AttributeError: 'str' object has no attribute '_meta'
apps/trees/models.py:89: : note: use --pdb to drop into pdb
```
I did some debugging and found that when this `'str' object has no attribute '_meta'` error happens, the value of self.remote_field is `'self'`.

**To Reproduce**

Here's the abridged content of my models.py:
```
"""Based on https://github.com/peopledoc/django-ltree-demo."""

from typing import Type

from django.core.exceptions import ValidationError
from django.utils.translation import ugettext_lazy as _

from core.models.model import Model


class TreeModel(Model):
    """Implements Postgres ltree for self-referencing hierarchy."""

    parent = models.ForeignKey(
        to='self',
        null=True,
        blank=True,
        related_name='children',
        on_delete=models.CASCADE,
        verbose_name=_('parent'),
    )
    name = models.TextField(verbose_name=_('name'))

    # https://docs.djangoproject.com/en/dev/ref/models/options/#model-meta-options
    class Meta:
        abstract = True

    @property
    def siblings(self) -> QuerySet['TreeModel']:
        """Return the model instances's siblings."""
        # Here is where it fails:
        return self.__class__.objects.exclude(id=self.id).filter(parent_id=self.parent_id)
```

UnknownPlatypus:
# I have made things!
Followup PR to https://github.com/typeddjango/django-stubs/pull/1436.
I found a reproducible exemple, the first commit should make test fail.
The second should fix it.

## Related issues
- Closes #626

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -141,8 +141,6 @@ class DjangoContext:
     ) -> MypyType:
         if isinstance(field, (RelatedField, ForeignObjectRel)):
             related_model_cls = self.get_field_related_model_cls(field)
-            if related_model_cls is None:
-                return AnyType(TypeOfAny.from_error)
             primary_key_field = self.get_primary_key_field(related_model_cls)
             primary_key_type = self.get_field_get_type(api, primary_key_field, method="init")
 
@@ -225,9 +223,6 @@ class DjangoContext:
                         continue
 
                     related_model = self.get_field_related_model_cls(field)
-                    if related_model is None:
-                        expected_types[field_name] = AnyType(TypeOfAny.from_error)
-                        continue
 
                     if related_model._meta.proxy_for_model is not None:
                         related_model = related_model._meta.proxy_for_model
@@ -327,8 +322,6 @@ class DjangoContext:
         is_nullable = self.get_field_nullability(field, method)
         if isinstance(field, RelatedField):
             related_model_cls = self.get_field_related_model_cls(field)
-            if related_model_cls is None:
-                return AnyType(TypeOfAny.from_error)
 
             if method in ("values", "values_list"):
                 primary_key_field = self.get_primary_key_field(related_model_cls)
@@ -354,7 +347,7 @@ class DjangoContext:
                 related_model_cls = field.model
             elif "." not in related_model_cls:
                 # same file model
-                related_model_fullname = field.model.__module__ + "." + related_model_cls
+                related_model_fullname = f"{field.model.__module__}.{related_model_cls}"
                 related_model_cls = self.get_model_class_by_fullname(related_model_fullname)
                 if related_model_cls is None:
                     raise UnregisteredModelError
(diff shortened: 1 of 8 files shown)


UnknownPlatypus:
Marking as draft for tonight, will check the typing issue later this weekend

mypy_django_plugin/django/context.py
```diff
@@ -364,13 +366,19 @@ def _resolve_field_from_parts(
 
             field = currently_observed_model._meta.get_field(field_part)
             if isinstance(field, RelatedField):
-                currently_observed_model = field.related_model
+                related_model = self.get_field_related_model_cls(field)
```
sobolevn: The code duplication is quite high. Can you please refactor it to be a method / function?

UnknownPlatypus: wasn't a big fan of that too, I can do that yes :+1:

UnknownPlatypus: I was also considering actually raising `LookupError` (or re-raising a custom `UnregisteredModelError` or similar) in `get_field_related_model_cls` and handle them in places we actually return mypy types. (`return AnyType(TypeOfAny.from_error)`)

Currently this is quite cumbersome to pass `None` values all the way up the intermediary functions/methods.
I'm not sure this is really better thought and maybe the intent is less clear?

UnknownPlatypus:
Sorry, couldn't get back to this earlier, got caught up with other things.

I tried the approach of raising an exception for unregistered model in `get_field_related_model_cls`, this lead to some nice simplifications but I'm still not sure this is the best way to deal with this.

The diff is a bit bigger because of that, let me know if this seems reasonable to you.
I'll be more available in the following days to take any feedback

EDIT: hum, test are passing locally, will investigate further.
Edit2: I had django 3 instead of django 4 installed, hence the test passing locally

sobolevn (approved):
Looks like a solid solution! Thank you!

intgr:
FWIW, I only have little experience working on plugin code, but I had the same feeling: checking for and passing `None` values up the stack felt unwieldy. I prefer the exception approach. 👍

PR #1497, opened 2023-05-15
Make argument optional in admin.sites.get_app_list()

AkaZecik:
The second argument has a default value of `None` and therefore is optional, but the signature currently reports it as required


intgr (approved):
Thanks!

PR #1498, opened 2023-05-15
Update the name of lookup_needs_distinct() function

AkaZecik:
In django 4.0, the name of `lookup_needs_distinct` function was changed to `lookup_spawns_duplicates` (see [https://docs.djangoproject.com/en/4.2/releases/4.0/#miscellaneous](https://docs.djangoproject.com/en/4.2/releases/4.0/#miscellaneous), 10th point)


intgr (approved):
:+1:

PR #1499, opened 2023-05-16
Add logo

adamchainz:
# I have made things!

Add new project logo.

## Related issues

As discussed in #1398.

Issue #1398, opened 2023-03-16
TypedDjango now has financial contributors


sobolevn:
Right now we have 200$ in our OpenCollective account: https://opencollective.com/typeddjangoorg

I think it might be fair to split these funds among active maintainers.
So, I want to raise a discussion about it.

What do you think? @intgr @adamchainz @flaeppe 

(I cannot have my share due to the political stuff, because both OC and PayPal have blocked all transactions in my country. PayPal even stole my money I had on my account. So, I volunteer to share my money with other devs).

intgr:
Awesome! Thanks to Pinecast for the sponsorship, and also individual contributors.

If there are some other expenses we could be making with these funds to help the project, I would happily contribute my share towards those. But I can't think of anything right now.

flaeppe:
Thank you Pinecast for the donation.

I'd rather use my share to move the project forward too. In whatever way it could help.

Could it be worth to put a bounty on some issue(s)?

sobolevn:
I don't have any experience with bounties. If anyone has, please share your feedback :)

adamchainz:
I'm going to print some stickers of my book mascots for Djangocon Europe. Perhaps we could also do some django-stubs stickers?

sobolevn:
@adamchainz sounds great! If others agree, you can send a payment request to https://opencollective.com/typeddjangoorg and I will approve it.

But, I am very bad at graphics: this is the best I can do https://avatars.githubusercontent.com/u/52535716?s=200&v=4 I won't be able to help here :(

P.S. I would be extra grateful to have one sticker by mail, but if it is too complex - I 100% understand :)

intgr:
I had a look at the [Django Trademark License](https://www.djangoproject.com/trademarks/). Our remixing of the Django logo probably isn't allowed, this is the closest thing that only permits names of localities together with the logo:

> 3. Groups and Events
>
> [...]
> A group or event may use a logo that incorporates the Django logo by adding a word or picture that identifies the locality (for example, the name of a city, or a stylized image of an identifiable city landmark), audience, or activities of the group or event.

So we may need to redesign the logo anyway. If it's just a bunch of volunteers, this probably wouldn't be an issue, but now that real money has entered the picture, we should take it more seriously.

And this part is also relevant:

> 4. Merchandise
> 
> You may produce merchandise displaying the Django name or logo, provided that:
> 
> 1. Artwork to be printed on merchandise has been formally approved by the DSF;
[...]

PR #1500, opened 2023-05-16
Fix logo URL

adamchainz:
# I have made things!

Woops, forgot to update ref before merging #1499.

## Related issues

#1398.


intgr:
Nope, we still use `master` branch here 😆

PR #1501, opened 2023-05-16
Fix logo URL with 'master'

adamchainz:
# I have made things!

Damn, in #1500 I missed that this project still uses a `master` branch not `main`.

## Related issues

#1398


adamchainz:
Actually checked the preview link at https://github.com/typeddjango/django-stubs/blob/logo3/README.md this time, it looks good.

PR #1502, opened 2023-05-17
Modify ticket in logo

adamchainz:
# I have made things!

@lolbat mentioned on Mastodon that it looks like a razor blade: https://fosstodon.org/@lolbat/110379380071734218 . Now I've seen it, I can't unsee it.

I think we should also avoid the razor blade appearance because it may trigger those who have some relationship with suicide.

I've modified the icon to use clipped-off corners, which is another common ticket design (see [image search for "ticket"](https://duckduckgo.com/?ia=images&iax=images&iar=images&q=ticket)).

## Related issues

Discussed in #1398 , originally added in #1499.


adamchainz:
Org icons @sobolevn 

django-stubs org icon
django-stubs org icon

sobolevn (approved):
TIL about mastodont

Thanks!

intgr (approved):
:+1:

adamchainz:
> TIL about mastodont

Is that a genuine typo, or is "mastodont" some kind of movement of saying "don't do that" on Mastodon? 😅

sobolevn:
TIL about both "mastodont" and Mastodon 😆

PR #1503, opened 2023-05-17
fix typo in mypy.ini error message

asottile:
copy pasting the error message leads to the same error message again -- only after changing it to a dash does it work correctly


sobolevn (approved):
Confirmed, thanks! 

https://github.com/typeddjango/django-stubs/blob/0015d35eb814661f078b1b2a8190f774fd5014a3/mypy_django_plugin/config.py#L95

PR #1505, opened 2023-05-18
Add QuerySet._result_cache and _fetch_all()

adamchainz:
# I have made things!

These two attributes are required for typing django-stubs correctly: https://github.com/tolomea/django-auto-prefetch/pull/217

I'm not sure about including all the other private attributes on `QuerySet` right now, adding them on a case-by-case basis seems fine?

## Related issues

n/a


intgr:
I remember there have been pull requests to explicitly **remove** stubs of private members from this project.

@sobolevn  What do you think, should we reconsider that policy?

Personally I'm in favor of adding stubs for private members as necessary. While ugly, accessing private members is sometimes justified. For example, when subclassing to extend logic of a class, private members seem like fair game, since the usage remains internal to that class hierarchy (similar to Java `protected` attributes).

intgr:
For example `QuerySet._result_cache` was removed in this commit: [fa718b8e55e4ac74b309b0068eeced6f89f48728](https://github.com/typeddjango/django-stubs/commit/fa718b8e55e4ac74b309b0068eeced6f89f48728#diff-389fe7d7dcfb77ddaac87c1f3bdc196f3ef0ca33b8047c6638a15f3a3db21b56L41-L51)

I think I've seen removals of underscore members more recently than that.

intgr:
@sobolevn What do you think? Yes/no/need more time to consider?

> Personally I'm in favor of adding stubs for private members as necessary.

intgr (approved):
I want to make a release today and I'm inclined to merge this.

Haven't received a response from sobolevn. But in https://github.com/typeddjango/django-stubs/pull/1517#discussion_r1211947894 he suggested the criteria for accepting underscore attributes is "if there are real use-cases", and `django-auto-prefetch` is one use case demonstrated here.

intgr:
Pff. I wanted to merge this for 4.2.1, but got distracted by other things that needed coordinating for the release, and forgot about this. Sorry! 🙄

adamchainz:
Thanks! Yeah, I agree with the policy of adding private members as use cases arise. Should we document that somewhere?

intgr:
> Should we document that somewhere?

Ah yes, I think we should. Then contributors have a better chance of avoiding wasted effort of hinting private methods.

PR #1507, opened 2023-05-20
CI: Replace isort with Ruff import sorting

intgr:
This is the first step in adopting Ruff: https://beta.ruff.rs/

Ruff is faster and will allow us to perform more fixes and checks in the future (e.g. replacing flake8, pyupgrade, auto-removing unused imports, etc).

There are some import formatting changes around places that use renamed imports (`import X import Y as Z`) but I think it's not any worse.

## Related issues

Ruff was suggested in https://github.com/typeddjango/djangorestframework-stubs/pull/399#issuecomment-1531102802, I checked it out and it's absolutely amazing! Thanks @orsinium.


.pre-commit-config.yaml
```diff
@@ -17,14 +17,11 @@ repos:
...
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.0.269
     hooks:
-      - id: isort
-        name: isort (python)
-      - id: isort
-        name: isort (pyi)
-        types: [pyi]
+      - id: ruff
+        args: ["--fix", "--fixable=I001", "--exit-non-zero-on-fix"]
```
intgr, line 24: `--exit-non-zero-on-fix` is suggested at https://github.com/charliermarsh/ruff-pre-commit

pyproject.toml
```diff
@@ -2,12 +2,21 @@
...
+    "out",
+]
+line-length = 120
+target-version = "py38"
+select = [
+    "I",    # isort
+]
+
+[tool.ruff.isort]
+split-on-trailing-comma = false
```
intgr, line 19: This option makes Ruff collapse multiline imports back to a single line if they're short enough to fit. E.g.

```python
from typing import (
    Any,
)
```
to
```python
from typing import Any
```

intgr: Related: https://github.com/charliermarsh/ruff/issues/4153

sobolevn:
Please, don't merge it just yet. I want to have a proper look in ~4 days.

orsinium (approved):

intgr:
@sobolevn  Ping :)

sobolevn (commented):
Thanks, I forgot about it.

What exactly are we winning here? 
We can change one setting in `isort` to have the same result.

intgr:
Like I said in the first comment:

> Ruff is faster and will allow us to perform more fixes and checks in the future (e.g. replacing flake8, pyupgrade, auto-removing unused imports, etc).

This is only the first step, the "win" in this PR is small still. Basically, Ruff is faster than isort and already has a [plugin for PyCharm/IntellIJ IDEA](https://plugins.jetbrains.com/plugin/20574-ruff) (which I use for developing django-stubs).

But Ruff's feature set covers so much and its speed is so good that, having used it in my personal projects, it seems like a no-brainer to adopt it. I used to hate pre-commit hooks with a passion because they make committing excruciatingly slow. But using Ruff is hardly noticeable.

One particular example, we had the problem with `pre-commit` that it won't allow running `pyupgrade` checks on `.pyi` files (https://github.com/typeddjango/djangorestframework-stubs/pull/399#issuecomment-1531053894). But with Ruff we can easily enable that.

PRs #1508 and #1509 have a taste of what's to come, but it can do more things. Exhaustive list of checks/fixes here: https://beta.ruff.rs/docs/rules/ but I would enable them in small steps.

sobolevn (approved):
Ok, let's try it!

sobolevn:
Maybe `ruff` also has `wemake` profile as `isort`?

intgr:
> Maybe ruff also has wemake profile as isort?

Looks like it doesn't.

But django-stubs isn't using using isort's `wemake` profile anyway, we were using the `black` profile.

sobolevn:
I am going to contribute `wemake` profile to `ruff`, because this is the best `isort` feature 😆

PR #1508, opened 2023-05-20
CI: Auto-remove unused imports using Ruff

intgr:
Ruff can automagically clean up unused imports.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
     rev: v0.0.270
     hooks:
       - id: ruff
-        args: ["--fix", "--fixable=I001", "--exit-non-zero-on-fix"]
+        args: ["--fix", "--fixable=I001,F401", "--exit-non-zero-on-fix"]
   - repo: https://github.com/psf/black
     rev: 23.3.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
     rev: v0.0.270
     hooks:
       - id: ruff
-        args: ["--fix", "--fixable=I001,F401", "--exit-non-zero-on-fix"]
+        args: ["--fix", "--exit-non-zero-on-fix"]
   - repo: https://github.com/psf/black
     rev: 23.3.0
     hooks:
--- a/django-stubs/contrib/syndication/views.pyi
+++ b/django-stubs/contrib/syndication/views.pyi
@@ -1,7 +1,6 @@
 from typing import Any, Generic, TypeVar
 
 from django.core.exceptions import ObjectDoesNotExist
-from django.db.models import Model
 from django.http.request import HttpRequest
 from django.http.response import HttpResponse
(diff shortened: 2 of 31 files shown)


django-stubs/contrib/syndication/views.pyi
```diff
@@ -1,7 +1,6 @@
 from typing import Any, Generic, TypeVar
 
 from django.core.exceptions import ObjectDoesNotExist
-from django.db.models import Model
```
sobolevn, line 4: Changes like this can actually affect our users who use `mypy` without strict re-export option enabled. However, I think that this removing unused stuff is a correct thing to do.

.pre-commit-config.yaml
```diff
@@ -21,7 +21,7 @@ repos:
     rev: v0.0.269
     hooks:
       - id: ruff
-        args: ["--fix", "--fixable=I001", "--exit-non-zero-on-fix"]
+        args: ["--fix", "--fixable=I001,F401", "--exit-non-zero-on-fix"]
```
sobolevn: Can we only use `pyproject.toml`?

intgr: Sure. I could move `fixable` to `pyproject.toml`. ~~The downside is that this would also apply when a developer runs explicit `ruff --fix` locally for other fixes that we haven't explicitly whitelisted.~~

Or we could get rid of `fixable` entirely, which would apply all fixes in selected categories (`select = [...]`)

For my own projects I've adopted a more cautious approach and explicitly named the autofixable rules. But I haven't seen any issues so far with the autofixes.

What do you think?

django-stubs/contrib/syndication/views.pyi
intgr: I think that's not the case for `.pyi` files.

[Mypy docs](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport) says about `--no-implicit-reexport`: "Note this is always treated as enabled for stub files". So if stubs don't use `from X import Y as Y`, this should have no effect.

.pre-commit-config.yaml
intgr: For CLI uses it's easy to override `pyproject.toml` with `ruff --fix --fixable=ALL`

sobolevn: Let's see what happens when we try to use `--fixable=ALL`? Let's try this

intgr: Sure, I've removed `--fixable` then, which applies all fixes (when used with `--fix`).

django-stubs/contrib/syndication/views.pyi
sobolevn: Ok! 👍

sobolevn (commented):
Nothing has changed when you removed `--fixable`. Is it correct?

intgr:
Yep. There was only one change, but that was because I rebased onto latest master (https://github.com/typeddjango/django-stubs/pull/1508/commits/2274c060e380dcab1afb1a6b1e47186b469d4137)

PR #1509, opened 2023-05-20
CI: Enable Ruff pyupgrade fixes

intgr:
As mentioned in https://github.com/typeddjango/djangorestframework-stubs/pull/399#issuecomment-1531053894, previously `pre-commit` was only running `pyupgrade` on `.py` files and not `.pyi` files.

Running via Ruff will apply to all files.


flaeppe (approved):

sobolevn (approved):
Thanks!

PR #1517, opened 2023-05-29
Add SessionBase methods, change SessionStore.get_model_class() return

realsuayip:
https://github.com/django/django/blob/main/django/contrib/sessions/backends/base.py
https://github.com/django/django/blob/main/django/contrib/sessions/backends/db.py


django-stubs/contrib/sessions/backends/base.pyi
```diff
@@ -13,17 +13,19 @@ class SessionBase(dict[str, Any]):
...
     def encode(self, session_dict: dict[str, Any]) -> str: ...
     def decode(self, session_data: bytes | str) -> dict[str, Any]: ...
-    def has_key(self, key: Any) -> Any: ...
-    def keys(self) -> Any: ...
-    def values(self) -> Any: ...
-    def items(self) -> Any: ...
+    def has_key(self, key: Any) -> bool: ...
     def clear(self) -> None: ...
     def is_empty(self) -> bool: ...
+    def _get_new_session_key(self) -> str: ...
```
sobolevn: we usually don't type protected methods, why do you need this?

realsuayip: `_get_or_create_session_key` may needed to be used when overriding `SessionStore`. For example:

https://github.com/QueraTeam/django-qsessions/blob/ecfd7327ebe32483746704b59fca72f05be07284/qsessions/backends/db.py#LL43C14-L43C14 

I don't have any use cases for the others, typed them just in case. I could remove those.

sobolevn: Please, remove them :)
We can always bring them back if there are real use-cases.

django-stubs/contrib/sessions/backends/base.pyi
```diff
@@ -13,15 +13,14 @@ class SessionBase(dict[str, Any]):
...
+    def key_salt(self) -> str: ...
     def set_test_cookie(self) -> None: ...
     def test_cookie_worked(self) -> bool: ...
     def delete_test_cookie(self) -> None: ...
     def encode(self, session_dict: dict[str, Any]) -> str: ...
     def decode(self, session_data: bytes | str) -> dict[str, Any]: ...
-    def has_key(self, key: Any) -> Any: ...
-    def keys(self) -> Any: ...
-    def values(self) -> Any: ...
-    def items(self) -> Any: ...
```
intgr: Why remove these 3 methods? Looks like they're present still: https://github.com/django/django/blob/main/django/contrib/sessions/backends/base.py#L121C1-L129

django-stubs/contrib/sessions/backends/base.pyi
intgr: I've pushed a change to remove the underscore methods.

django-stubs/contrib/sessions/backends/base.pyi
intgr: I've added them back for now, so I can merge this for the 4.2.1 release.

If there was a good reason to remove them, please let me know.

intgr (approved):

django-stubs/contrib/sessions/backends/base.pyi
realsuayip: In stubs, `SessionBase` is defined as subclass of `dict`:

https://github.com/typeddjango/django-stubs/blob/21e541f8f588f4a42963543c7d115bb4cb95f4c5/django-stubs/contrib/sessions/backends/base.pyi#L9

not in Django though:

https://github.com/django/django/blob/27fed08e5ff597c813de25b6e89dc300a1f4231b/django/contrib/sessions/backends/base.py#LL33C9-L33C9

Subclassing from dict should cover for `keys`, `values` and `items` (and with proper return values, not just `Any`)

I'm guessing the subclassing is probably done so as not type all the dict related dunder methods, which are present in Django to ensure dict-like behavior.

PR #1521, opened 2023-06-02
Fixed `create_model_instance` incorrect data argument type

namper:
Fixed `create_model_instance` incorrect data argument type
Session data is dynamic and it doesn't have to have `Model` values

Here is documentation of Django explaining how to override SessionStore https://docs.djangoproject.com/en/4.2/topics/http/sessions/#example


sobolevn (approved):
Thank you!

Issue #1519, opened 2023-06-01
Django 4.2 Tracking Issue


Alexerson:
This is a tracking issue for changes that are needed for Django 4.2, based on [release notes](https://docs.djangoproject.com/en/dev/releases/4.2/).

- [ ] Support for psycopg3 (`django.db.backends.postgresql.psycopg_any`)
- [ ] [Registration lookups](https://docs.djangoproject.com/en/dev/ref/models/lookups/#lookup-registration-api) on `Field` instances
- [ ] `django.contrib.postgres` `arrayfield.overlap` lookup now supports `values` and `values_list`
- [ ] (?) `django.db.models.query.QuerySet.prefetch_related` now supports `Prefetch` objects with sliced querysets
- [x] `django.contrib.postgres.TrigramStrictWordSimilarity` and `TrigramStrictWordDistance` (and lookup?) (#1536)
- [x] New `encoder` parameter for `django.utils.html.json_script` (#1536)
- [x] `django.forms.ModelForm` now accepts the new `Meta` option `formfield_callback` to customize form fields. (#1536)
- [x] `django.forms.models.modelform_factory` respects `formfield_callback` attribute (#1536)
- [x] New `KT` expression for `django.db.models.JSONField ` (#1536)
- [x] `F` expression that output `BooleanField` can now be negated (#1536)
- [x] `django.contrib.gis.gdal.GDALRaster` now supports `pathlib.Path` (#1525)
- [x] `django.contrib.gis.models.functions.IsEmpty` expression (and lookup?) (#1525)
- [x] `django.contrib.gis.models.functions.FromWKB` and `FromWKT` functions (#1525)
- [x] `django.test.RequestFactory`, `AsyncRequestFactory`, `Client`, and `AsyncClient` classes now support the `headers` parameter (#1529)
- [x] Comments on columns and tables: `django.db.models.Field.db_comment`, `django.db.models.Options.db_table_comment` & `django.db.migrations.opertions.AlterModelTableComment`
- [x] New storage `django.core.files.storage.InMemoryStorage`
- [x] Custom file storages and related changes
- [x] New `django.contrib.auth.forms.BaseUserCreationForm`
- [x] New `django.contrib.sitemaps.Sitemap.get_languages_for_item()` (#1527)
- [x] New `django.contrib.staticfiles.storage.ManifestStaticFilesStorage.manifest_hash` and `support_js_module_import_aggregation` (#1528)
[...]

PR #1523, opened 2023-06-05
Implement Django 4.2 deprecations

Alexerson:
This PR is the first easy step for the 4.2 issue tracker.

It is simply removing the dropped things (i.e. the last 5 points in the list)

## Related issues

- Refs #1519


sobolevn (approved):
Thank you, all these things were removed indeed.

PR #1524, opened 2023-06-05
Update global settings and appconfig to match Django 4.2

Alexerson:
I figured one way to see where we’re at in regards to Django 4.2 was to use the `stubgen`  script and see the diffs. 
I’m still not really convinced it’s such a good idea at this point because the diff is huge :sweat_smile: . But for some files it’s pretty useful, for instance the settings.

Please note that some of those settings are deprecated in 5.0 (or later), but I guess we should still support them for now? (feel free to let me know if not, I’ll keep only the proper ones).


sobolevn (approved):
> Please note that some of those settings are deprecated in 5.0 (or later), but I guess we should still support them for now?

Yes, we support what runtime supports, not what it recommends.

PR #1525, opened 2023-06-05
Implement GDAL-related updates in Django 4.2

Alexerson:
This is the implementation of the Django 4.2 updates related to GIS. 
I feel this is not the most useful PR, because the GIS module is overall mostly not typed (except some places, it’s a lot of `Any`, but at least it’s adding the new class and methods.

Also, because I very rarely used this module, I didn’t try to get fancy with the typing.

## Related issues
- Refs #1519 (GIS-related points)

### DIFF
--- a/django-stubs/contrib/gis/gdal/raster/source.pyi
+++ b/django-stubs/contrib/gis/gdal/raster/source.pyi
@@ -1,4 +1,6 @@
 from collections.abc import Sequence
+from ctypes import c_void_p
+from pathlib import Path
 from typing import Any
 
 from django.contrib.gis.gdal.driver import Driver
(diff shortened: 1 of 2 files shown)


django-stubs/contrib/gis/gdal/raster/source.pyi
```diff
@@ -20,7 +22,7 @@ class TransformPoint(list[Sequence[float]]):
 
 class GDALRaster(GDALRasterBase):
     destructor: Any
-    def __init__(self, ds_input: Any, write: bool = ...) -> None: ...
+    def __init__(self, ds_input: str | Path | bytes | dict | c_void_p, write: bool = ...) -> None: ...
```
Alexerson, line 25: I just noticed we have a shortcut for `str | Path`, but regardless, I have a question here: it’s unclear to me if the `c_void_p` type is allowed in the documentation or not. The django code tests `isinstance(…, c_void_p)` in the function, but I know it’s not the proper decision viewpoint.

See https://docs.djangoproject.com/en/4.2/ref/contrib/gis/gdal/#django.contrib.gis.gdal.GDALRaster

intgr:
Maybe @mjakob or @ciscorn as previous contributors to geo types can have a look.

django-stubs/contrib/gis/gdal/raster/source.pyi
intgr: I would keep `c_void_p` here since there's explicit support on Django side.

adamchainz (approved):
IMO it's fine as-is, another contribution can refine the `Any`s at another point.

PR #1526, opened 2023-06-05
Fix various checks

Alexerson:
With some new settings in 4.2, came some new checks. Others were introduced earlier and never added, and finally, one of them was removed in a previous version of Django.

Please note that I kept the signature `-> Sequence[…]` because that’s what’s used already, but I’m wondering if it wouldn’t be more accurate to use `list[…]` in this case.
My logic being that you want to be as large as possible for args, but as precise as possible for return values. Let me know what you think and I can also fix all the existing checks to implement this logic.


sobolevn (approved):
Thanks!

PR #1527, opened 2023-06-05
Add new method to Sitemap

Alexerson:
Pretty small PR: adding new method to `Sitemap`.
I noticed also that the `Sitemap` class has 2 methods that are not documented (they are use internally by `get_urls()`:
```python
def get_protocol(self, protocol: str | None | None = ...) -> str: ...
def get_domain(self, site: Site | RequestSite | None = ...) -> str: ...
```
IIRC, we don’t want those in the stubs because they are not explicitly documented, but let me know if I’m wrong and I’ll add them.

## Related issues
- Refs #1519


sobolevn (commented):
> IIRC, we don’t want those in the stubs because they are not explicitly documented, but let me know if I’m wrong and I’ll add them.

We tend to add all public methods there are. We don't look at the docs to decide.

Alexerson:
Sounds good. Adding them then.

sobolevn (approved):
Thank you!

PR #1529, opened 2023-06-05
Add new headers parameter to test classes

Alexerson:
Adding the new `headers` parameter to `django.test.RequestFactory`, `AsyncRequestFactory`, `Client,` and `AsyncClient` classes

## Related issues
- Refs #1519

Also, `extra` was not defined on `Client`, but it was on `AsyncClient`, so I assumed it was a mistake and added it.

### DIFF
--- a/django-stubs/test/client.pyi
+++ b/django-stubs/test/client.pyi
@@ -72,18 +72,79 @@ class _RequestFactory(Generic[_T]):
     errors: BytesIO
     def __init__(self, *, json_encoder: type[JSONEncoder] = ..., **defaults: Any) -> None: ...
     def request(self, **request: Any) -> _T: ...
-    def get(self, path: str, data: _GetDataType = ..., secure: bool = ..., **extra: Any) -> _T: ...
-    def post(self, path: str, data: Any = ..., content_type: str = ..., secure: bool = ..., **extra: Any) -> _T: ...
-    def head(self, path: str, data: Any = ..., secure: bool = ..., **extra: Any) -> _T: ...
-    def trace(self, path: str, secure: bool = ..., **extra: Any) -> _T: ...
-    def options(
-        self, path: str, data: dict[str, str] | str = ..., content_type: str = ..., secure: bool = ..., **extra: Any
+    def get(
+        self,
+        path: str,
+        data: _GetDataType = ...,
+        secure: bool = ...,
+        *,
+        headers: dict[str, Any] | None = ...,
+        **extra: Any
     ) -> _T: ...
-    def put(self, path: str, data: Any = ..., content_type: str = ..., secure: bool = ..., **extra: Any) -> _T: ...
-    def patch(self, path: str, data: Any = ..., content_type: str = ..., secure: bool = ..., **extra: Any) -> _T: ...
-    def delete(self, path: str, data: Any = ..., content_type: str = ..., secure: bool = ..., **extra: Any) -> _T: ...
+    def post(
+        self,
+        path: str,
+        data: Any = ...,
+        content_type: str = ...,
+        secure: bool = ...,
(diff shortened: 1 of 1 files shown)


django-stubs/test/client.pyi
```diff
@@ -130,39 +199,104 @@ class Client(ClientMixin, _RequestFactory[_MonkeyPatchedWSGIResponse]):
     handler: ClientHandler
     raise_request_exception: bool
     exc_info: tuple[type[BaseException], BaseException, TracebackType] | None
+    extra: Any
```
adamchainz: I think this could be tightened to:
```suggestion
    extra: dict[str, Any] | None
```

Here, and in `AsyncClient`.

adamchainz (approved):
Thanks, looks good. Happy to see this feature I suggested getting proper type hints. I would like another maintainer to double check though.

django-stubs/test/client.pyi
```diff
@@ -72,26 +72,95 @@ class _RequestFactory(Generic[_T]):
...
+        **extra: Any
     ) -> _T: ...
 
 class RequestFactory(_RequestFactory[WSGIRequest]): ...
 
 class _AsyncRequestFactory(_RequestFactory[_T]):
     def request(self, **request: Any) -> _T: ...
     def generic(
-        self, method: str, path: str, data: Any = ..., content_type: str | None = ..., secure: bool = ..., **extra: Any
+        self,
```
sobolevn: Why do we have this method? It is inherited from `_RequestFactory` and they are identical. Maybe we can remove this one?

sobolevn: And `def request` as well.

Alexerson: It’s most likely here because Django redefines it, and when the stubs are generated, the stub is added. 
It would probably make sense to remove both and rely on the types in `RequestFactory`.

Alexerson: See https://github.com/django/django/blob/51fdea68415c580cd753990e3238ae2e0f015263/django/test/client.py#L610

sobolevn (approved):
LOOKS SOLID! 👍 
Thanks!

django-stubs/test/client.pyi
```diff
@@ -72,28 +72,83 @@ class _RequestFactory(Generic[_T]):
...
-    def post(self, path: str, data: Any = ..., content_type: str = ..., secure: bool = ..., **extra: Any) -> _T: ...
-    def head(self, path: str, data: Any = ..., secure: bool = ..., **extra: Any) -> _T: ...
-    def trace(self, path: str, secure: bool = ..., **extra: Any) -> _T: ...
+    def get(
+        self,
+        path: str,
+        data: _GetDataType = ...,
+        secure: bool = ...,
+        *,
+        headers: dict[str, Any] | None = ...,
```
intgr, line 81: Any reason to type the dict value as `Any`? Is there any type besides `str` that makes sense?

adamchainz: Django supports anything that can be cast to a string, such as `headers={"x-user-id": user.id}`.

intgr: 1. Could be solved easily with `str(user.id)` in user code
2. I tested this, I don't think there's any "casting to string" that happens. Django's Client just passes the value verbatim to the view function.
   image

   So this actually misbehaves, since `request.headers[x]` is defined to return `str`, but here it returned `list`. And it can result code behaving differently during testing and with real world requests, causing more hair pulling for users.

3. For testing really weird cases there's always the `**extra` kwarg escape hatch.

intgr: Anyway let's continue this discussion in #1534.

PR #1530, opened 2023-06-05
`RequestSite.{save,delete}` methods are always raising

sobolevn:
Source: https://github.com/django/django/blob/51fdea68415c580cd753990e3238ae2e0f015263/django/contrib/sites/requests.py#L4


intgr (approved):
👍

PR #1535, opened 2023-06-06
Update patch.py to fix generic UserCreationForm

adambirds:
# I have made things!

This fixes the below error I'm getting:

```
 | apps_tenants/ticket_system/forms.py:41: error: Missing type parameters for generic type "UserCreationForm"  [type-arg]
mypy      | apps_tenants/ticket_system/forms.py:103: error: Missing type parameters for generic type "UserCreationForm"  [type-arg]
mypy      | 
mypy      | See https://zulip.readthedocs.io/en/latest/testing/mypy.html for debugging tips.
```

### DIFF
--- a/django_stubs_ext/django_stubs_ext/patch.py
+++ b/django_stubs_ext/django_stubs_ext/patch.py
@@ -4,7 +4,6 @@ from typing import Any, Generic, Iterable, List, Optional, Tuple, Type, TypeVar
 from django import VERSION
 from django.contrib.admin import ModelAdmin
 from django.contrib.admin.options import BaseModelAdmin
-from django.contrib.auth.forms import UserCreationForm
 from django.contrib.sitemaps import Sitemap
 from django.contrib.syndication.views import Feed
 from django.core.files.utils import FileProxyMixin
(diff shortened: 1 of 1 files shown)


adambirds:
Can anyone advise why the tests have failed?

sobolevn:
Because it needs django settings to import this model. You can:
1. Use fake settings
2. Import it differently, using `getattr_static` or something similar
3. Your ideas :)

adambirds:
I seem to be doing something wrong when trying to get this to work when trying to configure standard settings. Any help would be appreciated. This wasn't an issue with UserCreationForm until I updated Django. All the other generics ive done in the past never needed settings.

sobolevn:
I don't have any ideas right now, @adamchainz what about you?

UnknownPlatypus:
#2659 added a test that ensure we patch every class that requires it, the workaround for `UserCreationForm` is also documented in the [README](https://github.com/typeddjango/django-stubs?tab=readme-ov-file#i-cannot-use-queryset-or-manager-with-type-annotations) now

PR #1536, opened 2023-06-07
Various fixes for Django 4.2

Alexerson:
I’m adding more Django 4.2 new types. I’m picking up the easier ones, and rather than having 5 one-line PR, I figured I would have them all in there. Let me know if you see things to fix I missed.

My strategy is to start from the stubgen output, so I’m also adding the new fields from 4.1 (or before) that are missing in the files I’m editing.

## Related issues
- Refs #1519


Alexerson:
I’m trying to improve the signature for `NegatedExpression`, but the current proposition is the best I can do, and it’s a bit disappointing. I’ll try again tomorrow if noone jumped in.

Also, I wanted `Combinable.__invert__` to be `-> NegatedExpression`, because that would be logical with the implementation, but it causes all sort of issues that I don’t know how to fix.

django-stubs/db/models/expressions.pyi
```diff
@@ -199,8 +202,14 @@ class ExpressionList(Func):
...
 
-class ExpressionWrapper(Expression):
-    def __init__(self, expression: Q | Combinable, output_field: Field) -> None: ...
+_E = TypeVar("_E", bound=Q | Combinable)
+
+class ExpressionWrapper(Expression, Generic[_E]):
+    def __init__(self, expression: _E, output_field: Field) -> None: ...
+    expression: _E
+
+class NegatedExpression(ExpressionWrapper):
```
sobolevn: Shouldn't it also be generic? Since its parent is generic.

sobolevn (commented):
Thanks!

We also tend to add "stub" generics to `django-stubs-ext` to monkeypatch them.

django-stubs/db/models/expressions.pyi
Alexerson: Locally, I also made the following changes: 

```diff
diff --git a/django-stubs/db/models/expressions.pyi b/django-stubs/db/models/expressions.pyi
index 823b436b..296f0b3c 100644
--- a/django-stubs/db/models/expressions.pyi
+++ b/django-stubs/db/models/expressions.pyi
@@ -1,7 +1,7 @@
 import datetime
 from collections.abc import Callable, Iterable, Iterator, Sequence
 from decimal import Decimal
-from typing import Any, Generic, Literal, TypeVar
+from typing import Any, Generic, Literal, TypeVar, overload
 
 from _typeshed import Self
 from django.db.backends.base.base import BaseDatabaseWrapper
@@ -54,7 +54,7 @@ class Combinable:
     def __rand__(self, other: Any) -> Combinable: ...
     def __ror__(self, other: Any) -> Combinable: ...
     def __rxor__(self, other: Any) -> Combinable: ...
-    def __invert__(self) -> Combinable: ...
+    def __invert__(self: Self) -> NegatedExpression[Self]: ...
 
 class BaseExpression:
     is_summary: bool
@@ -208,8 +208,12 @@ class ExpressionWrapper(Expression, Generic[_E]):
     def __init__(self, expression: _E, output_field: Field) -> None: ...
     expression: _E
 
-class NegatedExpression(ExpressionWrapper):
-    def __init__(self, expression: Q | Combinable) -> None: ...
+class NegatedExpression(ExpressionWrapper, Generic[_E]):
+    def __init__(self, expression: _E) -> None: ...
+    @overload
+    def __invert__(self: NegatedExpression[Q]) -> Q: ...
+    @overload
+    def __invert__(self: NegatedExpression[Combinable]) -> Combinable: ...
 
 class When(Expression):
     template: str
```

but then, I get:
```bash
$ mypy --cache-dir=/dev/null --no-incremental django-stubs
django-stubs/db/models/expressions.pyi:57: error: Type argument "Self" of "NegatedExpression" must be a subtype of "Union[Q, Combinable]"
django-stubs/db/models/expressions.pyi:213: error: Signature of "__invert__" incompatible with supertype "Combinable"
django-stubs/db/models/expressions.pyi:213: note:      Superclass:
django-stubs/db/models/expressions.pyi:213: note:          def __invert__(self) -> NegatedExpression[NegatedExpression[_E]]
django-stubs/db/models/expressions.pyi:213: note:      Subclass:
django-stubs/db/models/expressions.pyi:213: note:          @overload
django-stubs/db/models/expressions.pyi:213: note:          def __invert__(self) -> Q
django-stubs/db/models/expressions.pyi:213: note:          @overload
django-stubs/db/models/expressions.pyi:213: note:          def __invert__(self) -> Combinable
django-stubs/contrib/postgres/search.pyi:62: error: Return type "SearchQuery" of "__invert__" incompatible with return type "NegatedExpression[SearchQuery]" in supertype "Combinable"
Found 3 errors in 2 files (checked 683 source files)
```
and couldn’t find a way to solve it.

Alexerson:
> We also tend to add "stub" generics to `django-stubs-ext` to monkeypatch them.

Ok, I need to look into this.

django-stubs/db/models/expressions.pyi
sobolevn: You can always add `type: ignore[override]`

Alexerson: Maybe just this:
```diff
diff --git a/django-stubs/db/models/expressions.pyi b/django-stubs/db/models/expressions.pyi
index 823b436b..60748ee6 100644
--- a/django-stubs/db/models/expressions.pyi
+++ b/django-stubs/db/models/expressions.pyi
@@ -1,7 +1,7 @@
 import datetime
 from collections.abc import Callable, Iterable, Iterator, Sequence
 from decimal import Decimal
-from typing import Any, Generic, Literal, TypeVar
+from typing import Any, Generic, Literal, TypeVar, overload
 
 from _typeshed import Self
 from django.db.backends.base.base import BaseDatabaseWrapper
@@ -208,8 +208,12 @@ class ExpressionWrapper(Expression, Generic[_E]):
     def __init__(self, expression: _E, output_field: Field) -> None: ...
     expression: _E
 
-class NegatedExpression(ExpressionWrapper):
-    def __init__(self, expression: Q | Combinable) -> None: ...
+class NegatedExpression(ExpressionWrapper, Generic[_E]):
+    def __init__(self, expression: _E) -> None: ...
+    @overload
+    def __invert__(self: NegatedExpression[Q]) -> Q: ...
+    @overload
+    def __invert__(self: NegatedExpression[Combinable]) -> Combinable: ...
 
 class When(Expression):
     template: str
```
and then `mypy` is ok. But then I feel we loose the relation between `Combinable` and `NegatedExpression` and it feels a bit sad :shrug:

intgr: ~~I haven't looked into this discussion in detail. But I've found that `typing_extensions.Self` sometimes fixes issues that occur when using `_typeshed.Self`, maybe something to try.~~

intgr: Disregard what I wrote above :)

I looked into this and I agree, `# type: ignore` seems the best option here. It's technically a violation of LSP, but I think it's more useful than keeping `Combinable.__invert__() -> Combinable`.

intgr: Any drawbacks to typehinting this as `def __invert__(self) -> _E: ...`?

> `def __invert__(self: NegatedExpression[Q]) -> Q: ...`

Can this actually occur?

ISTM `NegatedExpression` is not constructed with typevar _E=Q. I think _E can only be `Combinable`.

Because `Q.__invert__()` returns `Q`, not `NegatedExpression[Q]`

intgr:
Not to rush you, but I'd like to make a release soonish (#1550), and most changes in this MR look good.

Maybe we can defer NegatedExpression and ExpressionWrapper to a separate PR and merge this?

Alexerson:
I'll have some time this morning. Either I finalize it, or I'll move it to a separate PR!

django-stubs/db/models/expressions.pyi
Alexerson: I think you’re right, `def __invert__(self) -> _E: ...` is good enough, I’m not sure why I’m complicating things. I’m pushing this and will do a self review.

Alexerson:
Ok, so I added some `type: ignore`. I believe they make sense. I’m not 100% sure how to add the generics to `django-stubs-ext` though, if someone can show me the way (maybe in another PR so we can close this?).

intgr:
For monkeypatching you just need to add the classes to this list: https://github.com/typeddjango/django-stubs/blob/master/django_stubs_ext/django_stubs_ext/patch.py#L49-L74

django-stubs/db/models/expressions.pyi
```diff
@@ -199,8 +202,15 @@ class ExpressionList(Func):
...
 
-class ExpressionWrapper(Expression):
-    def __init__(self, expression: Q | Combinable, output_field: Field) -> None: ...
+_E = TypeVar("_E", bound=Q | Combinable)
+
+class ExpressionWrapper(Expression, Generic[_E]):
+    def __init__(self, expression: _E, output_field: Field) -> None: ...
+    expression: _E
+
+class NegatedExpression(ExpressionWrapper, Generic[_E]):
```
intgr: If you do it like this, then the _E argument of ExpressionWrapper and _E argument of NegatedExpression won't be connected.

For example since `.expression` member comes from ExpressionWrapper
```python
from django.db.models.expressions import NegatedExpression, Exists
x: NegatedExpression[Exists]
reveal_type(x.expression)  # note: Revealed type is "Any"
```

This fixes it:
```suggestion
class NegatedExpression(ExpressionWrapper[_E]):
```

intgr: This kind of mistake would be harder to make if we used `disallow_any_generics` (#1339)

intgr (approved):
Awesome! One minor issue, looks minor enough that I can apply the fix myself and then it's good to go.

django_stubs_ext/django_stubs_ext/patch.py
```diff
@@ -8,6 +8,7 @@
 from django.contrib.syndication.views import Feed
 from django.core.files.utils import FileProxyMixin
 from django.core.paginator import Paginator
+from django.db.models.expressions import ExpressionWrapper, NegatedExpression
```
intgr: Oof, this import will fail with Django 4.1 and earlier, since NegatedExpression class only appeared in 4.2.

Also we clearly have a blind spot in our CI since it didn't detect this.

@sobolevn Any ideas how to handle backwards compatibility in django-stubs-ext monkeypatching?

intgr: PR #1552 will add CI coverage for such surprises.

intgr: Anyway I'd remove `NegatedExpression` monkeypatching for now, to unblock this PR again.

I've created issue #1555 to solve it in the future.

intgr (approved):
Let's get this merged finally.

PR #1538, opened 2023-06-09
Fix base Field and GenericIPAddressField attribute default_error_messages type

asottile:
the actual implementation has `dict[str, _StrPromise]` and subclasses matching that produce an error:

```
src/sentry/db/models/fields/jsonfield.py:53: error: Dict entry 0 has incompatible type "str": "_StrPromise"; expected "str": "str"  [dict-item]
```

### DIFF
--- a/django-stubs/db/models/fields/__init__.pyi
+++ b/django-stubs/db/models/fields/__init__.pyi
@@ -148,7 +148,7 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
     creation_counter: int
     auto_creation_counter: int
     default_validators: Sequence[validators._ValidatorCallable]
-    default_error_messages: dict[str, str]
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Note for self: `_ErrorMessagesT: TypeAlias = dict[str, Any]`

@intgr you had some opinions about using `Any`, do you want to take a second look?

sobolevn:
Thank you! 👍

intgr (approved):
> @intgr you had some opinions about using Any, do you want to take a second look?

Sorry, ping notification got lost among other ones. Yeah, thanks for pointing that out, I think we should redefine `_ErrorMessagesT` type. Though it requires some testing, whether we can get away with `dict[str, _StrOrPromise]`, or it causes issues due to mutability/non-covariance.

But that's out of scope of this PR, I think this change is an improvement regardless.

PR #1541, opened 2023-06-12
Annotate `AppConfig.default_auto_field` as an attribute

mthuurne:
In Django `AppConfig.default_auto_field` is a `cached_property`. I think the fact that it's implemented as a property isn't really relevant, as it is documented as an attribute and it's redefined as an attribute in [Django's test application](https://github.com/django/django/blob/1699f8b52ac15f687cc39088401c2641022c71cd/tests/apps/apps.py#L38).

## Related issues

Documentation: https://docs.djangoproject.com/en/4.2/ref/applications/#django.apps.AppConfig.default_auto_field


mthuurne:
I don't know why the mypy-self-check for 3.9 was cancelled automatically, but it looks like a glitch in the CI execution rather than a problem with the patch itself.

sobolevn (approved):
Thank you!

PabloAlexis611:
Hey there @sobolevn! I assume this change hasn't been released in an official version yet, right? 👀 

Currently `mypy` gives me this error after bumping to version `4.2.2`:
> some_app/apps.py:9: error: Incompatible types in assignment (expression has type "str", base class "AppConfig" defined the type as "Callable[[AppConfig], str]")  [assignment]

adamchainz:
>  I assume this change hasn't been released in an official version yet, right? 👀

You can see for yourself that it has been released. Click the commit in the record above your comment:

> @[sobolevn](https://github.com/sobolevn) sobolevn (Nikita Sobolev) merged commit [deae5ee](https://github.com/typeddjango/django-stubs/commit/deae5eebd10b88c66bdca25e8fc6f7e332fcc510) ...

You can see the tag on the commit:

Xnapper-2023-06-28-10 15 00

It sounds like you have a fresh bug to report, please open a new issue with more details.

adamchainz:
Ah, there's already an issue: #1593. This commit was undone.

PR #1544, opened 2023-06-14
Improve spatialite DatabaseWrapper attributes

filbasi:
# I have made things!

## Related issues
While working on custom DatabaseWrapper i encountered mypy error claiming that "DatabaseOperations" has no attribute "spatial_version". Fixes were achieved by adding ops type, which is similar to: https://github.com/typeddjango/django-stubs/blob/b81b1bf3868eaab9fecebdd56ed817ffcb07fad8/django-stubs/db/backends/postgresql/base.pyi#L23

### DIFF
--- a/django-stubs/contrib/gis/db/backends/spatialite/base.pyi
+++ b/django-stubs/contrib/gis/db/backends/spatialite/base.pyi
@@ -1,15 +1,20 @@
 from typing import Any
 
-from django.contrib.gis.db.backends.spatialite.operations import SpatiaLiteOperations
 from django.db.backends.sqlite3.base import DatabaseWrapper as SQLiteDatabaseWrapper
 
+from .client import SpatiaLiteClient
+from .features import DatabaseFeatures
+from .introspection import SpatiaLiteIntrospection
+from .operations import SpatiaLiteOperations
+from .schema import SpatialiteSchemaEditor
+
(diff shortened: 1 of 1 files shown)


intgr (approved):
Thanks!

This is already an improvement, but would be great to fix other types too in this file. E.g. `ops_class` should be `type[SpatiaLiteOperations]`

filbasi:
Done

intgr (approved):

django-stubs/contrib/gis/db/backends/spatialite/base.pyi
```diff
@@ -2,13 +2,20 @@ from typing import Any
...
+from .schema import SpatialiteSchemaEditor
+
 class DatabaseWrapper(SQLiteDatabaseWrapper):
-    SchemaEditorClass: Any
-    client_class: Any
-    features_class: Any
-    introspection_class: Any
-    ops_class: Any
-    lib_spatialite_paths: Any
+    SchemaEditorClass: SpatialiteSchemaEditor
```
intgr: Fails in CI. I think you need `type[]` around this.

filbasi:
Sorry about that. Now it should run fine.

intgr (approved):
Thanks!

PR #1545, opened 2023-06-14
Add ModelAdmin.get_formset_kwargs()

adamchainz:
# I have made things!

Added to Django in 4.0, https://github.com/django/django/commit/3119a6decab7788eca662b10e8c18351d20df212 .

https://docs.djangoproject.com/en/4.0/releases/4.0/#django-contrib-admin :

> The new [ModelAdmin.get_formset_kwargs()](https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_formset_kwargs) method allows customizing the keyword arguments passed to the constructor of a formset.

## Related issues

None, went straight to the PR.


intgr (approved):

django-stubs/contrib/admin/options.pyi
```diff
@@ -279,6 +279,9 @@ class ModelAdmin(BaseModelAdmin[_ModelT]):
     def history_view(
         self, request: HttpRequest, object_id: str, extra_context: dict[str, Any] | None = ...
     ) -> HttpResponse: ...
+    def get_formset_kwargs(
+        self, request: HttpRequest, obj: _ModelT, inline: InlineModelAdmin, prefix: str
```
adamchainz: ah wait a sec, the `InlineModelAdmin` needs type parameters. Should it be: `InlineModelAdmin[Any, _ModelT]` ?

adamchainz: By `Any` I meant `Model`, as in "any model", because all inlines are iterated over and passed to this function. Changed.

intgr: We've used `Any` instead of `Model` in places where users expect a concrete model instance with its fields to be present, but inference doesn't know what model exactly.

Here it seems less likely that users need to interact with `InlineModelAdmin`'s model, but I'm not sure, I couldn't find many exaples on the web using `get_formset_kwargs`. I would be fine with either.

intgr (approved):

django-stubs/contrib/admin/options.pyi
adamchainz: Okay that's a good point about the concrete fields. I'll use `Any`.

PR #1547, opened 2023-06-15
Add console EmailBackend.write_message()

adamchainz:
# I have made things!

Today I'm creating a subclass of this backend that displays messages more prettily, and I realized the stub for this method was missing.

Source: https://github.com/django/django/blob/5d20e0207867615f6989714261ea3c0a3eb85d88/django/core/mail/backends/console.py#L16

## Related issues

None


intgr (approved):
Thanks. I don't think this requires two reviewers 😉

adamchainz:
Yes, I just thought I’d explicitly ping you both, whoever came first :)

PR #1551, opened 2023-06-16
Add Signal._live_receivers()

adamchainz:
# I have made things!

Found a test that uses this private method to ensure that expected receivers are connected to a given signal, so adding it.

Source: https://github.com/django/django/blob/4b433ef236bf1607d5612d4c30f095ec9f6e63dc/django/dispatch/dispatcher.py#L234

## Related issues

None.


django-stubs/dispatch/dispatcher.pyi
```diff
@@ -20,5 +20,6 @@ class Signal:
     def has_listeners(self, sender: Any = ...) -> bool: ...
     def send(self, sender: Any, **named: Any) -> list[tuple[Callable, str | None]]: ...
     def send_robust(self, sender: Any, **named: Any) -> list[tuple[Callable, Exception | str]]: ...
+    def _live_receivers(self, sender: Any) -> list[Callable]: ...
```
intgr, line 23: Would be great to typehint the arguments of `Callable[]` here. But offtopic for this PR.

intgr (approved):
Looks good to me.

PR #1552, opened 2023-06-18
CI: Run django-stubs-ext tests in full build matrix

intgr:
This would help detect some issues earlier, such as adding imports to django-stubs-ext that don't exist in all Django or Python versions. This happened in https://github.com/typeddjango/django-stubs/pull/1536/files#r1233319757 and caught me by surprise that it wasn't being tested.

* Renamed 'typecheck' job to 'matrix-test' as it now runs both django-stubs-ext test suite and typecheck in full Django/Python version matrix.
* Dropped long-EOL Django 2.2 from CI. Our readme didn't claim to support 2.2 and django-stubs-ext monkeypatching didn't work with it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,8 +86,9 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -U pip setuptools wheel
-          sed -i -E "s/^Django==.+$/Django==${{ matrix.django-version }}.*/" ./requirements.txt
           pip install -r ./requirements.txt
+          pip install "Django==${{ matrix.django-version }}"
+          pip check
 
       - name: Run django-stubs-ext tests
         run: PYTHONPATH='.' pytest django_stubs_ext
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,6 +88,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -U pip setuptools wheel
+          sed -i -E "s/^Django==.+$/Django==${{ matrix.django-version }}.*/" ./requirements.txt
           pip install -r ./requirements.txt
 
       - name: Run django-stubs-ext tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,7 @@ jobs:
       - name: Run tests
         run: PYTHONPATH='.' pytest
 
-  typecheck:
+  matrix-test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -88,7 +86,11 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -U pip setuptools wheel
+          sed -i -E "s/^Django==.+$/Django==${{ matrix.django-version }}.*/" ./requirements.txt
```
intgr: Not sure if there's a nicer way to override Django version from requirements.txt?

sobolevn: There is: https://github.com/wemake-services/django-test-migrations/blob/1f002997b50619d39e10b32693a8f8727c04f175/.github/workflows/test.yml#L110-L112

sobolevn (commented):
Good idea!

.github/workflows/test.yml
intgr: I don't understand how I would use that here. That example is not using a `requirements.txt`.

sobolevn: It is using `pyproject.toml`, which is the same thing. Basically, you install `requirements.txt` as usual, then you run `pip install -U $DJANGO_VERSION_OVERRIDE` to change the django version for this build.

intgr: Ah OK, the `poetry install` line wasn't highlighted in your link, it wasn't clear that approach was what you meant.

It bothers me a little that Django would then be installed twice. And I suspect when using multiple `pip` invocations, it won't error on dependency conflicts.

Do you think it's an improvement regardless of these?

intgr: > when using multiple pip invocations, it won't error on dependency conflicts.

Actually this can be solved by using `pip check`, which returns nonzero on conflicts.

intgr: Yep, I think it looks better this way. Installing Django twice is ugly but the better of two evils :)

sobolevn (approved):
A good question is: do we still need this typecheck test?
Maybe we can drop it and use `stubtest` instead?

intgr:
> A good question is: do we still need this typecheck test?

Actually I'm not sure it does anything at all. I suspect not! 🤷

In djangorestframework-stubs I have needed to tweak the ignores lists regularly, but I don't remember the last time I saw it fail in django-stubs. Even when adding typecheck for Django 4.2, there weren't any changes.

Adopting stubtest definitely sounds like a good idea. Is it something you're interested in implementing?

One concern is, who will be responsible for fixing the issues from stubtest? Our maintainer bandwidth is limited. But if we could somehow get more contributors involved in fixing issues, then that would be awesome.

PR #1556, opened 2023-06-19
Remove typecheck test and clean things up

sobolevn:


sobolevn:
Refs https://github.com/typeddjango/django-stubs/pull/1552

intgr:
You didn't remove the `typecheck_tests.py` script itself?

sobolevn:
Good catch :)

scripts/git_helpers.py
```diff
@@ -1,34 +0,0 @@
-import shutil
-from typing import Optional, Union
-
-from git.remote import RemoteProgress
-from git.repo import Repo
```
intgr, line 5: We can also get rid of the `gitpython` dependency.

intgr (commented):
I feel a bit bad to delete all this code, people clearly went through a fair bit of effort to create this system.

Maybe we should keep it around until we have a new system to replace it? It hasn't really been a big maintenace burden, unless I'm missing something?

intgr:
> It hasn't really been a big maintenace burden, unless I'm missing something?

Oh OK, I see the errors coming from mypy 1.4 update now. That seems like a compelling reason.

scripts/git_helpers.py
sobolevn: > Maybe we should keep it around 

We have it: in our git history :)

> until we have a new system to replace it?

I am working on `stubtest` PR right now.

intgr (approved):
👍

PR #1557, opened 2023-06-20
Add `strict_settings` option, allow runtime fallbacks for custom settings

### LINKED ISSUE #417: Incompatible with django-configurations due to the way the settings are imported

### LINKED ISSUE #1276: Dynamic settings fail as of 1.13.0

### LINKED ISSUE #1410: Custom django settings generate mypy errors
## What's wrong

We have a project that is based on the wemake template. We created the project years ago some stuff might have changed with respect to the latest template

We have added some custom settings to server/settings/components/common.py. :
```
# common.py
MIN_GH_VERSION = "1.0.0"
```

Since we updated to latest version of django-stubs, we get mypy errors when we reference these errors in other parts of the code:

```
server/runner/logic/versioning.py:29: error: 'Settings' object has no attribute 'MIN_GH_VERSION'
```` 

We come from django-stubs 1.12.0 and are now on 1.16.0. We are on mypy version 1.1.1

django-stubs in setup.cfg:
```
[mypy.plugins.django-stubs]
django_settings_module = server.settings
```

- OS:
- `python` version: 3.10
- `django` version: 3.2.18
- `mypy` version: 1.1.1

edit: traced it back to this pull request I think: https://github.com/typeddjango/django-stubs/pull/1175

### PR DESCRIPTION
**Pull request by sobolevn**
This PR adds a new setting to our plugin called `strict_settings`.
When set to `True` (the default), it raises for any statically unknown setting value.

But, when set to `False`, it checks for the runtime setting value and if it is present, it returns `Any` in all cases.

This will change will make it possible to use `django-stubs` with [`django-split-settings`](https://github.com/wemake-services/django-split-settings) (which I maintain and use for all my projects) and similar tools like `django-configurations`.

After https://github.com/typeddjango/django-stubs/pull/1163 it was very hard to use settings in dynamic workflows.

Closes https://github.com/typeddjango/django-stubs/issues/1410
Closes https://github.com/typeddjango/django-stubs/issues/1276
Closes https://github.com/typeddjango/django-stubs/issues/417

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -311,6 +311,52 @@ If you encounter this error in your own code, you can either cast the `Promise`
...
+So, mypy would not like this code:
+
+```python
+from django.conf import settings
+
+settings.CUSTOM_VALUE  # E: 'Settings' object has no attribute 'CUSTOM_SETTING'
+```
+
+To handle this corner case we have a special setting `strict_settings` (`True` by default),
+you can switch it to `False` to always return `Any` and not raise any errors if runtime settings module has the given value:
```
intgr: Could use clarifying, which example is for ini and which for toml.

```suggestion
you can switch it to `False` to always return `Any` and not raise any errors if runtime settings module has the given value,
for example `pyproject.toml`:
```

README.md
```diff
@@ -311,6 +311,52 @@ If you encounter this error in your own code, you can either cast the `Promise`
...
+
+```toml
+[tool.django-stubs]
+strict_settings = false
+```
+
+or
```
intgr:
```suggestion
or `mypy.ini`:
```

tests/typecheck/test_settings.yml
```diff
@@ -58,3 +58,32 @@
...
       main:5: error: 'Settings' object has no attribute 'NON_EXISTANT_SETTING'
       main:5: note: Revealed type is "Any"
+
+
+-   case: settings_loaded_from_runtime_magic
+    disable_cache: true
+    main: |
+        from django.conf import settings
+        reveal_type(settings.A)  # N: Revealed type is "Any"
+        reveal_type(settings.B)  # E: 'Settings' object has no attribute 'B' # N: Revealed type is "Any"
```
intgr, line 73: What about Django builtin settings like `settings.SECRET_KEY`, I think those should still have concrete types?

README.md
```diff
@@ -311,6 +311,52 @@ If you encounter this error in your own code, you can either cast the `Promise`
...
+
+```ini
+[mypy.plugins.django-stubs]
+strict_settings = false
+```
```
intgr, line 349: Although there are just 2 settings for now, I would prefer it if all django-stubs settings were documented in one place. Maybe add another README section for settings and then link to it from here?

mypy_django_plugin/config.py
```diff
@@ -14,21 +14,23 @@
...
+strict_settings = bool (default: true)
 ...
 """
 INVALID_FILE = "mypy config file is not specified or found"
 COULD_NOT_LOAD_FILE = "could not load configuration file"
-MISSING_SECTION = "no section [{section}] found".format
+MISSING_SECTION = "no section [{section}] found"
 MISSING_DJANGO_SETTINGS = "missing required 'django_settings_module' config"
-INVALID_SETTING = "invalid {key!r}: the setting must be a boolean".format
+INVALID_BOOL_SETTING = "invalid {key!r}: the setting must be a boolean"
```
intgr, line 33: This was weird, thanks for cleaning up.

tests/typecheck/test_settings.yml
sobolevn: They are checked above, this setting does not change the default behaviour.

README.md
sobolevn: Yes, we can do this later. Good idea 👍

tests/typecheck/test_settings.yml
intgr: Let's add one such setting to this `strict_settings = false` test case too.

intgr:
When I try to run this in one of my projects, I get the following error. Haven't investigated at all.

```
mypy 1.4.0 (compiled: yes)
  File ".../bin/mypy", line 8, in <module>
Error constructing plugin instance of NewSemanalDRFPlugin

    sys.exit(console_entry())
             ^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/mypy/__main__.py", line 15, in console_entry
    main()
  File "mypy/main.py", line 95, in main
  ...
  File "mypy/build.py", line 484, in load_plugins_from_config
  File ".../lib/python3.11/site-packages/mypy_drf_plugin/main.py", line 26, in __init__
    self.django_context = DjangoContext(django_settings_module)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/mypy_django_plugin/django/context.py", line 85, in __init__
    apps, settings = initialize_django(self.plugin_config.django_settings_module)
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'django_settings_module'
```

intgr:
Looks like some interaction between `djangorestframework-stubs` and `django-stubs`.

sobolevn:
Oh, good catch! I've change the signature of `DjangoContext`, so we would need to change it here as well: https://github.com/typeddjango/djangorestframework-stubs/blob/8709db5d2af980919a46072725ab54ca637a8703/mypy_drf_plugin/main.py#L25-L26

I've opened https://github.com/typeddjango/djangorestframework-stubs/issues/432

UnknownPlatypus:
If I understand well, this makes it possible to use django-stubs along `django-configuration` or `django-split-settings` but the types inferred will be `Any` right ?

To have more accurate typing, one should make his own plugin ?

sobolevn:
> If I understand well, this makes it possible to use django-stubs along django-configuration or django-split-settings but the types inferred will be Any right ?

Yes!

> To have more accurate typing, one should make his own plugin ?

I can work on this later on, I need a way to infer mypy types from runtime types. I have a prototype that does just this. But, I need to polish it for a bit more.

mypy_django_plugin/django/context.py
```diff
@@ -71,13 +72,20 @@ class LookupsAreUnsupported(Exception):
 
 
 class DjangoContext:
-    def __init__(self, django_settings_module: str) -> None:
-        self.django_settings_module = django_settings_module
+    def __init__(self, plugin_config: DjangoPluginConfig) -> None:
+        if isinstance(plugin_config, str):
+            raise RuntimeError("Old API is used: we now require DjangoPluginConfig instance")
```
intgr: So people who have older djangorestframework-stubs installed and update django-stubs will receive this message?

Seems a bit cryptic. How about "Older djangorestframework-stubs versions are no longer supported, please update djangorestframework-stubs" or similar?

I doubt there are any other projects that import djagno-stubs internals, anyway.

sobolevn: The only project is `drf-stubs`. I don't really like this solution now. See https://github.com/typeddjango/djangorestframework-stubs/pull/433

I propose this plan:
- [x] I remove this error
- [ ] We release a new version of `drf-stubs` with the upper bound of `django-stubs` set to the current version of `django-stubs` (`>=4.2,<=4.2.1`)
- [ ] We merge this and release a new version of `django-stubs`
- [ ] We merge https://github.com/typeddjango/djangorestframework-stubs/pull/433 and release a new version of `drf-stubs` with no upper bound

sobolevn: On the other hand, it won't solve the problem of only upgrading `django-stubs` with old `drf-stubs` 😞 

Ideas?

sobolevn: Ok, I was able to solve all the compat problems in Python code. The easiest way for some reason is always the hardest one to find.

sobolevn:
I want to merge this today, so I can work on https://github.com/typeddjango/django-stubs/issues/1568

@intgr any other comments? :)

intgr (commented):
Sorry, I was away for a few days. But looks good to me now.

sobolevn:
No worries, happy to see that you like my fix :)

kaldown:
This do not work with `django-configurations`.

```
[tool.mypy]
mypy_path = "./src"

[tool.django-stubs]
django_settings_module = "myproj.settings"
strict_settings = false
```

returns an Error:
```
...
raise ImproperlyConfigured(install_failure)
django.core.exceptions.ImproperlyConfigured: django-configurations settings importer wasn't correctly installed. Please use one of the starter functions to install it as mentioned in the docs: https://django-configurations.readthedocs.io/
```

Whilst commenting usage of **django-stubs** in **[tool.mypy]** directive of **pyproject.toml** leading to expected issue of **mypy** not being able to catch _django_ types from _models.py_ (for example).

As I see this PR https://github.com/typeddjango/django-stubs/pull/180 is still open.
Would it be better to remove **django-configurations** entry from _README.md_ since it could cause confusion?

intgr:
@kaldown Please open a separate issue about this.

kaldown:
@intgr thank you, for the fast reply. Do you think it's better to open another one in addition to https://github.com/typeddjango/django-stubs/pull/180?

intgr:
Yes.

kaldown:
Done: https://github.com/typeddjango/django-stubs/issues/1709

PR #1558, opened 2023-06-20
Remove try-except around import of `ArrayField`

sobolevn:
Since we no longer support older versions of `Django`, we can say that `ArrayField` is always there.


adamchainz (approved):

intgr (approved):
Thanks, cleanups are always nice to see.

PR #1560, opened 2023-06-21
Add stubtest with lots of errors (currently)

sobolevn:
This is big: we introduce a new tool to check django / django-stubs inconsistencies.
Basically, this is the very first step with just all errors in one big pile of errors.

But, we can work on these errors later on: one by one.

Reference: https://github.com/python/typeshed/blob/main/tests/stubtest_third_party.py

What do you think? :)


.github/workflows/test.yml
```diff
@@ -64,6 +64,29 @@ jobs:
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          pip install -U pip setuptools wheel
+          pip install -r ./requirements.txt
+
+      - name: Run stubtest
+        run: bash ./scripts/stubtest.sh
```
adamchainz, line 88: Make the script executable?

```suggestion
        run: ./scripts/stubtest.sh
```

scripts/stubtest/allowlist_generated.txt
adamchainz, line 1: Can we call it something else like `to_fix` or `todo`? The current name sounds kinda ignorable.

CONTRIBUTING.md
```diff
@@ -76,23 +77,19 @@ If you get some unexpected results or want to be sure that tests run is not affe
...
-```
-
-You can also pass an optional commit hash as a second kwarg to checkout a specific commit, e.g.
+You might also want to disable `incremental` mode while working on `stubtest` changes.
+This mode leads to several known problems (stubs do not show up or have strange errors).
 
-```bash
-python ./scripts/stubgen-django.py --django_version 3.2 --commit_sha <commit_sha>
-```
+**Important**: right now we only run `stubtest` on Python3.9, any other versions might generate different outputs. Any work to create per-version allowlists is welcome.
```
adamchainz: can you explain why 3.9? It's not the minimum or maximum version supported by Django 4.2, seems like an odd choice.

adamchainz (commented):
Wow, this IS big. Nice one.

.github/workflows/test.yml
sobolevn: I prefer to use explicit `bash` here, because there are *multiple* problems with shebangs / permissions / etc. It is just not worth it :)

(I have a long and frightful story maintaining pure bash project 😅 https://github.com/sobolevn/git-secret/)

scripts/stubtest/allowlist_generated.txt
sobolevn: `todo` seems great! 👍

CONTRIBUTING.md
intgr: 3.9 is also the Python version we run our main tests on. But I agree, sounds arbitrary.

Shall we use the newest supported Python? It's faster as well.

adamchainz: Latest sounds like a good policy for now. I think we can merge this using 3.11 and upgrade the other workflows in another PR.

sobolevn: Agreed, 3.11 is better in the long run. Updated!

sobolevn:
Done! 

No worries, we only have `Found 2865 errors (checked 858 modules)` to fix! 😆

CONTRIBUTING.md
```diff
@@ -76,23 +77,19 @@ If you get some unexpected results or want to be sure that tests run is not affe
...
+### Testing stubs with `stubtest`
 
-### Generating Stubs using Stubgen
+Run `bash ./scripts/stubtest.sh` to test that stubs and sources are in-line.
 
-The stubs are based on auto-generated code created by Mypy's stubgen tool (see: [the stubgen docs](https://mypy.readthedocs.io/en/stable/stubgen.html)).
-To make life easier we have a helper script that auto generates these stubs. To use it you can run:
+We have two special files to allow errors:
+1. `scripts/stubtest/allowlist.txt` where we store things that we really don't care about: hacks, django internal utility modules, things that are handled by our plugin, things that are not representable by type system, etc
+2. `scripts/stubtest/allowlist_generate.txt` where we store all errors there are right now. Basically, this is a TODO list: we need to work throught this list and fix things (or move entries to real `allowlist.txt`). In the end, ideally we can remove this file
```
flaeppe:
```suggestion
2. `scripts/stubtest/allowlist_generate.txt` where we store all errors there are right now. Basically, this is a TODO list: we need to work through this list and fix things (or move entries to real `allowlist.txt`). In the end, ideally we can remove this file
```

flaeppe (commented):
Looks great 🙂

sobolevn:
I am going to merge this, any further changes can be done in the next PRs! Thanks everyone for the feedback.

adamchainz:
> No worries, we only have `Found 2865 errors (checked 858 modules)` to fix! 😆

💪 Challenge accepted.

PR #1562, opened 2023-06-22
Add memcache_key_warnings()

### LINKED ISSUE #1561: missing memcache_key_warnings
# Bug report

## What's wrong

memcache_key_warnings isn't typed in django-stubs.

from django.core.cache.backends.base import memcache_key_warnings

## How is that should be

A typed version of `memcache_key_warnings `

## System information

- OS: debian 11 (bullseye)
- `python` version: 3.9.17
- `django` version: 3.2.19
- `mypy` version: 1.1.1

rvanlaar:
Add `django.core.cache.backends.base.memcache_key_warnings()`

## Related issues
- Closes #1561


django-stubs/core/cache/backends/base.pyi
```diff
@@ -65,3 +65,5 @@ class BaseCache:
     async def adecr_version(self, key: Any, delta: int = ..., version: int | None = ...) -> int: ...
     def close(self, **kwargs: Any) -> None: ...
     async def aclose(self, **kwargs: Any) -> None: ...
+
+def memcache_key_warnings(key: str) -> Generator[str, None, None]: ...
```
sobolevn:
```suggestion
def memcache_key_warnings(key: str) -> Iterator[str]: ...
```

django-stubs/core/cache/backends/base.pyi
```diff
@@ -1,5 +1,5 @@
 from collections.abc import Callable, Iterable
-from typing import Any
+from typing import Any, Generator
```
sobolevn:
```suggestion
from typing import Any
from collections.abc import Iterator
```

sobolevn (commented):
Thanks!

django-stubs/core/cache/backends/base.pyi
rvanlaar: A question, could you explain to me why the choice for Iterator, and not Generator here?

rvanlaar:
Much appreciated for the review.

django-stubs/core/cache/backends/base.pyi
sobolevn: Because `Generator` is only needed when you use `.send` and `.throw` methods. Otherwise, this is just an `Iterator`

flaeppe: Although one have to be careful about the not so obvious pitfall found here(regarding context managers): https://github.com/python/typeshed/pull/7430

rvanlaar: I implemented the changes.

django-stubs/core/cache/backends/base.pyi
rvanlaar: I implemented the changes.

django-stubs/core/cache/backends/base.pyi
adamchainz: I guess you didn't push your changes or something, since they didn't appear here. I've just applied the suggestions.

sobolevn (approved):
Thanks everyone!

adamchainz:
Looking at the function's source inspired me to optimize it: https://code.djangoproject.com/ticket/34681

sobolevn:
Off-topic: `code.djangoproject.com` gives me https://bugs.python.org/ vibes.
I am so happy that we have moved away from it in CPython development (we are using GitHub issues now). Do you have similar plans for Django?

adamchainz:
Not at the moment. I think if someone came forward with a solid plan, it could be done.

PR #1563, opened 2023-06-22
Fix plugin exception "KeyError: 'model_bases'" and related errors

### LINKED ISSUE #1383: Non-deterministic crash "KeyError: 'model_bases'" with auth models
# Bug report

I'm able to reproduce this crash mostly with a *clean* cache (`--no-incremental` or `rm -rf .mypy_cache`). If I have a warm cache from scanning some other files, this does not occur.

Just a small annoyance, I don't have time to debug right now, but figured I should report it.

```
.../site-packages/django-stubs/contrib/auth/models.pyi:21: error: INTERNAL ERROR -- Please try using mypy master on GitHub:
version: 1.0.1
  File "mypy/semanal.py", line 6210, in accept
  File "mypy/nodes.py", line 1133, in accept
  File "mypy/semanal.py", line 1545, in visit_class_def
  File "mypy/semanal.py", line 1625, in analyze_class
  File "mypy/semanal.py", line 1652, in analyze_class_body_common
  File "mypy/semanal.py", line 1735, in apply_class_plugin_hooks
  File ".../site-packages/mypy_django_plugin/main.py", line 44, in transform_model_class
    helpers.get_django_metadata(sym.node)["model_bases"][ctx.cls.fullname] = 1
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'model_bases'
.../site-packages/django-stubs/contrib/auth/models.pyi:21: : note: use --pdb to drop into pdb
```

## System information

- OS: macOS
- `python` version: 3.11.2
- `django` version: 4.1.7
- `mypy` version: 1.0.1

intgr:
This should solve the error without too many changes. I still don't understand what these data structures do, but this seems to have been the original intent of the developer.

## Related issues

Fixes #1383.


mypy_django_plugin/main.py
```diff
@@ -42,7 +42,8 @@ def transform_model_class(ctx: ClassDefContext, django_context: DjangoContext) -
     sym = ctx.api.lookup_fully_qualified_or_none(fullnames.MODEL_CLASS_FULLNAME)
 
     if sym is not None and isinstance(sym.node, TypeInfo):
-        helpers.get_django_metadata(sym.node)["model_bases"][ctx.cls.fullname] = 1
+        bases = helpers.get_django_metadata(sym.node).setdefault("model_bases", {})
+        bases[ctx.cls.fullname] = 1
```
flaeppe: I'm not sure why there's a number(`1`) set as value for these. All I've encountered is they existence of the fullname as key and value is ignored.

Not saying we should change the type, but what do you think about hiding the `= 1` behind some method interface?

e.g.

```python
class DjangoTypeMetadata(...):
    ...

    def add(self, fullname: str, /) -> None:
        self[fullname] = 1
```

Just thinking that might allow later refactoring where e.g. the `1` could be removed.

intgr: Can't add methods to a TypedDict.

These should be converted to `set` instead of `dict` anyway, which already has an `add` method, so the additional method would end up unnecessary.

intgr: I could refactor these to `set`, but without understanding of this code, I feel that increases risk to introduce new bugs.

Does this metadata end up serialized into `.mypy_cache`? Seems like yes. Then that would cause backwards incompatibility unless we somehow also flush the cache. Then how do I do that?

I'd rather not deal with this scope creep, I'm just trying to fix an obvious bug here.

mypy_django_plugin/lib/helpers.py
```diff
@@ -43,8 +44,17 @@
...
 
-def get_django_metadata(model_info: TypeInfo) -> Dict[str, Any]:
-    return model_info.metadata.setdefault("django", {})
+class DjangoTypeMetadata(TypedDict, total=False):
+    from_queryset_manager: str
+    reverse_managers: Dict[str, str]
+    baseform_bases: Dict[str, int]
+    manager_bases: Dict[str, int]
+    model_bases: Dict[str, int]
+    queryset_bases: Dict[str, int]
```
flaeppe, line 53: I also think these different mappings are individually used. What I mean is that we use mypy's api to lookup a symbol(`sym`) instance where we mutate its metadata.

We always look for the same symbol instance, depending on if there's a model(`MODEL_CLASS_FULLNAME`), manager(`MANAGER_CLASS_FULLNAME`), etc

What I mean is that we won't add metadata for forms on the symbol instance of the base model.

I'm not totally sure if we could represent that in a nicer way. But I think what you have here is still an improvement.

Something contextually bound like (warning, pseudo):

```python
T = TypeVar("T", Literal["baseform_bases"], Literal["manager_bases"], ...)
class DjangoTypeMetadata(dict[T, int]):
    ...

def get_django_metadata(context: T, info: TypeInfo) -> DjangoTypeMetadata[T]:
    ...
```

Is that understandable? That we'll never work with more than 1 key at a time from the typed dict you declared here.

mypy_django_plugin/main.py
flaeppe: > Can't add methods to a TypedDict.
> ...

> I could refactor these to `set`, but without understanding of this code, I feel that increases risk to introduce new bugs.
> 
> Does this metadata end up serialized into `.mypy_cache`? Seems like yes. Then that would cause backwards incompatibility unless we somehow also flush the cache. Then how do I do that?
> 
> I'd rather not deal with this scope creep, I'm just trying to fix an obvious bug here.

No need to refactor to a `set`. I still meant to keep the dict type with an int. I only meant for later changes, and perhaps get rid of seeing/setting the `1` explicitly now.

Just an additional comment in regards to cache, I would suspect you'd have to explicitly tell people to flush cache.
Though I'm not 100% sure, but I have previously gotten the impression that mypy taints the whole cache when a plugin version has changed. But I have nothing to back that up with.

flaeppe (approved):
I just wanted to dump my thoughts while I was looking here. Lets merge it if you want to.

We had no regression case for this, right?

mypy_django_plugin/lib/helpers.py
intgr: Hmm, seems simpler to create a second helper function like this:

```python
def get_django_metadata_bases(
    model_info: TypeInfo, key: Literal["baseform_bases", "manager_bases", "model_bases", "queryset_bases"]
) -> Dict[str, int]:
    return get_django_metadata(model_info).setdefault(key, {})
```

intgr: I've pushed this change, I think it does something similar to what you were thinking?

flaeppe: Yes, precisely

intgr:
> We had no regression case for this, right?

I'll try if I can isolate a test case for this. It's something about the ordering of imports.

intgr:
> I'll try if I can isolate a test case for this. It's something about the ordering of imports.

I gave it a shot, but it seemed to be affected by a specific ordering of many imports across a dozen or so files, even minor changes would prevent the issue. I don't feel like putting more work into that. :)

If @jonodrew manages to isolate a simpler test case, then that would be great. But for now I'll merge this without a test.

PR #1567, opened 2023-06-22
Improve types for dispatch.dispatcher (Signal)

GabDug:
# I have made things!

Improved types for the dispatch module and Signals:

- Remove providing_args (removed in Django 4.0)
- Loosen send_robust tuple types (signals can return anything)
- Explicit kwargs for receiver (kwargs for connect)
- Better (not perfect) types for attributes

Any feedback is welcome, as it's my first contribution in this project. Thanks!


django-stubs/dispatch/__init__.pyi
```diff
@@ -1,2 +1,2 @@
-from django.dispatch.dispatcher import Signal as Signal
```
sobolevn: Any specific reason to change this?

django-stubs/dispatch/dispatcher.pyi
```diff
@@ -1,16 +1,22 @@
...
-    providing_args: Any
-    lock: Any
-    use_caching: Any
-    sender_receivers_cache: Any
-    def __init__(self, providing_args: list[str] = ..., use_caching: bool = ...) -> None: ...
+    receivers: list[Any]
+    lock: threading.Lock
+    use_caching: bool
+    sender_receivers_cache: WeakKeyDictionary[Any, Any] | dict[Any, Any]
+    _dead_receivers: bool
```
sobolevn: Usually we don't annotate protected attributes, any reason to make an exception?

django-stubs/dispatch/dispatcher.pyi
```diff
@@ -1,16 +1,22 @@
...
-    receivers: Any
-    providing_args: Any
-    lock: Any
-    use_caching: Any
-    sender_receivers_cache: Any
-    def __init__(self, providing_args: list[str] = ..., use_caching: bool = ...) -> None: ...
+    receivers: list[Any]
+    lock: threading.Lock
+    use_caching: bool
+    sender_receivers_cache: WeakKeyDictionary[Any, Any] | dict[Any, Any]
```
sobolevn: I think `WeakKeyDictionary` is just a runtime hack. What do you think about just `dict` or `MutableMapping`?

django-stubs/dispatch/dispatcher.pyi
```diff
@@ -19,7 +25,14 @@ class Signal:
...
     def send(self, sender: Any, **named: Any) -> list[tuple[Callable, str | None]]: ...
-    def send_robust(self, sender: Any, **named: Any) -> list[tuple[Callable, Exception | str]]: ...
+    def send_robust(self, sender: Any, **named: Any) -> list[tuple[Callable, Exception | Any]]: ...
     def _live_receivers(self, sender: Any) -> list[Callable]: ...
 
-def receiver(signal: list[Signal] | Signal, **kwargs: Any) -> Callable: ...
+def receiver(
+    signal: list[Signal] | tuple[Signal, ...] | Signal,
+    *,
+    receiver: Callable,
```
sobolevn: Can we annotate arg types? Maybe using `ParamSpec`?

sobolevn (commented):
Thank you! This is the first PR to remove something from `allowlist_todo`! Congrats 🎉

django-stubs/dispatch/dispatcher.pyi
GabDug: No specific reason, but it's type was easily inferable. I'll remove it for consistency.

django-stubs/dispatch/dispatcher.pyi
GabDug: Seems like a better fit indeed :)

django-stubs/dispatch/dispatcher.pyi
GabDug: Good idea! I also removed the receiver argument, which should not have been there.

django-stubs/dispatch/__init__.pyi
GabDug: None, rollback!

django-stubs/dispatch/dispatcher.pyi
```diff
@@ -1,16 +1,20 @@
...
+import threading
+from collections.abc import Callable, MutableMapping
+from logging import Logger
+from typing import Any, ParamSpec, TypeVar
 
-NONE_ID: Any
+NONE_ID: int
 NO_RECEIVERS: Any
 
+logger: Logger
```
flaeppe, line 9: I don't think we should include loggers in stubs?

```suggestion
```

IMO this could be a valid ignore for the allowlist, in all files

GabDug: Seems like we have a others `logger: Logger` in multiple files, could this be more suited for a different PR that makes everything consistent?

flaeppe: No need to clean up the others but I'm not seeing why we should add to the pile in this PR. If we agree to remove these later on anyways

sobolevn: In typeshed we include loggers, but I am also fine with allowlisting them.

flaeppe: Any particular reason or just for transparency?

sobolevn: You never know how the code is going to be used, but adding a variable is simple enough to follow the same route runtime does ¯\_(ツ)_/¯

adamchainz: I vote for adding them too, just because it's simple and we can never tell if users are importing them.

adamchainz (commented):
CI failures are due to use of `ParamSpec`:

```
E     ../../home/runner/work/django-stubs/django-stubs/django-stubs/dispatch/dispatcher.pyi:4: error: Module "typing" has no attribute "ParamSpec"; maybe "_ParamSpec"? (diff)
E     ../../home/runner/work/django-stubs/django-stubs/django-stubs/dispatch/dispatcher.pyi:4: note: Use `from typing_extensions import ParamSpec` instead (diff)
E     ../../home/runner/work/django-stubs/django-stubs/django-stubs/dispatch/dispatcher.pyi:4: note: See https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module (diff)
E     ../../home/runner/work/django-stubs/django-stubs/django-stubs/dispatch/dispatcher.pyi:38: error: The first argument to Callable must be a list of types, parameter specification, or "..." (diff)
E     ../../home/runner/work/django-stubs/django-stubs/django-stubs/dispatch/dispatcher.pyi:38: note: See https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas (diff)
```

Is it possible to use a `TypeVar` bound to `Callable` instead?

GabDug:
> CI failures are due to use of `ParamSpec`:
> 
> ```
> E     ../../home/runner/work/django-stubs/django-stubs/django-stubs/dispatch/dispatcher.pyi:4: error: Module "typing" has no attribute "ParamSpec"; maybe "_ParamSpec"? (diff)
> E     ../../home/runner/work/django-stubs/django-stubs/django-stubs/dispatch/dispatcher.pyi:4: note: Use `from typing_extensions import ParamSpec` instead (diff)
> E     ../../home/runner/work/django-stubs/django-stubs/django-stubs/dispatch/dispatcher.pyi:4: note: See https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module (diff)
> E     ../../home/runner/work/django-stubs/django-stubs/django-stubs/dispatch/dispatcher.pyi:38: error: The first argument to Callable must be a list of types, parameter specification, or "..." (diff)
> E     ../../home/runner/work/django-stubs/django-stubs/django-stubs/dispatch/dispatcher.pyi:38: note: See https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas (diff)
> ```
> 
> Is it possible to use a `TypeVar` bound to `Callable` instead?

Seems to yield the same returns for Mypy and Pyright, so I changed it, thanks!

django-stubs/dispatch/dispatcher.pyi
```diff
@@ -19,7 +23,15 @@ class Signal:
...
 
-def receiver(signal: list[Signal] | Signal, **kwargs: Any) -> Callable: ...
+_F = TypeVar("_F", bound=Callable[..., Any])
+
+def receiver(
+    signal: list[Signal] | tuple[Signal, ...] | Signal,
+    *,
+    sender: object | None = ...,
+    weak: bool = ...,
+    dispatch_uid: str | None = ...
```
sobolevn:
```
dispatch_uid
                An identifier used to uniquely identify a particular instance of
                a receiver. This will usually be a string, though it may be
                anything hashable.
```

So, let's use `Hashable` here.

GabDug: Updated it for `receiver`, `connect` and `disconnect`.

sobolevn (approved):
Thank you!

PR #1570, opened 2023-06-23
Migrate setuptools configuration to pyproject.toml + PEP621

### LINKED ISSUE #1781: setup.py install is deprecated
Will need to deal with this at one point. https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

This warning shows up when making a release:

```
.../site-packages/setuptools/_distutils/cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated.
!!

        ********************************************************************************
        Please avoid running ``setup.py`` directly.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.

        See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
        ********************************************************************************

!!
```

GabDug:
# I have made things!

This PR is an experiment to migrate from setuptools `setup.py` configuration to the more declarative `pyproject.toml` format, following [PEP 621](https://peps.python.org/pep-0621/).

It allows to remove custom discovery functions in `setup.py`, that are also apparently not needed, and moving to `pyproject.toml` may also enable the project (or its contributors) to use PEP621 compliant project managers like [PDM](https://pdm.fming.dev/) or [hatch](https://hatch.pypa.io/latest/).

> Note: while using `pyproject.toml` is now recommended for new projects by setuptools, the `[tool.setuptools]` configuration is in beta and may change, as indicated by the [docs](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html). 

## Changes to the wheels dist_info

I did not find any differences in included files or modules, but only changes to the metadata:

1. Author and maintainers format
```
 "author": "Maksim Kurnikov",
 "author_email": "maxim.kurnikov@gmail.com",
 "maintainer": "Marti Raudsepp",
 "maintainer_email": "marti@juffo.org",
```
to 
```
"author_email": "Maksim Kurnikov <maxim.kurnikov@gmail.com>",
"maintainer_email": "Marti Raudsepp <marti@juffo.org>",
```
2. Added two urls
```
 {
  "label": "Homepage",
  "url": "https://github.com/typeddjango/django-stubs"
 },
 {
   "label": "Repository",
   "url": "https://github.com/typeddjango/django-stubs"
}
```
3. Removed `home_page` from metadata 

Regarding urls/homepage, it may not be issues, see [this discussion on setuptools project](https://github.com/pypa/setuptools/discussions/3318).

I'm willing to investigate/see upstream if you consider it might be important :)

## Related issues

* Fixes #1781


sobolevn (commented):
Let's please separate comma and `setup.py` changes

GabDug:
> Let's please separate comma and `setup.py` changes

Yes, unfortunately it seems pre-commit.ci wanted to make the changes.

pyproject.toml
```diff
@@ -21,4 +21,59 @@ select = [
...
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Typing :: Typed",
+    "Framework :: Django",
+    "Framework :: Django :: 3.2",
+    "Framework :: Django :: 4.1",
+    "Framework :: Django :: 4.2",
+]
+requires-python = ">=3.8"
+maintainers = [{ name = "Marti Raudsepp", email = "marti@juffo.org" }]
```
sobolevn, line 58: Oh, we can now add all team as maintainers :)
Will do later.

sobolevn (commented):
Thanks! Looks great!

sobolevn:
You can create a PR with `,` changes first. Later we can rebase this one.

pyproject.toml
```diff
@@ -21,4 +21,59 @@ select = [
 split-on-trailing-comma = false
 
 [build-system]
-requires = ["setuptools<64", "wheel"]
+requires = ["setuptools>=68,<69", "wheel"]
```
adamchainz: The [recommendation](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use) now is to use:

```suggestion
requires = ["setuptools"]
```

adamchainz (changes requested):

GabDug:
Seems like mypy has trouble detecting the source/stubs with this new way of declaring the project. I've tried a few things with no luck, any insights are appreciated :)

sobolevn (commented):
Looking into it

sobolevn:
I have no idea what is going on :)

GabDug:
> I have no idea what is going on :)

I'm puzzled as well!

I'll probably try to use another build-backend like hatchling, we never know.

intgr:
- [ ] Please update `django_stubs_ext/setup.py` as well.

intgr:
> Please update `django_stubs_ext/setup.py` as well.

Ugh, that can be a bit problematic. Black for example only looks for the closest `pyproject.toml`, so we would need to duplicate Black configuration? Not sure about other tools.

intgr:
> Seems like mypy has trouble detecting the source/stubs with this new way of declaring the project.

This may be related to the changes regarding editable installs in recent setuptools. We've since added `SETUPTOOLS_ENABLE_FEATURES=legacy-editable` in #1759 which may help with this.

pyproject.toml
```diff
@@ -31,15 +31,16 @@ known-first-party = ["django_stubs_ext", "mypy_django_plugin"]
 split-on-trailing-comma = false
 
 [build-system]
-requires = ["setuptools"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling"]
```
sobolevn, line 34: No idea what `hatchling` is. 

If other maintainers also don't know what this is, I prefer not to rely on it :(

GabDug: I'll test again with setuptools then :)

flaeppe: I've used `hatchling` for other projects, a totally fair build system that does its job. It also comes with a sort of ecosystem for package management called `hatch`, which is completely __optional__ and I've _only_ used the build system named `hatchling`.

One thing I did like about it was that it validates the `project` section in your `pyproject.toml`, which was quite nice.  I found that the documentation for `hatchling` was quite simple to pick up on too, here: https://hatch.pypa.io/latest/config/build/

It's maintained by https://github.com/pypa/ and e.g. https://github.com/pypa/virtualenv/ is packaged with `hatchling`: https://github.com/pypa/virtualenv/blob/9eac8a6b85c047e2287c95bdb9caf6f0fbf35041/pyproject.toml#L1-L6

It's nice to extend `hatchling` with `hatchling-vcs` to get versioning via version control(`git`). Should be similar to `setuptools-vcs`. Which is 1 less thing to remember/do when crafting a release.

---

I'm +1 on using `hatchling`

flaeppe: Here are their own reasons why to use `hatchling` as build backend instead of setuptools:

https://hatch.pypa.io/latest/why/#build-backend

flaeppe:
Is it possible to revive this PR? Would be nice if we could upgrade to a `pyproject.toml`

PR #1571, opened 2023-06-23
feat: type db.backends introspection and operations on 4.2

### LINKED ISSUE #1554: Database operations method signatures are not updated for Django >= 4.1
# Bug report

Database operations method signatures are not updated for Django >= 4.1

## What's wrong

According to the documentation, there is a backwards incompatibility related change:
https://docs.djangoproject.com/en/4.2/releases/4.1/#backwards-incompatible-changes-in-4-1

## How is that should be
The listed methods signatures should be updated accordingly, based on the methods signatures in Django:
https://github.com/django/django/blob/main/django/db/backends/base/operations.py

## System information

- OS:
- `python` version: 3.9.16
- `django` version: 4.2.2
- `mypy` version: 1.3.0

GabDug:
# I have made things!

Hey!

This PR attemps to improve the typings for `django.db.backends.*.operations` and `django.db.backends.*.introspection` .

I also deleted some overrides of the base that were just typed as `Any`, which was sometime less specific than the base. I was wondering if there were some type-checker issues that forced us to copy typings on subclasses? If not, we may delete even more duplicates.

Typings can still be improved and I need to dig a little deeper to remove some `Any`s and check things, feedback is very much appreciated!

## Related issues

- Closes #1554


sobolevn:
Why is it marked as a draft? :)

GabDug:
> Why is it marked as a draft? :)

Put it in ready for review -- sorry about that.

django-stubs/db/backends/base/operations.pyi
```diff
@@ -102,5 +106,8 @@ class BaseDatabaseOperations:
     def window_frame_rows_start_end(self, start: int | None = ..., end: int | None = ...) -> tuple[str, str]: ...
     def window_frame_range_start_end(self, start: int | None = ..., end: int | None = ...) -> tuple[str, str]: ...
     def explain_query_prefix(self, format: str | None = ..., **options: Any) -> str: ...
-    def insert_statement(self, ignore_conflicts: bool = ...) -> str: ...
-    def ignore_conflicts_suffix_sql(self, ignore_conflicts: Any | None = ...) -> str: ...
+    def insert_statement(self, on_conflict: OnConflict | None = ...) -> str: ...
+    def on_conflict_suffix_sql(
+        self, fields: Any, on_conflict: Any, update_fields: Any, unique_fields: Any
+    ) -> str | Any: ...
+    def format_for_duration_arithmetic(self, sql: str) -> str | NoReturn: ...
```
sobolevn:
```suggestion
    def format_for_duration_arithmetic(self, sql: str) -> str: ...
```

`str` is equal to `str | NoReturn`

sobolevn (commented):
Thanks!

sobolevn (approved):
I will merge this, we can always fix some issues later! Thanks!

PR #1572, opened 2023-06-23
Upgrade to Mypy 1.4.0

christianbundy:
## Problem

We have tests that make assertions around Mypy output, and Mypy 1.4.0 contains changes to that output.

## Solution

Use the new `force_uppercase_builtins` and `force_union_syntax` settings to keep our tests passing without mass edits to test files.

## Related issues

* https://github.com/typeddjango/django-stubs/pull/1559


mypy.ini
```diff
@@ -15,6 +15,8 @@ disallow_untyped_defs = true
 disallow_incomplete_defs = true
 show_error_codes = false
 disable_error_code = empty-body
+force_uppercase_builtins = true
```
sobolevn, line 19:
```suggestion
# TODO: update our output assertions to match a new syntax
force_uppercase_builtins = true
```

sobolevn (approved):
Thank you!

christianbundy:
Thanks! Longer-term, how do we want to solve this? Originally I started fixing all of the errors, but I realized that the Mypy output is different depending on which Python version you're using since we don't specify [`python_version`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-python_version) in `mypy.ini`.

I'm on the fence between: 

- `python_version: 3.8` using the lowest supported Python version ensures that Mypy is using the more compatible semantics and will yell at us for doing things that won't work in Python 3.9
- `python_version: 3.11` using the highest supported Python version provides the latest and greatest in error messages / etc., and doesn't actually require that tests are run with Python 3.11

Do you have a preference? I'm happy to put together a PR that updates our output assertions, but I didn't want to put much time into it before aligning on a strategy.

christianbundy:
(I had a bit of free time this morning and opened https://github.com/typeddjango/django-stubs/pull/1577 as a proof of concept to consider.)

setup.py
```diff
@@ -32,7 +32,7 @@ def find_stub_files(name: str) -> List[str]:
 ]
 
 extras_require = {
-    "compatible-mypy": ["mypy>=1.3.0,<1.4"],
+    "compatible-mypy": ["mypy>=1.3.0,<1.5"],
```
intgr, line 35: We have always updated the `compatible-mypy` lower bound as well.

PR #1573, opened 2023-06-24
Chore: set Black Python target to 3.8+ explicitely

GabDug:
# I have made things!

In #1570, setting the `project.requires-python` to `>=3.8` made Black change its configuration.

Here's a PR to update its configuration explicitely, and reformat the files according to the newer configuration.


sobolevn (approved):
Thanks!

PR #1574, opened 2023-06-24
fix: Window.order_by accepts str field_names, asc/desc accepts None in 4.1+

### LINKED ISSUE #1512: django.db.models.Window order_by argument does not accept F objects but accepts strings
# Bug report

## What's wrong

When instantiating a `Window` object, its `order_by` argument is mistyped.

MyPy reports the following error when I have `Window(..., order_by="field_name")`
```
error: Argument "order_by" to "Window" has incompatible type "str"; expected "Union[Sequence[Union[BaseExpression, F]], BaseExpression, F, None]"  [arg-type]
```

## How is that should be

But if I change my code to be `Window(..., order_by=F("field_name"))`, I get the following Django error:
```
ValueError: Window.order_by must be either a string reference to a field, an expression, or a list or tuple of them.
```

## System information

- OS:
- `python` version: 3.10.11
- `django` version: 4.1.8
- `mypy` version: mypy 1.2.0 (compiled: yes)

GabDug:
# I have made things!

Change window.order_by signature to allow field names (`str`), added in Django 4.1.

FIx `asc`/`desc` signatures as well, which now defaults to None for some args, see  https://docs.djangoproject.com/en/4.2/ref/models/expressions/#django.db.models.Expression.asc

## Related issues

Closes #1512


sobolevn (approved):
Thanks!

PR #1575, opened 2023-06-24
Type django.utils.* against Django 4.2

### LINKED ISSUE #1474: conditional_escape() should return SafeString
# Bug report

## What's wrong

The [stub definition](https://github.com/typeddjango/django-stubs/blob/b81b1bf3868eaab9fecebdd56ed817ffcb07fad8/django-stubs/utils/html.pyi) for [`conditional_escape`](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.html.conditional_escape) says that it returns a `str`

The [django code](https://github.com/django/django/blob/4bf4222010fd8e413963c6c873e4088614332ef9/django/utils/html.py#L82) follows one of two paths:
- if already escaped, then call `__html__` (this calls into [`SafeString`/`SafeData`](https://github.com/django/django/blob/4bf4222010fd8e413963c6c873e4088614332ef9/django/utils/safestring.py#L13) which just returns the same object back).
- it not escaped then call [`escape`](https://github.com/django/django/blob/4bf4222010fd8e413963c6c873e4088614332ef9/django/utils/html.py#L17) which returns a new `SafeString`

Running `mypy` on `test.py`:
```
from django.utils.html import conditional_escape
from django.utils.safestring import mark_safe
from django.utils.safestring import SafeString

a = SafeString("<")
b = conditional_escape("<")
c = mark_safe("<")

if TYPE_CHECKING:
    reveal_type(a)
    reveal_type(b)
    reveal_type(c)

print(a.__class__)
print(b.__class__)
print(c.__class__)
```

mypy output is:
```
test.py:11: note: Revealed type is "django.utils.safestring.SafeString"
test.py:12: note: Revealed type is "builtins.str"
test.py:13: note: Revealed type is "django.utils.safestring.SafeString"
```

runtime output is:
```
<class 'django.utils.safestring.SafeString'>
<class 'django.utils.safestring.SafeString'>
<class 'django.utils.safestring.SafeString'>
```

## How is that should be

mypy output:
```
test.py:11: note: Revealed type is "django.utils.safestring.SafeString"
test.py:12: note: Revealed type is "django.utils.safestring.SafeString"
test.py:13: note: Revealed type is "django.utils.safestring.SafeString"
```

## System information

- OS: MacOS
- `python` version: 3.10
- `django` version: 3.2
- `mypy` version: 1.2.0

## Pedantic Note
Strictly speaking [django allows](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.html.html_safe) for objects that implement `__html__` and are _not_ instances of `SafeData`.

To be 100% correct the declarations would have to create a protocol for objects that implement `__html__`, but IMO that's a somewhat obscure edge case. I note also that the existing [mark_safe](https://github.com/typeddjango/django-stubs/blob/b81b1bf3868eaab9fecebdd56ed817ffcb07fad8/django-stubs/utils/safestring.pyi#L23) type definition already assumes `SafeData` so just marking `conditional_escape` as returning `SafeString` wouldn't be introducing an issue that doesn't already exist.

GabDug:
# I have made things!

This MR synchronises all changes of Django 4.2 to the stubs for the `django.utils.*` packages.

Remove all utils from the allowlist-todo, but adds some to the regular allowlist, notably all the lazy regexes. 

While working on this PR, I noticed that stubtest does not return errors when passed the `--generate-allowlist` argument, I have not changed this behaviour as I was unsure if it was intentional.

## Related issues

Closes #1474


sobolevn:
@GabDug my bad, sent https://github.com/typeddjango/django-stubs/pull/1576

tests/typecheck/utils/test_functional.yml
```diff
@@ -8,7 +8,7 @@
           def attr(self) -> List[str]: ...
 
           reveal_type(attr)      # N: Revealed type is "django.utils.functional.cached_property[builtins.list[builtins.str]]"
-          reveal_type(attr.name) # N: Revealed type is "builtins.str"
+          reveal_type(attr.name) # N: Revealed type is "Union[builtins.str, None]"
```
GabDug, line 11: Implementation indicates:
"The name argument is deprecated as it's unnecessary as of Python 3.6.",
  RemovedInDjango50Warning,

django-stubs/utils/connection.pyi
```diff
@@ -17,13 +19,13 @@ class BaseConnectionHandler(Generic[_T]):
     settings_name: str | None
     exception_class: type[Exception]
     thread_critical: bool
+    settings: cached_property[dict[str, Any]]
```
sobolevn: Maybe `@cached_property def settings`?

django-stubs/utils/datastructures.pyi
```diff
@@ -83,11 +84,13 @@ class MultiValueDict(dict[_K, _V]):
     def __iter__(self) -> Iterator[_K]: ...
     # Fake to make `values` work properly
     def values(self) -> Iterator[_V | list[object]]: ...  # type: ignore[override]
+    def __copy__(self) -> MultiValueDict: ...
```
sobolevn: No, it returns `Self` type: see https://github.com/django/django/blob/fcc5091b4a7c29e2912f6c8b2e0ebc64f4853457/django/utils/datastructures.py#L95-L105

django-stubs/utils/datastructures.pyi
```diff
@@ -83,11 +84,13 @@ class MultiValueDict(dict[_K, _V]):
...
+    def __deepcopy__(self, memo: MutableMapping[int, Incomplete]) -> MultiValueDict: ...
```
sobolevn: The same here.

django-stubs/utils/decorators.pyi
```diff
@@ -16,7 +15,7 @@ def method_decorator(decorator: Callable | Iterable[Callable], name: str = ...)
 def decorator_from_middleware_with_args(middleware_class: type) -> Callable: ...
 def decorator_from_middleware(middleware_class: type) -> Callable: ...
 def make_middleware_decorator(middleware_class: type[MiddlewareMixin]) -> Callable: ...
-
+@type_check_only
```
sobolevn: Two things:
1. Looks like this should be internal: `_AsyncGetResponseCallable`
2. Can you please also fix https://github.com/typeddjango/django-stubs/issues/1269 while we are at it?

django-stubs/utils/deprecation.pyi
```diff
@@ -29,10 +30,15 @@ class DeprecationInstanceCheck(type):
     deprecation_warning: type[Warning]
     def __instancecheck__(self, instance: Any) -> bool: ...
 
+@type_check_only
 class GetResponseCallable(Protocol):
```
sobolevn: It should be protected instead.

```suggestion
class _GetResponseCallable(Protocol):
```

django-stubs/utils/functional.pyi
```diff
@@ -63,6 +64,7 @@ class LazyObject:
     def __delattr__(self, name: str) -> None: ...
     def __reduce__(self) -> tuple[Callable, tuple[Model]]: ...
     def __copy__(self) -> LazyObject: ...
```
sobolevn, line 66: It can also return wrapped object. So, basically we need to make `LazyObject` generic.
But, for now you can make it return `Any` with a `TODO` note.

See https://github.com/django/django/blob/fcc5091b4a7c29e2912f6c8b2e0ebc64f4853457/django/utils/functional.py#L332-L339

django-stubs/utils/tree.pyi
```diff
@@ -14,10 +14,14 @@ class Node:
     def __init__(
         self, children: _NodeChildren | None = ..., connector: str | None = ..., negated: bool = ...
     ) -> None: ...
+    @classmethod
+    def create(cls, children: _NodeChildren | None = ..., connector: str | None = ..., negated: bool = ...) -> Node: ...
+    def copy(self) -> Node: ...
```
sobolevn, line 19: Yes, this is `Node`: https://github.com/django/django/blob/fcc5091b4a7c29e2912f6c8b2e0ebc64f4853457/django/utils/tree.py#L36 👍

scripts/stubtest/allowlist.txt
```diff
@@ -13,3 +13,66 @@ django.contrib.admin.migrations.*
 django.contrib.auth.migrations.*
 django.contrib.flatpages.migrations.*
 django.contrib.contenttypes.migrations.*
+
+# BaseArchive abstract methods that take no argument, but typed with arguments to match the Archive and TarArchive Implementations
+django.utils.archive.BaseArchive.list
+django.utils.archive.BaseArchive.extract
+
+# ImmutableList always raise when calling these,
```
sobolevn, line 21:
```suggestion
# ImmutableList always raise when calling these,
# later they can be added with `-> NoReturn` overrides.
```

scripts/stubtest/allowlist.txt
```diff
@@ -13,3 +13,66 @@ django.contrib.admin.migrations.*
...
+django.utils.translation.template.plural_re
+django.utils.translation.template.inline_re
+django.utils.translation.template.endblock_re
+django.utils.translation.template.dot_re
+django.utils.translation.template.context_re
+django.utils.translation.template.constant_re
+django.utils.translation.template.block_re
+django.utils.version.version_component_re
+
+# `Trans` class is deleted in Django
```
sobolevn: We can also delete it, we try to match Django@4.2

sobolevn (commented):
Great work! Thanks a lot!

django-stubs/utils/datastructures.pyi
GabDug: Fixed `Self` in `datastructures.py`, used the typing_extensions one good catch! Was there a specific reason to import `Self` from `_typeshed`, which is a TypeVar instead of `Self` from typing_extensions?

scripts/stubtest/allowlist.txt
GabDug: My comment here may be incomplete -- the class is deleted, but there's an instance still present, thus the class may be used for typing.

Should we still delete it from the stubs, or maybe rename it to something type-specific?

django-stubs/utils/connection.pyi
GabDug: We could indeed use
```python
@cached_property
def settings(self) -> dict[str, Any]: ...
```
but not 
```python
@cached_property[dict[str, Any]]
def settings(self) -> dict[str, Any]: ...
```
due to the generic decorator syntax in earlier Python version.

The one-line approach seemed to work fine on all version tho, with proper types for mypy and pyright, but may not be needed here, as we don't need a union like `cached_property[dict[str, Any]] | dict[str, Any]`.

What do you think?

django-stubs/utils/functional.pyi
GabDug: Added the note, Generic LazyObject would indeed be a nice addition!

django-stubs/utils/datastructures.pyi
sobolevn: Legacy stuff :)

GabDug: Okay, I will move the import to typing_extensions and remove redundant `self: Self` / `cls: type[Cls]` from this file then :)

scripts/stubtest/allowlist.txt
sobolevn: We can replace it with a protocol, no need to have something that runtime does not have: https://github.com/django/django/blob/879e5d587b84e6fc961829611999431778eb9f6a/django/utils/translation/__init__.py#L49

GabDug: I'll delete the `Trans` stub, as it only seems to be used in this module with `_trans`.

django-stubs/utils/decorators.pyi
GabDug: Made it internal, moved it alongside a usage on MiddlewareMixin.

tests/typecheck/contrib/auth/test_misc.yml
```diff
@@ -3,9 +3,15 @@
       from django.contrib.auth import get_user
       from django.contrib.sessions.middleware import SessionMiddleware
       from django.http import HttpRequest
+      from django.http.response import HttpResponse
+      from django.http.request import HttpRequest
+      from typing import Callable
```
sobolevn:
```suggestion
```

tests/typecheck/contrib/auth/test_misc.yml
```diff
@@ -3,9 +3,15 @@
       from django.contrib.auth import get_user
       from django.contrib.sessions.middleware import SessionMiddleware
       from django.http import HttpRequest
+      from django.http.response import HttpResponse
+      from django.http.request import HttpRequest
```
sobolevn:
```suggestion
      from django.http import HttpRequest, HttpResponse
```

django-stubs/utils/connection.pyi
sobolevn: Let's use

```python
@cached_property 
def settings(self) -> dict[str, Any]: ...
```

django-stubs/utils/deprecation.pyi
```diff
@@ -29,10 +30,19 @@ class DeprecationInstanceCheck(type):
     deprecation_warning: type[Warning]
     def __instancecheck__(self, instance: Any) -> bool: ...
 
-class GetResponseCallable(Protocol):
+@type_check_only
```
sobolevn: We generally don't use `@type_check_only` for protected protocols.

```suggestion
```

Let's do it consistently.

django-stubs/utils/deprecation.pyi
```diff
@@ -29,10 +30,19 @@ class DeprecationInstanceCheck(type):
...
+class _GetResponseCallable(Protocol):
     def __call__(self, __request: HttpRequest) -> HttpResponseBase: ...
 
+@type_check_only
```
sobolevn:
```suggestion
```

sobolevn (commented):
CC @intgr 

There are a lot of changes, I would highly appreciate a second pair of eyes :)

django-stubs/utils/decorators.pyi
```diff
@@ -16,10 +13,6 @@ def method_decorator(decorator: Callable | Iterable[Callable], name: str = ...)
 def decorator_from_middleware_with_args(middleware_class: type) -> Callable: ...
 def decorator_from_middleware(middleware_class: type) -> Callable: ...
 def make_middleware_decorator(middleware_class: type[MiddlewareMixin]) -> Callable: ...
-
-class AsyncGetResponseCallable(Protocol):
-    def __call__(self, __request: HttpRequest) -> Awaitable[HttpResponseBase]: ...
```
adamchainz, line 21: This was just added in #1425, I don't think we should be removing it?

adamchainz (changes requested):

django-stubs/utils/decorators.pyi
GabDug: Renamed it to `_AsyncGetResponseCallable` with regard to @sobolevn recommendation, and moved it in `deprecation.pyi`, where I added a usage.

I must say I don't have the full context for #1425 and #1269, so I'd be happy to rollback the changes and leave it for later.

sobolevn: This is my mistake: adding public protocols is not a good idea: because they do not exist at runtime. We must not promote importing them.

As far as I understand, this protocol is only used in our `django.utils` package.
So, it should be protected.

adamchainz (approved):
Alrighty

sobolevn (approved):

django-stubs/utils/decorators.pyi
intgr: Thanks for cleaning this up, I had already forgotten.

PR #1577, opened 2023-06-24
Set `python_version` and fix uppercase builtins

christianbundy:
**Problem**

When upgrading to Mypy 1.4 (#1572) I added a setting to force uppercase builtins to ensure that tests would pass. In [a comment](https://github.com/typeddjango/django-stubs/pull/1572#issuecomment-1605606030) I highlighted that we may not be able to remove this setting without changing `python_version` otherwise tests would only pass on some versions of Python.

I tried setting `python_version = 3.8`, but that caused [test failures](https://github.com/typeddjango/django-stubs/actions/runs/5365122788). Candidly, I didn't take very much time to investigate these.

**Solution**

Configure `python_version = 3.11` to match our highest supported Python version, and fix uppercase builtins. I can fix the union types in a future PR, but I want to keep this one as minimal as possible. If this is the direction we want to go, we can merge this PR, but I'm open to other options.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -15,7 +15,7 @@ disallow_untyped_defs = true
 disallow_incomplete_defs = true
 show_error_codes = false
 disable_error_code = empty-body
-python_version = 3.8
+python_version = 3.11
 
 plugins =
     mypy_django_plugin.main
--- a/mypy.ini
+++ b/mypy.ini
@@ -16,6 +16,9 @@ disallow_incomplete_defs = true
 show_error_codes = false
 disable_error_code = empty-body
 python_version = 3.11
+# TODO: update our output assertions to match a new syntax
+force_union_syntax = true
+
 
 plugins =
     mypy_django_plugin.main
--- a/mypy.ini
+++ b/mypy.ini
@@ -15,10 +15,7 @@ disallow_untyped_defs = true
 disallow_incomplete_defs = true
 show_error_codes = false
 disable_error_code = empty-body
-# TODO: update our output assertions to match a new syntax
-force_uppercase_builtins = true
-force_union_syntax = true
-
+python_version = 3.8
 
 plugins =
     mypy_django_plugin.main
--- a/tests/typecheck/contrib/admin/test_options.yml
+++ b/tests/typecheck/contrib/admin/test_options.yml
@@ -126,7 +126,7 @@
 
       class A(admin.ModelAdmin):
           formfield_overrides = {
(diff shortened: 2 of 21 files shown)


christianbundy:
Hmm. This is failing locally too -- I'm not sure why.

```
_______________________ models_triple_circular_reference _______________________
/home/runner/work/django-stubs/django-stubs/tests/typecheck/fields/test_related.yml:252: 
E   Actual:
E     ...
E     main:3: error: "User" has no attribute "profile" (diff)
E     main:3: note: Revealed type is "Any"          (diff)
E   Expected:
E     ...
E     main:3: note: Revealed type is "myapp.models.profile.Profile" (diff)
E     E: main:3: note: Revealed type is "myapp.models.profile.Profile"
E     A: main:3: error: "User" has no attribute "profile"
E                ^
```

mypy.ini
```diff
@@ -15,8 +15,8 @@ disallow_untyped_defs = true
 disallow_incomplete_defs = true
 show_error_codes = false
```
intgr, line 16: Note that `show_error_codes = false` also exists here as legacy, and should be fixed eventually.

You can move it below the `# TODO` comment you have added.

mypy.ini
```diff
@@ -15,8 +15,8 @@ disallow_untyped_defs = true
...
 disable_error_code = empty-body
+python_version = 3.11
```
intgr, line 18: `python_version` also determines the value of `sys.version_info` used when interpreting typeshed stubs (for conditionals such as https://github.com/python/typeshed/blob/main/stdlib/typing.pyi#L93-L113)

So if we override this setting, we lose test coverage with different variations of `typeshed` definitions, and would no longer catch subtle compatibility issues in CI.

I very much prefer the modern formatting of types in error messages, but still I would err on the side of caution and prioritize test coverage here.

I wish mypy had a way to override `force_uppercase_builtins` in the opposite direction, e.g. `force_lowercase_builtins = true`. Maybe they would be willing to entertain such a PR if someone is interested doing the work. And same for `force_union_syntax`.

intgr (commented):
Thanks for the effort and sorry that this PR has languished without review for so long. Clearly there is room for improvement in how we handle PRs.

mypy.ini
sobolevn: Yes, I agree - we cannot touch `python_version` :(

UnknownPlatypus:
Superseded by #2744

PR #1578, opened 2023-06-25
Use `report_config_data` mypy plugin hook to reset cache on changes

### LINKED ISSUE #1568: Support `report_config_data` mypy plugin hook
When the configuration of our plugin changes, mypy cache is not reset. So, settings are not changed.
There might be other side-effects of this.

I will send a PR shortly.

Docs: https://mypy.readthedocs.io/en/stable/extending_mypy.html?highlight=report_config_data#current-list-of-plugin-hooks

sobolevn:
Now, when any of our settings will change, mypy will be clear the existing cache. This is especially useful for stateful checks: `incremental`, `dmypy`, etc.

However, I don't think that it is possible to easily test this. This is how it is tested in mypy itself:

```ini
[case testPluginConfigData]
# flags: --config-file tmp/mypy.ini
import a
import b
[file a.py]
[file b.py]
[file test.json]
{"a": false, "b": false}
[file test.json.2]
{"a": true, "b": false}

[file mypy.ini]
\[mypy]
plugins=<ROOT>/test-data/unit/plugins/config_data.py

# The config change will force a to be rechecked but not b.
[rechecked a]
```

The only testing I've done is manual. It seems to work.

Closes https://github.com/typeddjango/django-stubs/issues/1568

### DIFF
--- a/mypy_django_plugin/config.py
+++ b/mypy_django_plugin/config.py
@@ -112,3 +112,10 @@ class DjangoPluginConfig:
             self.strict_settings = parser.getboolean(section, "strict_settings", fallback=True)
         except ValueError:
             exit_with_error(INVALID_BOOL_SETTING.format(key="strict_settings"))
+
+    def to_json(self) -> dict[str, Any]:
+        """We use this method to reset mypy cache via `report_config_data` hook."""
+        return {
+            "django_settings_module": self.django_settings_module,
+            "strict_settings": self.strict_settings,
+        }
--- a/mypy_django_plugin/config.py
+++ b/mypy_django_plugin/config.py
@@ -113,7 +113,7 @@ class DjangoPluginConfig:
         except ValueError:
             exit_with_error(INVALID_BOOL_SETTING.format(key="strict_settings"))
 
-    def to_json(self) -> dict[str, Any]:
+    def to_json(self) -> Dict[str, Any]:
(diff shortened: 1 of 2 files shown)


flaeppe (approved):
Cool stuff 👍

flaeppe:
Would be interesting to experiment with models, etc

sobolevn:
@flaeppe any feedback from real-life projects is welcome 😉

flaeppe:
Sure, I'll try it out.

Additional thought: would this invalidate the whole cache if settings changes? Perhaps that is a bit drastic? Wouldn't it be enough to purge the cache for just settings file(s)?

I suppose it could be problematic to try to purge cache of any imports in a settings file though.

First thing coming to mind: a configuration declaring additional settings files to purge. Although theoretically it should be possible to automatically figure out additional settings modules to purge

sobolevn:
> Additional thought: would this invalidate the whole cache if settings changes? Perhaps that is a bit drastic? Wouldn't it be enough to purge the cache for just settings file(s)?

No, because settings can change installed apps, models, urls, etc.

flaeppe:
Yeah, I suppose

flaeppe:
I was looking at this like: if any _contents_ of settings file(s) changes. But I realised now when testing it out that wasn't the case.

PR #1582, opened 2023-06-26
CI: Run tests and pre-commit using newest Python version

intgr:
The choice of Python 3.9 was arbitrary. As agreed in other pull request https://github.com/typeddjango/django-stubs/pull/1560#discussion_r1237101876, we should use the newest supported Python version.

Python 3.11 is also slightly faster.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,7 +45,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.9']
+        python-version: ['3.11']
     steps:
       - uses: actions/checkout@v3
       - name: Setup system dependencies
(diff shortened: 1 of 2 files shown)


sobolevn (approved):

intgr:
For some reason the `test (3.11)` step was actually slower (15 minutes, https://github.com/typeddjango/django-stubs/actions/runs/5376468673/jobs/9753662335) than previously (~11 minutes or so): 

Maybe variability in GitHub runners. Let's see if retry changes anything.

adamchainz:
> Maybe variability in GitHub runners. Let's see if retry changes anything.

There is *wild* variability in GitHub runners, I wouldn't read too much into it.

intgr:
Yep indeed, after merging here it runs in 11 minutes again: https://github.com/typeddjango/django-stubs/actions/runs/5377562795/jobs/9756188667

adamchainz:
Interesting... 🤷‍♂️

PR #1583, opened 2023-06-26
4.2 typings for db.migrations.operations

GabDug:
# I have made things!

Hello again!

Sorry to make such a big PR again, this one is focused on `django.db.migrations.operations`, but also improves some misc types, to make them on par with Django 4.2 

It removes 85 elements from the allowlist-todo.


django-stubs/db/migrations/operations/base.pyi
```diff
@@ -13,6 +14,7 @@ class Operation:
     elidable: bool
     serialization_expand_args: list[str]
     _constructor_args: tuple[Sequence[Any], dict[str, Any]]
+    def __new__(cls, *args: Any, **kwargs: Any) -> Self: ...
```
adamchainz, line 17: this seems kinda redundant? 🤔 I mean, I guess if stubtest says it's needed...

django-stubs/db/migrations/state.pyi
```diff
@@ -7,8 +7,11 @@ from django.apps.registry import Apps
 from django.db.models.base import Model
 from django.db.models.fields import Field
 from django.db.models.manager import Manager
+from django.utils.functional import cached_property
 
-class AppConfigStub(AppConfig): ...
+class AppConfigStub(AppConfig):
+    def __init__(self, label: str) -> None: ...
+    def import_models(self) -> None: ...
```
adamchainz: This method exists in the base class with the same signature?
```suggestion
```

GabDug: Good catch thanks :)

This kind of testing would actually be very nice, to identify useless overrides!

adamchainz (approved):
Overall LGTM

django-stubs/db/migrations/operations/base.pyi
GabDug: Yep, stubtest complains, seems like the typeshed for object.__new__ is `def __new__(cls) -> Self: ...`

sobolevn (approved):
Great work! Thanks so much!

adamchainz:
Awesome!

django-stubs/apps/config.pyi
```diff
@@ -17,7 +18,8 @@ class AppConfig:
     models_module: str | None
     models: dict[str, type[Model]]
     def __init__(self, app_name: str, app_module: types.ModuleType | None) -> None: ...
-    default_auto_field: str
+    @cached_property
+    def default_auto_field(self) -> str: ...
```
intgr, line 22: Heh this was just changed recently: https://github.com/typeddjango/django-stubs/pull/1541/files

intgr: This looks like the cause of issue #1593

GabDug: Should we ignore stubtest and set s `str`, or try to cover both with `default_auto_field: cached_property[str] | str `?

intgr: Just `str` seems fine, overengineering seems more likely to cause issues.

PR #1586, opened 2023-06-27
Add SQLCompiler._order_by_pairs()

adamchainz:
# I have made things!

I’ve used this private function in this glorious hack of a pytest fixture:

```python
@pytest.fixture(scope="session", autouse=True)
def make_unordered_queries_randomly_ordered():
    """Patch Django’s ORM to randomly order all queries without a specified sort order.

    This discovers problems where code or tests assume a sort order, but the
    database doesn’t guarantee any.
    """
    from django.db.models.sql.compiler import SQLCompiler

    patchy.patch(
        SQLCompiler._order_by_pairs,  # type: ignore[attr-defined]
        """\
        @@ -9,7 +9,7 @@
                 ordering = meta.ordering
                 self._meta_ordering = ordering
             else:
        -        ordering = []
        +        ordering = ["?"]
             if self.query.standard_ordering:
                 default_order, _ = ORDER_DIR["ASC"]
             else:
        """,
    )
```

I guess that means we can add it?

## Related issues

None


flaeppe (approved):
What a piece of beauty!

sobolevn (commented):
I don't oppose adding it, but this use-case seems like it is using very internal stuff :)

By the way, I would love to add this plugin to my projects. Is it ready?
PRs to https://github.com/wemake-services/wemake-django-template/ are welcome! 👍

adamchainz:
So far I've only tried it on one project. Was thinking of writing a blog post rather than making a package.

adamchainz:
Post: https://adamj.eu/tech/2023/07/04/django-test-random-order-querysets/

PR #1588, opened 2023-06-27
Update compatible-mypy to 1.4.x

intgr:
The lower bound of `compatible-mypy` was not bumped in #1572.

While at it, I've also updated CI to use mypy 1.4.1.

----

Honestly the `compatible-mypy` requirement has never made much sense to me. I don't use it in any of my projects. I update mypy as soon as a new version comes out and usually it works fine.

How can we determine which mypy versions are compatible? We currently run one mypy version in CI, and our test suite almost always needs changes for newer mypy versions, because it's sensitive to mypy's error messages that change frequently. If we wanted to test with multiple mypy versions in a test matrix, that would require rethinking our testing approach.

So my approach has been to pin `compatible-mypy` to the major.minor version we run in CI, at least we know for a fact that it works.

----

So @sobolevn @christianbundy or other maintainers, please chime in, what should `compatible-mypy` mean? Or should we just punt and carry on pinning to latest mypy minor like before?

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ We rely on different `django` and `mypy` versions:
 
 | django-stubs   | Mypy version | Django version | Django partial support | Python version |
 |----------------|--------------|----------------|------------------------|----------------|
+| (next release) | 1.4.x        | 4.2            | 4.1, 3.2               | 3.8 - 3.11     |
 | 4.2.1          | 1.3.x        | 4.2            | 4.1, 3.2               | 3.8 - 3.11     |
 | 4.2.0          | 1.2.x        | 4.2            | 4.1, 4.0, 3.2          | 3.7 - 3.11     |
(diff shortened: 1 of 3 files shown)


intgr:
> So my approach has been to pin compatible-mypy to the minor version we run in CI

Although this itself causes issues: I go through a fair bit of effort to keep `compatible-mypy` in sync between django-stubs and djangorestframework-stubs. It's a chore that I would prefer to not having to worry about.

sobolevn:
I also don't use this extra. I care about the mypy version I use up to patch difference (so, I want to be able to update it independently). But, I think that many people don't. They just want any mypy version that works with django-stubs.

So, we can use the same we run in CI.

sobolevn (approved):

intgr:
OK, I will document that with a code comment in `setup.py` for other developers.

PR #1592, opened 2023-06-27
Make error_messages and default_error_messages types stricter

intgr:
The values of these dicts/mappings should only be `str` or `_StrPromise` from lazy translation. Previously `Any` was accepted.

* Typed `error_messages=` parameters as `Mapping[str, _StrOrPromise] | None`
* Typed `error_messages` and `default_error_messages` class attributes as `dict[str, _StrOrPromise]`

## Related issues

Previously discussed in https://github.com/typeddjango/django-stubs/pull/1538#pullrequestreview-1472848331


flaeppe (approved):
Just ran across your mention for review. I think the changes here are fine.

PR #1596, opened 2023-06-28
Fix overriding AppConfig.default_auto_field with string value

### LINKED ISSUE #1593: `AppConfig.default_auto_field` -- Incompatible types in assignment
# Bug report

```python3
class FooConfig(AppConfig):
    default_auto_field = "django.db.models.BigAutoField"
    name = "foo"
```

## What's wrong

> Incompatible types in assignment (expression has type "str", base class "AppConfig" defined the type as "Callable[[AppConfig], str]")  [assignment]

```python3
default_auto_field = "django.db.models.BigAutoField"
```

## How is that should be

No error.

## System information

- OS: N/A
- `python` version: 3.11
- `django` version: 4.1
- `mypy` version: 1.4
- `django-stubs` version: 4.2.2

GabDug:
# I have made things!

Loosen the typing for `AppConfig.default_auto_field` to allow `str` or `@property`-like

## Related issues

Format is:
- Closes #1593


tests/typecheck/apps/test_config.yml
```diff
@@ -0,0 +1,24 @@
+-   case: test_appconfig_can_be_str_or_property
```
adamchainz: I'd prefer these to be three separate tests that can fail independently :)

GabDug: Done! 

Just checked, and the tests are wrong -- they don't fail with the old config.pyi, would you have any insights?

adamchainz: I've just tried locally and also cannot reproduce the issue. I wonder if there's some Mypy config?

adamchainz: Ah the problem was the tests used `django.app.config` instead of `django.apps.config`. Enabling strict mode temporarily in the tests uncovered this, because it showed:

```
E     main:4: error: Class cannot subclass "AppConfig" (has type "Any") (diff)
```

I've just pushed a commit fixing the import path, and the calls, but the tests now fail in other ways.

christianbundy (approved):
I don't have approve perms, but this LGTM. Thanks for the fix!

tests/typecheck/apps/test_config.yml
intgr: > I'd prefer these to be three separate tests

We have a technical limitation here that each individual test case adds significant time overhead. The test suite takes 11-15 minutes, which is already way too long.

But I think anything except the first test here is a vanishingly small use case and doesn't really need to be tested.

> that can fail independently

Not really a concern for stub tests IMO because one test case can report multiple mypy errors.

intgr:
We should make a release with this fix.  Unfortunately, I have issues with my computer and won't be able to release a fixed django-stubs until tomorrow evening. Anyone else up for it?

tests/typecheck/apps/test_config.yml
adamchainz: Okay totally fair, let’s combine them back.

GabDug: Merged the tests.

intgr: Whoa, the test suite slowness was already fixed and merged in #1603.

But still I think the test is fine as is right now. I'll merge and do a quick release to close the regression.

intgr (approved):
Thanks for contributing the fix so quickly!

I'll go ahead and do a quick release it to close the regression.

intgr:
This was fixed in django-stubs 4.2.3.

WhyNotHugo:
Thanks!

PR #1598, opened 2023-06-28
BigIntegerField has MAX_BIGINT

asottile:
https://github.com/django/django/blob/21e686471598d6c755d93a570de20dc00d8d27f6/django/db/models/fields/__init__.py#L2146


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -239,7 +239,10 @@ class PositiveIntegerRelDbTypeMixin:
     def rel_db_type(self, connection: BaseDatabaseWrapper) -> str: ...
 
 class SmallIntegerField(IntegerField[_ST, _GT]): ...
-class BigIntegerField(IntegerField[_ST, _GT]): ...
+
+class BigIntegerField(IntegerField[_ST, _GT]):
+    MAX_BIGINT: int
```
sobolevn: Since it used as a class prop (https://github.com/django/django/blob/21e686471598d6c755d93a570de20dc00d8d27f6/django/db/models/fields/__init__.py#L2154), we need to mark it as a `ClassVar`:

```suggestion
    MAX_BIGINT: ClassVar[int]
```

sobolevn (commented):
Thanks! -3 allowlist entries 👍

sobolevn (approved):

PR #1600, opened 2023-06-28
Allow `None` as redirect_field_name value for `user_passes_test` and `login_required`

JoeHitchen:
According to the [Django documentation](https://docs.djangoproject.com/en/4.2/topics/auth/default/#django.contrib.auth.decorators.user_passes_test), the `redirect_field_name` argument to `user_passes_test` accepts `None` as a value as well as a string, but Django Stubs currently only recognises strings as acceptable values. Specifically, passing a `None` indicates that the redirect field should be omitted.

The [documentation](https://docs.djangoproject.com/en/4.2/topics/auth/default/#the-login-required-decorator) does not explicitly state whether `redirect_field_name` of `login_required` allows or does not allow `None` values, but my own usage indicates that it does and it behaves in the same way.

I haven't attempted to make this change globally for all instances of `redirect_field_name` since I'm not familiar with all their usages, but note that some cases already allow `None`s while others that still won't.

- Precommit checks run successfully
- `pytest` runs with 31 failures (344 passes), however this is the same as I get on `master`
- I don't have a Linux system to run `stubtest.sh`

There is not an open issue related to this as far as I am aware - I noticed the problem and since it was a simple solution I went straight to fixing.


sobolevn (approved):
Yes, you are right.
1. Source: https://github.com/django/django/blob/21e686471598d6c755d93a570de20dc00d8d27f6/django/contrib/auth/views.py#L186
2. We already have `redirect_field_name` annotated as `Optional`: https://github.com/typeddjango/django-stubs/blob/8c08d3536f28d48805067cdddeef6b75308e50c7/django-stubs/contrib/auth/views.pyi#L32-L34

JoeHitchen:
Cheers!

PR #1601, opened 2023-06-29
Bump `pytest-mypy-plugins` and improve `jinja2` stubs

sobolevn:
Refs https://github.com/typeddjango/pytest-mypy-plugins/issues/118

### DIFF
--- a/django-stubs/template/backends/jinja2.pyi
+++ b/django-stubs/template/backends/jinja2.pyi
@@ -1,3 +1,4 @@
+from _typeshed import Incomplete
 from collections.abc import Callable
 from typing import Any
 
@@ -16,5 +17,12 @@ class Origin:
     name: str
     template_name: str | None
     def __init__(self, name: str, template_name: str | None) -> None: ...
+
+class Template:
+    template: Incomplete
+    backend: Jinja2
+    origin: Origin
+    def __init__(self, template: Incomplete, backend: Jinja2) -> None: ...
+    def render(self, context: Incomplete | None = ..., request: Incomplete | None = ...) -> Incomplete: ...
 
 def get_exception_info(exception: TemplateSyntaxError) -> dict[str, Any]: ...
--- a/django-stubs/template/backends/jinja2.pyi
+++ b/django-stubs/template/backends/jinja2.pyi
@@ -1,7 +1,7 @@
-from _typeshed import Incomplete
 from collections.abc import Callable
 from typing import Any
 
+from _typeshed import Incomplete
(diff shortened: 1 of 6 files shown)


mypy_django_plugin/django/context.py
```diff
@@ -60,12 +60,11 @@ def initialize_django(settings_module: str) -> Tuple["Apps", "LazySettings"]:
         from django.apps import apps
         from django.conf import settings
 
-        apps.get_models.cache_clear()  # type: ignore
         apps.get_swappable_settings_name.cache_clear()  # type: ignore
+        apps.clear_cache()
```
sobolevn, line 64: Btw, why `apps.get_swappable_settings_name.cache_clear()` is not a part of `app.clear_cache()`? Should I commit this to Django?

adamchainz: Make a ticket!

sobolevn: Done https://code.djangoproject.com/ticket/34687#ticket

django-stubs/template/backends/jinja2.pyi
```diff
@@ -17,4 +18,11 @@ class Origin:
     template_name: str | None
     def __init__(self, name: str, template_name: str | None) -> None: ...
 
+class Template:
+    template: Incomplete
+    backend: Jinja2
+    origin: Origin
+    def __init__(self, template: Incomplete, backend: Jinja2) -> None: ...
+    def render(self, context: Incomplete | None = ..., request: Incomplete | None = ...) -> Incomplete: ...
```
intgr, line 26: Is there a reason why these Jinja changes are in this PR? Seems like a separate topic

intgr (commented):
Both changes look good to me, thanks, but I'd prefer if they were separate PRs

django-stubs/template/backends/jinja2.pyi
sobolevn: Yes, new version of `pytest-mypy-plugins` requires `jinja2`. Now the environment has changed, new errors are detected by `stubtest`. See https://github.com/typeddjango/django-stubs/issues/1602

intgr: Oh I see. But we could merge splitted Jinja2 changes first and then this one separately.

intgr: Or if you don't think it's worth the effort, at least mention both changes in the PR title and commit message lead line

PR #1603, opened 2023-06-29
Run tests in multiple shards and all Python versions

sobolevn:
Now we can tests all python version (it won't be too long, but can catch a lot of potential bugs).

We will have additional 15 (16 - 1 existing) new CI jobs, but my experience in `typeshed` prooves that it is fine.

Or we can increase the number of shards to make this even faster.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.8', '3.9', '3.10', '3.11']
+      fail-fast: false
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
@@ -47,6 +48,7 @@ jobs:
       matrix:
         python-version: ['3.8', '3.9', '3.10', '3.11']
         shard: [0, 1, 2, 3]
+      fail-fast: false
     steps:
       - uses: actions/checkout@v3
       - name: Setup system dependencies
@@ -71,6 +73,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.11']
+      fail-fast: false
     steps:
       - uses: actions/checkout@v3
       - name: Setup system dependencies
@@ -92,6 +95,7 @@ jobs:
   matrix-test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: ['3.8', '3.9', '3.10']
         django-version: ['3.2', '4.2']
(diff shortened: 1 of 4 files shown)


sobolevn:
Current test execution (1 job) takes: 376 passed in 815.10s (0:13:35)
https://github.com/typeddjango/django-stubs/actions/runs/5409850829/jobs/9830509376?pr=1601

New test execution (each job) should ideally take: 14 / 4 = ~3.5 minutes

Real data (3.11 as well, so we can have a direct comparision):
- 0: 95 passed in 217.29s (0:03:37)
- 1: 104 passed in 184.53s (0:03:04)
- 2: 92 passed in 205.61s (0:03:25)
- 3: 85 passed in 136.82s (0:02:16)

All in all, we have the same amount of tests (376) and they are 4 times faster 🎉

.github/workflows/test.yml
```diff
@@ -45,7 +46,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
```
intgr, line 49: Just a thought, I don't have strong feelings about this, but maybe testing the lowest and highest Python version would be enough?

We're not paying for compute, but it might be a waste of CO2 😄

Would apply to other tests too.

intgr (approved):
Awesome, I had no idea this was so easy to do

.github/workflows/test.yml
sobolevn: I think that it still makes sense, because there are lots of typing changes between versions.

intgr:
I tweaked the PR title to mention testing multiple Python versions. Feel free to change if you're not happy with it

sobolevn:
Now to `stubtest`, I also want to run it on all versions, because I am pretty sure there will be noticable differences :)

sobolevn:
Thanks for the review!

intgr:
Why would stubtests give different results with different Python versions? I think Django's classes/functions/methods/parameters don't differ depending on Python.

I don't think typeshed differences should affect these either, unless I'm missing something about how stubtest works. To be fair I haven't had the time to play with it yet

intgr:
It seems GitHub is throttling the number of concurrent running jobs.

I just merged a PR to master, the CI for master branch is still running, and the CI of another PR branch has most of its jobs in "queued" status.

> image

But anyway it's still better than before.

UnknownPlatypus:
There is also [pytest-split](https://github.com/jerry-git/pytest-split) that does something similar but based on test duration. It should distribute more evenly the test runtime so we could get to something around 3:05min per shard. I can send a PR for that if you think it's worth trying

sobolevn:
No, thanks. The quantity is pretty simple and good enough.

adamchainz:
Nice one.

PR #1604, opened 2023-06-29
Version 4.2.3 release (django-stubs only)

intgr:
This release has multiple bugfixes, fixing a significant regression from version 4.2.2.

I'll just release `django-stubs` this time and skip `django-stubs-ext` since there are no changes there (I explained this approach in https://github.com/typeddjango/django-stubs/issues/1450#issuecomment-1525851358)


intgr:
Closes https://github.com/typeddjango/django-stubs/issues/1605

PR #1607, opened 2023-06-30
Add _body attribute to HttpRequest

asottile:
this is created on access to self.body -- or usable as a shortcut to create a request body in tests

### DIFF
--- a/django-stubs/http/request.pyi
+++ b/django-stubs/http/request.pyi
@@ -39,6 +39,7 @@ class HttpRequest(BytesIO):
     resolver_match: ResolverMatch | None
     content_type: str | None
     content_params: dict[str, str] | None
+    _body: bytes
     _stream: BinaryIO
     # Attributes added by optional parts of Django
     # django.contrib.admin views:


adamchainz:
Can you share an example from the tests? Nomrally tests use a `RequestFactory` to create `HttpRequest` objects, so the WSGI environment is correctly filled in for the body contents: https://github.com/django/django/blob/047844270bd5cfc4a2716146616970ac1407d2a2/django/test/client.py#L593-L600 .

I guess we could add `_body` anyway, but just wanna check it's not a completely off-piste usage...

asottile:
yeah it's a little tricky to find it all but it's basically these:

- https://github.com/getsentry/sentry/blob/0612318b218b052baa80bbc6cf253d10bf70b2c3/src/sentry/testutils/cases.py#L241
- https://github.com/search?q=repo%3Agetsentry%2Fsentry%20._body&type=code

unless I am reading it wrong I can't really use the test request factory because it's not actually making the requests -- just building those and then testing functions taking a `HttpRequest`

adamchainz:
[`RequestFactory`](https://docs.djangoproject.com/en/4.2/topics/testing/advanced/#django.test.RequestFactory) builds request objects only, it doesn't process them like `Client`. I think the linked functions would all be improved by using `RequestFactory` to build their requests, rather than instantiating `HttpRequest` directly. Doing so would be shorter and more accurate.

That said, I think we can merge this PR. I found a bunch more usage on GitHub search: https://github.com/search?q=django+%22request._body%22+path%3A*.py&type=code

PR #1608, opened 2023-07-03
Allow translated URL patterns in path(), re_path()

imankulov:
URL patterns can be marked translatable using the gettext_lazy() function.

Ref: https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#translating-url-patterns


imankulov:
👋 I would appreciate it if someone could review this PR. My current workaround is using `cast(str, _("admin/"))`, which is acceptable but unnecessarily verbose. Thanks!

tests/typecheck/urls/test_conf.yml
```diff
@@ -17,6 +17,17 @@
...
 
+-   case: test_path_accepts_translated_patterns
+    main: |
+      from typing import List
+      from django.urls import path, _AnyURL
+      from django.utils.translation import gettext_lazy as _
+      from django.contrib import admin
+
+      urlpatterns: List[_AnyURL] = [
+          path(_("admin/"), admin.site.urls, name="admin"),
```
intgr, line 30: If you already added a test, let's add a test case for `re_path()` too.

intgr: Just add another line to this test case.

intgr (approved):
Lookg good to me, small tweak to tests and then this is ready.

tests/typecheck/urls/test_conf.yml
imankulov: Thanks for the review, @intgr! Addressed in 0546071824520a0af35d869b2644fc01c8ce09a0

intgr (approved):

PR #1616, opened 2023-07-08
Reveal missing stubtest errors

UnknownPlatypus:
# I have made things!

Hi, I was checking the stubtest output and noticed a lot of `RuntimeError` reported such as this one:

```
error: django.contrib.sites.models failed to import, RuntimeError: Model class django.contrib.sites.models.Site doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
Stub: in file /home/thibaut/workspace/django-stubs/django-stubs/contrib/sites/models.pyi:1
MypyFile:1(
  /home/thibaut/workspace/django-stubs/django-stubs/contrib/sites/models.pyi)
Runtime:
MISSING
```

Turns out because we don't have any installed apps in `django_tests_settings.py`, many stubtest checks were failing because of import runtime errors. This MR adds the missing django apps to the `INSTALLED_APPS` list to fix this.

This solved ~25 allowlist_todo issues but unfortunately revealed ~200 other ones.


sobolevn (approved):
Good catch!

sobolevn:
Thanks!

PR #1618, opened 2023-07-10
Declare View.http_method_names as Sequence

### LINKED ISSUE #1615: View.http_method_names should be Iterable
# Bug report

## What's wrong
This is valid code:
```python
from django.views.generic.base import View
class MyView(View):
    http_method_names = ('get',)
```

## How is that should be
`http_method_names` should be `Iterable[str]` instead of `list[str]`.

## System information

- OS: Linux
- `python` version: 3.11.4
- `django` version: 4.2.3
- `mypy` version: 1.4.1

93578237:
# I have made things!

## Related issues
Closes https://github.com/typeddjango/django-stubs/issues/1615

### DIFF
--- a/django-stubs/views/generic/base.pyi
+++ b/django-stubs/views/generic/base.pyi
@@ -1,4 +1,4 @@
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from typing import Any
 
 from django.http.request import HttpRequest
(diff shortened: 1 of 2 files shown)


adamchainz (approved):

sobolevn (approved):
Thanks :)

PR #1622, opened 2023-07-11
Fix input type of conditional_escape

ljodal:
This should accept plain strings. It sends the argument on to `escape`, which forces the argument into a string, so sending in a string to begin with is perfectly allowed.


sobolevn (approved):
Good catch! Thanks

PR #1625, opened 2023-07-12
Forbid instantiating abstract models

### LINKED ISSUE #1624: Instantiating an abstract model yields no error
# Bug report

Simple example where Django raises, while `django-stubs` doesn't

```python
from django.contrib.auth.models import AbstractUser
AbstractUser()
# Traceback (most recent call last):
#  File "/usr/local/lib/python3.10/code.py", line 90, in runcode
#    exec(code, self.locals)
#  File "<console>", line 1, in <module>
#  File "/app/venv/lib/python3.10/site-packages/django/db/models/base.py", line 467, in __init__
#    raise TypeError("Abstract models cannot be instantiated.")
# TypeError: Abstract models cannot be instantiated.
AbstractUser.objects.create()
# Traceback (most recent call last):
#  File "/usr/local/lib/python3.10/code.py", line 90, in runcode
#    exec(code, self.locals)
#  File "<console>", line 1, in <module>
#  File "/app/venv/lib/python3.10/site-packages/django/db/models/manager.py", line 191, in __get__
#    raise AttributeError(
# AttributeError: Manager isn't available; AbstractUser is abstract
```

This isn't really any serious issue though, since one will get smacked with a runtime error quite quickly. Just thought it was worth reporting.

If anything, this could clean up our test suite in regards to abstract models..

Ref: #1393 

## What's wrong

There's no complaints from `django-stubs`

## How is that should be

There should be errors on both initialisation and usage of manager of an abstract model.

## System information

- OS:
- `python` version: 3.10
- `django` version: 4.2.2
- `mypy` version: 1.4.1
- `django-stubs` version: 4.2.2

sobolevn:
I went ahead with another approach, because setting `TypeInfo.is_abstract` can have different side-effects, because it is widely used in mypy itself.

I think that checking `.create` and `__init__` is good enough for our use-case.

Closes https://github.com/typeddjango/django-stubs/issues/1624

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -231,9 +231,11 @@ class NewSemanalDjangoPlugin(Plugin):
             if info and info.has_base(fullnames.OPTIONS_CLASS_FULLNAME):
                 return partial(meta.return_proper_field_type_from_get_field, django_context=self.django_context)
 
-        elif class_fullname in manager_classes and method_name == "create":
-            return partial(init_create.redefine_and_typecheck_model_create, django_context=self.django_context)
-        elif class_fullname in manager_classes and method_name in {"filter", "get", "exclude"}:
+        elif method_name == "create":
+            # We need `BASE_MANAGER_CLASS_FULLNAME` to check abstract models.
+            if class_fullname in manager_classes or class_fullname == fullnames.BASE_MANAGER_CLASS_FULLNAME:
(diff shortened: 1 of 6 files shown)


flaeppe (approved):
I think Django raises on access of a manager attribute(?) but I don't think mimicking that is trivial, as the attribute name is varying.

Let's start out with initialization and create

PR #1631, opened 2023-07-24
get_group_by_cols() -> list[BaseExpression]

### LINKED ISSUE #1623: get_group_by_cols does not seem to have the correct type
# Bug report

it seems to always have had `-> list[Self]` (but in various forms through the history)

## What's wrong

however looking at the implementation it is finding sub-expressions and attempting to extract columns from it? 

https://github.com/django/django/blob/7f2bc365b3652eaa8a8c19e58ddd35f062c68dbc/django/db/models/expressions.py#L434-L435

## How is that should be

I suspect it should be `-> list[BaseExpression]` or really `list[Any]` since the return is a union of everything with `contains_aggregate` being `False`

## System information

- OS:
- `python` version: irrelevant
- `django` version: HEAD
- `mypy` version: irrelevant
- `django-stubs` version: HEAD
- `django-stubs-ext` version: HEAD

asottile:
resolves #1623


asottile:
CI appears stuck, will need to rerun that one

sobolevn (approved):
LGTM! I've created https://github.com/typeddjango/django-stubs/pull/1632 to add timeouts to CI. Right now I've stoped it after 2+ hours of waiting.

adamchainz (approved):

adamchainz:
Cheers m’dears

PR #1635, opened 2023-07-31
fix(Options): pk cannot be None

gg-mmill:
Fix the `Options.pk` typing to always be a `Field`

# I have made things!

Hello, I am no Django expert so I may be wrong, but it seems to me a Model always define a primary key (either explicitly or implicitly), which means `Model._meta.pk` always references a `Field`
 

## Related issues

No issue, I can create one if needs be .

### DIFF
--- a/django-stubs/db/models/options.pyi
+++ b/django-stubs/db/models/options.pyi
@@ -73,7 +73,7 @@ class Options(Generic[_M]):
     required_db_features: list[str]
     required_db_vendor: Literal["sqlite", "postgresql", "mysql", "oracle"] | None
     meta: type | None
-    pk: Field | None
+    pk: Field
     auto_field: AutoField | None
     abstract: bool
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
It is set to `None` by default: https://github.com/django/django/blob/22b0b73c7732ba67db4e69fd9fa75aad84c8e5c4/django/db/models/options.py#L133

So, `Options(meta).pk` is `None`, it is only set to not-`None` after `setup_pk`: https://github.com/django/django/blob/22b0b73c7732ba67db4e69fd9fa75aad84c8e5c4/django/db/models/options.py#L370

sobolevn:
I think that I will close this PR, I hope that my explanation is good enough :)

Anyways, thanks a lot for taking your time to contribute! 👍 
Please, feel free to send more PRs :)

PR #1639, opened 2023-08-02
Make WSGIHandler compatible with stdlib wsgi definition

### LINKED ISSUE #1053: Incompatibility between WSGIHandler and stdlib types in wsgiref
# Bug report

## What's wrong

Changes to typing of `core.handlers.WSGIHandler` introduced in #909 are incompatible with the stdlib annotations under `wsgiref`, raising a error when checked with mypy:

```python
from wsgiref.simple_server import make_server

from django.core.handlers.wsgi import WSGIHandler

make_server(
    "0.0.0.0",
    8080,
    WSGIHandler(),
)
```
```
example.py:8: error: Argument 3 to "make_server" has incompatible type "WSGIHandler"; expected "Callable[[Dict[str, Any], StartResponse], Iterable[bytes]]"  [arg-type]
scce.py:8: note: "WSGIHandler.__call__" has type "Callable[[Arg(Dict[str, Any], 'environ'), Arg(Callable[[str, Sequence[Tuple[str, str]], Optional[Any]], None], 'start_response')], HttpResponseBase]"
```

`wsgiref` types are in typeshed Python <= 3.10, and in `wsgiref.types` [as of Python 3.11](https://github.com/python/cpython/commit/0ddc63b240340a952692b11dfe0810973393ed11#diff-ba6db771673b0d7c4e29c4bf9634494e4d756da6bc51576e29fb7744e253f0ba).

## How is that should be

Types should be compatible with stdlib as per PEP 3333.

## System information

- OS:
- `python` version: 3.10
- `django` version: 4.0.5
- `mypy` version: 0.961

asottile:
# I have made things!

## Related issues

Resolves #1053


intgr (approved):
Looks good to me, thanks

sobolevn:
Thanks!

PR #1641, opened 2023-08-03
Rename SuccessURLAllowedHostsMixin to RedirectURLMixin and hint missing attributes

si14:
Just a simple stub for `contrib.auth.views.RedirectURLMixin`.

I noticed that it was in the todolist and I needed it, so here it is.


django-stubs/contrib/auth/views.pyi
```diff
@@ -9,6 +9,15 @@ from django.views.generic.edit import FormView
...
+class RedirectURLMixin:
+    next_page: str | None
+    redirect_field_name: str
+    success_url_allowed_hosts: set[str]
+    def get_success_url(self) -> str: ...
+    def get_redirect_url(self) -> str: ...
+    def get_success_url_allowed_hosts(self) -> set[str]: ...
+    def get_default_redirect_url(self) -> str: ...
+
 class SuccessURLAllowedHostsMixin:
```
intgr: Some more changes look warranted...

`SuccessURLAllowedHostsMixin` no longer exists in Django 4.2, `LoginView` and `LogoutView` should inherit from `RedirectURLMixin` instead.

intgr (changes requested):
Please remove `SuccessURLAllowedHostsMixin` and replace it with the new `RedirectURLMixin`.

si14:
@intgr fixed, all tests are green

intgr (approved):
Thank you!

PR #1649, opened 2023-08-10
Move ModelBase.objects declaration to Model.objects, for mypy 1.5.0

### LINKED ISSUE #1648: Custom managers broken with mypy 1.5.0
# Bug report

## What's wrong

Mypy 1.5.0 was just released and breaks lots of django-stubs tests. Obviously that’s semi-expected since 1.5.0 is outside the `django-stubs[compatible-mypy]` bound of `mypy==1.4.*`. But upon investigating anyway, I think mypy is correct that there’s a django-stubs problem here. The failures began with

- python/mypy#14988

which fixes mypy to understand that metaclass attributes take precedence over attributes defined in regular classes, matching Python’s runtime behavior. Unfortunately, django-stubs was relying on the previous misunderstanding:

https://github.com/typeddjango/django-stubs/blob/325006ccd72a9b838cf841b1c61ba4092b084f99/django-stubs/db/models/base.pyi#L21-L29

https://github.com/typeddjango/django-stubs/blob/325006ccd72a9b838cf841b1c61ba4092b084f99/django-stubs/contrib/contenttypes/models.pyi#L14-L18

mypy 1.4.1 thinks `ContentType.objects` has type `ContentTypeManager`, but mypy 1.5.0 thinks `ContentType.objects` has type `BaseManager[ContentType]` from the metaclass.

This affects real projects too, such as Zulip (after bumping its pinned version of mypy):

```
corporate/models.py:80: error: "BaseManager[ContentType]" has no attribute "get_for_model"  [attr-defined]
zerver/actions/invites.py:458: error: "BaseManager[ContentType]" has no attribute "get_for_model"  [attr-defined]
zerver/actions/invites.py:469: error: "BaseManager[ContentType]" has no attribute "get_for_model"  [attr-defined]
zerver/tests/test_users.py:769: error: "BaseManager[ContentType]" has no attribute "clear_cache"  [attr-defined]
zerver/tests/test_signup.py:929: error: "BaseManager[ContentType]" has no attribute "clear_cache"  [attr-defined]
corporate/views/webhook.py:70: error: "BaseManager[ContentType]" has no attribute "get_for_model"  [attr-defined]
corporate/views/webhook.py:85: error: "BaseManager[ContentType]" has no attribute "get_for_model"  [attr-defined]
corporate/views/webhook.py:100: error: "BaseManager[ContentType]" has no attribute "get_for_model"  [attr-defined]
```

## How is that should be

We need to somehow convince mypy 1.5.0 that `ContentType.objects` has type `ContentTypeManager`.

## System information

- OS: Linux (NixOS 23.11 x86-64)
- `python` version: 3.10.2
- `django` version: 4.2.4
- `mypy` version: 1.5.0

andersk:
# I have made things!

mypy 1.5.0 was fixed to understand that metaclass attributes take precedence over attributes in the regular class.  So we need to declare `objects` in the regular class to allow it to be overridden in subclasses.

## Related issues

- Fixes #1648.

### DIFF
--- a/django-stubs/db/models/base.pyi
+++ b/django-stubs/db/models/base.pyi
@@ -20,13 +20,13 @@ class ModelState:
 
 class ModelBase(type):
     @property
-    def objects(cls: type[_Self]) -> BaseManager[_Self]: ...  # type: ignore[misc]
-    @property
     def _default_manager(cls: type[_Self]) -> BaseManager[_Self]: ...  # type: ignore[misc]
     @property
     def _base_manager(cls: type[_Self]) -> BaseManager[_Self]: ...  # type: ignore[misc]
 
 class Model(metaclass=ModelBase):
+    objects: BaseManager[Self]
+
     # Note: these two metaclass generated attributes don't really exist on the 'Model'
     # class, runtime they are only added on concrete subclasses of 'Model'. The
     # metaclass also sets up correct inheritance from concrete parent models exceptions.
--- a/tests/typecheck/models/test_abstract.yml
+++ b/tests/typecheck/models/test_abstract.yml
@@ -51,6 +51,7 @@
     Recursive(parent=Recursive(parent=None))
     Concrete(parent=Concrete(parent=None))
   out: |
+    main:4: error: Access to generic instance variables via class is ambiguous
     main:4: error: Unexpected attribute "parent" for model "Recursive"
     main:5: error: Cannot instantiate abstract class "Recursive" with abstract attributes "DoesNotExist" and "MultipleObjectsReturned"
     main:5: error: Unexpected attribute "parent" for model "Recursive"
@@ -209,4 +210,4 @@
 
             def create_animal_generic(klass: Type[T], name: str) -> T:
                 reveal_type(klass)  # N: Revealed type is "Type[T`-1]"
-                return klass.objects.create(name=name)
+                return klass.objects.create(name=name)  # E: Incompatible return value type (got "Animal", expected "T")
--- a/scripts/stubtest/allowlist_todo.txt
+++ b/scripts/stubtest/allowlist_todo.txt
@@ -315,17 +315,13 @@ django.contrib.gis.db.backends.oracle.features.DatabaseFeatures.supports_toleran
 django.contrib.gis.db.backends.oracle.features.DatabaseFeatures.unsupported_geojson_options
 django.contrib.gis.db.backends.oracle.introspection
 django.contrib.gis.db.backends.oracle.models.OracleGeometryColumns.Meta
-django.contrib.gis.db.backends.oracle.models.OracleGeometryColumns.objects
 django.contrib.gis.db.backends.oracle.models.OracleSpatialRefSys.Meta
-django.contrib.gis.db.backends.oracle.models.OracleSpatialRefSys.objects
 django.contrib.gis.db.backends.oracle.operations
 django.contrib.gis.db.backends.postgis.adapter.PostGISAdapter.prepare
 django.contrib.gis.db.backends.postgis.features.DatabaseFeatures.empty_intersection_returns_none
 django.contrib.gis.db.backends.postgis.features.DatabaseFeatures.supports_geography
 django.contrib.gis.db.backends.postgis.models.PostGISGeometryColumns.Meta
-django.contrib.gis.db.backends.postgis.models.PostGISGeometryColumns.objects
 django.contrib.gis.db.backends.postgis.models.PostGISSpatialRefSys.Meta
-django.contrib.gis.db.backends.postgis.models.PostGISSpatialRefSys.objects
 django.contrib.gis.db.backends.postgis.operations.PostGISOperations.convert_extent
 django.contrib.gis.db.backends.postgis.operations.PostGISOperations.convert_extent3d
 django.contrib.gis.db.backends.postgis.operations.PostGISOperator.check_geography
@@ -335,9 +331,7 @@ django.contrib.gis.db.backends.spatialite.features.DatabaseFeatures.can_alter_ge
 django.contrib.gis.db.backends.spatialite.features.DatabaseFeatures.django_test_skips
 django.contrib.gis.db.backends.spatialite.features.DatabaseFeatures.supports_area_geodetic
 django.contrib.gis.db.backends.spatialite.models.SpatialiteGeometryColumns.Meta
-django.contrib.gis.db.backends.spatialite.models.SpatialiteGeometryColumns.objects
 django.contrib.gis.db.backends.spatialite.models.SpatialiteSpatialRefSys.Meta
-django.contrib.gis.db.backends.spatialite.models.SpatialiteSpatialRefSys.objects
 django.contrib.gis.db.backends.spatialite.operations.SpatiaLiteOperations.convert_extent
 django.contrib.gis.db.backends.spatialite.operations.SpatiaLiteOperations.from_text
 django.contrib.gis.db.backends.spatialite.operations.SpatiaLiteOperations.geom_lib_version
@@ -1276,7 +1270,6 @@ django.db.migrations.questioner.NonInteractiveMigrationQuestioner.log_lack_of_mi
 django.db.migrations.recorder.MigrationRecorder.Migration.get_next_by_applied
 django.db.migrations.recorder.MigrationRecorder.Migration.get_previous_by_applied
 django.db.migrations.recorder.MigrationRecorder.Migration.id
-django.db.migrations.recorder.MigrationRecorder.Migration.objects
 django.db.migrations.serializer.ChoicesSerializer
 django.db.migrations.utils.FieldReference
 django.db.migrations.utils.field_is_referenced
@@ -1534,7 +1527,6 @@ django.db.models.base.Model.Meta
 django.db.models.base.Model.add_to_class
 django.db.models.base.ModelBase.__new__
 django.db.models.base.ModelBase.add_to_class
-django.db.models.base.ModelBase.objects
 django.db.models.base.ModelStateFieldsCacheDescriptor.__get__
 django.db.models.base.make_foreign_order_accessors
 django.db.models.base.method_get_order
--- a/tests/typecheck/models/test_meta_options.yml
+++ b/tests/typecheck/models/test_meta_options.yml
@@ -80,7 +80,7 @@
 
         # Errors:
         AbstractModel()  # E: Cannot instantiate abstract class "AbstractModel" with abstract attributes "DoesNotExist" and "MultipleObjectsReturned"
-        AbstractModel.objects.create()
+        AbstractModel.objects.create()  # E: Access to generic instance variables via class is ambiguous
     installed_apps:
         - myapp
     files:


django-stubs/db/models/base.pyi
```diff
@@ -19,14 +19,14 @@ class ModelState:
...
 class ModelBase(type):
-    @property
-    def objects(cls: type[_Self]) -> BaseManager[_Self]: ...  # type: ignore[misc]
     @property
     def _default_manager(cls: type[_Self]) -> BaseManager[_Self]: ...  # type: ignore[misc]
     @property
     def _base_manager(cls: type[_Self]) -> BaseManager[_Self]: ...  # type: ignore[misc]
 
 class Model(metaclass=ModelBase):
+    objects: BaseManager[Self] = ...
```
sobolevn:
```suggestion
    objects: BaseManager[Self]
```

sobolevn (commented):
Looks like `ModelBase` only has `_default_manager` and `_base_manager`: https://github.com/django/django/blob/9946f0b0d9356b55e819f861b31615fa5b548f99/django/db/models/base.py#L430-L436

And `objects` is really not found 🤔

intgr:
> Looks like ModelBase only has _default_manager and _base_manager

The `objects` attribute is dynamically created in `ModelBase._prepare` method:

https://github.com/django/django/blob/9946f0b0d9356b55e819f861b31615fa5b548f99/django/db/models/base.py#L417-L419

But the change seems good to me. :+1: Thanks!

django-stubs/db/models/base.pyi
andersk: Adjusted.

tests/typecheck/models/test_abstract.yml
```diff
@@ -51,6 +51,7 @@
     Recursive(parent=Recursive(parent=None))
     Concrete(parent=Concrete(parent=None))
   out: |
+    main:4: error: Access to generic instance variables via class is ambiguous
```
intgr, line 54: Do we know why this new error appears in abstract models?

Are there any use cases for accessing `objects` of an abstract model? I can't think of any.

andersk: I think the mypy plugin is why we only get this for abstract models, but I haven’t looked at the details.

An abstract model does not have an `objects` attribute at all at runtime, so extra errors for that are fine.

flaeppe: > Are there any use cases for accessing `objects` of an abstract model? I can't think of any.

I added this test in #1393. Just want to point out that accessing `objects` on an abstract model wasn't part of any use case, but rather an attempt for completeness of coverage in the plugin code

intgr: I tried this out: actually one of my projects relies on this in several places, and new `django-stubs` now causes errors. Let me open an issue describing the use case and we can discuss it there.

intgr: Please have a look at issue https://github.com/typeddjango/django-stubs/issues/1653

django-stubs/db/models/base.pyi
```diff
@@ -19,14 +19,14 @@ class ModelState:
...
 class ModelBase(type):
-    @property
-    def objects(cls: type[_Self]) -> BaseManager[_Self]: ...  # type: ignore[misc]
     @property
     def _default_manager(cls: type[_Self]) -> BaseManager[_Self]: ...  # type: ignore[misc]
     @property
     def _base_manager(cls: type[_Self]) -> BaseManager[_Self]: ...  # type: ignore[misc]
 
 class Model(metaclass=ModelBase):
+    objects: BaseManager[Self]
```
intgr, line 28: Changing this to `@classproperty` seems to fix the `create_animal_generic` part of issue #1653:

```suggestion
    @classproperty
    def objects(cls: type[_Self]) -> BaseManager[_Self]: ...
```

With import
```python
from django.utils.functional import classproperty
```

intgr: DOH! That causes new errors to appear...

> django-stubs/django-stubs/contrib/contenttypes/models.pyi:18: error: Incompatible types in assignment (expression has type "ContentTypeManager", base class "Model" defined the type as "Callable[[Type[_Self]], BaseManager[_Self]]")

andersk: `classproperty` can’t help us with typing this: its constructor annotates the `cls` parameter as `Any`, so there’s nothing to connect `_Self` to the actual model type.

https://github.com/typeddjango/django-stubs/blob/325006ccd72a9b838cf841b1c61ba4092b084f99/django-stubs/utils/functional.pyi#L99-L101

intgr: It does help. The `Any` doesn't matter here.

Before `classproperty` even comes into play, mypy has already resolved the relationship between the parameter and return value. The method signature `def objects(cls: type[_Self]) -> BaseManager[_Self]: ...` is evaluated first, before decorators.

But as I point out above, this breaks overriding `objects` attribute in a custom model.

intgr:
@andersk This PR solves some of the issues with mypy 1.5, but there are other test failures that remain involving the `objects` manager.

Did you look at the remaining errors? https://github.com/typeddjango/django-stubs/actions/runs/5888209826

andersk:
Yeah I looked for a while but didn’t make further progress.

I did note that some of the tests seem clearly bogus—for example, in the `managers_that_defined_on_other_models_do_not_influence` case, `myapp.models.PublishedBookManager[myapp.models.Book]` cannot possibly be a valid type to reveal because the `PublishedBookManager` class has no generic parameters. But for some reason that test isn’t failing, so I didn’t investigate further.

https://github.com/typeddjango/django-stubs/blob/3c77add89ebc861fbafae9ea1a4e268122d81dbe/tests/typecheck/managers/test_managers.yml#L172

My proposal for now would be to release this without bumping `compatible-mypy`, so that at least some users are able to upgrade their projects to mypy 1.5 and give us more data about what real problems may remain.

django-stubs/db/models/base.pyi
andersk: Ignoring the decorator is clearly a mypy bug. And it still doesn’t help us, because ignoring the decorator makes mypy incorrectly assume the first parameter is an instance of `Model`, not any kind of `type`.

```python
from django.utils.functional import classproperty

class Model:
    @classproperty
    def foo(cls: "Model") -> "Model":  # incorrectly allowed
        return cls

    @classproperty
    def bar(cls: "type[Model]") -> "type[Model]":  # incorrect error: The erased type of self "Type[__main__.Model]" is not a supertype of its class "__main__.Model"
        return cls

print(Model.foo is Model)  # True at runtime
reveal_type(Model)  # Revealed type is "def () -> __main__.Model"
reveal_type(Model.foo)  # Revealed type is "__main__.Model"
```

See also

- python/mypy#2563

tests/typecheck/models/test_abstract.yml
flaeppe: Hm, the error also appears for e.g. `mypy==1.4.1`. Error is emitted by mypy [here](https://github.com/python/mypy/blob/3534bacc4c0d3c4b1983a533e2a36cce43f2ec9d/mypy/checkmember.py#L994) and they have [an example](https://github.com/python/mypy/blob/3534bacc4c0d3c4b1983a533e2a36cce43f2ec9d/mypy/checkmember.py#L980-L985) in a comment close by looking like this:

```python
# Check if original variable type has type variables. For example:
class C(Generic[T]):
    x: T
C.x  # Error, ambiguous access
C[int].x  # Also an error, since C[int] is same as C at runtime
```

I think `C.x` should be the case that we run in to here(?)

The plugin is populating managers based on the runtime, but since abstract models don't have any, it stays as `objects: BaseManager[Self]`. Otherwise `objects` is (normally) overwritten by the plugin, that's why we're only seeing it for abstract models.

https://github.com/typeddjango/django-stubs/blob/85e1f3f0e14cf64bde9c00da5de6cef87cb01b38/mypy_django_plugin/transformers/models.py#L316-L339

So, I would suspect that if you try to annotate any manager to a model, this error would appear. e.g.

```python
from django.db import models

class MyModel(models.Model):
    other_manager: models.Manager[Self]

reveal_type(MyModel.other_manager)
```

Is there something up with `typing_extensions.Self` here, or do we need to change something?

tylerlaprade (approved):

tests/typecheck/models/test_abstract.yml
andersk: You can make mypy accept your example using [`ClassVar`](https://docs.python.org/3/library/typing.html#typing.ClassVar) ([PEP 526](https://peps.python.org/pep-0526/)).

```python
class MyModel(models.Model):
    other_manager: ClassVar[models.Manager[Self]]
```

Another way that more accurately reflects what’s happening at runtime might be

```python
ModelT = TypeVar("ModelT", bound=models.Model)

class ManagerDescriptor:
    def __get__(self, instance: None, owner: type[ModelT]) -> models.Manager[ModelT]: ...

class MyModel(models.Model):
    other_manager: ManagerDescriptor = models.Manager()
```

This PR has no effect on your `other_manager`, of course. I didn’t annotate `django.db.models.Model.objects` itself as a `ClassVar` or descriptor because that would seem to break more things than it fixes: every subclass of `Model` that overrides `objects` would then be required to annotate it as a `ClassVar` or descriptor too.

(Maybe it would be better to delete the `objects` annotation from the stub and rely entirely on the plugin for it?)

flaeppe: > Another way that more accurately reflects what’s happening at runtime might be
> 
> ```python
> ModelT = TypeVar("ModelT", bound=models.Model)
> 
> class ManagerDescriptor:
>     def __get__(self, instance: None, owner: type[ModelT]) -> models.Manager[ModelT]: ...
> 
> class MyModel(models.Model):
>     other_manager: ManagerDescriptor = models.Manager()
> ```

~I don't think this is correctly reflected? The runtime type of `objects` is actually `models.Manager`, so it's rather that `BaseManager.__get__` that should be implemented? You'd never get hold of the descriptor class itself. Anyways, adding tyst method doesn't change anything either..~ Edit: never mind, it's the descriptor doing the work

This is a side note but I think that should be an `@overload` method with anything other than class access having `-> NoReturn:`

> (Maybe it would be better to delete the `objects` annotation from the stub and rely entirely on the plugin for it?)

Yes, I've been thinking about this too and I think it's something we _should_ do. I did it just recently with `DoesNotExist` and `MultipleObjectsReturned` in #1663. Though we're leaving other type checkers behind by removing `objects: ` from the `.pyi`. Would've been nice if that attribute would stay, even though it's a bit incorrect, mainly for their quality of life situation.

flaeppe: Although, perhaps it's fair to say that it should be declared as a `ClassVar`? The plugin can auto generate properly etc. But doing it explicitly on a subclass one would have to declare it a `ClassVar`?

flaeppe: I took the liberty to extend the work here and try out changing managers to `ClassVar`s and adjust `objects` to one too. While the plugin will only add `objects` to models where it exists runtime.

It can be found in #1672

flaeppe:
> mypy 1.5.0 was fixed to understand that metaclass attributes take precedence over attributes in the regular class. So we need to declare `objects` in the regular class to allow it to be overridden in subclasses.

I think I hunted down the PR in mypy that is affecting us, just wanted to paste it here for traceability: https://github.com/python/mypy/pull/14988.

andersk:
Yeah that’s the one I linked from the issue #1648.

flaeppe:
> Yeah that’s the one I linked from the issue #1648.

Oh, I completely missed that.

flaeppe:
This was merged via #1672

PR #1659, opened 2023-08-22
Fix `F.resolve_expression()` return type

schinckel:
# I have made things!

## Related issues

No issues raised.

When attempting to write a non-trivial django Field subclass, I needed to override the resolve_expression method, but discovered at that time that the stubs for that method are incorrect.

This method calls `query.resolve_ref`, which in turn returns either a Ref or something else, however I cannot see that it can ever return an F object itself (as that would defeat the purpose of the method, which is to resolve the references, and F is by definition an unresolved reference).


django-stubs/db/models/expressions.pyi
```diff
@@ -133,7 +133,7 @@ class F(Combinable):
         reuse: set[str] | None = ...,
         summarize: bool = ...,
         for_save: bool = ...,
-    ) -> F: ...
+    ) -> Ref: ...
```
intgr: Are you 100% sure it always returns `Ref` and not other `Expression` subclasses?

`Query.resolve_ref()` is hinted to return `Expression`, so seems safer to go with that here. But I'm not at all familiar with this code.

intgr (commented):
Sorry that this has been sitting without review.

django-stubs/db/models/expressions.pyi
intgr: Since I'm planning to do a release soon, I'll go ahead and merge this with `-> Expression` for now, which is already an improvement.

If you believe this should be constrained further to `Ref`, please do open another pull request.

```suggestion
    ) -> Expression: ...
```

PR #1663, opened 2023-08-28
Mark abstract model classes as abstract classes (in mypy)

### LINKED ISSUE #1653: Restore ability to instantiate, access `objects` manager, of abstract models
Continuing the discussion started at https://github.com/typeddjango/django-stubs/pull/1649#discussion_r1291095534

There are two similar use cases affected by two different PRs:

* #1393
* #1649

Since the changes in #1393, django-stubs will raise errors on some kinds of generic code that handles abstract models. For example:

```python
# models.py
from typing import TypeVar
from django.db import models
from django_stubs_ext.db.models import TypedModelMeta

class Animal(models.Model):

    class Meta(TypedModelMeta):
        abstract = True

class Cat(Animal):  # Concrete model
    pass

def create_animal(klass: type[Animal], name: str) -> Animal:
    obj = klass(name=name)
    obj.save()
    return obj

T = TypeVar("T", bound=Animal)

def create_animal_generic(klass: type[T], name: str) -> T:
    return klass.objects.create(name=name)

# Usage example:
# create_animal(Cat, "Garfield")
# create_animal_generic(Cat, "Grumpy")
```

These two functions are allowed and work with django-stubs 4.2.3. Function `create_animal` causes errors in git master version of django-stubs:

```
apidoc/models.py:16:11: error: Cannot instantiate abstract model "Animal"  [misc]
```

But after the unmerged changes in PR #1649, the second function `create_animal_generic` will also start failing:

```
models.py:16:11: error: Cannot instantiate abstract model "Animal"  [misc]
models.py:23:12: error: Cannot instantiate abstract model "Animal"  [misc]
models.py:23:12: error: Incompatible return value type (got "Animal", expected "T")  [return-value]
```

## System information

- `python` version: 3.11.4
- `mypy` version: 1.4.1
- `django-stubs` version: 4.2.3 / git

flaeppe:
Attempt to mark abstract model classes as abstract classes in mypy.

_Note:_ Please feel free to chime in with additional test cases/scenarios we should add coverage for.

__Note that with the changes here `Abstract.objects.create(...)` no longer yields any instantiation error__

# I have made things!

Changes here should (quite naively) allow an abstract model class to be used in annotations e.g.

```python
class Abstract(models.Model):
    class Meta:
        abstract = True

def x(klass: type[Abstract]) -> None:
    klass()

T = TypeVar("T", bound=Abstract)
def y(klass: type[T]) -> None:
    klass()
```

This is similar to how mypy behaves in regards to abstract base classes.

Additionally Mypy requires some attributes to be abstract in order to make the class abstract, so we mark model attributes `<Model>.DoesNotExist` and `<Model>.MultipleObjectsReturned` to be abstract.

## Related issues

Closes #1653


flaeppe:
The errors  in CI seem to come due to `get_method_hook`. I'm suspecting that there's a bug in mypy, as it seems that when a noop plugin hook is called, returning the default return type, there's some discrepancy.

e.g.

```python
class MyPlugin(Plugin):
    def get_method_hook(self, fullname: str) -> Callable[[MethodContext], Type]:
        return self.return_default_type

    def return_default_type(self, ctx: MethodContext) -> Type:
        return ctx.default_return_type
```

Optimally we'd probably be happiest if we didn't return a hook for every function/method call. Perhaps;

- `get_method_hook` should sort out if something is a model or not.
- `get_function_hook` could e.g. check if function signature contains any `type[<AbstractModel>]` argument.

_Note to self:_

Think there might be something with a `.copy_modified` being called here: https://github.com/python/mypy/blob/dc7344539bd6e40825486dfaaa3d0bc34a64784e/mypy/checkexpr.py#L1745

mypy_django_plugin/main.py
```diff
@@ -186,12 +187,17 @@ def get_function_hook(self, fullname: str) -> Optional[Callable[[FunctionContext
 
             if helpers.is_model_subclass_info(info, self.django_context):
                 return partial(init_create.redefine_and_typecheck_model_init, django_context=self.django_context)
-        return None
+
+        return partial(self.check_function_call, fullname=fullname)
```
sobolevn: We cannot do that :(
It will slow down type-checking for all function calls for all users.

flaeppe: I know, it's the best option I could find though.. We could e.g. inspect function args and see if there's any abstract model class expected (basically move a part of `check_function_call`) here. Apart from that, I'm out of ideas.

sobolevn: We can try adding `is_abstract = True` to `TypeInfo` of the model as well.

flaeppe: Yeah, I could try that out.

It's not going to cover for the manager calls. But might be a different story, as Django adds no manager attribute to an abstract model, it might work better to handle that differently.

flaeppe: > We can try adding `is_abstract = True` to `TypeInfo` of the model as well.

Right, so I made that happen. But had to make something an abstract attribute. So I opted for `DoesNotExist` and `MultipleObjectsReturned` since those are only added on concrete model classes by the metaclass

sobolevn (commented):
Don't forget to also set:
- https://github.com/python/mypy/blob/a7e0f6f8b0ec5de2fe7b804c9ac7160893ae5bf8/mypy/nodes.py#L2890
- https://github.com/python/mypy/blob/a7e0f6f8b0ec5de2fe7b804c9ac7160893ae5bf8/mypy/nodes.py#L2895

flaeppe:
> Don't forget to also set:
> 
> * https://github.com/python/mypy/blob/a7e0f6f8b0ec5de2fe7b804c9ac7160893ae5bf8/mypy/nodes.py#L2890
> * https://github.com/python/mypy/blob/a7e0f6f8b0ec5de2fe7b804c9ac7160893ae5bf8/mypy/nodes.py#L2895

We could, but those are actually set/overwritten by mypy

Refs:

- https://github.com/python/mypy/blob/010da0b2f48dc92be2f79495fd4551c92351868f/mypy/semanal_classprop.py#L82
- https://github.com/python/mypy/blob/010da0b2f48dc92be2f79495fd4551c92351868f/mypy/semanal_classprop.py#L88
- https://github.com/python/mypy/blob/010da0b2f48dc92be2f79495fd4551c92351868f/mypy/semanal_classprop.py#L96

 What is needed for mypy to populate them is to find class symbols, either `Var` or `FuncDef` that are marked as abstract. Thus it's the plugin adding 2 `Var`s - `DoesNotExist` and `MultipleObjectsReturned` marked as abstract that makes the class become abstract.

django-stubs/contrib/auth/models.pyi
```diff
@@ -80,6 +83,9 @@ class AbstractUser(AbstractBaseUser, PermissionsMixin):
 
     EMAIL_FIELD: str
     USERNAME_FIELD: str
+
+    class Meta:
+        abstract: bool = True
```
flaeppe: I've made the plugin code to look for `abstract = True`. But perhaps we should also look for `abstract: Literal[True]`?

Not sure what effect that has on/if inheriting e.g. 

```python
class Meta(AbstractUser.Meta):
    abstract = False
```

sobolevn: I think we should support both.

mypy_django_plugin/transformers/models.py
```diff
@@ -640,19 +641,41 @@ def get_exception_bases(self, name: str) -> List[Instance]:
...
+        return False
+
+    def add_exception_classes(self) -> None:
         """
         Adds exception classes 'DoesNotExist' and 'MultipleObjectsReturned' to a model
         type, aligned with how the model metaclass does it runtime.
 
         If the model is abstract, exceptions will be added as abstract attributes.
         """
-        is_abstract = model_cls._meta.abstract
```
sobolevn: Can you please explain why do you prefer static check over runtime one?

flaeppe: Runtime one didn't work. As abstract models aren't returned by `apps.get_models()`, we just happen to append any abstract model bases we run in to when looking through `apps.get_models()`.

But that becomes an issue with e.g. `AbstractUser` as it is then dependant on having a concrete model (i.e. in `apps.get_models()`) that has subclassed `AbstractUser` to be able to process `AbstractUser`.

Ref:

https://github.com/typeddjango/django-stubs/blob/a8e42cbcca7a2ce3d890789223d3043412e51c4c/mypy_django_plugin/django/context.py#L88-L98

sobolevn: Hm, maybe we should modify this plugin hook: https://github.com/typeddjango/django-stubs/blob/a8e42cbcca7a2ce3d890789223d3043412e51c4c/mypy_django_plugin/main.py#L259-L263 to work with all classes that have `Model` base class in its `mro`?

This way we can still use the runtime API which is the most reliable way to analyze things.

flaeppe: I think it's shorter/better to look for `isinstance(type_info.metaclass_type, Instance) and type_info.metaclass_type.type.fullname == "django.db.models.base.ModelBase"` to figure out what is a model class.

But changing the condition you mention won't affect if we find a model or not. We won't be given all abstract models from the runtime API.

sobolevn: Yes, we won't get abstract models from runtime, we will get them from mypy. Mypy calls `get_base_classes` hook for all classes. We can check them and decide which ones are models.

flaeppe: Alright, what you suggest we change is just that the hook is called for each class inheriting from `Model` (or as I suggested has `ModelBase` as metaclass)?

sobolevn: Yes :)

flaeppe: Changing the hook condition seems to propagate into removing some more stuff. Perhaps we could do that in a separate PR?

django-stubs/contrib/auth/models.pyi
flaeppe: I've added support for both `abstract: Literal[True]` and `abstract = True` now (in 68a29592bff31f69db7dea8484c5979eeac03655)

flaeppe:
I'm thinking it could also be appropriate to add `django.core.exceptions` as an additional dependency via `get_additional_deps` for models, `django.contrib.auth` etc.

mypy_django_plugin/transformers/models.py
flaeppe: I've opened #1668 just now

sobolevn:
> I'm thinking it could also be appropriate to add django.core.exceptions as an additional dependency via get_additional_deps for models, django.contrib.auth etc.

I think that should be done in a separate PR, it makes sense.

flaeppe:
> > I'm thinking it could also be appropriate to add django.core.exceptions as an additional dependency via get_additional_deps for models, django.contrib.auth etc.
> 
> I think that should be done in a separate PR, it makes sense.

Alright, I'll break 911b960dc7836d006f957b6c73cb1697c364c142 out a bit later

flaeppe:
> > > I'm thinking it could also be appropriate to add django.core.exceptions as an additional dependency via get_additional_deps for models, django.contrib.auth etc.
> > 
> > 
> > I think that should be done in a separate PR, it makes sense.
> 
> Alright, I'll break [911b960](https://github.com/typeddjango/django-stubs/commit/911b960dc7836d006f957b6c73cb1697c364c142) out a bit later

I dropped those changes for now.

flaeppe:
@sobolevn what do you think. Is this ready?

tests/typecheck/models/test_metaclass_additions.yml
flaeppe, line 1: Perhaps this file should be named `test_metaclass.yml`

mypy_django_plugin/transformers/models.py
```diff
@@ -587,6 +599,133 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+        if meta is not None and isinstance(meta.node, TypeInfo) and "abstract" in meta.node.names:
+            for stmt in meta.node.defn.defs.body:
+                if (
+                    # abstract =
+                    isinstance(stmt, AssignmentStmt)
+                    and len(stmt.lvalues) == 1
+                    and isinstance(stmt.lvalues[0], NameExpr)
+                    and stmt.lvalues[0].name == "abstract"
+                    and (
+                        (
```
sobolevn: Let's break this down into separate conditions to have more readability.

1. Ending with `and stmt.lvalues[0].name == "abstract"`
2. About `abstract = True`
3. About `abstract: Literal[True]`

tests/typecheck/models/test_abstract.yml
```diff
@@ -71,3 +70,135 @@
...
+
+- case: test_abstract_model_type_cannot_be_used_as_runtime_arg
+  main: |
+      from myapp.models import Abstract, Concrete
+      from typing import Generic, Type, TypeVar, overload, Union
+
+      def f(klass: Type[Abstract]) -> None:
+          return None
+
+      f(Abstract)  # E: Only concrete class can be given where "Type[Abstract]" is expected
```
sobolevn, line 86: `f(Concerete)` is missing

tests/typecheck/models/test_abstract.yml
```diff
@@ -71,3 +70,135 @@
...
+
+      second_arg("abc", Abstract)  # E: Only concrete class can be given where "Type[Abstract]" is expected
+      second_arg(Abstract, Concrete)  # E: Argument 1 to "second_arg" has incompatible type "Type[Abstract]"; expected "str"
+
+      T = TypeVar("T", bound=Abstract)
+
+      def g(klass: Type[T]) -> None:
+          return None
+
+      g(Abstract)  # E: Only concrete class can be given where "Type[Abstract]" is expected
```
sobolevn, line 100: `g(Concrete)`

tests/typecheck/models/test_abstract.yml
```diff
@@ -71,3 +70,135 @@
...
+
+      A = Abstract
+      A()  # E: Cannot instantiate abstract class "Abstract" with abstract attributes "DoesNotExist" and "MultipleObjectsReturned"
```
sobolevn:
```suggestion
      Abstract()  # E: Cannot instantiate abstract class "Abstract" with abstract attributes "DoesNotExist" and "MultipleObjectsReturned"
```

tests/typecheck/models/test_abstract.yml
```diff
@@ -71,3 +70,135 @@
...
+      second_arg(Abstract, Concrete)  # E: Argument 1 to "second_arg" has incompatible type "Type[Abstract]"; expected "str"
+
+      T = TypeVar("T", bound=Abstract)
+
+      def g(klass: Type[T]) -> None:
+          return None
+
+      g(Abstract)  # E: Only concrete class can be given where "Type[Abstract]" is expected
+
+      A = Abstract
```
sobolevn:
```suggestion
```

tests/typecheck/models/test_abstract.yml
```diff
@@ -71,3 +70,135 @@
...
+
+      @overload
+      def o(klass: Type[T], arg: str) -> None: ...
+      @overload
+      def o(klass: Type[T], arg: int) -> None: ...
+      def o(klass: Type[T], arg: Union[str, int]) -> None:
+          return None
+
+      o(Abstract, "a")  # E: Only concrete class can be given where "Type[Abstract]" is expected
+      o(Abstract, 1)  # E: Only concrete class can be given where "Type[Abstract]" is expected
```
sobolevn, line 111: `o(Concrete, 1)`

tests/typecheck/models/test_metaclass_additions.yml
sobolevn: Both are fine :)

sobolevn (commented):
Also, please add an epxlicit test with `Literal[True]`, right now we only test `= True`

mypy_django_plugin/transformers/models.py
flaeppe: I've split it up a bit. Have a look again and see if it reads better

sobolevn (approved):
Thanks for doing this! This is great! 👍

sjdemartini:
fyi I've opened a new issue https://github.com/typeddjango/django-stubs/issues/1697 related to `models.Model.DoesNotExist` and `models.Model.MultipleObjectsReturned` being "not defined", since this breaks mypy validation of generic code that relies upon those base exception classes.

PR #1665, opened 2023-08-30
Make "default_storage" produce a "Storage" object

### LINKED ISSUE #1610: DefaultStorage has no attribute "url"
# Bug report

## What's wrong

I am seeing a regression after updating from ``django-stubs==4.2.0`` to ``django-stubs==4.2.3``. This PR appears to be the source of the issue: https://github.com/typeddjango/django-stubs/pull/1488

All of the attributes for the `default_storage` django object now raise type errors.

```python
from django.core.files.storage import default_storage

default_storage.url("filename")
...
```

```
error: "DefaultStorage" has no attribute "url"  [attr-defined]
error: "DefaultStorage" has no attribute "listdir"  [attr-defined]
error: "DefaultStorage" has no attribute "get_modified_time"  [attr-defined]
error: "DefaultStorage" has no attribute "delete"  [attr-defined]
...
```

I understand that this object is difficult to annotate because it's lazy loaded from the user's settings, but I would at least expect the attributes and methods on the base `Storage` class to be allowed.

https://docs.djangoproject.com/en/4.2/ref/files/storage/#the-storage-class

## How is that should be

Using the `default_storage` object, as described by the django docs, should not raise type errors.

## System information

- `python` version: 3.10
- `django` version: 3.1.10
- `mypy` version: 1.4.1

brianhelba:
The actual `DefaultStorage` storage class is a `LazyObject` proxy, which ought to be a generic `LazyObject[Storage]`, but doesn't support functionality as a proxy. Currently, `default_storage` doesn't have a useful annotation.

## Related issues
- Closes #1610

### DIFF
--- a/django-stubs/core/files/storage/__init__.pyi
+++ b/django-stubs/core/files/storage/__init__.pyi
@@ -24,4 +24,5 @@ def get_storage_class(import_path: str | None = ...) -> type[Storage]: ...
 class DefaultStorage(LazyObject): ...
 
 storages: StorageHandler
-default_storage: DefaultStorage
+# default_storage is actually an instance of DefaultStorage, but it proxies through to a Storage
+default_storage: Storage
--- a/scripts/stubtest/allowlist.txt
+++ b/scripts/stubtest/allowlist.txt
@@ -14,6 +14,9 @@ django.contrib.auth.migrations.*
 django.contrib.flatpages.migrations.*
 django.contrib.contenttypes.migrations.*
 
+# default_storage is actually an instance of DefaultStorage, but it proxies through to a Storage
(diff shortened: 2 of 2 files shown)


brianhelba:
I think we need to add an exception to `stubtest/allowlist.txt`, but I'd appreciate another person's opinion.

sobolevn (commented):
Yes, we can probably lie about it. Please:
- Add a comment near `default_storage`
- Add a `allowlist` entry with a comment as well

brianhelba:
@sobolevn @intgr I've just updated this again to address feedback.

sobolevn (approved):
Thanks!

PR #1666, opened 2023-08-30
Add TypeVar resolution for as_manager querysets (typeddjango#1646)

### LINKED ISSUE #1646: create_new_manager_class_from_as_manager_method : TypeVar QuerySet does not seems to work.
# Bug report

I have the following code (this is a simplified version of my problem):
```
from typing import TypeVar, TYPE_CHECKING

from django.db import models

_CTE = TypeVar("_CTE", bound=models.Model)

class _MyModelQuerySet(models.QuerySet[_CTE]):

    if TYPE_CHECKING:
        def example(self) -> _CTE: ...
    
    def test(self) -> "Self":
        return self

class MyModelQuerySet(_MyModelQuerySet["MyModel"]):
    pass

class MyModel(models.Model):
    objects = MyModelQuerySet.as_manager()


def use_my_model():
    foo: MyModel = MyModel.objects.example() # error (1)
    foo2: int = MyModelQuerySet().example()  # error (2)
    MyModel.objects.test().example()   # error (3)

```

I have the mypy report :

(1)  `error: Incompatible types in assignment (expression has type "_CTE", variable has type "MyModel")  [assignment] `
(2)  `error: Incompatible types in assignment (expression has type "MyModel", variable has type "int")  [assignment] `
(3)  `error: <nothing> has no attribute "example" `
The (2) is right, but the (1) is wrong.

Moreover Self types seems not to work (with (3))

## What's wrong

It seems that when ManagerFromMyModelQuerySet is created, it does not work with a TypeVar, nor Self type.

## System information

    OS: Ubuntu
    python version: 3.9.10
    django version: 3.2.19
    mypy version: 1.4.1
    django-stubs version: 4.2.3
    django-stubs-ext version: 4.2.2

moranabadie:
# I have made things!

In this pull request, I introduce a new utility function called _find_type_var, which aims to identify the _CTE associated with _MyModelQuerySet. This change is demonstrated in the accompanying test.

I've strived to make the implementation as generic as possible. I believe this approach should not introduce any regressions, although I welcome feedback on whether there might be a simpler way to achieve the same result.

Please refer to the included test for a detailed demonstration of the issue that this pull request seeks to resolve.

## Related issues

- Partially Closes #1646

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -19,7 +19,7 @@ from mypy.nodes import (
 from mypy.plugin import AttributeContext, ClassDefContext, DynamicClassDefContext
 from mypy.semanal import SemanticAnalyzer
 from mypy.semanal_shared import has_placeholder
-from mypy.types import AnyType, CallableType, FunctionLike, Instance, Overloaded, ProperType, TypeOfAny
+from mypy.types import AnyType, CallableType, FunctionLike, Instance, Overloaded, ProperType, TypeOfAny, TypeVarType
 from mypy.types import Type as MypyType
 from mypy.typevars import fill_typevars
 
@@ -74,7 +74,10 @@ def get_method_type_from_dynamic_manager(
 
     is_fallback_queryset = queryset_info.metadata.get("django", {}).get("any_fallback_queryset", False)
 
-    method_type = _get_funcdef_type(queryset_info.get_method(method_name))
+    base_that_has_method = queryset_info.get_containing_type_info(method_name)
+    if base_that_has_method is None:
+        return None
+    method_type = _get_funcdef_type(base_that_has_method.names[method_name].node)
     if not isinstance(method_type, FunctionLike):
         return method_type
 
@@ -84,6 +87,7 @@ def get_method_type_from_dynamic_manager(
             _process_dynamic_method(
                 method_name,
                 item,
+                base_that_has_method=base_that_has_method,
                 queryset_info=queryset_info,
                 manager_instance=manager_instance,
                 is_fallback_queryset=is_fallback_queryset,
@@ -97,6 +101,7 @@ def _process_dynamic_method(
     method_type: CallableType,
     *,
     queryset_info: TypeInfo,
+    base_that_has_method: TypeInfo,
     manager_instance: Instance,
     is_fallback_queryset: bool,
 ) -> CallableType:
@@ -117,10 +122,19 @@ def _process_dynamic_method(
             # only needed for pluign-generated querysets.
             ret_type = Instance(queryset_info, [manager_instance.args[0], manager_instance.args[0]])
         variables = []
+    args_types = method_type.arg_types[1:]
+    if _has_compatible_type_vars(base_that_has_method):
+        ret_type = _replace_type_var(
+            ret_type, base_that_has_method.defn.type_vars[0].fullname, manager_instance.args[0]
+        )
+        args_types = [
+            _replace_type_var(arg_type, base_that_has_method.defn.type_vars[0].fullname, manager_instance.args[0])
+            for arg_type in args_types
+        ]
 
     # Drop any 'self' argument as our manager is already initialized
     return method_type.copy_modified(
-        arg_types=method_type.arg_types[1:],
+        arg_types=args_types,
         arg_kinds=method_type.arg_kinds[1:],
         arg_names=method_type.arg_names[1:],
         variables=variables,
@@ -134,6 +148,91 @@ def _get_funcdef_type(definition: Union[Node, None]) -> Optional[ProperType]:
     elif isinstance(definition, Decorator):
         return definition.func.type
     return None
+
+
+def _has_compatible_type_vars(type_info: TypeInfo) -> bool:
+    """
+    Determines whether the provided 'type_info',
+    is a generically parameterized subclass of models.QuerySet[T], with exactly
+    one type variable.
+
+    Criteria for compatibility:
+    1. 'type_info' must be a generic class with exactly one type variable.
+    2. All superclasses of 'type_info', up to and including models.QuerySet,
+       must also be generic classes with exactly one type variable.
+
+    Examples:
+
+    Compatible:
+        class _MyModelQuerySet(models.QuerySet[T]): ...
+        class MyModelQuerySet(_MyModelQuerySet[T_2]):
+            def example(self) -> T_2: ...
+
+    Incompatible:
+        class MyModelQuerySet(models.QuerySet[T], Generic[T, T2]):
+            def example(self, a: T2) -> T_2: ...
+
+    Returns:
+        True if the 'base' meets the criteria, otherwise False.
+    """
+    args = type_info.defn.type_vars
+    if not args or len(args) > 1 or not isinstance(args[0], TypeVarType):
+        # No type var to manage, or too many
+        return False
+    type_var: Optional[MypyType] = None
+    # check that for all the bases it has only one type vars
+    for sub_base in type_info.bases:
+        unic_args = list(set(sub_base.args))
+        if not unic_args or len(unic_args) > 1:
+            # No type var for the sub_base, skipping
+            continue
+        if type_var and unic_args and type_var != unic_args[0]:
+            # There is two different type vars in the bases, we are not compatible
+            return False
+        type_var = unic_args[0]
+    if not type_var:
+        # No type var found in the bases.
+        return False
+
+    if type_info.has_base(fullnames.QUERYSET_CLASS_FULLNAME):
+        # If it is a subclass of _QuerySet, it is compatible.
+        return True
+    # check that at least one base is a subclass of queryset with Generic type vars
+    return any(_has_compatible_type_vars(sub_base.type) for sub_base in type_info.bases)
+
+
+def _replace_type_var(ret_type: MypyType, to_replace: str, replace_by: MypyType) -> MypyType:
+    """
+    Substitutes a specified type variable within a Mypy type expression with an actual type.
+
+    This function is recursive, and it operates on various kinds of Mypy types like Instance,
+    ProperType, etc., to deeply replace the specified type variable.
+
+    Parameters:
+    - ret_type: A Mypy type expression where the substitution should occur.
+    - to_replace: The type variable to be replaced, specified as its full name.
+    - replace_by: The actual Mypy type to substitute in place of 'to_replace'.
+
+    Example:
+    Given:
+        ret_type = "typing.Collection[T]"
+        to_replace = "T"
+        replace_by = "myapp.models.MyModel"
+    Result:
+        "typing.Collection[myapp.models.MyModel]"
+    """
+    if isinstance(ret_type, TypeVarType) and ret_type.fullname == to_replace:
+        return replace_by
+    elif isinstance(ret_type, Instance):
+        # Since it is an instance, recursively find the type var for all its args.
+        ret_type.args = tuple(_replace_type_var(item, to_replace, replace_by) for item in ret_type.args)
+    if isinstance(ret_type, ProperType) and hasattr(ret_type, "item"):
+        # For example TypeType has an item. find the type_var for this item
+        ret_type.item = _replace_type_var(ret_type.item, to_replace, replace_by)
+    if isinstance(ret_type, ProperType) and hasattr(ret_type, "items"):
+        # For example TypeList has items. find recursively type_var for its items
+        ret_type.items = [_replace_type_var(item, to_replace, replace_by) for item in ret_type.items]
+    return ret_type
 
 
 def get_method_type_from_reverse_manager(
--- a/tests/typecheck/managers/querysets/test_as_manager.yml
+++ b/tests/typecheck/managers/querysets/test_as_manager.yml
@@ -146,6 +146,122 @@
                 class MyModel(models.Model):
                     objects = ManagerFromModelQuerySet
 
+-   case: handles_type_var_in_subclasses_of_subclasses_of_queryset
+    main: |
+        from myapp.models import MyModel, MyOtherModel
+        reveal_type(MyModel.objects.example_2())  # N: Revealed type is "myapp.models.MyModel"
+        reveal_type(MyModel.objects.example())  # N: Revealed type is "myapp.models.MyModel"
+        reveal_type(MyModel.objects.example_2())  # N: Revealed type is "myapp.models.MyModel"
+        reveal_type(MyModel.objects.override())  # N: Revealed type is "myapp.models.MyModel"
+        reveal_type(MyModel.objects.override2())  # N: Revealed type is "myapp.models.MyModel"
+        reveal_type(MyModel.objects.dummy_override())  # N: Revealed type is "myapp.models.MyModel"
+        reveal_type(MyModel.objects.example_mixin(MyModel()))  # N: Revealed type is "myapp.models.MyModel"
+        reveal_type(MyModel.objects.example_other_mixin())  # N: Revealed type is "myapp.models.MyModel"
+        reveal_type(MyOtherModel.objects.example())  # N: Revealed type is "myapp.models.MyOtherModel"
+        reveal_type(MyOtherModel.objects.example_2())  # N: Revealed type is "myapp.models.MyOtherModel"
+        reveal_type(MyOtherModel.objects.override())  # N: Revealed type is "myapp.models.MyOtherModel"
+        reveal_type(MyOtherModel.objects.override2())  # N: Revealed type is "myapp.models.MyOtherModel"
+        reveal_type(MyOtherModel.objects.dummy_override())  # N: Revealed type is "myapp.models.MyOtherModel"
+        reveal_type(MyOtherModel.objects.example_mixin(MyOtherModel()))  # N: Revealed type is "myapp.models.MyOtherModel"
+        reveal_type(MyOtherModel.objects.example_other_mixin())  # N: Revealed type is "myapp.models.MyOtherModel"
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+              from typing import TypeVar, Generic
+              from django.db import models
+
+              T = TypeVar("T", bound=models.Model)
+              T_2 = TypeVar("T_2", bound=models.Model)
+
+              class SomeMixin:
+                  def example_mixin(self, a: T) -> T: ...
+
+              class OtherMixin(models.QuerySet[T]):
+                  def example_other_mixin(self) -> T: ...
+
+              class _MyModelQuerySet(OtherMixin[T], models.QuerySet[T], Generic[T]):
+                  def example(self) -> T: ...
+                  def override(self) -> T: ...
+                  def override2(self) -> T: ...
+                  def dummy_override(self) -> int: ...
+
+              class _MyModelQuerySet2(SomeMixin, _MyModelQuerySet[T_2]):
+                  def example_2(self) -> T_2: ...
+                  def override(self) -> T_2: ...
+                  def override2(self) -> T_2: ...
+                  def dummy_override(self) -> T_2: ...  # type: ignore[override]
+
+              class MyModelQuerySet(_MyModelQuerySet2["MyModel"]):
+                  def override(self) -> "MyModel": ...
+
+              class MyModel(models.Model):
+                  objects = MyModelQuerySet.as_manager()
+
+              class MyOtherModel(models.Model):
+                  objects = _MyModelQuerySet2.as_manager()  # type: ignore
+
+-   case: handles_type_vars
+    main: |
+        from myapp.models import MyModel, BaseQuerySet
+        reveal_type(MyModel.objects.example())  # N: Revealed type is "myapp.models.MyModel"
+        reveal_type(MyModel.objects.example_list())  # N: Revealed type is "builtins.list[myapp.models.MyModel]"
+        reveal_type(MyModel.objects.example_type())  # N: Revealed type is "Type[myapp.models.MyModel]"
+        reveal_type(MyModel.objects.example_tuple_simple())  # N: Revealed type is "Tuple[myapp.models.MyModel]"
+        reveal_type(MyModel.objects.example_tuple_list())  # N: Revealed type is "builtins.tuple[myapp.models.MyModel, ...]"
+        reveal_type(MyModel.objects.example_tuple_double())  # N: Revealed type is "Tuple[builtins.int, myapp.models.MyModel]"
+        reveal_type(MyModel.objects.example_class())  # N: Revealed type is "myapp.models.Example[myapp.models.MyModel]"
+        reveal_type(MyModel.objects.example_type_class())  # N: Revealed type is "Type[myapp.models.Example[myapp.models.MyModel]]"
+        reveal_type(MyModel.objects.example_collection())  # N: Revealed type is "typing.Collection[myapp.models.MyModel]"
+        reveal_type(MyModel.objects.example_set())  # N: Revealed type is "builtins.set[myapp.models.MyModel]"
+        reveal_type(MyModel.objects.example_dict())  # N: Revealed type is "builtins.dict[builtins.str, myapp.models.MyModel]"
+        reveal_type(MyModel.objects.example_list_dict())  # N: Revealed type is "builtins.list[builtins.dict[builtins.str, myapp.models.MyModel]]"
+        class TestQuerySet(BaseQuerySet[str]): ... # E: Type argument "str" of "BaseQuerySet" must be a subtype of "Model"
+        reveal_type(MyModel.objects.example_t(5))  # N: Revealed type is "builtins.int"
+        MyModel.objects.example_arg(5, "5")  # E: Argument 1 to "example_arg" of "BaseQuerySet" has incompatible type "int"; expected "MyModel"
+        reveal_type(MyModel.objects.example_arg(MyModel(), "5"))  # N: Revealed type is "myapp.models.MyModel"
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from __future__ import annotations
+                from typing import TypeVar, Generic, Collection
+
+                from django.db import models
+
+                _CTE = TypeVar("_CTE", bound=models.Model)
+                T = TypeVar("T")
+
+                class Example(Generic[_CTE]): ...
+
+                class BaseQuerySet(models.QuerySet[_CTE], Generic[_CTE]):
+
+                    def example(self) -> _CTE: ...
+                    def example_list(self) -> list[_CTE]: ...
+                    def example_type(self) -> type[_CTE]: ...
+                    def example_tuple_simple(self) -> tuple[_CTE]: ...
+                    def example_tuple_list(self) -> tuple[_CTE, ...]: ...
+                    def example_tuple_double(self) -> tuple[int, _CTE]: ...
+                    def example_class(self) -> Example[_CTE]: ...
+                    def example_type_class(self) -> type[Example[_CTE]]: ...
+                    def example_collection(self) -> Collection[_CTE]: ...
+                    def example_set(self) -> set[_CTE]: ...
+                    def example_dict(self) -> dict[str, _CTE]: ...
+                    def example_list_dict(self) -> list[dict[str, _CTE]]: ...
+                    def example_t(self, a: T) -> T: ...
+                    def example_arg(self, a: _CTE, b: str) -> _CTE: ...
+
+
+                class MyModelQuerySet(BaseQuerySet["MyModel"]):
+                    ...
+
+                class MyModel(models.Model):
+                    objects = MyModelQuerySet.as_manager()
+
 -   case: reuses_generated_type_when_called_identically_for_multiple_managers
     main: |
         from myapp.models import MyModel
--- a/tests/typecheck/managers/querysets/test_from_queryset.yml
+++ b/tests/typecheck/managers/querysets/test_from_queryset.yml
@@ -70,6 +70,41 @@
                 NewManager = BaseManager.from_queryset(ModelQuerySet)
                 class MyModel(models.Model):
                     objects = NewManager()
+-   case: handles_subclasses_of_queryset
+    main: |
+        from myapp.models import MyModel
+        reveal_type(MyModel.objects.example())  # N: Revealed type is "myapp.models.MyModel"
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/querysets.py
+            content: |
+                from typing import TypeVar, TYPE_CHECKING
+
+                from django.db import models
+                from django.db.models.manager import BaseManager
+                if TYPE_CHECKING:
+                    from .models import MyModel
+
+                _CTE = TypeVar("_CTE", bound=models.Model)
+
+                class _MyModelQuerySet(models.QuerySet[_CTE]):
+
+                    def example(self) -> _CTE: ...
+
+                class MyModelQuerySet(_MyModelQuerySet["MyModel"]):
+                    ...
+
+        - path: myapp/models.py
+          content: |
+            from django.db import models
+            from django.db.models.manager import BaseManager
+            from .querysets import MyModelQuerySet
+
+            NewManager = BaseManager.from_queryset(MyModelQuerySet)
+            class MyModel(models.Model):
+                objects = NewManager()
 
 -   case: from_queryset_generated_manager_imported_from_other_module
     main: |


sobolevn (commented):
ALternative idea: how har it would be to resolve these type vars when we create a type with `.as_manager()` and `.as_queryset()`?

moranabadie:
> ALternative idea: how har it would be to resolve these type vars when we create a type with `.as_manager()` and `.as_queryset()`?

If I understand well you want this to make it work for from_queryset as well. It seems to work (I have added a test).

sobolevn (commented):
Nice that it works, but I was asking about these lines: https://github.com/typeddjango/django-stubs/blob/a8e42cbcca7a2ce3d890789223d3043412e51c4c/mypy_django_plugin/main.py#L326-L333

Can we do it somewhere there? Or will it be too hard?

moranabadie:
It seems that all the methods of the generated manager are all purposely implemented with the  `sym_type=AnyType(TypeOfAny.implementation_artifact)`, according the comment :

https://github.com/typeddjango/django-stubs/blob/a8e42cbcca7a2ce3d890789223d3043412e51c4c/mypy_django_plugin/transformers/managers.py#L343.py#L352-L362.

`populate_manager_from_queryset` is called by `create_manager_info_from_from_queryset_call` and `create_new_manager_class_from_as_manager_method`.

This I am not sure if it is a good idea to make a special case for the issue of this pull request. Maybe I'm wrong.

sobolevn (approved):
Ok then, but I would like to have a second opinion as well :)

tests/typecheck/managers/querysets/test_as_manager.yml
```diff
@@ -146,6 +146,32 @@
...
+
+                class _MyModelQuerySet(models.QuerySet[_CTE]):
+
+                    def example(self) -> _CTE: ...
+
+                class MyModelQuerySet(_MyModelQuerySet["MyModel"]):
+                    ...
+
+                class MyModel(models.Model):
+                    objects = MyModelQuerySet.as_manager()
```
flaeppe: I think there's a missing `Generic` on `_MyModelQuerySet` in this test case? It looks odd with:

```python
class MyModelQuerySet(_MyModelQuerySet["MyModel"]):
    ...
```

Consider how mypy acts on generic classes not touched by the plugin:

```python
from typing import Generic, TypeVar

T = TypeVar("T")

class A(Generic[T]):
    def m(self, arg: T) -> T:
        return arg
        
        
class B(A[int]): ...


class C(B[str]): ...  # main.py:13: error: "B" expects no type arguments, but 1 given  [type-arg]
```

https://mypy-play.net/?mypy=latest&python=3.11&gist=5fc71cabc2fe3dc45cf3d379e6919ca0

So, in general, something is off here in regards to arguments..

Ref: https://github.com/typeddjango/django-stubs/pull/1649#issuecomment-1681846078

moranabadie:
I can also add a more generic approach :
```
    if isinstance(ret_type, TypeVarType):
        ret_type = _find_type_var(queryset_info, ret_type) or ret_type
    elif isinstance(ret_type, Instance):
        ret_type.args = tuple(
            _find_type_var(queryset_info, item) or item
            if isinstance(item, TypeVarType) else item for item in ret_type.args
        )
    elif isinstance(ret_type, TypeType) and isinstance(ret_type.item, TypeVarType):
        ret_type.item = _find_type_var(queryset_info, ret_type.item) or ret_type.item
    elif isinstance(ret_type, ProperType) and hasattr(ret_type, "items"):
        ret_type.items = [
            _find_type_var(queryset_info, item) or item
            if isinstance(item, TypeVarType) else item for item in ret_type.items
        ]

```

for 

```
    def example2(self) -> list[_CTE]: ...


    def example3(self) -> type[_CTE]: ...

    def example4(self) -> tuple[_CTE]: ...

    def example5(self) -> List[_CTE]: ...

```

if you think this is useful and not overkill.

tests/typecheck/managers/querysets/test_as_manager.yml
moranabadie: If you mean 
`class _MyModelQuerySet(models.QuerySet[_CTE]):` instead of `class _MyModelQuerySet(models.QuerySet[_CTE], Generic[_CTE)):` ? I am not sure to see the difference for the test.

I can Add the Generic if you think this is more understandable.

And in your example, B, as you point out, is not generic. This is what I want for MyModelQuerySet, for not being generic :

MyModelQuerySet is a queryset of MyModel and nothing else.

moranabadie: In both cases I have the same mypy output :
```
from typing import Generic, TypeVar

T = TypeVar("T")

class A(Generic[T]):
    def m(self, arg: T) -> T:
        return arg
class B(A[T]): ...
class C(B[int]): ...
reveal_type(C().m(3))
```
test1.py:14: note: Revealed type is "builtins.int"

```
from typing import Generic, TypeVar

T = TypeVar("T")

class A(Generic[T]):
    def m(self, arg: T) -> T:
        return arg

class B(A[T], Generic[T]): ...

class C(B[int]): ...


reveal_type(C().m(3))
```
test1.py:14: note: Revealed type is "builtins.int

flaeppe: Thanks for clearing that up. It's me not being used to not using a `Generic`.

mypy_django_plugin/transformers/managers.py
```diff
@@ -115,6 +118,22 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
+    Example:
+    Suppose queryset_info is based on a generic class `QuerySet[T]` and ret_type corresponds
+    to `T`, then this function will return the concrete type that `T` was instantiated with.
+    """
+    for base in queryset_info.bases:
+        for i, type_var in enumerate(base.type.type_vars):
+            type_var_path = f"{base.type.module_name}.{type_var}"
+            if type_var_path == ret_type.fullname and i < len(base.args):
+                return base.args[i]
+    return None
```
flaeppe: As you've touched in https://github.com/typeddjango/django-stubs/pull/1666#issuecomment-1700700518 I think we rather need to figure out the correct way to "forward" the `TypeVar`. As the argument could be used for another generic class, e.g. `list`, `Collection`, `Sequence`, `dict`, my custom generic class, etc.

We should cover such cases in in your test.

mypy_django_plugin/transformers/managers.py
```diff
@@ -115,6 +118,22 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
+    """
+    Attempts to find the concrete type corresponding to a TypeVarType in a queryset's base types.
+
+    Example:
+    Suppose queryset_info is based on a generic class `QuerySet[T]` and ret_type corresponds
+    to `T`, then this function will return the concrete type that `T` was instantiated with.
+    """
+    for base in queryset_info.bases:
+        for i, type_var in enumerate(base.type.type_vars):
+            type_var_path = f"{base.type.module_name}.{type_var}"
```
flaeppe: This limits the `TypeVar` to be declared in the same module as the base class, right? That could mismatch if someone imports their `TypeVar` I think?

tests/typecheck/managers/querysets/test_as_manager.yml
```diff
@@ -146,6 +146,32 @@
...
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from typing import TypeVar
+
+                from django.db import models
+
+                _CTE = TypeVar("_CTE", bound=models.Model)
```
flaeppe, line 236: I also think we should ensure the upper bound of `QuerySet`s `TypeVar` is honored. So that we know it'll emit an error correctly.

e.g. use a `TypeVar` without upper bound and test pass e.g. `int` as argument

moranabadie: I have set : 
class OtherQuerySet(_MyModelQuerySet[str, "MyOtherModel"]):

and I have the error :
myapp/models:30: error: Type argument "str" of "_MyModelQuerySet" must be a subtype of "Model" (diff

Should I create a test for this ?

mypy_django_plugin/transformers/managers.py
moranabadie: you are right, I found a way to make it cleaner  with `base.type.defn.type_vars` instead of `base.type.type_vars` in my last commit

mypy_django_plugin/transformers/managers.py
moranabadie: I have added the code and the test in my last commit

tests/typecheck/managers/querysets/test_as_manager.yml
flaeppe: I would be fine if you'd include that error in your tests

mypy_django_plugin/transformers/managers.py
```diff
@@ -115,6 +117,48 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
+
+def _find_type_var(queryset_info: TypeInfo, ret_type: TypeVarType) -> Optional[MypyType]:
+    """
+    Attempts to find the concrete type corresponding to a TypeVarType in a queryset's base types.
+
+    Example:
+    Suppose queryset_info is based on a generic class `QuerySet[T]` and ret_type corresponds
+    to `T`, then this function will return the concrete type that `T` was instantiated with.
+    """
+    for base in queryset_info.bases:
```
flaeppe: Hm, using `TypeInfo.bases` would only include direct base classes while also not necessarily honoring the mro, I think.

I think a breaking case then is if you put half of your custom methods in your test in another parent class, under a `TypeVar` named differently, it won't be found. e.g.

```python
class _MyModelQuerySet(models.QuerySet[_CTE]):
    def example(self) -> _CTE: ...

class _MyModelQuerySet2(_MyModelQuerySet[_CTE2]):
    def example_2(self) -> _CTE_2: ...

class MyModelQuerySet(_MyModelQuerySet2["MyModel"]):
    ...

...

reveal_type(MyModel.objects.example())
```

mypy_django_plugin/transformers/managers.py
```diff
@@ -105,6 +105,8 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
             ret_type = Instance(queryset_info, [manager_instance.args[0], manager_instance.args[0]])
         variables = []
 
+    ret_type = _manage_type_var(ret_type, queryset_info)
```
flaeppe: I think your helper methods are doing a bit of duplicated work. While brute forcing a bit, trying to figure out which base class is holding the argument. I'm thinking that since we already have the method name, and with how type vars work we should then _always_ be able to find the type var on the same base that has the method.

We could, instead of _only_ fetching the method on the line below, fetch the base that declares the method, while also honoring the mro.

https://github.com/typeddjango/django-stubs/blob/81a543a070aa9596e3296db88563057f31b1e5df/mypy_django_plugin/transformers/managers.py#L82

So we could adjust to something like:
```python
# Note: Important that we look through bases of bases, according to the mro.
# Make sure to look for method in <base.type.names>
# Note: There's no `get_base_containing_method`, you have to implement that loop.
base_that_has_method: Instance | None = <queryset_info.get_base_containing_method(method_name)>
if base_that_has_method is None:
    return None
method_type = get_funcdef_type(base_that_has_method.type.names[method_name].node)
```

`get_funcdef_type` has to change its signature a little bit, but that's totally fine.

Now, we shouldn't need to dig through any more bases and their arguments. The argument we need _should_ exist on the `base_that_has_method.args` if we can't find it we just `return method_type.ret_type`

Which leads to an additional case for your test having a method with an argument that is not found in the class' args

```python
T = TypeVar("T")
class _MyModelQuerySet(models.QuerySet[_CTE]):
    ...
    def type_var_outside_class(self) -> T: ...
```

`.type_var_outside_class()` should resolve to `T`

flaeppe (commented):
I think I found a failing case but hopefully an approach that should fix it.

tests/typecheck/managers/querysets/test_as_manager.yml
moranabadie: yes, done :)

mypy_django_plugin/transformers/managers.py
moranabadie: Thank you for the nice feedback.

I have tried your approach, but in your example https://github.com/typeddjango/django-stubs/pull/1666#discussion_r1311703917, `base_that_has_method.args` has `_CTE2`. which means that `reveal_type(MyModel.objects.example())` returns `_CTE2`.

moranabadie: I have pushed a solution and added your two examples ( `def type_var_outside_class(self) -> T: ...` and your `_MyModelQuerySet2`) to the tests.

I have also added base_that_has_method, but I think this is useless, with the new way of getting the mro.

But I understand if you think that there is duplicated work, but I do not see where it is and how using it.

mypy_django_plugin/transformers/managers.py
```diff
@@ -105,6 +109,8 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
             ret_type = Instance(queryset_info, [manager_instance.args[0], manager_instance.args[0]])
         variables = []
 
+    ret_type = _find_type_var_from_mro(ret_type, queryset_info)
```
flaeppe: I'm wondering why we need to dig through the mro again, once we know which base has the method?

I think it should suffice to look through `base_that_has_method.args` for `ret_type`? Or what am I missing?

What I mean is that if one were to write it out. The type var in a method signature would be found in the arguments of the class it's declared in.

flaeppe: I checked out your branch and my suspicions were correct, your tests pass if we just call `_manage_type_var` for `base_that_has_method`

```suggestion
    ret_type = _manage_type_var(ret_type, base_that_has_method)
```

We've already dug through the mro to find everything. There was no need to do it again

mypy_django_plugin/transformers/managers.py
```diff
@@ -115,6 +121,66 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
+
+def _find_type_var(base: Instance, ret_type: TypeVarType) -> Optional[MypyType]:
+    """
+    Attempts to find the concrete type corresponding to a TypeVarType in a queryset's base types.
+
+    Example:
+    Suppose queryset_info is based on a generic class `QuerySet[T]` and ret_type corresponds
+    to `T`, then this function will return the concrete type that `T` was instantiated with.
+    """
+    for i, type_var in enumerate(base.type.defn.type_vars):
```
```suggestion
    for i, type_var in enumerate(base.type.type_vars):
```

No need to use the definition here, I think.

mypy_django_plugin/transformers/managers.py
```diff
@@ -115,6 +121,66 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
+    if isinstance(ret_type, ProperType) and hasattr(ret_type, "item"):
+        # For example TypeType has an item. find the type_var for this item
+        ret_type.item = _manage_type_var(ret_type.item, base)
+    if isinstance(ret_type, ProperType) and hasattr(ret_type, "items"):
+        # For example TypeList has items. find recursively type_var for its items
+        ret_type.items = [_manage_type_var(item, base) for item in ret_type.items]
+    return ret_type
+
+
+def _find_type_var(base: Instance, ret_type: TypeVarType) -> Optional[MypyType]:
```
```suggestion
def _find_type_var(base: Instance, target: TypeVarType) -> Optional[MypyType]:
```

mypy_django_plugin/transformers/managers.py
```diff
@@ -115,6 +121,66 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
+    """
+    Attempts to find the concrete type corresponding to a TypeVarType in a queryset's base types.
```
```suggestion
    Attempts to find the concrete type corresponding to a TypeVarType in a class's base types.
```

mypy_django_plugin/transformers/managers.py
moranabadie: In this example :
```
class _MyModelQuerySet(models.QuerySet[_CTE]):
    def example(self) -> _CTE: ...

class _MyModelQuerySet2(_MyModelQuerySet[_CTE2]):
    def example_2(self) -> _CTE_2: ...

class MyModelQuerySet(_MyModelQuerySet2["MyModel"]):
    ...
```

base_that_has_method will be `myscript._MyModelQuerySet[_CTE_2`1]`.


base_that_has_method.args is thus `(_CTE_2,)`.

So we do not have anywere `MyModel` :).

Or maybe I missed something.

mypy_django_plugin/transformers/managers.py
moranabadie: `base.type.type_vars` are strings. I can get the fullname with `base.type.defn.type_vars`

flaeppe:
I also realised we have to resolve any `TypeVar` arguments of the methods. As it's the same case for those as for the return type.

I'm getting a feeling there should be some other way to get the resolved types of a type var. Not at all sure how mypy does it, I'm hoping there could be something to piggyback from there..

mypy_django_plugin/transformers/managers.py
moranabadie: > _manage_type_var(ret_type, base_that_has_method)

It does not work for me :

image

You launched the test `handles_subclasses_of_subclasses_of_queryset` ?

flaeppe: No, it doesn't work for me either.

moranabadie:
> I also realised we have to resolve any `TypeVar` arguments of the methods. As it's the same case for those as for the return type.
> 
> I'm getting a feeling there should be some other way to get the resolved types of a type var. Not at all sure how mypy does it, I'm hoping there could be something to piggyback from there..

Do you know why those things (setting the return type of the methods) are not set here https://github.com/typeddjango/django-stubs/blob/master/mypy_django_plugin/transformers/managers.py#L351, like @sobolevn said ?

I think it would be easier to use mypy here.
I have also no clue how mypy does it. But I do not think intuitively mypy make this kind of thing by managing the methods and the class separately like it is done in this mypy plugin (populate_manager_from_queryset for creating the manager, and get_method_type_from_dynamic_manager for the methods).

flaeppe:
Hm, a different way to look at it could be that since;

- We (should) require the original `Manager` and `QuerySet` arguments to align
- We create a subclass manager instance from those 2 classes
- We get the manager _instance_ from the attribute hook context

We could just populate any type vars with `manager_instance.args[0]`(?). So what I think I'm saying is that we should fall back on everything else during creation of the new manager type to be correct and just put that in? At least for the model argument..

flaeppe:
> Do you know why those things (setting the return type of the methods) are not set here

Yes, it's because that's happening during semantic analysis and the attribute hook is coming in during type checking. Main practical difference (in my very own words) is that not all types is or probably will be resolved during semantic analysis. So we wait for the type checker to have come around, as we then should have more types to look up and work with.

> [...] But I do not think intuitively mypy make this kind of thing by managing the methods and the class separately like it is done in this mypy plugin (populate_manager_from_queryset for creating the manager, and get_method_type_from_dynamic_manager for the methods).

You can backtrack to my (very old) PR as to why we do this: #820 . What Django is doing is that it dynamically creates a new type, based on the manager and queryset class. And we need to help mypy understand that happened and how.

moranabadie:
> Hm, a different way to look at it could be that since;
> 
>     * We (should) require the original `Manager` and `QuerySet` arguments to align
> 
>     * We create a subclass manager instance from those 2 classes
> 
>     * We get the manager _instance_ from the attribute hook context
> 
> 
> We could just populate any type vars with `manager_instance.args[0]`(?). So what I think I'm saying is that we should fall back on everything else during creation of the new manager type to be correct and just put that in? At least for the model argument..

I am not sure to follow :

with this example :

```
from typing import TypeVar
from django.db import models

_CTE = TypeVar("_CTE", bound=models.Model)
_CTE_2 = TypeVar("_CTE_2", bound=models.Model)
_CTE_3 = TypeVar("_CTE_3")

class _MyModelQuerySet(models.QuerySet[_CTE], Generic[_CTE, _CTE_3):
    def example(self) -> _CTE: ...
    def example3(self) -> _CTE_3: ...
class _MyModelQuerySet2(_MyModelQuerySet[_CTE_2, str]):
    def example_2(self) -> _CTE_2: ...

class MyModelQuerySet(_MyModelQuerySet2["MyModel"]): ....

class MyModel(models.Model):
    objects = MyModelQuerySet.as_manager()
```

in the actual context of `get_method_type_from_dynamic_manager`, we have `def example(self) -> _CTE: ...`. How can we know that _CTE is related to manager_instance.args[0] ? 
For example `def example3(self) -> _CTE_3: ...` has nothing to do with manager_instance.args[0], i.e. "MyModel"].


Or maybe I did not understand want you want to do :)

moranabadie:
> I also realised we have to resolve any `TypeVar` arguments of the methods. As it's the same case for those as for the return type.

I added the code and a test to make it work whit the original way.


But indeed it is seems like recreating some mypy code. But with this architecture of django-stubs, I am not sure how to make it without "duplicate code"

mypy_django_plugin/transformers/managers.py
```diff
@@ -105,16 +105,70 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
+
+def _find_type_var_from_mro(ret_type: MypyType, queryset_info: TypeInfo) -> Optional[MypyType]:
+    """
+    Attempts to find the type var by looking at the mro elements (reversed).
+    """
+    new_ret_type = ret_type
+    for i, mro in enumerate(reversed(queryset_info.mro)):
+        for base in mro.bases:
+            new_ret_type = _manage_type_var(new_ret_type, base)
+    return new_ret_type or ret_type
```
flaeppe: Hm, there's 2 problems I can see here;

1. The mro isn't honored, since we're iterating over bases of the mro. This would do some sort of depth first iteration and not a breadth first.
2. We never look at given `queryset_info`, only it's bases. Which requires all queryset classes used as manager to be declared _without_ any arguments.

This code _requires_ this
```python
class Parent(models.QuerySet[T]):
    def method(self) -> T:
        ...
class Child(Parent["MyModel"]):  # Here `Child` gets a `mypy.types.Instance` of `Parent` as base
    ...

class MyModel(models.Model):
    objects = Child.as_manager()
```

But it should also be totally fine to do below, which isn't working right now for multiple reasons, I think..

```python
class MyModel(models.Model):
    objects = Parent.as_manager()  # There's no `mypy.types.Instance` of `Parent` here
```

The thing is, that we _never_ have a `mypy.types.Instance` of the queryset type, we only (for sure) have a `mypy.types.Instance` of the manager type. And the code goes looking for an instance of the queryset type.

flaeppe (commented):
> in the actual context of `get_method_type_from_dynamic_manager`, we have `def example(self) -> _CTE: ....` How can we know that _CTE is related to manager_instance.args[0] ?
For example `def example3(self) -> _CTE_3: ...` has nothing to do with manager_instance.args[0], i.e. "MyModel"].

I'll try to explain what I'm thinking as best as I can. It's rather lengthy though..

So, the first part is to look at [what the runtime does](https://github.com/django/django/blob/4.2.4/django/db/models/manager.py#L107-L118). In short words; it creates a new _`Manager`_ type, which subclasses e.g. `django.db.models.Manager` and also "copies over" methods from the given queryset. As such, the resulting class is a sort of fusion between the specified manager and queryset.

e.g.

```python
    def method(self):
        ...

class MyManager(models.Manager):
    ...

MyModelManager = MyManager.from_queryset(MyQuerySet)
class MyModel(models.Model):
    objects = MyModelManager()
```

`MyModelManager` becomes

```python
class MyModelManager(MyManager):
    method = MyQuerySet.method
```

The plugin mimics this behaviour.

The second part is to look at the definition of `models.BaseManager` and `models.QuerySet` in the stubs.

`BaseManager`:
https://github.com/typeddjango/django-stubs/blob/3a1f1398aac4ca352b69beffe2be030921e9b952/django-stubs/db/models/manager.pyi#L12-L14

`QuerySet`:
https://github.com/typeddjango/django-stubs/blob/3a1f1398aac4ca352b69beffe2be030921e9b952/django-stubs/db/models/query.pyi#L11-L12
https://github.com/typeddjango/django-stubs/blob/3a1f1398aac4ca352b69beffe2be030921e9b952/django-stubs/db/models/query.pyi#L43
https://github.com/typeddjango/django-stubs/blob/3a1f1398aac4ca352b69beffe2be030921e9b952/django-stubs/db/models/query.pyi#L219

`QuerySet` and `BaseManager` are both generic over a form of `models.Model`. As a subclass of a `BaseManager`, our generated manager class (`MyModelManager`) is also generic over a form of `models.Model`.

Third part is that when a manager instance is assigned to an attribute of a `models.Model`, the generic argument of our manager is populated(i.e. this is where we get `MyModelManager[MyModel]`). This is done by the plugin. Also note that this is where we get the manager _instance_ from and the model argument type.

https://github.com/typeddjango/django-stubs/blob/3a1f1398aac4ca352b69beffe2be030921e9b952/mypy_django_plugin/transformers/models.py#L320-L328

Now, if our custom `def method` on our queryset is annotated to return a form of model e.g.

```python
class MyQuerySet(models.QuerySet[T]):
    def method(self) -> T:
        ...
```

At runtime, when is the first time we can say _anything_ about model `T`? -> It's not until our manager has been instantiated and added as an attribute to a model (which is happening with a `setattr` found [here](https://github.com/django/django/blob/6f4c7c124a03f7b01d868cb7d1fa55564b5f2bf8/django/db/models/manager.py#L120-L126). The call chain to get here is quite lengthy and circular so I'll just link this one place) and also gotten its model attribute (you can see the `self.model = cls`).

Essentially our plugin runs in to the same situation. Now, with the copies of the queryset methods on our dynamically created manager(`MyModelManager`), what should `T` be? -> It _has_ to come from the manager instance, they will always align runtime. To declare that statically we need compatible manager and queryset model arguments. It's a prerequisite to get things right and we should introduce a check that controls that compatibility.

Otherwise the provided manager and queryset classes are incompatible.

But once we reach the code that should resolve/populate `T`, that code should just replace `T` with the model argument of the manager instance, as that what is happening runtime. There is no queryset instance.

The check mentioned above should uphold the prerequisite and our code inside `get_method_type_from_dynamic_manager` can instead be naive and just use the model argument, whatever type it has.

What do you think? Does this make sense?

---

An additional thing with this is if one decides to add _more_ generic arguments to a subclass of a `QuerySet`(like you gave an example of). But I think we could do stuff in a step by step manner and see what needs to be done once that becomes a real use case.

mypy_django_plugin/transformers/managers.py
moranabadie: OK, I think I made it work with your way, I have commited the code and the new tests.

Firstly, I used `_get_base_containing_method` as you said the other day to get the base_containing_method.

Then, I use `_has_compatible_type_vars`, to check if is as type vars, if it is a subclass of **generic querysets**, and thus if it is compatible with the created manager.
It is kind of more complicated than it looks to manage for examples mixins in the subclasses (multiple bases).

If it is compatible, replace for the `ret_type` and the `args_types` with `_replace_type_var`.

Do you prefer this approach ?

Of course if a QuerySet has two type var, it is not compatible anymore, hence I have removed the tests with two type vars.

If something is not clear, or if I did not understood something, please let me know :)

mypy_django_plugin/transformers/managers.py
```diff
@@ -104,17 +108,122 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
             # only needed for pluign-generated querysets.
             ret_type = Instance(queryset_info, [manager_instance.args[0], manager_instance.args[0]])
         variables = []
+    args_types = method_type.arg_types[1:]
+    if _has_compatible_type_vars(base_that_has_method):
```
flaeppe, line 126: Can we move this check to where the manager and queryset are "fused" together?

https://github.com/typeddjango/django-stubs/blob/218d5fd1fb19dcc68811d3568b2f47b4ecaecabb/mypy_django_plugin/transformers/managers.py#L326-L329

That would be more efficient, since it'll only happen once, while this is called for each accessed manager/queryset method. We would also get better context if we'd emit an error message. Then we can only do replacement of type vars in here.

We should also ensure that upper bounds of any `TypeVar`s are compatible (i.e. `mypy.nodes.TypeVarExpr.upper_bound` or `mypy.types.TypeVarType.upper_bound`)

mypy_django_plugin/transformers/managers.py
```diff
@@ -71,15 +72,18 @@ def get_method_type_from_dynamic_manager(
...
+    def get_funcdef_type(definition: Node | None) -> Optional[ProperType]:
         # TODO: Handle @overload?
         if isinstance(definition, FuncBase) and not isinstance(definition, OverloadedFuncDef):
             return definition.type
         elif isinstance(definition, Decorator):
             return definition.func.type
         return None
 
-    method_type = get_funcdef_type(queryset_info.get_method(method_name))
+    base_that_has_method = _get_base_containing_method(queryset_info, method_name)
```
```suggestion
    base_that_has_method = queryset_info.get_containing_type_info(method_name)
```

`_get_base_containing_method` is now a reimplementation of `TypeInfo.get_containing_type_info`

mypy_django_plugin/transformers/managers.py
```diff
@@ -104,17 +108,122 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
+    """Finding the base that contains the method."""
+    if method_name in queryset_info.names:
+        return queryset_info
+    for mro in queryset_info.mro:
+        for base in mro.bases:
+            if method_name in base.type.names:
+                return base.type
+    return None
+
+
```
```suggestion
```

Mypy is shipped with this; `TypeInfo.get_containing_type_info`

mypy_django_plugin/transformers/managers.py
```diff
@@ -104,17 +108,122 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
+def _has_compatible_type_vars(type_info: TypeInfo) -> bool:
```
flaeppe, line 153: Can we yield an error message on incompatibility?

flaeppe (commented):
Nice work! I had a few more comments

mypy_django_plugin/transformers/managers.py
flaeppe: Hm, thinking twice about this, it's quite branched off and self contained. It's also only new additions and shouldn't really break anything. While it solves a couple of simple use cases.

We could try to improve in subsequent PR:s instead, if you'd prefer that? But I'd say we at least remove `_get_base_containing_method` in favor of mypy's implementation.

mypy_django_plugin/transformers/managers.py
```diff
@@ -104,17 +108,122 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
...
+            continue
+        if type_var and unic_args and type_var != unic_args[0]:
+            # There is two different type vars in the bases, we are not compatible
+            return False
+        type_var = unic_args[0]
+    if not type_var:
+        # No type var found in the bases.
+        return False
+
+    if any(sub_base.type.fullname == "django.db.models.query._QuerySet" for sub_base in type_info.bases):
```
```suggestion
    if type_info.has_base(fullnames.QUERYSET_CLASS_FULLNAME):
```

mypy_django_plugin/transformers/managers.py
moranabadie: yes maybe let's do that in another PR.

mypy_django_plugin/transformers/managers.py
moranabadie: we will do that in another PR right ?

mypy_django_plugin/transformers/managers.py
moranabadie: oups, fixed in the last commit

mypy_django_plugin/transformers/managers.py
moranabadie: oups you are right, fixed in my last commit

mypy_django_plugin/transformers/managers.py
flaeppe: Yes, sure

flaeppe (approved):
Nice work 👍 Lets merge this

flaeppe:
> Nice work 👍 Lets merge this

Thought CI had run, seems to be a few syntax fixes for older python versions to fix first 🙂

moranabadie:
> > Nice work 👍 Lets merge this
> 
> Thought CI had run, seems to be a few syntax fixes for older python versions to fix first 🙂

Arf, I forgot about the `|` for py<3.10, fixed :)

moranabadie:
I do not understand why I do have this error https://github.com/typeddjango/django-stubs/actions/runs/6077373631/job/16486992087?pr=1666. I added  `from __future__ import annotations` but I do not understand why I have it only with 3.8 and not 3.7 for example

flaeppe:
> I do not understand why I do have this error https://github.com/typeddjango/django-stubs/actions/runs/6077373631/job/16486992087?pr=1666. I added `from __future__ import annotations` but I do not understand why I have it only with 3.8 and not 3.7 for example

I'm not sure why actually

PR #1668, opened 2023-08-31
Change plugin hook condition for analyzing model base class

flaeppe:
Should simplify the hook requirement and remove some unnecessary code.

## Related issues

Refs: https://github.com/typeddjango/django-stubs/pull/1663#discussion_r1310137808


sobolevn (approved):
Great! It worked 🎉

sobolevn:
One idea: what about models with `ModelBase` subclasses used as metaclasses?
Do we care to support them?

flaeppe:
> One idea: what about models with `ModelBase` subclasses used as metaclasses? Do we care to support them?

Hm, interesting. Perhaps we should wait for that case to begin with?

I suppose we have to iterate the `metaclass_type.info.mro` too in that case.

```python
(
    metaclass_type.type.fullname == fullnames.MODEL_METACLASS_FULLNAME
    or metaclass_type.type.has_base(fullnames.MODEL_METACLASS_FULLNAME)
)
```

?

Though hard to say if we can or cannot support those..

sobolevn:
I think that it is pretty easy to support, so why not?

PR #1669, opened 2023-08-31
Add get_attname_column method of models.Field

gg-mmill:
Add type declaration of the get_attname_column of models.Field

See source in django: https://github.com/django/django/blob/main/django/db/models/fields/__init__.py#L973


sobolevn (commented):
Please, adjust https://github.com/typeddjango/django-stubs/blob/master/scripts/stubtest/allowlist_todo.txt

scripts/stubtest/allowlist_todo.txt
```diff
@@ -1,4 +1,3 @@
-# Autogenerated by `stubtest`
```
sobolevn: Please, keep this line :)

scripts/stubtest/allowlist_todo.txt
```diff
@@ -1,4 +1,3 @@
...
 # Unsorted: there are real problems and things we can really ignore.
```
sobolevn:
```suggestion
# Autogenerated by `stubtest`
# Unsorted: there are real problems and things we can really ignore.
```

scripts/stubtest/allowlist_todo.txt
gg-mmill: Indeed, wasn't on purpose :/

sobolevn (approved):
Thank you! 👍

PR #1671, opened 2023-09-02
Use dict as data structure for `model_modules`

flaeppe:
Should give a slight performance improvement as we don't need to iterate over all models declared in a file, but can lookup by name instead.


sobolevn (approved):
Yes, looks reasonable. I don't really think that we need to benchmark this, because it *should* be faster.

PR #1672, opened 2023-09-03
Declare manager class attributes on models as `ClassVar`s

flaeppe:
# I have made things!

This is an extension of the work in #1649, including:

- Adjustments for the plugin to make generated managers `ClassVar`s
- Changes the default 'objects' to 'ClassVar' and controls it via the plugin
- Plugin ensures to only add the 'objects' manager to models it exists on during runtime


sobolevn (commented):
By "extension" you mean:
1. We should merge the original PR first, then this one
2. We should only merge this one

? :)

flaeppe:
> By "extension" you mean:
> 
> 1. We should merge the original PR first, then this one
> 2. We should only merge this one
> 
> ? :)

I haven't look thoroughly enough to see exactly what I changed from fc3ef75480ba34c3148fb3833f6b062653194752. I just wanted to use it at base but push it separately.

If we decide to use the changes here I'd vote for a squash the PRs before merge. Instead of merging them one by one. But we could always have: 3. cherry pick commit here to #1649 as well 🙂

Any is fine for me. I just wanted to try to move things forward

flaeppe:
I think the changes here should be compatible with `mypy==1.5.x` now. I had a successful run locally at least

mypy_django_plugin/lib/helpers.py
```diff
@@ -379,14 +379,17 @@ def check_types_compatible(
     api.check_subtype(actual_type, expected_type, ctx.context, error_message, "got", "expected")
 
 
-def add_new_sym_for_info(info: TypeInfo, *, name: str, sym_type: MypyType, no_serialize: bool = False) -> None:
+def add_new_sym_for_info(
+    info: TypeInfo, *, name: str, sym_type: MypyType, no_serialize: bool = False, is_classvar: bool = False
```
```suggestion
    info: TypeInfo, name: str, sym_type: MypyType, *, no_serialize: bool = False, is_classvar: bool = False
```

mypy_django_plugin/transformers/models.py
```diff
@@ -70,8 +70,13 @@ def create_new_var(self, name: str, typ: MypyType) -> Var:
         var.is_inferred = True
         return var
 
-    def add_new_node_to_model_class(self, name: str, typ: MypyType, no_serialize: bool = False) -> None:
-        helpers.add_new_sym_for_info(self.model_classdef.info, name=name, sym_type=typ, no_serialize=no_serialize)
+    def add_new_node_to_model_class(
+        self, name: str, typ: MypyType, no_serialize: bool = False, is_classvar: bool = False
```
sobolevn:
```suggestion
        self, name: str, typ: MypyType, *,  no_serialize: bool = False, is_classvar: bool = False
```

sobolevn (commented):
Please, update `mypy` in `setup.py`/`requirements.txt` to test it.

Other than that - it looks good to me!

flaeppe:
> Please, update `mypy` in `setup.py`/`requirements.txt` to test it.
> 
> Other than that - it looks good to me!

Haha, I didn't check stubtest though. Quite a bit of new stuff there: https://github.com/typeddjango/django-stubs/actions/runs/6077153236/job/16486367676?pr=1672

tests/typecheck/models/test_contrib_models.yml
```diff
@@ -31,9 +31,11 @@
 
 -   case: can_override_abstract_user_manager
     main: |
-        from myapp.models import User
-        reveal_type(User.objects)  # N: Revealed type is "myapp.models.UserManager[myapp.models.User]"
-        reveal_type(User.objects.all())  # N: Revealed type is "django.db.models.query._QuerySet[myapp.models.User, myapp.models.User]"
+        from myapp.models import MyBaseUser, MyUser
+        reveal_type(MyBaseUser.objects)  # N: Revealed type is "myapp.models.MyBaseUserManager[myapp.models.MyBaseUser]"
```
andersk, line 35: Similarly to my [note](https://github.com/typeddjango/django-stubs/pull/1649#issuecomment-1681846078) on the other PR, this is a bogus type, since `MyBaseUserManager` is not a generic class and does not accept type parameters. There are many existing instances of this problem, so maybe we don’t care about adding new ones to the test suite, but I figure it’s worth pointing out.

flaeppe: I thought so too, but I think mypy is happy about that: https://github.com/typeddjango/django-stubs/pull/1666#discussion_r1311390430

Here it is under a playground link.
https://mypy-play.net/?mypy=latest&python=3.11&gist=1fc5abb96d6e8c606849f3f1af181196

andersk: Everything you wrote in that playground is fine. `A` is generic (explicitly), `B` is generic (because it inherits from `A` and does not bind the free type parameter `T`), `C` is not generic (because it inherits from `B` but **does** bind the free type parameter `T` to `int`).

What would not be fine is writing `C[int]`; that would be similar to the problem I’m pointing out here.

flaeppe: Hm, or maybe that's off. Since the type is already declared for the argument i.e. `MyUser`.

This should really only be `MyBaseUserManager`. The plugin should refrain from add in model arguments when there's no `TypeVar`

flaeppe: Anyways, I wanted to add these reveals to follow what happened with managers on `AbstractUser`.

I'll try to have a follow up round to fix sprinkling of model arguments.

sobolevn (approved):
@andersk any other comments? :)

tests/typecheck/models/test_contrib_models.yml
flaeppe: I have opened a draft for this just now, based on the work here: #1683

sobolevn:
I am going to make a new release with new mypy support! 🎉

PR #1676, opened 2023-09-04
Support `@overload` methods in `from_queryset`

sobolevn:


mypy_django_plugin/transformers/managers.py
```diff
@@ -115,6 +128,14 @@ def get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[P
     )
 
 
+def _get_funcdef_type(definition: Union[FuncBase, Decorator, None]) -> Optional[ProperType]:
```
flaeppe: We could do `mypy.nodes.Node` here I think.

```suggestion
def _get_funcdef_type(definition: Union[Node, None]) -> Optional[ProperType]:
```

mypy_django_plugin/transformers/managers.py
```diff
@@ -71,25 +71,38 @@ def get_method_type_from_dynamic_manager(
...
-        return None
+    is_fallback_queryset = queryset_info.metadata.get("django", {}).get("any_fallback_queryset", False)
 
-    method_type = get_funcdef_type(queryset_info.get_method(method_name))
+    method_type = _get_funcdef_type(queryset_info.get_method(method_name))
     if method_type is None:
         return None
 
-    assert isinstance(method_type, CallableType)
+    assert isinstance(method_type, FunctionLike)
```
flaeppe: Can `_get_funcdef_type` return `FunctionLike`? To get rid of the assert, or can we branch on it instead of asserting?

sobolevn: Changed to `if`. Because it can return `Any` for sure.

mypy_django_plugin/transformers/managers.py
```diff
@@ -71,25 +72,37 @@ def get_method_type_from_dynamic_manager(
...
-        elif isinstance(definition, Decorator):
-            return definition.func.type
-        return None
+    is_fallback_queryset = queryset_info.metadata.get("django", {}).get("any_fallback_queryset", False)
 
-    method_type = get_funcdef_type(queryset_info.get_method(method_name))
-    if method_type is None:
-        return None
+    method_type = _get_funcdef_type(queryset_info.get_method(method_name))
+    if method_type is None or not isinstance(method_type, FunctionLike):
```
flaeppe: I'm always split about this. This should work, but is it reasonable?

```suggestion
    if not isinstance(method_type, FunctionLike):
```

mypy_django_plugin/transformers/managers.py
```diff
@@ -71,25 +72,37 @@ def get_method_type_from_dynamic_manager(
...
+            )
+        )
+    return Overloaded(items) if len(items) > 1 else items[0]
 
-    assert isinstance(method_type, CallableType)
 
+def _process_dynamic_method(
+    method_name: str,
+    method_type: CallableType,
+    *,
```
flaeppe, line 98: Oh, nice, I just assumed that we still supported too old versions that we couldn't force named arguments.

I'm quite certain I've seen a couple of boolean traps in the plugin code

flaeppe (approved):
👍

mypy_django_plugin/transformers/managers.py
sobolevn: Feel free to refactor them!

flaeppe: I will hunt them down. Might even try to lint them away with ruff: https://beta.ruff.rs/docs/rules/#flake8-boolean-trap-fbt

sobolevn: Some many rules stolen from [the one true linter](https://github.com/wemake-services/wemake-python-styleguide) 😆

PR #1678, opened 2023-09-04
MultivalueDict items should return only last item

### LINKED ISSUE #1677: request.FILES items are resolved as UploadedFile | list[object] instead of UploadedFile
# Bug report

It seems that multidict items are not resolved correctly.

```
if request.FILES:
    for key, uploaded in request.FILES.items():
        reveal_type(uploaded)
```

## What's wrong

Type is resolved as:
```
UploadedFile | list[object]
```

instead of just UploadedFile.

```
    def items(self):
        """
        Yield (key, value) pairs, where value is the last item in the list
        associated with the key.
        """
        for key in self:
            yield key, self[key]
```

So if there are multi values under one key ... only last item in the list is returned.

## How is that should be

Type should be just UploadedFile.

## System information

- OS:
- `python` version: 3.11.2
- `django` version: 4.2.5
- `mypy` version: 1.5.1

matejsp:
# I have made things!

I have change the signature according to:
```
    def items(self):
        """
        Yield (key, value) pairs, where value is the last item in the list
        associated with the key.
        """
        for key in self:
            yield key, self[key]
```

## Related issues

Closes https://github.com/typeddjango/django-stubs/issues/1677


sobolevn (changes requested):
When value is empty, it can return an empty list: https://github.com/django/django/blob/6c6606aa014862f1a5c112d688d5e91c0cd9a8d8/django/utils/datastructures.py#L90

matejsp:
Ah I see ... yeah this edge case is really bad for typing :D Any idea how to get around this? or are we stuck with empty list in typing.

adamchainz:
I think we’re left with this things the way they are. It’s probably too disruptive to try change things in Django. You can adopt a pattern like:

```py
for k, v in mvd.items():
    if not isinstance(v, list):
        ...
```

sobolevn:
I don't like the status-quo either, but there's nothing django-stubs can do here :(

adamchainz:
The best we could do might be to propose a new API in Django, like `mvd.items(empty=False)`, but that would probably need fleshing out to include all methods.

PR #1683, opened 2023-09-05
Only populate a model argument for generic managers

flaeppe:
# I have made things!

Change the plugin so that non generic managers (like below) won't get an implicit model argument

```python
class MyManager(models.Manager["MyModel"]):
    ...

class MyModel(models.Model):
    objects = MyManager()
```

Previously these(`MyModel.objects`) were assigned a type like: `MyManager["MyModel"]`

### TODO:

- [ ] Ensure compatibility between a __manager__ and the __model__ it is assigned to. e.g. the model argument of a manager needs to be compatible with the model the manager is instantiated on
- [ ] Ensure compatibility between a __manager__ and __queryset__ e.g. the model argument of a manager needs to be compatible with the model argument of the queryset it's combined with

## Related issues

- Refs: https://github.com/typeddjango/django-stubs/pull/1672#discussion_r1315182802
- Refs: https://github.com/typeddjango/django-stubs/pull/1666#discussion_r1314489628

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -314,8 +314,11 @@ class AddManagers(ModelClassInitializer):
             return
 
         assert manager_info is not None
-        # Reparameterize dynamically created manager with model type
-        manager_type = Instance(manager_info, [Instance(self.model_classdef.info, [])])
+        if manager_info.is_generic():
+            # Reparameterize dynamically created manager with model type
+            manager_type = Instance(manager_info, [Instance(self.model_classdef.info, [])])
+        else:
+            manager_type = Instance(manager_info, [])
         self.add_new_node_to_model_class(manager_name, manager_type, is_classvar=True)
 
     def run_with_model_cls(self, model_cls: Type[Model]) -> None:
@@ -340,7 +343,10 @@ class AddManagers(ModelClassInitializer):
                 incomplete_manager_defs.add(manager_name)
                 continue
 
-            manager_type = Instance(manager_info, [Instance(self.model_classdef.info, [])])
+            if manager_info.is_generic():
+                manager_type = Instance(manager_info, [Instance(self.model_classdef.info, [])])
+            else:
+                manager_type = Instance(manager_info, [])
             self.add_new_node_to_model_class(manager_name, manager_type, is_classvar=True)
 
         if incomplete_manager_defs:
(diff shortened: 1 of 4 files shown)


mypy_django_plugin/transformers/models.py
```diff
@@ -340,7 +343,10 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
                 incomplete_manager_defs.add(manager_name)
                 continue
 
-            manager_type = Instance(manager_info, [Instance(self.model_classdef.info, [])])
+            if manager_info.is_generic():
```
sobolevn, line 346: Maybe refactor into helper?

sobolevn (commented):
This is more correct! Thanks!

tests/typecheck/managers/querysets/test_from_queryset.yml
```diff
@@ -675,3 +675,92 @@
...
+                  bm_generic_from_qs_generic = BaseManagerGenericFromQuerySetGeneric()
+
+                  m_populated_from_qs_populated = ManagerPopulatedFromQuerySetPopulated()
+                  m_generic_from_qs_populated = ManagerGenericFromQuerySetPopulated()
+
+                  m_populated_from_qs_generic = ManagerPopulatedFromQuerySetGeneric()
+                  m_generic_from_qs_generic = ManagerGenericFromQuerySetGeneric()
+
+                  qs_populated_as_manager = QuerySetPopulated.as_manager()
+                  qs_generic_as_manager = QuerySetGeneric["MyModel"].as_manager()
```
flaeppe, line 766: I think I found a mypy bug here, but I'm not sure. But the `get_dynamic_class_hook` isn't called for `QuerySetGeneric["MyModel"].as_manager()` due to `["MyModel"]`.

If I remove `["MyModel"]` it is triggered. But that means the hook isn't compatible with a classmethod on a generic class.

I have no idea if it's unreasonable to expect it to be triggered, it just feels like a bug that adding type parameters affects the triggering..

sobolevn: Please, report it anyways :)

sobolevn: (and tag me, since I don't watch all issues)

flaeppe: It was already discovered, some time ago. That time from django stubs as well: https://github.com/python/mypy/issues/8359

UnknownPlatypus:
Hey, just chiming in to say I really appreciate all the work you all do on the plugin code to make it more reliable. Keep up the good work, would have loved to help you a bit but it's holidays season for me currently ahah.
 
Anyway, thanks!

tests/typecheck/managers/querysets/test_from_queryset.yml
sobolevn: We can add a test-case with `xfail` for now. After `mypy@1.6.0` we will just remove `xfail` and move on.

UnknownPlatypus:
I believe this was done in #2314

PR #1689, opened 2023-09-07
fix: Fixed compiler.pui file

### LINKED ISSUE #1688: SQLCompiler.setup_query - incorrect signature
# Incorrect method signature in stubs

## What's wrong
Incorrect signature of `SQLCompiler.setup_query()` for Django 4.2.5. Signature is specified as following: `def setup_query(self) -> None: ...`

## How is that should be
Actual signature is following:
`def setup_query(self, with_col_aliases=False):`

## System information

- OS: os x 13.5.1
- `python` version: 3.11.4
- `django` version: 4.2.5
- `mypy` version: 1.5.1

ghost:
# I have made things!
I have updated the methods of the SQLCompiler class in order to avoid problems similar to those described in [Issues](https://github.com/typeddjango/django-stubs/issues/1688)

## Related issues
Closes #1688


sobolevn (approved):

sobolevn:
Thank you!

PR #1692, opened 2023-09-07
Remove "django.forms.Field.deconstruct"

brianhelba:
This method doesn't exist on upstream Django 4.2.


brianhelba:
Incorporated into #1695

PR #1693, opened 2023-09-07
Remove redefinition of inherited "deconstruct" methods.

brianhelba:
The base class `Operation` already defines a `deconstruct` method (which we have type annotations for). The child classes `AlterModelTableComment` and `AlterTogetherOptionOperation` do not need to redefine this.


sobolevn (approved):

brianhelba:
I'm not clear on the policy of django-stubs here. It seems sensible that unless child classes change the signature of an overridden method (which they're allowed to do by the Liskov substitution principle), we should try to avoid the clutter of adding type stubs of child class methods where the override only changes the behavior (which is very common). That makes it much easier to maintain the limited number of override stubs with different signatures.

brianhelba:
Since it looks like this change is acceptable, I'm going to push one more commit to this PR to remove pointless redefinitions from one other place.

brianhelba:
Ah, I'll just open another PR.

PR #1694, opened 2023-09-07
Remove redefinition of inherited "deconstruct" methods.

brianhelba:
The base class `Field` already defines a `deconstruct` method (which we have type annotations for). The child classes `BaseSpatialField` and `GeometryField` do not need to redefine this.


brianhelba:
Follows #1693

sobolevn (approved):

PR #1695, opened 2023-09-07
Annotate return value of all "deconstruct" methods

brianhelba:
This is typically some number of bool and string values, followed by the `*args` and `**kwargs` of the object being deconstructed.

[`BaseManager.deconstruct`](https://github.com/django/django/blob/4.2/django/db/models/manager.py#L42) has an even more complicated signature, where most return elements can also be `None`.

Use `Sequence` as the type of `*args`, because multiple places in Django actually return this as a `list` (even though Python always treats `*args` as a `tuple`).


sobolevn (approved):
Thanks!

PR #1699, opened 2023-09-10
Support inheriting manager classes created via `.from_queryset`

flaeppe:
# I have made things!

The changes here allows creation of managers looking like below

```python
    ...

MyManager = models.Manager.from_queryset(MyQuerySet)

class MyOtherManager(MyManager):
    ...
```

I'll be honest here and will be saying that this is a bit controversial, because we kind of enforce mypy to interpret this kind of line:

```python
MyManager = models.Manager.from_queryset(MyQuerySet)
```

As a `TypeAlias`. The only thing I have to back up the changes here, apart from that runtime wise it's actually a type alias that is created, is here: https://github.com/python/mypy/issues/8897#issuecomment-634241682

> This is generally too dynamic for mypy to understand. It would probably be quite difficult to make it work in mypy except in simple cases where all the arguments are static.

In our case it's not _too_ dynamic as our plugin can resolve the `.from_queryset` call. But I haven't found any other tool that we can work with to make the generated type inheritable.

### DIFF
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -642,8 +642,8 @@
         reveal_type(user.article_set) # N: Revealed type is "myapp.models.Article_RelatedManager"
         reveal_type(user.book_set.add)  # N: Revealed type is "def (*objs: Union[myapp.models.Book, builtins.int], *, bulk: builtins.bool =)"
         reveal_type(user.article_set.add)  # N: Revealed type is "def (*objs: Union[myapp.models.Article, builtins.int], *, bulk: builtins.bool =)"
-        reveal_type(user.book_set.filter)  # N: Revealed type is "def (*args: Any, **kwargs: Any) -> myapp.models.LibraryEntityQuerySet[myapp.models.Book]"
-        reveal_type(user.article_set.filter)  # N: Revealed type is "def (*args: Any, **kwargs: Any) -> myapp.models.LibraryEntityQuerySet[myapp.models.Article]"
+        reveal_type(user.book_set.filter)  # N: Revealed type is "def (*args: Any, **kwargs: Any) -> myapp.models.LibraryEntityQuerySet"
+        reveal_type(user.article_set.filter)  # N: Revealed type is "def (*args: Any, **kwargs: Any) -> myapp.models.LibraryEntityQuerySet"
         reveal_type(user.book_set.queryset_method())  # N: Revealed type is "builtins.int"
         reveal_type(user.article_set.queryset_method())  # N: Revealed type is "builtins.int"
     installed_apps:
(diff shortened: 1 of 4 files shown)


tests/typecheck/fields/test_related.yml
```diff
@@ -642,8 +642,8 @@
         reveal_type(user.article_set) # N: Revealed type is "myapp.models.Article_RelatedManager"
         reveal_type(user.book_set.add)  # N: Revealed type is "def (*objs: Union[myapp.models.Book, builtins.int], *, bulk: builtins.bool =)"
         reveal_type(user.article_set.add)  # N: Revealed type is "def (*objs: Union[myapp.models.Article, builtins.int], *, bulk: builtins.bool =)"
-        reveal_type(user.book_set.filter)  # N: Revealed type is "def (*args: Any, **kwargs: Any) -> myapp.models.LibraryEntityQuerySet[myapp.models.Book]"
-        reveal_type(user.article_set.filter)  # N: Revealed type is "def (*args: Any, **kwargs: Any) -> myapp.models.LibraryEntityQuerySet[myapp.models.Article]"
+        reveal_type(user.book_set.filter)  # N: Revealed type is "def (*args: Any, **kwargs: Any) -> myapp.models.LibraryEntityQuerySet"
```
sobolevn, line 645: Is it correct?

tests/typecheck/managers/querysets/test_as_manager.yml
```diff
@@ -310,7 +310,7 @@
 -   case: handles_type_collision_with_from_queryset
     main: |
         from myapp.models import MyModel, FromQuerySet
-        reveal_type(FromQuerySet)  # N: Revealed type is "def [_T <: django.db.models.base.Model] () -> myapp.models.ManagerFromModelQuerySet[_T`1]"
+        reveal_type(FromQuerySet)  # N: Revealed type is "def () -> myapp.models.ManagerFromModelQuerySet[_T`1]"
```
sobolevn, line 313: `T` will be unbound here

flaeppe: Hm, yeah, gotta look in to that. But I'm not so sure there should be any arguments here. It might be, if we implicitly copy the manager argument to the new subclass created. It's all a bit blurry, since it needs to align with the model argument of the queryset etc.

tests/typecheck/fields/test_related.yml
flaeppe: Yeah, I think this is a chain reaction of a bug that was revealed. Might be better to push it as a separate PR though.

The below change is the reason, in `mypy_django_plugin/transformers/managers.py`

```diff
- manager_type_info = manager_instance.type
+ manager_type_info = manager_instance.type.get_containing_type_info(method_name)
```

flaeppe: Right, think I figured it out. The change there takes a more correct path. There's nothing generic on the related managers, their model arguments are already populated on parent class(es). It becomes more obvious when using `.get` and seeing what model is returned. So I'll add that.

flaeppe: I've opened #1701

craigds:
Anyone know where this got to? I can't really tell from the string of PRs above :/

Current workarounds for this issue seem to be:

* use `CustomQueryset.as_manager()` instead of `Manager.from_queryset(CustomQueryset)()` where possible (i.e. when no custom Manager subclass is required)
* Add `# type: ignore[misc]` on the line that subclasses the dynamic manager type

PR #1701, opened 2023-09-11
Resolve dynamic manager methods through manager MRO

flaeppe:
This mainly affects reverse managers as we can't inherit dynamic managers in any other way.

It also renders a manual code path for method name lookup irrelevant as we can let mypy do that for us.

We can do this adjustment as reverse managers inherits the default manager on the related model. Which also includes any dynamically created managers. Reverse manager creation is done here, where bases can be seen too

https://github.com/typeddjango/django-stubs/blob/12422e64ad7a3660e92bccbaa7aef7590b3f5fe0/mypy_django_plugin/transformers/models.py#L534-L540

## Related issues

Ref: #1699

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -58,10 +58,11 @@ def get_method_type_from_dynamic_manager(
     Attempt to resolve a method on a manager that was built from '.from_queryset'
     """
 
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Yes, this seems correct. Thanks!

PR #1705, opened 2023-09-12
Fix wrong type hints for SyndicationFeed methods

WhyNotHugo:
- The correct type is actually mentioned in django's documentation in docs/ref/contrib/syndication.txt.
- The previously defined type doesn't have methods like `addQuickElement`, which are used in Django's implementation of `add_root_elements`.


WhyNotHugo:
I noticed that the exact type of the passed value is `django.utils.xmlutils.SimplerXMLGenerator`, which is actually a subtype of the documented one. I think that using the documented type is probably safer.

sobolevn:
@WhyNotHugo can you please point me to the exact place where it is documented? :)

WhyNotHugo:
I mentioned the source file where this is documented above.

The rendered version is at https://docs.djangoproject.com/en/4.2/ref/contrib/syndication/#custom-feed-generators

> SyndicationFeed.add_root_elements(self, handler)
> Callback to add elements inside the root feed element (feed/channel). handler is an > XMLGenerator from Python’s built-in SAX library; you’ll call methods on it to add to the > output 
> XML document in process.

sobolevn:
I cannot find any `addQuickElement` references in CPython at all 🤔

WhyNotHugo:
Indeed, `addQuickElement` is defined in `SimplerXMLGenerator` in `django/utils/xmlutils.py`.

My initial version of this fix used `SimplerXMLGenerator`, but I then switched it to `XMLGenerator` because the documentation mentions this last one. I guess the docs could be improved here too.

WhyNotHugo:
I've updated the PR based on this discussion.

WhyNotHugo:
v3: I've updated a few other methods in this file. No method actually uses `ContentHandler` here.

sobolevn (approved):
Thanks! Now it seems to be correct.

WhyNotHugo:
Is there any way to type-check django's codebase with these type hints? I think doing so would make this kind of inconsistency raise an error pretty quickly.

sobolevn:
Nope, it is not possible right now.

PR #1706, opened 2023-09-12
Annotate Migration fields as a ClassVar

brianhelba:
For the fields `initial` and `atomic`, this is strictly correct, as these are set on the class, not the instance.

For `operations`, `dependencies`, `run_before`, and `replaces`, [Django intentionally copies the class attributes to the instance during construction](https://github.com/django/django/blob/e15174983adf0ee5caa1b678c94bb1c6fbd40402/django/db/migrations/migration.py#L59-L62). However, annotating the attributes as either (using `|`) class or instance variables causes MyPy to issue the warning:
> Cannot override instance variable (previously declared on base class "Migration") with class variable

Since Django projects will nearly always override these as class variables, annotate them as such.

### DIFF
--- a/django-stubs/db/migrations/migration.pyi
+++ b/django-stubs/db/migrations/migration.pyi
@@ -1,4 +1,5 @@
 from collections.abc import Sequence
+from typing import ClassVar
 
 from django.db.backends.base.schema import BaseDatabaseSchemaEditor
 from django.db.migrations.operations.base import Operation
@@ -6,12 +7,14 @@ from django.db.migrations.state import ProjectState
 from typing_extensions import Self
 
 class Migration:
-    operations: Sequence[Operation]
-    dependencies: list[tuple[str, str]]
-    run_before: list[tuple[str, str]]
-    replaces: list[tuple[str, str]]
-    initial: bool | None
-    atomic: bool
+    # Django copies these 4 attributes from the class to the instance,
+    # but they're practically used as ClassVar
+    operations: ClassVar[Sequence[Operation]]
+    dependencies: ClassVar[list[tuple[str, str]]]
+    run_before: ClassVar[list[tuple[str, str]]]
+    replaces: ClassVar[list[tuple[str, str]]]
+    initial: ClassVar[bool | None]
+    atomic: ClassVar[bool]
     name: str
     app_label: str
     def __init__(self, name: str, app_label: str) -> None: ...
--- a/tests/typecheck/db/migrations/test_classvar.yml
+++ b/tests/typecheck/db/migrations/test_classvar.yml
@@ -1,0 +1,28 @@
+- case: explicit
+  main: |
+    from typing import ClassVar, List
+    from django.db import migrations
+    from django.db.migrations.operations.base import Operation
+
+    class Migration(migrations.Migration):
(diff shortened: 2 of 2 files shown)


brianhelba:
Interestingly, I originally discovered this issue in a downstream project via Ruff's [mutable-class-default (RUF012)](https://beta.ruff.rs/docs/rules/mutable-class-default/#mutable-class-default-ruf012) rule. It asked me to explicitly annotate the `Migration.operations` attributes, etc. as `ClassVar`. When I did so, I discovered that these upstream type definitions were incorrect.

sobolevn:
Can you please resolve conflicts?

brianhelba:
@sobolevn Fixed.

django-stubs/db/migrations/migration.pyi
```diff
@@ -1,17 +1,20 @@
...
 class Migration:
-    operations: Sequence[Operation]
-    dependencies: list[tuple[str, str]]
-    run_before: list[tuple[str, str]]
-    replaces: list[tuple[str, str]]
-    initial: bool | None
-    atomic: bool
+    # Django copies these 4 attributes from the class to the instance,
+    # but they're practically used as ClassVar
+    operations: ClassVar[Sequence[Operation]]
```
sobolevn, line 12: Do we have real evidence that instance variables *won't work*?

brianhelba: Yes. If you remove the `ClassVar` annotation on `operations`, the `test_classvar.yml::explicit` test case fails with:
>  error: Cannot override instance variable (previously declared on base class "Migration") with class variable (diff).

These attributes are nearly always defined by downstreams as `ClassVar`, so if we want to support the correct annotations (which tools like Ruff require), the upstream needs to match.

Of course, downstreams which do not explicitly annotate these attributes will be unaffected, as the `test_classvar.yml::implicit` test case shows.

sobolevn: Will users experience any bugs if they use these variables as instance attributes, not class variables?

A common problem that exists and that is solved by `ClassVar` is this pattern: `MyClass.my_var` - in this case we need to have `my_var: ClassVar`, otherwise it will fail with `AttributeError`

Viicos: > Will users experience any bugs if they use these variables as instance attributes, not class variables?

You are allowed to make use of class variables "as instance attributes", i.e.:

```python
class MyMigration(Migration):
    def some_method(self):
        if self.operations == ...:  # Fine
            ...
        self.operations = ...  # Error
```

If this is what you are asking

brianhelba:
@Viicos @sobolevn @zyv Looks like #2530 was more successful at getting this adopted for some reason.

Closing this.

zyv:
> @Viicos @sobolevn @zyv Looks like #2530 was more successful at getting this adopted for some reason.

Hey @brianhelba, I didn't notice your PR #1706 before submitting my own PR #2530, which was a follow-up to PR #2524. Thanks for mentioning it, though — I think your tests are valuable. I have resubmitted them as PR #2702. Hopefully they can be added as well!

brianhelba:
@zyv I'm just glad the change was made, thanks for your effort.

PR #1707, opened 2023-09-12
fix variance for migration operations

asottile:
# I have made things!

without this there are some cryptic errors when the operations are all the same type or they are defined as a tuple

### DIFF
--- a/django-stubs/db/migrations/migration.pyi
+++ b/django-stubs/db/migrations/migration.pyi
@@ -1,4 +1,4 @@
-from typing import Sequence
+from collections.abc import Sequence
 
 from django.db.backends.base.schema import BaseDatabaseSchemaEditor
 from django.db.migrations.operations.base import Operation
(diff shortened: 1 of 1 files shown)


django-stubs/db/migrations/migration.pyi
```diff
@@ -1,10 +1,12 @@
...
+
 from django.db.backends.base.schema import BaseDatabaseSchemaEditor
 from django.db.migrations.operations.base import Operation
 from django.db.migrations.state import ProjectState
 from typing_extensions import Self
 
 class Migration:
-    operations: list[Operation]
+    operations: Sequence[Operation]
     dependencies: list[tuple[str, str]]
```
sobolevn, line 10: Technically others can also be changed, because they are only used as `Iterable`. But, it is up to you.

sobolevn (approved):
Thanks!

django-stubs/db/migrations/migration.pyi
asottile: neat! I'll leave this as is for now since I didn't hit any issues with the others

brianhelba: @sobolevn @asottile I'll make another PR now to change the others.

brianhelba: See #1710

PR #1708, opened 2023-09-12
adjust bases in CreateModel to allow for mixins

asottile:
# I have made things!

we have a few migrations which create models that have additional base classes

### DIFF
--- a/django-stubs/db/migrations/operations/models.pyi
+++ b/django-stubs/db/migrations/operations/models.pyi
@@ -28,7 +28,7 @@ class CreateModel(ModelOperation):
         name: str,
         fields: list[tuple[str, Field]],
         options: dict[str, Any] | None = ...,
-        bases: Sequence[type[Model] | str] | None = ...,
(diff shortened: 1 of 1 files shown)


sobolevn:
Let me understand this a little bit better: so, you have, say, a model like `class MyModel(SomeMixin, models.Model): ...`

And this type does not allow to run a `CreateModel` opetation on it?
I suspect that this might be a mypy bug 🤔

sobolevn:
`MyModel` still should be `type[Model]`

asottile:
> Let me understand this a little bit better: so, you have, say, a model like `class MyModel(SomeMixin, models.Model): ...`
> 
> And this type does not allow to run a `CreateModel` opetation on it? I suspect that this might be a mypy bug 🤔

for example https://github.com/getsentry/sentry/pull/56087/files#diff-a3e504055395e3cb856d99f4bc647fd048e61a1d75fa75fcc601877f62b88903R2283

sobolevn (approved):
Oh, I see 👍

PR #1710, opened 2023-09-13
Fix variance for all Migration list attributes

brianhelba:


brianhelba:
This completes the changes started in #1707.

~This depends on #1706, so that should be merged first.~

brianhelba:
@sobolevn This is ready for review. No need to block this on #1706.

adamchainz (approved):
LGTM

PR #1714, opened 2023-09-15
Fix _Composable protocol for compatibility with psycopg2-stubs change

andersk:
# I have made things!

An implementor of `_Composable` is required to support `+` with its own type, not with an arbitrary other implementor of `_Composable`. This is now required for compatibility with
https://github.com/python/typeshed/pull/10630.

## Related issues

- Refs #1029


sobolevn (approved):
Thanks!

PR #1715, opened 2023-09-16
fix: missing max_length attribute on forms.FileField

GabDug:
# I have made things!

Add `max_length` to `forms.FileField`

Official docs: https://docs.djangoproject.com/en/4.2/ref/forms/fields/#filefield


sobolevn (approved):
Thanks!

PR #1716, opened 2023-09-17
Type django.views

GabDug:
# I have made things!

- Adds missing Warning classes
- Add missing methods
- Made `DatedItems` private
- Allow None for somme attributes
  - `template_name`, if we only set the type as str, users might wrongfully access it, and type checking would not warn them that it can be None.
  - `paginate_by` can be None per `get_paginate_by()` documentation
  - `packages` defaults to and supports None according to `JavaScriptCatalog.get()` implementation

All properties / cached properties typed as `_Getter[X] | X` to allow `@property`/`@cached_property` or direct value assignment.


sobolevn (approved):
Thanks!

PR #1717, opened 2023-09-17
chore: add some missing generic types

GabDug:
# I have made things!

- Add some missing generic types
- Correct `SyndicationFeed .categories` that defaults to `Tuple` but accepts `Sequence`


sobolevn (approved):
Thanks!

PR #1718, opened 2023-09-17
chore: Migrate from flake8 to ruff

GabDug:
# I have made things!

Ruff now includes `flake8-pyi` rules has well as the `flake8-no-pep420` rule.

To simplify the dev experience, I propose to use ruff to replace these flake8 rules, thus completely removing the need for flake8.

Migration notes:
- `PYI037` is not implemented but `UP007` checks for `Optional` usage instead of union type
- `PYI022` is equivalent to `UP035`


CONTRIBUTING.md
```diff
@@ -55,7 +55,7 @@ pre-commit install
 
 ### Testing and Linting
 
-We use `mypy`, `pytest`, `flake8`, and `black` for quality control. All tools except pytest are executed using pre-commit when you make a commit.
+We use `mypy`, `pytest`, `ruff`, and `black` for quality control. All tools except pytest are executed using pre-commit when you make a commit.
```
flaeppe: Can we update to say only ruff and black does?

```suggestion
We use `mypy`, `pytest`, `ruff`, and `black` for quality control. `ruff` and `black` are executed using pre-commit when you make a commit.
```

django-stubs/contrib/admin/options.pyi
```diff
@@ -1,5 +1,5 @@
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
-from typing import Any, Generic, Literal, Optional, TypeVar  # noqa: Y037  # https://github.com/python/mypy/issues/12211
+from typing import Any, Generic, Literal, Optional, TypeVar  # https://github.com/python/mypy/issues/12211
```
flaeppe:
```suggestion
from typing import Any, Generic, Literal, Optional, TypeVar
```

I don't think any link is relevant here?

django-stubs/db/models/options.pyi
```diff
@@ -1,11 +1,5 @@
 from collections.abc import Iterable, Sequence
-from typing import (  # noqa: Y037   # https://github.com/python/mypy/issues/12211
-    Any,
-    Generic,
-    Literal,
-    TypeVar,
-    overload,
-)
+from typing import Any, Generic, Literal, TypeVar, overload  # https://github.com/python/mypy/issues/12211
```
flaeppe:
```suggestion
from typing import Any, Generic, Literal, TypeVar, overload
```

django-stubs/forms/models.pyi
```diff
@@ -1,12 +1,5 @@
 from collections.abc import Callable, Collection, Container, Iterator, Mapping, Sequence
-from typing import (  # noqa: Y037  # https://github.com/python/mypy/issues/12211
-    Any,
-    ClassVar,
-    Generic,
-    Literal,
-    TypeVar,
-    overload,
-)
+from typing import Any, ClassVar, Generic, Literal, TypeVar, overload  # https://github.com/python/mypy/issues/12211
```
flaeppe:
```suggestion
from typing import Any, ClassVar, Generic, Literal, TypeVar, overload
```

django-stubs/template/response.pyi
```diff
@@ -1,7 +1,7 @@
 import functools
 from collections.abc import Callable, Iterator, Sequence
 from http.cookies import SimpleCookie
-from typing import Any  # noqa: Y037   # https://github.com/python/mypy/issues/12211
+from typing import Any  # https://github.com/python/mypy/issues/12211
```
flaeppe:
```suggestion
from typing import Any
```

pyproject.toml
```diff
@@ -17,11 +18,35 @@ line-length = 120
...
 ]
+ignore = ["PYI021", "PYI024", "PYI041", "PYI043"]
+
+[tool.ruff.per-file-ignores]
+"*.pyi" = [
+    "B",
+    "E501",
+    "E741",
+    "E743",
+    "F401", # Unused import
```
flaeppe: I'm not sure, but I don't think we should completely ignore `F401`.

As that wouldn't clear up places where there's no re-export and an import becomes unused.

django-stubs/contrib/admin/options.pyi
GabDug: Removed all the links to this issue, and `Optional` imports

intgr:
Last time I checked, Ruff `PYI*` group was still missing a few checks that `flake8-pyi` has.

Can you please compile a list of relevant checks that Ruff is missing? Then we can make an informed decision on whether we want to drop `flake8-pyi`.

GabDug:
Ruff implemented [PYI rules](https://docs.astral.sh/ruff/rules/#flake8-pyi-pyi).
flake8-pyi implemented [Y rules](https://github.com/PyCQA/flake8-pyi/blob/main/ERRORCODES.md).

| Missing rule | Description | Alternative |
| ------------- | ------------- |------------ |
| PYI022 | Do not use `typing` if available in stdlib  (e.g. prefer builtins.list over typing.List, collections.Counter over typing.Counter, etc.). | [UP035](https://docs.astral.sh/ruff/rules/deprecated-import/#why-is-this-bad) |
| PYI023 | Where there is no detriment to backwards compatibility, import objects such as ClassVar and NoReturn from typing rather than typing_extensions. | [UP035](https://docs.astral.sh/ruff/rules/deprecated-import/#why-is-this-bad) |
| PYI027 | Removed from flake8-pyi | PYI022, thus UP035 |
| PYI028 | Use class-based `NamedTuple` instead of assignment-based | [UP014](https://docs.astral.sh/ruff/rules/convert-named-tuple-functional-to-class/#example) |
| PYI031 | TypedDicts should use class-based syntax instead of assignment-based syntax wherever possible. (In situations where this is not possible, such as if a field is a Python keyword or an invalid identifier, this error will not be emitted.)  | [UP014](https://docs.astral.sh/ruff/rules/convert-named-tuple-functional-to-class/#example) |
| PYI037 | Use PEP 604 syntax instead of typing.Union and typing.Optional. E.g. use `str \| int` instead of Union[str, int], and use `str \| None` instead of Optional[str]. | [UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation/)
| PYI038 | Use from collections.abc import Set as AbstractSet instead of from typing import AbstractSet. | Not sure, may be covered in [UP035](https://docs.astral.sh/ruff/rules/deprecated-import/#why-is-this-bad)  |
| PYI039 | Use str instead of typing.Text. | [UP019](https://docs.astral.sh/ruff/rules/typing-text-str-alias/#why-is-this-bad) |
| PYI040 | Never explicitly inherit from object, as all classes implicitly inherit from object in Python 3. | [UP004](https://docs.astral.sh/ruff/rules/useless-object-inheritance/) |
| PYI057 | Do not use typing.ByteString or collections.abc.ByteString. These types have unclear semantics, and are deprecated; use typing_extensions.Buffer or a union such as `bytes \| bytearray \| memoryview` instead. See [PEP 688](https://peps.python.org/pep-0688/) for more details. | None, recently added rule |

We may need to tweak target-version and/or ignores, as UP may not be enabled with a 3.8 target.

intgr:
Thanks!

> We may need to tweak target-version and/or ignores, as UP may not be enabled with a 3.8 target.

Let's not do that. Bumping `target-version` can cause real compatibility issues in `.py` files like the mypy plugin or django_stubs_ext, which are executed in the Python runtime, and we still want to support Python 3.8.

While many of the missing `flake8-pyi` rules do exist as `UP*` rules in Ruff, `flake8-pyi` is more eager in adopting updates because type checkers support newer features than the Python runtime. It's still a bit unclear how much we lose from to that. Most common ones are already covered by the `target-version = "py38"` level, so it's not a big loss.

---

Still, I don't see a downside to keeping `flake8-pyi` in parallel with Ruff either. Typeshed project uses both as well. Any third opinions?

intgr (approved):
No-one has voiced any opinions or concerns. Let's keep this as proposed and drop `flake8-pyi` then. Worst case I see is that we'll be using some imports of compatibility aliases, no big deal.

Issue #1638, opened 2023-08-02
ManyToManyField is typed as `RelatedManager[<model>] | None` rather than just `RelatedManager[<model>]`


PureTryOut:
## What's wrong

I have the following model:

```
class Customer(User):
    addresses = models.ManyToManyField(
        Address,
        related_name="customers",
        null=True,
        blank=True,
    )
```

`addresses` here is typed as `RelatedManager[<model>] | None`. meaning any queryset call on it will make mypy complain about the attribute not existing. For example:

```
project/views/address.py:68: error: Item "None" of "RelatedManager[Address] | None" has no attribute "add"  [union-attr]
project/views/address.py:93: error: Item "None" of "RelatedManager[Address] | None" has no attribute "get"  [union-attr]
project/views/address.py:97: error: Item "None" of "RelatedManager[Address] | None" has no attribute "remove"  [union-attr]
```

However, it never is `None`, and I'm not sure how it could be either. If there are no addresses for a customer than `customer.addresses.exists()` would just return false rather than `customer.addresses` being `None`.

## How is that should be

It should be typed as just `RelatedManager[<model>]`.

## System information

- OS: Linux
- `python` version: 3.11
- `django` version: 4.2
- `mypy` version: 1.4.1
- `django-stubs` version: 4.2.3
- `django-stubs-ext` version: not installed

sobolevn:
Does `null=True` or `blank=True` affect `| None` part?

andersk:
The `ManyToManyField` documentation [says](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.ManyToManyField.swappable)

> [`null`](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.Field.null) has no effect since there is no way to require a relationship at the database level.

and in fact the system check framework raises the [`fields.W340` warning](https://docs.djangoproject.com/en/4.2/ref/checks/#related-fields) for this. django-stubs should ignore `null` here too, but meanwhile you can just remove it.

Issue #1454, opened 2023-04-26
Add ManyRelatedManager and distinguish from RelatedManager


intgr:
Methods of `RelatedManager` and `ManyRelatedManager` have different signatures.

I noticed while reviewing #1443.

PR #1719, opened 2023-09-17
Add better support for `ManyToManyField`'s `through` model

### LINKED ISSUE #883: Error typechecking ManyToManyField manager `request.user.groups.all()`
# Bug report

## What's wrong

If I use `request.user.groups.all()` in a view (request being an instance of `django.http.HttpRequest`), this runs fine but mypy complains with

`error: Item "ManyToManyField[Sequence[Group], RelatedManager[Group]]" of "Union[ManyToManyField[Sequence[Group], RelatedManager[Group]], EmptyManager]" has no attribute "all"`

## How is that should be

Typechecking should pass without errors.

## System information

- OS: Ubuntu 20.04
- `python` version: 3.9.10
- `django` version: 4.0.2
- `mypy` version: 0.910

### LINKED ISSUE #1549: Many-to-Many generated through model are not properly typed
# Bug report

The generated through model for many-to-many is not getting typed properly.

## What's wrong

In the following example:
```python
from django.db import models


class Topping(models.Model):
    label = models.CharField(max_length=100)


class Pizza(models.Model):
    toppings = models.ManyToManyField(Topping)


for toppings_through in Pizza.toppings.through.objects.all():
    reveal_type(toppings_through)
    reveal_type(toppings_through.pizza)
    reveal_type(toppings_through.topping)
```

 I would expect the revealed types to be:
 * Not sure, but at least `Model`
 * `Pizza`
 * `Topping`

but the result is:
```
demo.py:14: note: Revealed type is "Any"
demo.py:15: note: Revealed type is "Any"
demo.py:16: note: Revealed type is "Any"
```

`reveal_type(Pizza.toppings.through)` gives `Type[django.db.models.base.Model]`, so I’m assuming that’s where we could be more precise.

## System information

- OS:
- `python` version: 3.11.1
- `django` version: 4.2.2
- `mypy` version: 1.3.0

flaeppe:
# I have made things!

I've modified how `ManyToManyField` is generic and also made the `ManyToManyDescriptor` generic. This was to get a better description of the `ManyToManyField`, which requires less plugin interference and results in more accurate types.

The most important thing to notice here is a call without any explicit `through` argument makes the `ManyToManyField` instance only partially declared. This is where the plugin steps in and will detect the call argument is missing and generate a through model definition that aligns with what Django describes under the `ManyToManyField.through` section in their documentation. Here: https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.ManyToManyField.through.

`ManyToManyField.through` should now be properly resolved to either an implicitly generated through model, or an explicitly passed one. For example, expanding a generated through model looks like below:

```python
class MyModel_m2mfield(models.Model):
    id: <DefaultAutoField>
    mymodel: ForeignKey(MyModel)
    mymodel_id: MyModel.pk
    othermodel: ForeignKey(OtherModel)
    othermodel_id: OtherModel.pk

    objects: models.Manager["MyModel_m2mfield"]
```

`ManyToManyField()` calls with lazy references (e.g. `"myapp.MyModel"`) should also be supported, `"self"` too.

## Related issues

Closes: #1549 
Closes: #883
Refs: #1638
Refs: #1454 

---

An additional note here is that I'm getting a feeling that the there's a similar adjustment to be done for e.g. `ForeignKey` and `OneToOneField` and their descriptors, which could improve their definition in general, while requiring less plugin code.


django-stubs/db/models/fields/related.pyi
```diff
@@ -204,10 +205,9 @@ class OneToOneField(ForeignKey[_ST, _GT]):
     @overload
     def __get__(self, instance: Any, owner: Any) -> Self: ...
 
-class ManyToManyField(RelatedField[_ST, _GT]):
-    _pyi_private_set_type: Sequence[Any]
-    _pyi_private_get_type: RelatedManager[Any]
+_To = TypeVar("_To", bound=Model)
 
+class ManyToManyField(RelatedField[Any, Any], Generic[_To, _M]):
```
sobolevn, line 210: Can we make it generic on all the things and simplify our plugin a bit?

mypy_django_plugin/lib/helpers.py
```diff
@@ -385,3 +389,34 @@ def add_new_manager_base(api: SemanticAnalyzerPluginInterface, fullname: str) ->
     if sym is not None and isinstance(sym.node, TypeInfo):
         bases = get_django_metadata_bases(sym.node, "manager_bases")
         bases[fullname] = 1
+
+
+def is_abstract_model(model: TypeInfo, /) -> bool:
```
```suggestion
def is_abstract_model(model: TypeInfo) -> bool:
```

No reason to make it pos-only

mypy_django_plugin/transformers/fields.py
```diff
@@ -201,6 +205,155 @@ def drop_combinable(_type: MypyType) -> Optional[MypyType]:
...
+
+
+class M2MArguments(NamedTuple):
+    to: M2MTo
+    through: Optional[M2MThrough]
+
+
+def get_model_info_from_call_arg(
+    call_arg: CallArg,
+    /,
```
sobolevn:
```suggestion
```

mypy_django_plugin/transformers/models.py
```diff
@@ -597,6 +609,182 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+            through_model = get_model_info_from_call_arg(
+                through_arg,
+                current_module_name=self.model_classdef.info.module_name,
+                api=self.api,
+                django_context=self.django_context,
+            )
+            if through_model is not None:
+                through = M2MThrough(arg=through_arg, model=through_model)
+
+        return M2MArguments(to=to, through=through)
```
sobolevn, line 809: This part of the change seems like the most complex one to me.

tests/typecheck/fields/test_related.yml
```diff
@@ -278,9 +278,9 @@
 -   case: many_to_many_field_converts_to_queryset_of_model_type
     main: |
         from myapp.models import App, Member
-        reveal_type(Member().apps)  # N: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.App]"
-        reveal_type(App().members)  # N: Revealed type is "django.db.models.manager.RelatedManager[myapp.models.Member]"
-        reveal_type(Member.apps.field)  # N: Revealed type is "django.db.models.fields.related.ManyToManyField[typing.Sequence[myapp.models.App], django.db.models.manager.RelatedManager[myapp.models.App]]"
+        reveal_type(Member().apps.get())  # N: Revealed type is "myapp.models.App"
```
sobolevn, line 282: Let's have both assertions: with `Member().app` and with `.get()`

tests/typecheck/fields/test_related.yml
```diff
@@ -600,7 +608,7 @@
                     publisher = models.ForeignKey(Publisher, on_delete=models.CASCADE)
                 class Author(models.Model):
                     blogs = models.ManyToManyField(Blog)
-                    file = models.FileField()
+                    file = FileField()
```
sobolevn: Why is it changed?

sobolevn (commented):
This is not a full review, just some questions.

Thanks for your work!

tests/typecheck/fields/test_related.yml
flaeppe: I was fiddling around with some stuff in the beginning. This should be reverted.

django-stubs/db/models/fields/related.pyi
flaeppe: Which parts are you referring to?

The plugin only acts on what is `_T` and `_M` here, essentially the `to=` and `through=` argument on `ManyToManyField`.

It does however have to sort some stuff out when there was no explicit `through=` argument passed, as Django automatically generates a through model behind the scenes in that case.

Additionally it supports resolving lazy references e.g. `"myapp.MyModel"` and replaces either `_T` or `_M` with any found model type.

mypy_django_plugin/transformers/fields.py
```diff
@@ -201,6 +205,155 @@ def drop_combinable(_type: MypyType) -> Optional[MypyType]:
...
+                if (
+                    isinstance(defn, AssignmentStmt)
+                    and defn.rvalue is ctx.context
+                    and len(defn.lvalues) == 1
+                    and isinstance(defn.lvalues[0], NameExpr)
+                ):
+                    field_name = defn.lvalues[0].name
+                    break
+
+            if field_name is not None:
```
```suggestion
            else:
```

__Edit:__ Nope, nvm, that's the opposite

mypy_django_plugin/transformers/models.py
flaeppe: Yeah, it's messy. All it does is sort out how `ManyToManyField()` was called;

- Did it have `to=` and/or `through=`?
- Resolve any lazy reference for `to=` or `through=`
- Include the model (typeinfo) that `to=` and/or `through=` pointed to

I tried to encapsulate these results in the named tuple `M2MArguments`, and also to get some form of declarative description of how we can interpret a call to `ManyToManyField`.

e.g. required `to` but optional `through`, while `through` can also be an implicit model:

```python
M2MArguments(..., through=M2MThrough(call_arg=None, model=...))
```

django-stubs/db/models/fields/related.pyi
flaeppe: > Which parts are you referring to?

What I meant to say here was: which other parts should we make generic to simplify the plugin?

sobolevn: `_GT, _ST`? Will it help?

flaeppe: No, this changes so that those arguments aren't used at all any more, for `ManyToManyField`. It's just the parent class that wants them.

The generics of the descriptor class does almost all the work now instead (except when `through` is implicit, then the plugin does that part)

flaeppe: It's a bit hard to follow, but instead of using `_GT` and `_ST` and let our plugin do _all_ the work, I've updated the descriptor class `ManyToManyDescriptor` to become generic over a model, and that will be the `through=` model for a `ManyToManyField()`.

And if you now look a bit further, you find: `ManyToManyDescriptor.through -> type[_M]` which means that the `through=` argument align with the `MyModel.m2m_field.through` without any plugin work (again, except for special cases of lazy referencing or not passing the `through=` argument, which is all the new plugin code you see in this PR)

flaeppe: Corresponding part is the `to=` argument, plugin code will _only_ be running for lazy referencing. Previously this was completely managed by the plugin, now it's instead mypy doing the work.

mypy_django_plugin/lib/fullnames.py
```diff
@@ -37,7 +37,6 @@
         FOREIGN_OBJECT_FULLNAME,
         FOREIGN_KEY_FULLNAME,
         ONETOONE_FIELD_FULLNAME,
-        MANYTOMANY_FIELD_FULLNAME,
```
flaeppe, line 41: For extra clarification: the removal of this line detaches `ManyToManyField` to be handled by the `_GT` and `_ST` generic arguments and the attributes `_pyi_private_set_type` and `_pyi_private_get_type`

mypy_django_plugin/transformers/fields.py
```diff
@@ -201,6 +205,155 @@ def drop_combinable(_type: MypyType) -> Optional[MypyType]:
...
+                through_model_fullname = m2m_throughs.get(field_name)
+                if through_model_fullname is not None:
+                    through_model = helpers.lookup_fully_qualified_typeinfo(checker, through_model_fullname)
+                    if through_model is not None:
+                        through = M2MThrough(arg=through_arg, model=through_model)
+
+    return M2MArguments(to=to, through=through)
+
+
+def fill_model_args_for_many_to_many_field(
```
flaeppe: Might be worth moving this to some `transformers/manytomany.py`, if possible

mypy_django_plugin/transformers/fields.py
```diff
@@ -201,6 +205,156 @@ def drop_combinable(_type: MypyType) -> Optional[MypyType]:
...
+    """
+    Attempts to resolve an expression to a 'TypeInfo' instance. Any lazy reference
+    argument(e.g. "<app_label>.<object_name>") to a Django model is also attempted.
+    """
+    # TODO: Handle settings.AUTH_USER_MODEL?
+    if isinstance(expr, NameExpr) and isinstance(expr.node, TypeInfo):
+        if (
+            expr.node.metaclass_type is not None
+            and expr.node.metaclass_type.type.fullname == fullnames.MODEL_METACLASS_FULLNAME
+        ):
```
flaeppe: I suppose this could be broken out to a type guard helper, if we want?

```python
ModelTypeInfo = NewType("ModelTypeInfo", TypeInfo)
def is_model_info(info: TypeInfo) -> TypeGuard[ModelTypeInfo]:
    return (
        info.metaclass_type is not None
        and info.metaclass_type.type.fullname == fullnames.MODEL_METACLASS_FULLNAME
    )
```

mypy_django_plugin/transformers/fields.py
```diff
@@ -201,6 +205,156 @@ def drop_combinable(_type: MypyType) -> Optional[MypyType]:
...
+    expr: Expression,
+    *,
+    api: Union[TypeChecker, SemanticAnalyzer],
+    django_context: DjangoContext,
+) -> Optional[ProperType]:
+    """
+    Attempts to resolve an expression to a 'TypeInfo' instance. Any lazy reference
+    argument(e.g. "<app_label>.<object_name>") to a Django model is also attempted.
+    """
+    # TODO: Handle settings.AUTH_USER_MODEL?
```
flaeppe: To expand a bit more here; I have an idea that we could also attempt to resolve a `MemberExpr` pointing towards `settings.AUTH_USER_MODEL`. As such it's possible to pick up its lazy reference, something like

```python
if isinstance(expr, MemberExpr) and ...:
    label = django_context.settings.AUTH_USER_MODEL
    # Below here is same as when `isinstance(expr, StrExpr)`
    fullname = django_context.model_class_fullnames_by_label.get(label)
        if fullname is not None:
            model_info = helpers.lookup_fully_qualified_typeinfo(api, fullname)
            if model_info is not None:
                return Instance(model_info, [])
            elif isinstance(api, SemanticAnalyzer) and not api.final_iteration:
                # Getting this far, where Django matched the reference but we still can't
                # find it, we want to defer
                api.defer()
```

Like so, there could be another helper:

```python
def resolve_lazy_reference(expr: Expression) -> Optional[TypeInfo]:
    ...
```

Which focuses on converting any `StrExpr("<app_label>"."<object_name>")` to a `TypeInfo`. This could probably be reused in multiple other places e.g. `Apps.get_model(...)` and `to=` argument for `ForeignKey`, `OneToOneField`, etc.

mypy_django_plugin/transformers/fields.py
```diff
@@ -201,6 +205,156 @@ def drop_combinable(_type: MypyType) -> Optional[MypyType]:
...
+            api.fail("Could not resolve lazy reference without an app label", expr)
+            api.note(
+                ("Try to use a reference explicitly prefixed with app label:" f' "<app_label>.{expr.value}" instead'),
+                expr,
+            )
+            return None
+
+        # Expression conforms to the structure of a lazy reference: '<app_label>.<object_name>'
+        fullname = django_context.model_class_fullnames_by_label.get(expr.value)
+        if fullname is not None:
```
flaeppe: Hm, I think we could yield an error on the `else` clause here, saying that the lazy reference is invalid and doesn't point to a model

flaeppe:
@sobolevn I know that this is a large addition. I wanted to have it more in pieces but wasn't really able to, I pulled in one end and some stuff started to follow along.

Anyways, I've run this on one of our larger projects, hit a few issues that I resolved and it's now green except 1 part, namely:

```python
users = ManyToManyField(settings.AUTH_USER_MODEL)
```

But I wanted to keep that as a separate change to avoid piling up even more in here.

I'm asking for an opinion here, how could we proceed to make it mergable?

Additionally, @ljodal, I know you've previously had possibility to run changes on projects generating a bit more use cases, could you perhaps try this one out? See if we can touch a bit wider surface? Perhaps there's some cheap fixes to be found before anything is merged.

All in all this PR should unlock some `ManyToManyField`-box, as we haven't really had any good support for it previously. No idea what'll come out of it though. Just hoping that it's not gonna generate too many crashes..

ljodal:
I'll give it a go, we haven't updated django-stubs in a while though, so not sure how much unrelated will pop up

flaeppe:
Additionally I want to emphasize that what I've done here is actually to use __more__ of `mypy` and less plugin.

Starting to use the descriptor class is the trick, it does __half__ of the work when having an implicit `through=` table:

```python
others = ManyToManyField(Other)
```

And it does __all__ the work when having an explicit `through=` table:

```python
others = ManyToManyField(Other, through=InBetween)
```

- Then there's a bunch of plugin code helping out with resolving lazy references: `"app_label.Model"`. But all it does is replace those expressions with model instances, so that `mypy` can see it the same way that I've written it above (☝️).

- The exception is the implicit through table case, where the plugin builds and inserts that class, similar to what Django does. That code looks a bit intimidating but is nothing to be afraid of. It's flat out writing a model class declaration in "mypy code"..

mypy_django_plugin/lib/helpers.py
```diff
@@ -385,3 +390,61 @@ def add_new_manager_base(api: SemanticAnalyzerPluginInterface, fullname: str) ->
...
+    reference: str, *, api: Union[TypeChecker, SemanticAnalyzer], django_context: "DjangoContext", ctx: Context
+) -> Optional[TypeInfo]:
+    """
+    Attempts to resolve a lazy reference(e.g. "<app_label>.<object_name>") to a
+    'TypeInfo' instance.
+    """
+    if "." not in reference:
+        # <object_name> -- needs prefix of <app_label>. We can't implicitly solve
+        # what app label this should be, yet.
+        return None
```
ljodal, line 436: We have a case of this at least (the referenced model is defined right below in the file), which now causes "Needs type annotation" errors

mypy_django_plugin/transformers/manytomany.py
```diff
@@ -0,0 +1,144 @@
...
+    expr: Expression,
+    *,
+    api: Union[TypeChecker, SemanticAnalyzer],
+    django_context: DjangoContext,
+) -> Optional[ProperType]:
+    """
+    Attempts to resolve an expression to a 'TypeInfo' instance. Any lazy reference
+    argument(e.g. "<app_label>.<object_name>") to a Django model is also attempted.
+    """
+    # TODO: Handle settings.AUTH_USER_MODEL?
```
ljodal, line 132: We have a bunch of these cases as well

ljodal (commented):
Haven't looked too closely at the code, but it seems to run fine on our largest project. A couple of new warnings due to the limitations mentioned in separate comments, but nothing major

mypy_django_plugin/transformers/manytomany.py
flaeppe: Something tells me this one will hit quite many projects.

flaeppe:
> Haven't looked too closely at the code, but it seems to run fine on our largest project. A couple of new warnings due to the limitations mentioned in separate comments, but nothing major

Thank you for the help! Great to get more input

mypy_django_plugin/transformers/manytomany.py
ljodal: Guess it's fine to do in a separate PR, but maybe make sure to do it before the next release?

flaeppe: Yeah, that's what I'm thinking too

sobolevn (approved):
Thanks! But, merging this PR would block the release. Can you please open a new issue with new release planning, so we won't forget about it?

flaeppe:
> Thanks! But, merging this PR would block the release. Can you please open a new issue with new release planning, so we won't forget about it?

I've opened a next release issue: #1740

intgr:
Is this a known issue? When passing argument `through` to `ManyToManyField`: `through=CustomThrough` works, `through="myapp.CustomThrough"` works.

But `through="CustomThrough"` breaks. It should work since the model is defined in the same file. Test patch:

```diff
diff --git a/tests/typecheck/fields/test_related.yml b/tests/typecheck/fields/test_related.yml
index 8f21557e..3ac3117c 100644
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -1133,7 +1133,7 @@
                     auto_through = models.ManyToManyField(Other, related_name="autos")
                     # Have multiple M2Ms with implicit through
                     other_again = models.ManyToManyField(Other, related_name="others_again")
-                    custom_through = models.ManyToManyField(Other, through=CustomThrough, related_name="customs")
+                    custom_through = models.ManyToManyField(Other, through="CustomThrough", related_name="customs")
 
 -   case: test_many_to_many_with_lazy_references
     main: |
```

Output:

```
E   Actual:
E     ...
E     main:6: note: Revealed type is "Any"          (diff)
E    main:7: note: Revealed type is "myapp.models.Other"
E     main:8: note: Revealed type is "Any"          (diff)
```

ljodal:
Yes, see https://github.com/typeddjango/django-stubs/pull/1719#discussion_r1336317602 @intgr :)

flaeppe:
Yes, opting out of the lazy reference without "app_label" was intentional, I also implemented the lazy reference resolver for another PR, but had it erroring, and there was a small discussion regarding that:

https://github.com/typeddjango/django-stubs/pull/1725#discussion_r1332599765

Basically it's non-trivial to figure out the correct implicit app label. So I wanted to push it for later. There might perhaps be shortcuts to take when using the runtime model, but I also wanted to minimise touching the runtime.

It's easy to resolve it at least, as it's just to explicitly add the correct app label.

PR #1722, opened 2023-09-18
Update layer.pyi

niconoe:
Added missing `num_feat` property in the Layer class

This property is described [here in the doc](https://docs.djangoproject.com/en/4.2/ref/contrib/gis/gdal/#django.contrib.gis.gdal.Layer.num_feat) and I confirmed it's actually present at runtime (that's how I spotted it - my code worked but mypy complained).

Hope this helps!

Nico


sobolevn (approved):
Thanks!

Source: https://github.com/django/django/blob/190874eadd0c6dcaae0c244cc47e838cf0faf24d/django/contrib/gis/gdal/layer.py#L111-L114

sobolevn:
Please, remove `django.contrib.gis.gdal.layer.Layer.num_feat` from `scripts/stubtest/allowlist_generated.txt`

sobolevn (approved):

PR #1724, opened 2023-09-19
Improve `Field.formfield()`, `GeometryField.formfield()` methods

brianhelba:


brianhelba:
@sobolevn Ready for review.

intgr:
Thanks! This was already partially added in #1739, I have merged your changes.

django-stubs/contrib/gis/db/models/fields.pyi
```diff
@@ -108,7 +108,9 @@ class GeometryField(BaseSpatialField[_ST, _GT]):
         validators: Iterable[_ValidatorCallable] = ...,
         error_messages: _ErrorMessagesMapping | None = ...,
     ) -> None: ...
-    def formfield(self, **kwargs: Any) -> Any: ...  # type: ignore[override]
+    def formfield(  # type: ignore[override]
+        self, form_class: type[forms.GeometryField] | None = ..., geom_type: str = ..., srid: Any = ..., **kwargs: Any
```
intgr: Since these are handled from `**kwargs`, you should typehint them as keyword-only parameters. That should get rid of the allowlist ignores too.

```suggestion
        self, *, form_class: type[GeometryField] | None = ..., geom_type: str = ..., srid: Any = ..., **kwargs: Any
```

django-stubs/contrib/gis/db/models/fields.pyi
```diff
@@ -108,7 +108,9 @@ class GeometryField(BaseSpatialField[_ST, _GT]):
...
+    ) -> forms.GeometryField: ...
```
intgr: No need to qualify these types, they are defined in this very file.

```suggestion
        self, form_class: type[GeometryField] | None = ..., geom_type: str = ..., srid: Any = ..., **kwargs: Any
    ) -> GeometryField: ...
```

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -206,9 +206,9 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
     def formfield(
         self,
         form_class: type[forms.Field] | None = ...,
```
intgr:
```suggestion
        form_class: type[Field] | None = ...,
```

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -206,9 +206,9 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
...
-        choices_form_class: Any | None = ...,
+        choices_form_class: type[forms.ChoiceField] | None = ...,
         **kwargs: Any,
-    ) -> Field: ...
+    ) -> forms.Field: ...
```
intgr:
```suggestion
    ) -> Field: ...
```

scripts/stubtest/allowlist.txt
```diff
@@ -14,6 +14,10 @@ django.contrib.auth.migrations.*
 django.contrib.flatpages.migrations.*
 django.contrib.contenttypes.migrations.*
 
+# Annotate optional parameters which Django accepts as **kwargs
+django.contrib.gis.db.models.GeometryField.formfield
+django.contrib.gis.db.models.fields.GeometryField.formfield
+
```
intgr: With the addition of `*,` to `GeometryField.formfield(...)`, these are no longer needed.

```suggestion
```

django-stubs/contrib/gis/db/models/fields.pyi
intgr:
```suggestion
    ) -> GeometryField: ...
```

django-stubs/db/models/fields/__init__.pyi
brianhelba: @intgr No, these are different fields!

`forms.Field` is a form field. The local `Field` is a model field.

This change now breaks the annotation correctness. I'll open a supplemental PR to fix it now.

django-stubs/contrib/gis/db/models/fields.pyi
brianhelba: @intgr Likewise, these are not the same types.

`forms.GeometryField` is a form field. The local `GeometryField` is a model field.

django-stubs/db/models/fields/__init__.pyi
intgr: Oops, sorry for messing that up. And thanks for responding so quickly.

Sadly this screw-up is now part of the django-stubs 4.2.5 release. I'll probably wait a few days to see if any other bugs turn up, and make another hotfix release.

django-stubs/contrib/gis/db/models/fields.pyi
brianhelba: @intgr Technically, these are *not* keyword-only arguments in the upstream `Field.formfield` method: https://github.com/django/django/blob/stable/4.2.x/django/db/models/fields/__init__.py#L1037

However, since the subclasses (`CharField`, etc.) only accept keyword arguments and we don't want to reimplement this for every subclass, I think it's fine to be slightly more strict than Django might actually allow. 👍

It'll probably be safer for users to be required to use keyword arguments too.

intgr: The `*` keyword-only change only affects the `GeometryField` method, not `Field`. So it should be accurate AFAICT. (https://github.com/django/django/blob/stable/4.2.x/django/contrib/gis/db/models/fields.py#L303)

PR #1725, opened 2023-09-21
Add plugin support for `apps.get_model()`

### LINKED ISSUE #304: Too broad return value for apps.get_model()

flaeppe:
Attempts to resolve the provided lazy reference to a model type. As such it can now return a specific model depending on what it was called with

## Related issues

Closes: #1318 
Closes: #304 
Closes: #305 
Closes: #814 

High value comment/discussion: https://github.com/typeddjango/django-stubs/pull/814#issuecomment-1301957498


flaeppe:
Interestingly enough the test that is added doesn't fail locally...

mypy_django_plugin/lib/helpers.py
```diff
@@ -385,3 +386,36 @@ def add_new_manager_base(api: SemanticAnalyzerPluginInterface, fullname: str) ->
...
+        # what app label this should be, yet.
+        api.fail("Could not resolve lazy reference without an app label", ctx)
+        api.note(
+            ("Try to use a reference explicitly prefixed with app label:" f' "<app_label>.{reference}" instead'),
+            ctx,
+        )
+        return None
+
+    # Reference conforms to the structure of a lazy reference: '<app_label>.<object_name>'
+    fullname = django_context.model_class_fullnames_by_label_lower.get(reference.lower())
```
sobolevn, line 404: Question about `lower`: why is that needed?

mypy_django_plugin/lib/helpers.py
```diff
@@ -385,3 +386,36 @@ def add_new_manager_base(api: SemanticAnalyzerPluginInterface, fullname: str) ->
...
+    reference: str, *, api: Union[TypeChecker, SemanticAnalyzer], django_context: "DjangoContext", ctx: Context
+) -> Optional[TypeInfo]:
+    """
+    Attempts to resolve a lazy reference(e.g. "<app_label>.<object_name>") to a
+    'TypeInfo' instance.
+    """
+    if "." not in reference:
+        # <object_name> -- needs prefix of <app_label>. We can't implicitly solve
+        # what app label this should be, yet.
+        api.fail("Could not resolve lazy reference without an app label", ctx)
```
sobolevn: Question: do we need to turn this into an error? Maybe just return `Any`? What are the pros / cons / risks?

sobolevn (commented):
Thanks!

mypy_django_plugin/lib/helpers.py
flaeppe: The implementation is case insensitive:

> Returns the [Model](https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model) with the given __app_label__ and __model_name__. As a shortcut, this method also accepts a single argument in the form __app_label.model_name__. __model_name__ is case-insensitive.

Ref: https://docs.djangoproject.com/en/4.2/ref/applications/#django.apps.apps.get_model

mypy_django_plugin/lib/helpers.py
flaeppe: Base thing is that Django has support without `app_label`, in some cases. But it's not trivial to align with how Django figures out any "current" app label without touching the runtime quite a bit, of course.

Only reason is that I wanted to put a strain on the usage here as we don't always fully align with Django's support. So it's not an error from Django when omitting app label, but we we'll resolve the model as `Type[Any]`. Thing was that from what I could tell Django _always_ support the "full" lazy reference, the one including app label, where the partial one is supported. So intention was to push usage towards using a full one instead, but I guess it could be considered a bit intrusive.

Anyways, I'm not feeling too strong about erroring either so I'm totally fine with having partial support and returning `Type[Any]` without a message when we can't resolve.

flaeppe:
> Interestingly enough the test that is added doesn't fail locally...

I really can't figure out how to reproduce locally..

@sobolevn would you mind checking out and try it, whenever you'd get a chance?

mypy_django_plugin/lib/helpers.py
sobolevn: I think it is better to start without an error. Later we can add them if needed.

flaeppe: I've removed the error now

PR #1726, opened 2023-09-21
Specify error codes to ignore for plugin type ignores

flaeppe:


mypy_django_plugin/transformers/models.py
```diff
@@ -747,7 +747,7 @@ def set_auth_user_model_boolean_fields(ctx: AttributeContext, django_context: Dj
 def handle_annotated_type(ctx: AnalyzeTypeContext, django_context: DjangoContext) -> MypyType:
     args = ctx.type.args
     type_arg = ctx.api.analyze_type(args[0])
-    api = cast(SemanticAnalyzer, ctx.api.api)  # type: ignore
```
ljodal, line 756: Could we use `helpers.get_semanal_api` instead?

https://github.com/typeddjango/django-stubs/blob/5ba0b063738361c683b1577b5deeec9dee18a448/mypy_django_plugin/lib/helpers.py#L344-L347

ljodal (approved):
Nice cleanup 🧹

sobolevn (approved):
Thanks! This got me thinking: maybe we should also add error codes for our own errors as the next step? Right now we don't have any.

flaeppe:
> Thanks! This got me thinking: maybe we should also add error codes for our own errors as the next step? Right now we don't have any.

IIRC we had problems `# type: ignore[our-code]` those? As mypy didn't pick up on plugin specified ones. Not sure if that has changed in recent mypy versions?

mypy_django_plugin/transformers/models.py
flaeppe: Yeah, looks about right

sobolevn:
> Not sure if that has changed in recent mypy versions?

We totally can change this 😉

UnknownPlatypus:
Ruff can lint for this https://docs.astral.sh/ruff/rules/blanket-type-ignore/#blanket-type-ignore-pgh003

Now that ruff is part of the toolchain, maybe we could add this rule to the configuration?

flaeppe:
> Ruff can lint for this https://docs.astral.sh/ruff/rules/blanket-type-ignore/#blanket-type-ignore-pgh003
> 
> Now that ruff is part of the toolchain, maybe we could add this rule to the configuration?

Cool, but there's about 100 `# type: ignore` in the stubs files so I'll exclude `*.pyi` from being ruled then

UnknownPlatypus:
I have a few scripts to automate that, had to do the same at work. I can do a follow PR if you want ?

flaeppe:
> I have a few scripts to automate that, had to do the same at work. I can do a follow PR if you want ?

Sounds great!

PR #1732, opened 2023-09-23
Update `Model._meta` to `ClassVar[Options[Self]]`

### LINKED ISSUE #471: error: Definition of "_meta" in base class "PermissionsMixin" is incompatible with definition in base class "AbstractBaseUser"
# Bug report

## What's wrong

We have defined a custom `AuthUser` model in our project like:

```python

class AuthUser(PermissionsMixin, AbstractBaseUser):
    class Meta:
        verbose_name = _('Auth User')
        abstract = False
        db_table = 'auth_user'
```

mypy gives an error:
```
myproject/models/auth.py:3: error: Definition of "_meta" in base class "PermissionsMixin" is incompatible with definition in base class "AbstractBaseUser"
```
(from the `class AuthUser(PermissionsMixin, AbstractBaseUser):` line)

The slightly weird part is Django's `PermissionMixin` and `AbstractBaseUser` both have an identical `Meta`:
```python
    class Meta:
        abstract = True
```

## How is that should be

No error

## System information

- OS:
- `python` version: 3.8.3
- `django` version: 2.2.12
- `mypy` version: 0.782

flaeppe:
Runtime `_meta` attribute is available on a model type, abstract or not doesn't matter. Also drop the plugin code populating the `Options` argument as it isn't needed.

## Related issues

Closes: #471

### DIFF
--- a/django-stubs/db/models/base.pyi
+++ b/django-stubs/db/models/base.pyi
@@ -41,7 +41,7 @@ class Model(metaclass=ModelBase):
     objects: ClassVar[Manager[Self]]
 
     class Meta: ...
-    _meta: Options[Any]
+    _meta: ClassVar[Options[Self]]
     pk: Any
     _state: ModelState
     def __init__(self, *args: Any, **kwargs: Any) -> None: ...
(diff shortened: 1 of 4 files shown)


sobolevn (approved):
Awesome! TIL about `ClassVar[Self]`

PR #1733, opened 2023-09-24
Improve hints for `ReverseOneToOneDescriptor` and start using it

### LINKED ISSUE #649: Mypy unable to find out RelatedObjectDoesNotExist
# Bug report

Setup

```python
from django.db import models

class MyUser(User):
    pass

class Role(models.Model):
    name = models.CharField(max_length=200, db_index=True)
    user = models.OneToOneField(MyUser, on_delete=models.PROTECT)


try:
    user = MyUser.objects.get(pk=1)
    user.role.name
except MyUser.role.RelatedObjectDoesNotExist:
    print('user role mapping is missing')
```

output

```bash
$mypy filename.py
...
user_related.py:15: error: "Type[MyUser]" has no attribute "role"
```

After modifying the code

```python
class MyUser(User):
    role: Role
    pass

```

output

```
...
user_related.py:16: error: "Role" has no attribute "RelatedObjectDoesNotExist"
``` 
## What's wrong

when add the following code in the model and print `MyUser.role.RelatedObjectDoesNotExist` in the shell, you can find the output as `<App.MyUser.role.RelatedObjectDoesNotExist'>`. At runtime, the exception handler works as expected but during type checking the exception is missing to mypy. 

type of `Myuser.role` is `django.db.models.fields.related_descriptors.ForwardOneToOneDescriptor` at runtime.

## How is that should be

Mypy should not complain attribute is missing. 

## System information

- OS: Mac OSX
- `python` version: 3.9.4
- `django` version: 3.2
- `mypy` version: 0.812

### LINKED ISSUE #1390: "RelatedModel" has no attribute "RelatedObjectDoesNotExist"  [attr-defined]
# Bug report

## What's wrong

~Mypy prints `Module "django.db.models" has no attribute "ObjectDoesNotExist"  [attr-defined]` on the `from django.db.models import ObjectDoesNotExist` line, which is obviously wrong.~

Also, if I alternatively use `MyModel.field.RelatedObjectDoesNotExist`, I get the `"RelatedModel" has no attribute "RelatedObjectDoesNotExist"  [attr-defined]` message. 

`field` here is a reverse relation from `RelatedModel` to `MyModel` one-to-one field.

## How is that should be

No error messages.

## System information

- OS:
- `python` version: 3.10.6
- `django` version: 3.2.18
- `mypy` version: 1.0.1

flaeppe:
Plugin now sets the `ReverseOneToOneDescriptor` for reverse one to one relations while it takes 2 new type arguments `_From` and `_To`, both as models, to describe the relation

## Related issues

Closes: #649


django-stubs/db/models/fields/related_descriptors.pyi
```diff
@@ -36,19 +38,31 @@ class ForwardManyToOneDescriptor(Generic[_F]):
...
+    """
+
     related: OneToOneRel
     def __init__(self, related: OneToOneRel) -> None: ...
     @property
     def RelatedObjectDoesNotExist(self) -> type[ObjectDoesNotExist]: ...
-    def is_cached(self, instance: Model) -> bool: ...
-    def get_queryset(self, **hints: Any) -> QuerySet: ...
+    def is_cached(self, instance: _From) -> bool: ...
+    def get_queryset(self, **hints: Any) -> QuerySet[_To]: ...
```
flaeppe, line 56: The plugin should probably refine `def get_queryset()`s return type to be from the default manager of the related model. Can be seen here:

https://github.com/django/django/blob/574ee4023e15cfb195833edfbaed353f8021c62f/django/db/models/fields/related_descriptors.py#L443-L444

I wanted to keep the change set smaller, so we can do this in a subsequent PR. As we already do a bit of changes and everything in here should be improvements. In general I think the usage is quite small for `get_queryset` in here anyways..

mypy_django_plugin/transformers/models.py
```diff
@@ -440,7 +440,7 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
         self.add_new_node_to_model_class("_default_manager", default_manager, is_classvar=True)
 
 
-class AddRelatedManagers(ModelClassInitializer):
+class AddReverseLookups(ModelClassInitializer):
```
flaeppe, line 443: It took me quite some time to figure out that the reverse one to one was set by the plugin in this initializer.

I'm all up for an even better name that describes that this handles the reverse parts of relationships..

tests/typecheck/models/test_create.yml
```diff
@@ -53,7 +53,7 @@
                     id2 = models.AutoField(primary_key=True)
                     name2 = models.CharField(max_length=50)
 
-                class Child1(Parent1, Parent2):
+                class Child1(Parent1, Parent2):  # type: ignore[misc]
```
flaeppe: Now, I've got no idea how or if types can even be solved properly for multiple inheritance.

Given that e.g. `Child.child1` exists here, and I'm not sure what that should or can even be typed as.

Anyways, this fails on the descriptor types that the plugin adds:

```python
Parent1.child1: ReverseOneToOneDescriptor[Child1, Parent1]
Parent2.child1: ReverseOneToOneDescriptor[Child1, Parent2]
```

And since it's also a python inheritance, mypy end up seeing an override on `Child1.child1` that is not compatible.

While I'm writing this I realise it's probably better to capture the error in `out:` to make everything a bit more obvious

tests/typecheck/fields/test_related.yml
```diff
@@ -1011,3 +1011,39 @@
...
+        reveal_type(Other.mymodel.get_queryset())
+        reveal_type(Other.mymodel.RelatedObjectDoesNotExist)
+        try:
+            other = Other.objects.get()
+            reveal_type(other.mymodel)
+        except Other.mymodel.RelatedObjectDoesNotExist:
+            ...
+        else:
+            other.mymodel = MyModel()
+            other.mymodel = Other()
```
sobolevn, line 1030: `other.mymodel = None`?

flaeppe: Yes true, that should be here too

tests/typecheck/fields/test_related.yml
```diff
@@ -1011,3 +1011,40 @@
...
+        -   path: myapp/__init__.py
+        -   path: myapp/models/__init__.py
+            content: |
+                from django.db import models
+
+                class Other(models.Model):
+                    ...
+
+                class MyModel(models.Model):
+                    field = models.OneToOneField(Other, on_delete=models.CASCADE)
```
sobolevn: Just an idea: we can also test a model with explicit `related_name`?

sobolevn (approved):
Thanks!

tests/typecheck/fields/test_related.yml
flaeppe: Sure, I've included that now

PR #1734, opened 2023-09-24
Add no blanket type ignore

UnknownPlatypus:
Following discussions on #1726, this PR adds the ruff linter rule [PGH003](https://docs.astral.sh/ruff/rules/blanket-type-ignore/#blanket-type-ignore-pgh003) disallowing blanket `type: ignore` annotations.

I've added specific error codes for previous violations using a few custom scripts.


adamchainz (approved):
Nice one

flaeppe (approved):
👍

PR #1737, opened 2023-09-26
AbstractBaseUser.REQUIRED_FIELDS is a ClassVar

WhyNotHugo:
On a subclass, there's not correct way to annotate `REQUIRED_FIELDS`.

- Annotating as `list[str]` warns that this should be annotated as `ClassVar`.
- Annotating as `ClassVar` warns that the this doesn't match the parent class.

This field is a class variable anyway, so this is the correct annotation.


sobolevn (approved):
Good catch! Thanks

PR #1739, opened 2023-09-27
Narrow down annotations for fields.Field.formfield

### LINKED ISSUE #1738: Types for django.db.models.fields.formfield
# Bug report

For the type `django.db.models.fields.formfield.Field`, the function `formfield` currently has the following type.

```py
def formfield(self, form_class: Any | None = ..., choices_form_class: Any | None = ..., **kwargs: Any) -> Any: ...
```

## What's wrong

The `form_class` attribute is `type[forms.Field]`, and the return value is `forms.Field`:

```py
def formfield(self, form_class: type[forms.Field] | None = ..., choices_form_class: Any | None = ..., **kwargs: Any) -> forms.Field: ...
```

## How is that should be

This isn't specific enough: the return value is of the same type of `form_class`, so there is a relationship between the two.

However, I'm not sure how to represent this type relationship. Should I use a `Protocol` for the entire type?

## System information

Current master of this project. All else is out of scope here.

WhyNotHugo:
Fixes: https://github.com/typeddjango/django-stubs/issues/1738


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -202,7 +204,9 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
     def get_db_prep_save(self, value: Any, connection: BaseDatabaseWrapper) -> Any: ...
     def get_internal_type(self) -> str: ...
     # TODO: plugin support
-    def formfield(self, form_class: Any | None = ..., choices_form_class: Any | None = ..., **kwargs: Any) -> Any: ...
+    def formfield(
+        self, form_class: type[_Field] | None = ..., choices_form_class: Any | None = ..., **kwargs: Any
+    ) -> _Field: ...
```
```suggestion
    def formfield(
        self, form_class: type[Field] | None = ..., choices_form_class: Any | None = ..., **kwargs: Any
    ) -> Field: ...
```

I propose to type it like this, because you can change the type of the returned field in this method: https://github.com/django/django/blob/f9e9526800c921ae393ff58826daed51587b1727/django/db/models/fields/__init__.py#L1125

flaeppe: Isn't this an `@overload`? __Edit:__ Difficult with the `self.choices` though...

flaeppe: But yeah, this method is kind of designed to be overridden. Most fields customize it

WhyNotHugo: @sobolevn Agreed, I'll update this.

WhyNotHugo:
Updated.

sobolevn (approved):
Thanks!

PR #1743, opened 2023-09-28
fill out descriptor for ModelStateFieldsCacheDescriptor

asottile:
some example usage here (and surrounding lines) https://github.com/getsentry/sentry/blob/808f610f97f78a5ad5df2381efecad642c368f63/src/sentry/db/models/base.py#L92


sobolevn (approved):
LGTM, thanks!

Issue #1740, opened 2023-09-27
Next release planning (4.2.5)


flaeppe:
__TODO:__
- [x] #1719 
- [x] Add plugin support for `users = ManyToManyField(settings.AUTH_USER_MODEL)` #1746 

#1719 introduces plugin support for `ManyToManyField` but is lacking support for passing `AUTH_USER_MODEL` as `to=` arg e.g.:

```python
users = ManyToManyField(settings.AUTH_USER_MODEL)
```

Which we suspect is a normal use case and should support in any next release.

---

Please comment on this issue if there's anything else you know should, or want to be included.

flaeppe:
And for anyone feeling adventurous, please try out `HEAD` on your project(s) and report back here if you run in to anything you think is a new issue. Perhaps we can then resolve it before pushing out a release.

intgr:
Now that mypy 1.6.0 is out, we should consider making a release with the `compatible-mypy` bump. I can do the release.

I'd consider this a blocker, since it's a regression and we already have a fix PR:

* https://github.com/typeddjango/django-stubs/pull/1745

intgr:
I'm releasing djangorestframework-stubs too, any takers to review this for the release:
* https://github.com/typeddjango/djangorestframework-stubs/pull/482

flaeppe:
It'd be nice if we could also include #1767 in the next release

intgr:
I want to move forward with:

* https://github.com/typeddjango/django-stubs/issues/1628

Making no changes quite yet, but I propose mentioning in 4.2.5 release notes that the next next version (4.2.6) will remove hard mypy dependency. Something like:

<details>

<summary>(outdated text)</summary>

> django-stubs is taking steps to improve experience with other type checkers besides mypy. Next version (4.2.6) will remove explicit mypy dependency. If you are currently using mypy, please add an explicit dev dependency on `mypy` to your project, or depend on django-stubs with the `compatble-mypy` extra, e.g. `django-stubs[compatible-mypy]`.
> **Note:** this is not a guarantee that other type checkers will be fully supported, pull requests for improvements are welcome. See [#1628](https://github.com/typeddjango/django-stubs/issues/1628) for details.

</details>

If you have comments on this, please discuss in #1628.

intgr:
Should I cut the release or wait for the `cached_property` work to be completed?

* #1771
* #1761
* #1768
* #1769

flaeppe:
[...]

PR #1746, opened 2023-09-28
Resolve any `settings.AUTH_USER_MODEL` used as `to` in relation

flaeppe:
e.g.

```python
OneToOneField(settings.AUTH_USER_MODEL, ...)
ForeignKey(settings.AUTH_USER_MODEL, ...)
ManyToManyField(settings.AUTH_USER_MODEL, ...)
```

## Related issues

Ref: https://github.com/typeddjango/django-stubs/pull/1719#issuecomment-1734341183
Ref: #1740

### DIFF
--- a/tests/typecheck/models/test_contrib_models.yml
+++ b/tests/typecheck/models/test_contrib_models.yml
@@ -91,3 +91,71 @@
                   class Meta:
                       abstract = False
                       db_table = "auth_user"
+
+-   case: test_relation_specified_by_auth_user_model
+    main: |
+        from other.models import Other
+        reveal_type(Other().users.get())
+        reveal_type(Other.users.through)
+        reveal_type(Other.users.through().myuser)
+        reveal_type(Other.users.through.objects.get().myuser)
+
+        reveal_type(Other().user)
+
+        reveal_type(Other().unq_user)
+    out: |
+        main:2: note: Revealed type is "myapp.models.MyUser"
+        main:3: note: Revealed type is "Type[other.models.Other_users]"
+        main:4: note: Revealed type is "myapp.models.MyUser"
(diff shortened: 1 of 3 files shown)


tests/typecheck/models/test_contrib_models.yml
```diff
@@ -62,3 +62,45 @@
...
+        main:2: note: Revealed type is "myapp.models.MyUser"
+        main:3: note: Revealed type is "Type[other.models.Other_users]"
+        main:4: note: Revealed type is "myapp.models.MyUser"
+        main:5: note: Revealed type is "myapp.models.MyUser"
+
+        main:7: note: Revealed type is "myapp.models.MyUser"
+
+        main:9: note: Revealed type is "myapp.models.MyUser"
+    custom_settings: |
+        INSTALLED_APPS = ('django.contrib.contenttypes', 'django.contrib.auth', 'myapp', 'other')
```
sobolevn, line 116: What will happen if `django.contrib.auth` is not installed?

sobolevn (commented):
Thanks!

tests/typecheck/models/test_contrib_models.yml
flaeppe: You'd get an error asking you for explicit annotations/args of the field.

I can add a case for that too though

flaeppe: Actually, not necessarily. It purely depends on if `settings.AUTH_USER_MODEL` can be resolve to a model or not, since we don't really touch the runtime..

flaeppe: So it behaves the same as the runtime, since Django doesn't care if you put the lazy reference in a variable or not.

sobolevn (approved):

PR #1751, opened 2023-10-02
Fix `django.test.signals` types

UnknownPlatypus:
Fixes types in `django/test/signals.py` following [signature change in django 4.1](https://github.com/django/django/commit/b0d16d0129b7cc5978a8d55d2331a34cb369e6c7)

Trimming the todo list of 16 items


sobolevn (approved):
Thanks!

PR #1752, opened 2023-10-02
Misc test stubs types

UnknownPlatypus:
Some more misc test stubs update

### DIFF
--- a/django-stubs/test/html.pyi
+++ b/django-stubs/test/html.pyi
@@ -5,7 +5,7 @@ from typing import Any
 
 from typing_extensions import TypeAlias
 
-ASCII_WHITESPACE: Pattern
+ASCII_WHITESPACE: Pattern[str]
 BOOLEAN_ATTRIBUTES: set[str]
 
 def normalize_whitespace(string: str) -> str: ...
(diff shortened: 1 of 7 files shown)


django-stubs/test/html.pyi
```diff
@@ -1,12 +1,15 @@
 from collections.abc import Sequence
 from html.parser import HTMLParser
+from re import Pattern
 from typing import Any
 
 from typing_extensions import TypeAlias
 
-WHITESPACE: Any
+ASCII_WHITESPACE: Pattern
```
sobolevn:
```suggestion
ASCII_WHITESPACE: Pattern[str]
```

sobolevn (commented):
Thanks!

django-stubs/test/html.pyi
UnknownPlatypus: Thanks, I've updated other `Pattern` which I believe also miss this.

While at it, I also updated the `allowlist.txt` to reference every regex pattern in the `Lazy regex` section. 
I found some missing running `rg ": Pattern" -N | cut -d":" -f2 | xargs -I {} rg {} ./scripts/stubtest/allowlist_todo.txt`

sobolevn (approved):
Thanks!

PR #1759, opened 2023-10-05
CI: Enable testing with Python 3.12

intgr:
This requires removing the `setuptools<64` version constraint from `[build-system]` in `pyproject.toml`, introduced in #1123.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -U pip setuptools wheel
-          pip install -r ./requirements.txt
+          SETUPTOOLS_ENABLE_FEATURES=legacy-editable pip install -r ./requirements.txt
 
       - name: Run mypy on plugin code
         run: mypy --strict mypy_django_plugin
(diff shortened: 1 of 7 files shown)


intgr:
The solution from #1122 to use `SETUPTOOLS_ENABLE_FEATURES=legacy-editable` works. Will need to figure out what effect it has and if we can fix our package to work without it.

intgr:
@andersk Any chance you know/remember, why django-stubs needs `SETUPTOOLS_ENABLE_FEATURES=legacy-editable` with newer setuptools?

intgr:
OK I see, `setuptools` new approach for editable installs does not work with static analysis tools like mypy:
* https://github.com/python/mypy/issues/13392
* https://github.com/pypa/setuptools/issues/3518

As long as the legacy mode works, it's good enough for us.

I just wish it was possible to configure `SETUPTOOLS_ENABLE_FEATURES=legacy-editable` or `--config-settings editable_mode=compat` in `pyproject.toml` file without burdening users with this.

sobolevn (approved):
Great! Thank you!

PR #1766, opened 2023-10-11
Require 1 callable argument for `@cached_property`

flaeppe:
`django.utils.functional.cached_property` requires a single argument, not arbitrary `*args` and `**kwargs`.

Having more than 1 argument results in a runtime error when any decorated attribute is accessed (i.e. `cached_property` being evaluated)

### DIFF
--- a/django-stubs/utils/functional.pyi
+++ b/django-stubs/utils/functional.pyi
@@ -7,9 +7,9 @@ from typing_extensions import Self, TypeAlias
 _T = TypeVar("_T")
 
 class cached_property(Generic[_T]):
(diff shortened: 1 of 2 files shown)


flaeppe:
For anyone being curious, this is similar to how typeshed adds hints for `functools.cached_property`:

https://github.com/python/typeshed/blob/7e2a3413faa7cb1f43b7743675b7f3c0909cb129/stdlib/functools.pyi#L196-L208

intgr (approved):
:+1: Makes sense and works fine.

PR #1768, opened 2023-10-13
Update `@property` attributes to `@cached_property` part 2

UnknownPlatypus:
Followup to #1761

Replace a bunch of `@property` that are now `@cached_property`

This removes 129 entries from the todolist

PR #1769, opened 2023-10-13
Update Expression classes `output_field` to `@cached_property` or `ClassVar` and improve type

UnknownPlatypus:
Followup to #1761 and #1768.

Should also fix #1765 since the `output_field` type on `BaseExpression` is now accurate. 

This removes 6 entries from the todolist


django-stubs/contrib/gis/db/models/functions.pyi
```diff
@@ -1,9 +1,12 @@
...
+from typing import Any, ClassVar
 
+from django.contrib.gis.db.models.fields import GeometryField
+from django.contrib.gis.db.models.sql.conversion import DistanceField
 from django.db.backends.base.base import BaseDatabaseWrapper
-from django.db.models import Func
+from django.db.models import BinaryField, BooleanField, FloatField, Func, IntegerField, TextField
 from django.db.models import Transform as StandardTransform
 from django.db.models.sql.compiler import SQLCompiler, _AsSqlType
+from utils.functional import cached_property
```
intgr: I'm surprised this passed CI, shouldn't this be:

```suggestion
from django.utils.functional import cached_property
```

UnknownPlatypus: absolutely, I wonder why pycharm added the import like this, I didn't really paid attention

UnknownPlatypus:
This is unblocked by #1771 too I think, I've rebased

intgr (changes requested):
Thanks! A few omissions I noticed:

* `django.contrib.gis.db.models.functions.GeomOutputGeoFunc`, `Area` attribute `output_field` can be `cached_property` too.
* `django.contrib.gis.db.backends.postgis.operations.ST_Polygon`

UnknownPlatypus:
Thanks, nice catch. I did some more greping and I think it's good now

intgr (approved):
Awesome, thanks!

PR #1770, opened 2023-10-13
Update `django.middleware.csrf` types

UnknownPlatypus:
This MR updates the types in `django.middleware.csrf`.

I also added a section in the `todolist.txt` for the specific errors `<...> is not present at runtime`. I don't think we should remove these because it would cause incompatibility with older Django version we partially support.

But maybe worth keeping a track of for when we want to officially drop support of some django versions ?

This removes 11 entries from the todolist

### DIFF
--- a/django-stubs/middleware/csrf.pyi
+++ b/django-stubs/middleware/csrf.pyi
@@ -1,18 +1,30 @@
+from collections import defaultdict
 from collections.abc import Callable
 from logging import Logger
+from re import Pattern
 from typing import Any
 
 from django.http.request import HttpRequest
 from django.http.response import HttpResponseBase, HttpResponseForbidden
 from django.utils.deprecation import MiddlewareMixin
+from django.utils.functional import cached_property
 
 logger: Logger
+
+invalid_token_chars_re: Pattern[str]
+
+REASON_BAD_ORIGIN: str
 REASON_NO_REFERER: str
 REASON_BAD_REFERER: str
 REASON_NO_CSRF_COOKIE: str
+REASON_CSRF_TOKEN_MISSING: str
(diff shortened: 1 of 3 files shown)


django-stubs/middleware/csrf.pyi
```diff
@@ -1,32 +1,63 @@
...
 CSRF_SECRET_LENGTH: int
 CSRF_TOKEN_LENGTH: Any
 CSRF_ALLOWED_CHARS: Any
 CSRF_SESSION_KEY: str
 
+def _get_new_csrf_string() -> str: ...
+def _get_new_csrf_token() -> str: ...
+def _mask_cipher_secret(secret: str) -> str: ...
+def _unmask_cipher_token(token: str) -> str: ...
+def _add_new_csrf_cookie(request: HttpRequest) -> str: ...
```
UnknownPlatypus: The 2 first ones were annotated at the end of this file, i just moved them here because that's where they are in the Django source code.

I added types for other private top level functions around but was wondering if I really should.
I've only seen a few top level private function annotated and stubtest seems to ignore them.

So should I just remove them or do we keep them ?

sobolevn: > I added types for other private top level functions around but was wondering if I really should.

We only add protected stuff when people ask us to. So, please revert this change.

sobolevn (commented):
Other than protected functions, it looks good to me. Thanks!

> I added types for other private top level functions around but was wondering if I really should.

Yes, we totally can! If we track django versions there, it would be easy to drop legacy stuff.

sobolevn (approved):

sobolevn:
Thank you!

PR #1773, opened 2023-10-16
Tests: Add mypy error codes to typecheck

UnknownPlatypus:
# I have made things!

I've been doing some work with the stubs lately and was relying on mypy error code to add accurate `type: ignore` comments when necessary. However, I had to change the `show_error_codes` settings quite often for that.

I've seen [here](https://github.com/typeddjango/django-stubs/pull/1577/files#r1243393029) that this setting was legacy so this PR removes it and update the tests accordingly. 

I've used a small python script to do the update automatically.

## Related issues
I could probably reuse the script to do what was started in #1577 and update `force_union_syntax` setting too.
This just require a decision on the value to set for `python_version`. See https://github.com/typeddjango/django-stubs/pull/1572#issuecomment-1605606030
However, because we want to run tests on multiple python versions, maybe it's fine to keep those settings.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -13,7 +13,6 @@ warn_unused_configs = true
 warn_unreachable = true
 disallow_untyped_defs = true
 disallow_incomplete_defs = true
-show_error_codes = false
 disable_error_code = empty-body
 # TODO: update our output assertions to match a new syntax
 force_uppercase_builtins = true
--- a/tests/typecheck/contrib/admin/test_decorators.yml
+++ b/tests/typecheck/contrib/admin/test_decorators.yml
@@ -17,7 +17,7 @@
         @admin.display
         def display_property(self) -> str: ...
 
-        @admin.display  # E: Decorators on top of @property are not supported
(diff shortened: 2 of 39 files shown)


intgr (approved):
Thanks! I've been looking forward to this.

I'd also like to get rid of `force_uppercase_builtins`, `force_union_syntax`, which also means lots of changes to messages in tests. But let's keep that separate.

sobolevn (approved):
This is great, thanks a lot!

UnknownPlatypus:
> I'd also like to get rid of force_uppercase_builtins, force_union_syntax, which also means lots of changes to messages in tests. But let's keep that separate.

I can do that quite easily now but there is another [pending MR](https://github.com/typeddjango/django-stubs/pull/1577) on the topic that needs decision (See https://github.com/typeddjango/django-stubs/pull/1572#issuecomment-1605606030).
@intgr , do you have an opinion on the python-version thingy ?

intgr:
Thanks, that PR has fallen through the cracks. I have responded in https://github.com/typeddjango/django-stubs/pull/1577#discussion_r1362204604

PR #1774, opened 2023-10-17
Django core cache

UnknownPlatypus:
Some misc stubs update.

This removes 12 entries from the todolist


django-stubs/core/cache/backends/filebased.pyi
```diff
@@ -1,7 +1,8 @@
-from typing import Any
+from typing import Any, ClassVar
 
 from django.core.cache.backends.base import BaseCache
 
 class FileBasedCache(BaseCache):
     cache_suffix: str
```
sobolevn: It is also `ClassVar`

django-stubs/core/cache/backends/memcached.pyi
```diff
@@ -12,6 +12,8 @@ class BaseMemcachedCache(BaseCache):
         library: ModuleType,
         value_not_found_exception: type[BaseException],
     ) -> None: ...
+    @property
+    def client_servers(self) -> Sequence[str]: ...
```
sobolevn, line 16: Return types should be exact, no need for abstactions (unless explicitly documented).

sobolevn (commented):
Thanks!

django-stubs/core/cache/backends/filebased.pyi
UnknownPlatypus: indeeed! updated

django-stubs/core/cache/backends/memcached.pyi
UnknownPlatypus: I thinks it's still correct here.
`self._servers` is set in the init this way:
```python
        if isinstance(server, str):
            self._servers = re.split("[;,]", server)
        else:
            self._servers = server
```

And server is annotated as `server: str | Sequence[str]`. So in the else branch, we could have 
a Sequence. `PyMemcacheCache(("aa", "bb"), {})` is valid for ex.

However, in the `PyLibMCCache` override, it always returns `list[str]`.

I don't know if this distinction is worth it tho. I've updated the PR to match runtime types, let me know what you think

sobolevn (approved):
Thanks!

PR #1778, opened 2023-10-18
Fix `Field.formfield()`, `GeometryField.formfield()` methods

brianhelba:
This partially reverts 9010b5f8569255c55bc2253602a90eba2e57f46c.

See discussion on #1724.


brianhelba:
@intgr PTAL.

If possible, could we get a supplemental patch release to fix this? Otherwise, users will have to `#type: ignore` any usages of this API until this is released.

intgr:
Oops, sorry for messing that up. And thanks for responding so quickly.

Sadly this screw-up is now part of the django-stubs 4.2.5 release. I'll probably wait a few days to see if any other bugs turn up, and make another hotfix release.

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -205,10 +205,10 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
     # TODO: plugin support
     def formfield(
         self,
-        form_class: type[Field] | None = ...,
+        form_class: type[forms.Field] | None = ...,
         choices_form_class: type[forms.ChoiceField] | None = ...,
         **kwargs: Any,
```
intgr, line 210: If you're feeling adventurous, it would be useful to typehint additional keys supported in `kwargs`, in a separate PR. Looking at the `formfield()` method code, seems like you can copy certain arguments from `CharField` constructor:

```python
    def formfield(self, form_class=None, choices_form_class=None, **kwargs):
            ...
            for k in list(kwargs):
                if k not in (
                    "coerce",
                    "empty_value",
                    "choices",
                    "required",
                    "widget",
                    "label",
                    "initial",
                    "help_text",
                    "error_messages",
                    "show_hidden_initial",
                    "disabled",
                ):
                    del kwargs[k]
        defaults.update(kwargs)
        if form_class is None:
            form_class = forms.CharField
        return form_class(**defaults)
```

intgr (approved):
Looks good, thanks again!

intgr:
Released 4.2.6

* #1780

PR #1783, opened 2023-10-19
fix: UserAdmin accepts AbstractUser

GabDug:
# I have made things!

Fix an issue I introduced with UserAdmin: if you inherit from it with a custom `User` it would not work, only with Django base `User`.

If needed, can add a test case akin of:
```
@admin.register(CustomUser)
class CustomUserAdmin(UserAdmin[CustomUser]):
   ...
```
and 
```
@admin.register(CustomUser)
class CustomUserAdmin(UserAdmin):
   ...
```

Any feedback is appreciated!

### DIFF
--- a/django-stubs/contrib/auth/admin.pyi
+++ b/django-stubs/contrib/auth/admin.pyi
@@ -1,16 +1,17 @@
-from typing import Any
+from typing import Any, Generic, TypeVar
 
 from django.contrib import admin
-from django.contrib.auth.models import Group, User
+from django.contrib.auth.models import AbstractUser, Group
 from django.http.request import HttpRequest
 from django.http.response import HttpResponse
 
 csrf_protect_m: Any
 sensitive_post_parameters_m: Any
+_M = TypeVar("_M", bound=AbstractUser, covariant=False)
 
 class GroupAdmin(admin.ModelAdmin[Group]): ...
 
-class UserAdmin(admin.ModelAdmin[User]):
+class UserAdmin(admin.ModelAdmin[_M], Generic[_M]):
     change_user_password_template: Any
     add_fieldsets: Any
     add_form: Any


django-stubs/contrib/auth/admin.pyi
```diff
@@ -1,16 +1,17 @@
...
 from django.http.response import HttpResponse
 
 csrf_protect_m: Any
 sensitive_post_parameters_m: Any
+_M = TypeVar("_M", bound=AbstractUser, covariant=False)
 
 class GroupAdmin(admin.ModelAdmin[Group]): ...
 
-class UserAdmin(admin.ModelAdmin[User]):
+class UserAdmin(admin.ModelAdmin[_M], Generic[_M]):
```
jmfederico, line 14:
```suggestion
class UserAdmin(admin.ModelAdmin[_M]):
```

I do not think the Generic is needed.

django-stubs/contrib/auth/admin.pyi
```diff
@@ -1,16 +1,17 @@
...
 
 from django.contrib import admin
-from django.contrib.auth.models import Group, User
+from django.contrib.auth.models import AbstractUser, Group
 from django.http.request import HttpRequest
 from django.http.response import HttpResponse
 
 csrf_protect_m: Any
 sensitive_post_parameters_m: Any
+_M = TypeVar("_M", bound=AbstractUser, covariant=False)
```
jmfederico, line 10:
```suggestion
_M = TypeVar("_M", bound=AbstractUser)
```

covariant is False by default I think.

django-stubs/contrib/auth/admin.pyi
```diff
@@ -1,16 +1,17 @@
-from typing import Any
+from typing import Any, Generic, TypeVar
```
jmfederico, line 1:
```suggestion
from typing import Any, TypeVar
```

As per below comments, I do not think this is needed.

vadimshvetsov:
Hey @GabDug, thanks for your PR. Do you plan to address comments or I can do it for you and open new PR? I need this also.

kolanos:
Shouldn't this be checking for inheritance from `AbstractBaseUser` and not `AbstractUser`? A lot of people create a custom user model based on `AbstractBaseUser`, such as not including a `username` field.

PR #1789, opened 2023-10-21
fix as manager and from queryset self types (#1788)

### LINKED ISSUE #1788: As_manager: Self not taken into account
# Bug report
```
from django.db import models
class MyQuerySet(BaseQuerySet):
       def example_simple(self) -> Self: ...

class MyModel(models.Model):
reveal_type(MyModel.objects.example_simple())  # -> Revealed type is "<nothing>" 
````

## What's wrong

The issue lies in the type revelation within the custom QuerySet method that is supposed to return Self. Currently, it reveals the type as `<nothing>`, which indicates that the Self type isn't properly recognized or handled.

## How is that should be

The correct and expected behavior should reveal the type accurately, as per the following expectation:
```
reveal_type(MyModel.objects.example_simple())  # -> Revealed type is "myapp.models.ManagerFromMyQuerySet[myapp.models.MyModel]"
```

## System information

- OS:
- `python` 3.12
- `django` 3.2.12
- `mypy` 1.6.1
- `django-stubs` 4.2.5
- `django-stubs-ext` 4.2.5

### PR DESCRIPTION
**Pull request by moranabadie**
# I have made things!

I replace self_type by the manager itself

## Related issues

It fixes : https://github.com/typeddjango/django-stubs/issues/1788

### DIFF
--- a/tests/typecheck/managers/querysets/test_as_manager.yml
+++ b/tests/typecheck/managers/querysets/test_as_manager.yml
@@ -1,3 +1,31 @@
+-   case: self_return_management
+    main: |
+        from myapp.models import MyModel
+        reveal_type(MyModel.objects.example_simple())  # N: Revealed type is "myapp.models.ManagerFromMyQuerySet[myapp.models.MyModel]"
+        reveal_type(MyModel.objects.example_list())  # N: Revealed type is "builtins.list[myapp.models.ManagerFromMyQuerySet[myapp.models.MyModel]]"
+        reveal_type(MyModel.objects.example_simple().just_int())  # N: Revealed type is "builtins.int"
+        reveal_type(MyModel.objects.example_dict())  # N: Revealed type is "builtins.dict[builtins.str, myapp.models.ManagerFromMyQuerySet[myapp.models.MyModel]]"
+
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+                from typing import List, Dict
+                from typing_extensions import Self
+
+                class BaseQuerySet(models.QuerySet):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by sobolevn — APPROVED** [review]
Thanks!

File: tests/typecheck/managers/querysets/test_as_manager.yml [hunk 1]
```diff
@@ -1,3 +1,30 @@
...
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+                from typing_extensions import Self
+
+                class BaseQuerySet(models.QuerySet):
+                    def example_dict(self) -> dict[str, Self]: ...
```
sobolevn:
```suggestion
                    def example_dict(self) -> Dict[str, Self]: ...
```

tests/typecheck/managers/querysets/test_as_manager.yml
```diff
@@ -1,3 +1,30 @@
...
+
+                class MyQuerySet(BaseQuerySet):
+                    def example_simple(self) -> Self: ...
+                    def example_list(self) -> list[Self]: ...
```
sobolevn:
```suggestion
                    def example_list(self) -> List[Self]: ...
```

sobolevn (commented):
To fix the CI

tests/typecheck/managers/querysets/test_as_manager.yml
moranabadie: oh thank you

tests/typecheck/managers/querysets/test_as_manager.yml
moranabadie: thx, fixed

sobolevn (approved):

PR #1791, opened 2023-10-22
Update Django db backend base wrapper

UnknownPlatypus:
# I have made things!

This adds a bunch of missing boolean flags in`django.db.backend.base` base classes.
I've moved some issues in `BaseDatabaseWrapper` to the allowlist because I think it make more sense to keep the current types.
Let me know if you agree.

This removes 63 entries from the todolist.
Cheers


django-stubs/db/backends/base/base.pyi
```diff
@@ -11,15 +12,19 @@ from django.db.backends.base.operations import BaseDatabaseOperations
 from django.db.backends.base.schema import BaseDatabaseSchemaEditor
 from django.db.backends.base.validation import BaseDatabaseValidation
 from django.db.backends.utils import CursorDebugWrapper, CursorWrapper
+from django.utils.functional import cached_property
 from typing_extensions import Self, TypeAlias
 
 NO_DB_ALIAS: str
 RAN_DB_VERSION_CHECK: set[str]
 
+logger: Logger
```
sobolevn, line 20: nit

```suggestion
logger: Logger

```

django-stubs/db/backends/base/base.pyi
```diff
@@ -60,9 +65,9 @@ class BaseDatabaseWrapper:
     operators: MutableMapping[str, str]
     def __init__(self, settings_dict: dict[str, Any], alias: str = ...) -> None: ...
     def ensure_timezone(self) -> bool: ...
-    @property
+    @cached_property
```
sobolevn: Let's not touch these at the moment, there are several PRs to handle property/cached_property.

django-stubs/db/backends/base/features.pyi
```diff
@@ -120,14 +127,19 @@ class BaseDatabaseFeatures:
...
     test_now_utc_template: str | None
     django_test_expected_failures: set[str]
     django_test_skips: dict[str, set[str]]
     connection: BaseDatabaseWrapper
     def __init__(self, connection: BaseDatabaseWrapper) -> None: ...
-    @property
+    @cached_property
     def supports_explaining_query_execution(self) -> bool: ...
-    @property
+    @cached_property
```
sobolevn: And these as well.

django-stubs/db/backends/dummy/features.pyi
```diff
@@ -1,5 +1,5 @@
 from django.db.backends.base.features import BaseDatabaseFeatures
 
 class DummyDatabaseFeatures(BaseDatabaseFeatures):
-    supports_transactions: bool
+    supports_transactions: bool  # type: ignore[assignment]
```
sobolevn: Can you please explain this `ignore`? I don't get it :(

UnknownPlatypus: This is solved by the `cached_property` changes too so I'll remove this from the MR. The error is the following:
> django-stubs/db/backends/postgresql/features.pyi:23: error: Incompatible types in assignment (expression has type "bool", base class "BaseDatabaseFeatures" defined the type as "Callable[[BaseDatabaseFeatures], bool]")  [assignment]

sobolevn (approved):
Thank you!

PR #1803, opened 2023-10-26
Allow additional types for SQL parameters in `migrations.RunSQL()`

### LINKED ISSUE #1797: lists as raw sql parameters in a migration
# Bug report

## What's wrong

Sometimes parameters to a raw sql query are arrays, for example

```python
query = """
UPDATE posts
SET category = %s
WHERE category = ANY(%s);
"""
...
migrations.RunSQL([(query, ['new category', ['retired category', 'another retired category']])])

```

Howerver, it seems lists of strings are not accepted as valid types for parameters:

```
error: List item 0 has incompatible type "list[str]"; expected "str"  [list-item]
```

...
## How is that should be

This is valid, should not raise an error

## System information

- OS:
- `python` version: 3.11
- `django` version: 4.2.6
- `mypy` version:1.6.0

UnknownPlatypus:
# I have made things!

Depending on the backend, params proxied from `migrations.RunSQL` to the underlying  `BaseDatabaseSchemaEditor.execute` could also be lists, or any datastructure supported by the backend.

See various implementations of the `BaseDatabaseSchemaEditor` method `quote_value`, responsible of parsing params and convert them to strings:
 
https://github.com/django/django/blob/e67d3580edbee1a4b58d40875293714ac3fc6937/django/db/backends/base/schema.py#L193-L196

By doing this change, we now align more closely with `BaseDatabaseSchemaEditor.execute`stubs.

## Related issues

Fixes #1797


django-stubs/db/migrations/operations/special.pyi
```diff
@@ -17,14 +17,14 @@ class SeparateDatabaseAndState(Operation):
...
-    reverse_sql: str | None | _ListOrTuple[str | tuple[str, dict[str, Any] | _ListOrTuple[str] | None]]
+    sql: str | _ListOrTuple[str | tuple[str, dict[str, Any] | _ListOrTuple[Any] | None]]
+    reverse_sql: str | None | _ListOrTuple[str | tuple[str, dict[str, Any] | _ListOrTuple[Any] | None]]
     state_operations: Sequence[Operation]
     hints: Mapping[str, Any]
     def __init__(
         self,
-        sql: str | _ListOrTuple[str | tuple[str, dict[str, Any] | _ListOrTuple[str] | None]],
-        reverse_sql: str | None | _ListOrTuple[str | tuple[str, dict[str, Any] | _ListOrTuple[str] | None]] = ...,
+        sql: str | _ListOrTuple[str | tuple[str, dict[str, Any] | _ListOrTuple[Any] | None]],
```
sobolevn: We now don't need `_ListOrTuple[str | tuple[str, dict[str, Any]` part

UnknownPlatypus: Are you sure about that ? I feel like every variant is needed here: sql il either:
- `str`: -> `RunSQL(sql="SOME SQL")`
- `_ListOrTuple[str]` -> `RunSQL(sql=("SOME SQLS", "SOME SQLS"))` / `RunSQL(sql=["SOME SQLS", "SOME SQLS"])`
- `_ListOrTuple[tuple[str, dict[str, Any]]` -> `RunSQL(sql=[("SOME SQLS %(VAL)s", {"VAL": "FOO"})]`
- `_ListOrTuple[tuple[str, _ListOrTuple[Any]]` ->`RunSQL(sql=[("SOME SQLS %s, %s", [object(), "ANOTHER PARAM"])])`
- `_ListOrTuple[tuple[str, None]` -> `RunSQL([("INSERT INTO musician (name) VALUES ('Reinhardt');", None)])`

I don't really see how it can be simplified, am I missing something here ?

btw, I'v added another test case and a TypeAlias to reduce duplication

sobolevn: We have `_ListOrTuple[Any]`, `Any` matches all possible inner types.

UnknownPlatypus: This is only part of the union for the second member of the tuple, not for the whole thing
```python
str | 
_ListOrTuple[
    str |
    tuple[
        str, 
        dict[str, Any] | _ListOrTuple[Any] | None
    ]
]
```

sobolevn (approved):
Got it, thanks!

PR #1805, opened 2023-10-30
Improve `ManyToManyDescriptor` and fix `Model.<manytomany>.through` typing

### LINKED ISSUE #1800: `Model.<related_name>` has no attribute "through"
# Bug report

## What's wrong
I confess, I am quite new to typing, so if this is a mypy issue and not a django-stubs issue, I apologise. I think I am reporting this in the right place though. 

I have a ManyToManyField on a model:
```
class Post(models.Model):
   ...
    identifiers = models.ManyToManyField(Identifier, verbose_name=_("Identifiers"), related_name="posts")
```
And a `TabularInline` in admin for the `Identifier` model: 
```
class PostInline(admin.TabularInline):
    model = Identifier.posts.through
    ...
```
But mypy does not like this :
```
error: "RelatedManager[Post]" has no attribute "through"  [attr-defined]
```
## How is that should be

No error should be raised because `Identifier.posts.through` is valid

## System information

- OS: WSL2 under windows 11
`Linux LMAVSURFACE4 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`
- `python` version: 3.11
- `django` version: 4.2.6
- `mypy` version: 1.6.1
- `mypy-extensions` version:  1.0.0

flaeppe:
`ManyToManyDescriptor` is now extended to take 1 new type argument, which is the target model/other side of the relation.

The plugin is updated to:

- Set a `ManyToManyDescriptor` instance instead of a related manager, for reverse relations of a `ManyToManyField`
- Produce a manager class with `ManyRelatedManager` and a model's default manager as bases for both sides of a many-to-many relation

## Related issues

Closes #1800


sobolevn (commented):
Generally LGTM, just one idea: maybe we should add at least one test that checks an error from this new guarantee that we added?

django-stubs/db/models/fields/related_descriptors.pyi
```diff
@@ -103,13 +104,17 @@ class ManyToManyDescriptor(ReverseManyToOneDescriptor, Generic[_M]):
...
     @property
-    def through(self) -> type[_M]: ...
+    def through(self) -> type[_Through]: ...
     @property
-    def related_manager_cls(self) -> type[ManyRelatedManager[Any]]: ...  # type: ignore[override]
+    def related_manager_cls(self) -> type[ManyRelatedManager[_To]]: ...  # type: ignore[override]
+    @overload  # type: ignore[override]
+    def __get__(self, instance: None, cls: Any = ...) -> Self: ...
+    @overload
+    def __get__(self, instance: Model, cls: Any = ...) -> ManyRelatedManager[_To]: ...
```
flaeppe, line 131: Just realised that the plugin needs to make the returned `ManyRelatedManager` here more accurate, so that it's possible to find custom queryset/manager methods.

Quite surprised we're missing tests for that actually.

It probably would've been easiest if `ManyRelatedManager` was another type argument, but I don't think it can. Since it's not possible to pass any dynamically generated managers as type arguments (ref #1699).

flaeppe: I've adjusted so that we imitate Django's behaviour by having the plugin create a manager class with `ManyRelatedManager` and the `_To` model's default manager as bases.

Ref: https://github.com/django/django/blob/46df3ab244e1688bd186f0bfbfea6a354097a910/django/db/models/fields/related_descriptors.py#L1032

flaeppe:
> Generally LGTM, just one idea: maybe we should add at least one test that checks an error from this new guarantee that we added?

Some parts changed since you looked at it. But I've still only added "happy"/inclusive asserts, not sure what asserts on errors we'd like to add? I find it a bit hard since it kind of appears as having to test "everything else". e.g. accessing attributes unknown for `ManyToManyDescriptor`, but then I'd say we're testing attribute access in mypy.

intgr:
Btw this needs merge/rebase

@sobolevn Do you have time to re-review this?

flaeppe:
I've rebased and resolved conflicts just now.

tests/typecheck/fields/test_related.yml
```diff
@@ -1110,6 +1129,12 @@
         main:20: note: Revealed type is "django.db.models.fields.related_descriptors.ForwardManyToOneDescriptor[django.db.models.fields.related.ForeignKey[Union[myapp.models.MyModel, django.db.models.expressions.Combinable], myapp.models.MyModel]]"
         main:21: note: Revealed type is "django.db.models.manager.Manager[myapp.models.MyModel_auto_through]"
         main:23: note: Revealed type is "Type[myapp.models.MyModel_other_again]"
+        main:24: note: Revealed type is "django.db.models.fields.related_descriptors.ManyToManyDescriptor[myapp.models.MyModel, myapp.models.MyModel_auto_through]"
+        main:25: note: Revealed type is "Type[myapp.models.MyModel_auto_through]"
+        main:26: note: Revealed type is "myapp.models.MyModel_auto_through"
+        main:28: note: Revealed type is "builtins.str"
+        main:29: note: Revealed type is "builtins.str"
+        main:30: note: Revealed type is "def (*objs: Union[myapp.models.MyModel, builtins.int], *, bulk: builtins.bool =)"
```
intgr: mypy 1.7 fix

```suggestion
        main:30: note: Revealed type is "def (*objs: Union[myapp.models.MyModel, builtins.int], bulk: builtins.bool =)"
```

django-stubs/db/models/fields/related_descriptors.pyi
```diff
@@ -136,7 +141,7 @@ class ManyRelatedManager(BaseManager[_M], Generic[_M]):
     async def aset(self, objs: QuerySet[_M] | Iterable[_M | int], *, bulk: bool = ..., clear: bool = ...) -> None: ...
     def clear(self) -> None: ...
     async def aclear(self) -> None: ...
-    def __call__(self, *, manager: str) -> ManyRelatedManager[_M]: ...
+    def __call__(self, *, manager: str) -> Self: ...
```
intgr: This seems suspect to me. Looking at the code, it resolves another manager from the Model and creates a ManyRelatedManager from it.

AFAICT they may be different manager classes and even have different base classes. So the previous annotation would be more accurate.

flaeppe: Yeah, either isn't good enough. Getting this right is a whole other ticket (and then some).

I can't recall why I changed it to be honest.

I'll revert it.

flaeppe: @intgr totally forgot about this. I've reverted just now

intgr: No problem. Sorry, I haven't had time to review this thoroughly yet.

mypy_django_plugin/main.py
```diff
@@ -187,6 +187,11 @@ def get_method_hook(self, fullname: str) -> Optional[Callable[[MethodContext], M
             info = self._get_typeinfo_or_none(class_fullname)
             if info and info.has_base(fullnames.FORM_MIXIN_CLASS_FULLNAME):
                 return forms.extract_proper_type_for_get_form
+        elif method_name == "__get__" and class_fullname in {
```
intgr, line 191: Let's keep a blank line between each of these `if/elif` conditions, like the rest of them do.

```suggestion

        elif method_name == "__get__" and class_fullname in {
```

mypy_django_plugin/transformers/models.py
```diff
@@ -487,13 +479,25 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
                     Instance(
-                        reverse_one_to_one_descriptor,
+                        self.reverse_one_to_one_descriptor,
                         [Instance(self.model_classdef.info, []), Instance(related_model_info, [])],
                     ),
                 )
                 continue
-
-            if isinstance(relation, ForeignObjectRel):
+            elif isinstance(relation, ManyToManyRel):
```
intgr, line 488:
```suggestion

            elif isinstance(relation, ManyToManyRel):
```

mypy_django_plugin/transformers/models.py
```diff
@@ -487,13 +479,25 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+                assert relation.through is not None
+                through_fullname = helpers.get_class_fullname(relation.through)
+                through_model_info = self.lookup_typeinfo_or_incomplete_defn_error(through_fullname)
+                self.add_new_node_to_model_class(
+                    attname,
+                    Instance(
+                        self.many_to_many_descriptor, [Instance(to_model_info, []), Instance(through_model_info, [])]
+                    ),
+                )
+            else:
```
intgr, line 502:
```suggestion

            else:
```

mypy_django_plugin/transformers/models.py
```diff
@@ -487,13 +479,25 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
-                        reverse_one_to_one_descriptor,
+                        self.reverse_one_to_one_descriptor,
                         [Instance(self.model_classdef.info, []), Instance(related_model_info, [])],
                     ),
                 )
                 continue
-
-            if isinstance(relation, ForeignObjectRel):
+            elif isinstance(relation, ManyToManyRel):
+                # TODO: 'relation' should be based on `TypeInfo` instead of Django runtime..
```
intgr:
```suggestion
                # TODO: 'relation' should be based on `TypeInfo` instead of Django runtime.
```

mypy_django_plugin/transformers/manytomany.py
```diff
@@ -151,3 +151,52 @@ def get_model_from_expression(
...
+    # This is a call to '__get__' overload with a model instance of 'ManyToManyDescriptor'.
+    # Returning a 'ManyRelatedManager'. Which we want to, just like Django, build from the
+    # default manager of the related model.
+    many_related_manager = ctx.default_return_type
+    # Require first type argument of 'ManyRelatedManager' to be a model
+    if (
+        not many_related_manager.args
+        or not isinstance(many_related_manager.args[0], Instance)
+        or many_related_manager.args[0].type.metaclass_type is None
+        or many_related_manager.args[0].type.metaclass_type.type.fullname != fullnames.MODEL_METACLASS_FULLNAME
```
intgr: Nit: for some reason I find it difficult to reason about these conditions because they use "inverted logic"

Would make more sense to me as `if args and isinstance(..., Instance) and ... is not None and ...fullname == MODEL_METACLASS_FULLNAME`

Possibly you could refactor these into another function e.g. `related_model_instance = get_related_model_instance(ctx)` that returns `Optional[Type]` and avoid the "inverted logic", without nesting the code too deep. But I'll leave it up to you to decide whether it's worth it.

Could also reduce repetition of `return ctx.default_return_type` a little.

django-stubs/db/models/fields/related.pyi
```diff
@@ -282,7 +282,7 @@ class ManyToManyField(RelatedField[Any, Any], Generic[_To, _M]):
     ) -> None: ...
     # class access
     @overload
-    def __get__(self, instance: None, owner: Any) -> ManyToManyDescriptor[_M]: ...
+    def __get__(self, instance: None, owner: Any) -> ManyToManyDescriptor[_To, _M]: ...
```
intgr: The naming you used in ManyToManyDescriptor `_To, _Through` makes more sense to me than `_To, _M`.

Would be nice to introduce `_Through` TypeVar for `ManyToManyField` as well.

intgr (approved):
Looks good to me, only had some nitpick comments.

flaeppe:
Thanks for the review. I'll look in to the needed fixes as soon as I'm able to squeeze in some time

mypy_django_plugin/transformers/manytomany.py
```diff
@@ -151,3 +151,57 @@ def get_model_from_expression(
         if model_info is not None:
             return Instance(model_info, [])
     return None
+
+
+def get_related_manager_and_model(ctx: MethodContext) -> Optional[tuple[Instance, Instance]]:
```
intgr: You broke Python 3.8

```suggestion
def get_related_manager_and_model(ctx: MethodContext) -> Optional[Tuple[Instance, Instance]]:
```

intgr (approved):
Thanks, looks good. I like how you refactored `get_related_manager_and_model()`.

intgr:
This issue is likely related:

* #1861

PR #1806, opened 2023-10-30
Move plugin generated `<Model>_RelatedManager` entries to allowlist

flaeppe:
Cleaned up some allowlist entries that are plugin generated.

Ref:

https://github.com/typeddjango/django-stubs/blob/53d0c22575dd6493ffde39e91833e0a2edc8a213/mypy_django_plugin/transformers/models.py#L557-L563


intgr (approved):
Makes sense, thanks

sobolevn (approved):
Thanks!

PR #1811, opened 2023-11-01
Improve hints of `django.contrib.gis.gdal.libgdal`

flaeppe:
Found these while trying to fix `gis` for my local environment. Not sure anyone will ever benefit from them, but was a bit fun to try to dig up anyways.
So I thought I might add them

### DIFF
--- a/django-stubs/contrib/gis/gdal/libgdal.pyi
+++ b/django-stubs/contrib/gis/gdal/libgdal.pyi
@@ -7,7 +7,7 @@ from typing import Any
 from _ctypes import _CData
 
 logger: Logger
-lib_path: str
+lib_path: str | None
 lib_names: list[str] | None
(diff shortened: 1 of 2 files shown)


django-stubs/contrib/gis/gdal/libgdal.pyi
```diff
@@ -1,21 +1,35 @@
+import sys
+from collections.abc import Callable, Sequence
+from ctypes import CDLL, _FuncPointer, _NamedFuncPointer
 from logging import Logger
 from typing import Any
 
+from _ctypes import _CData
+
 logger: Logger
 lib_path: str
```
intgr: Looks like this can be None

```suggestion
lib_path: str | None
```

intgr (approved):
Thanks

PR #1812, opened 2023-11-01
Allow custom Form class in `SuccessMessageMixin.form_valid()`

### LINKED ISSUE #1809: `SuccessMessageMixin` does not allow `form_valid` to be retyped
# Bug report

## What's wrong

When using the `SuccessMessageMixin`, it is impossible to retype `form_valid`, because it is already defined by `SuccessMessage`. #86 seems somewhat related, however, this bug is about the `form` argument.

```python
from django import forms
from django.contrib.messages.views import SuccessMessageMixin
from django.forms import ModelForm


class MyForm(forms.Form):
    pass


class Test(SuccessMessageMixin, ModelForm):
    def form_valid(self, form: MyForm):
        pass
```
This results in 
> evap/staff/views.py:1068: error: Argument 1 of "form_valid" is incompatible with supertype "SuccessMessageMixin"; supertype defines the argument type as "BaseForm"  [override]  
evap/staff/views.py:1068: note: This violates the Liskov substitution principle  
evap/staff/views.py:1068: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides  
Found 1 error in 1 file (checked 122 source files)

## How it should be

It should be possible to narrow the type of `field` in `form_valid`. It should be possible, if `SuccessMessageMixin` becomes generic to supply the type of the filed:

```python
from django import forms
from django.contrib.messages.views import SuccessMessageMixin
from django.forms import ModelForm


class MyForm(forms.Form):
    pass


class Test(SuccessMessageMixin[MyForm], ModelForm):
    def form_valid(self, form: MyForm):
        pass
```

## System information

- OS: Arch Linux
- `python` version: 3.10
- `django` version: 4.2.0
- `mypy` version: 1.6.1

SukiCZ:
# I have made \<first contrib here\> things!

Allow to retype `form_valid` in `SuccessMessageMixin` as suggested in relate issue

```py
from django import forms
from django.contrib.messages.views import SuccessMessageMixin
from django.forms import ModelForm


class MyForm(forms.Form):
    pass


class Test(SuccessMessageMixin[MyForm], ModelForm):
    def form_valid(self, form: MyForm):
        pass
```

## Related issues
Closes #1809 - `SuccessMessageMixin` does not allow `form_valid` to be retyped

Feedback: Thanks for awesome library 😻

### DIFF
--- a/django-stubs/contrib/messages/views.pyi
+++ b/django-stubs/contrib/messages/views.pyi
@@ -1,8 +1,12 @@
+from typing import Generic, TypeVar
+
 from django.forms.forms import BaseForm
 from django.http.response import HttpResponse
 from django.utils.functional import _StrOrPromise
 
-class SuccessMessageMixin:
+_F = TypeVar("_F", bound=BaseForm)
+
+class SuccessMessageMixin(Generic[_F]):
     success_message: _StrOrPromise
-    def form_valid(self, form: BaseForm) -> HttpResponse: ...
+    def form_valid(self, form: _F) -> HttpResponse: ...
     def get_success_message(self, cleaned_data: dict[str, str]) -> _StrOrPromise: ...
--- a/ext/django_stubs_ext/patch.py
+++ b/ext/django_stubs_ext/patch.py
@@ -4,6 +4,7 @@ from typing import Any, Generic, Iterable, List, Optional, Tuple, Type, TypeVar
 from django import VERSION
 from django.contrib.admin import ModelAdmin
 from django.contrib.admin.options import BaseModelAdmin
+from django.contrib.messages.views import SuccessMessageMixin
 from django.contrib.sitemaps import Sitemap
 from django.contrib.syndication.views import Feed
 from django.core.files.utils import FileProxyMixin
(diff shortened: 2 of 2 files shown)


tests/typecheck/contrib/messages/test_views.yml
```diff
@@ -0,0 +1,15 @@
...
+      from django.forms import ModelForm
+      from django.http.response import HttpResponse
+
+
+      class MyForm(forms.Form):
+          pass
+
+
+      class Test(SuccessMessageMixin[MyForm], ModelForm):
+          def form_valid(self, form: MyForm) -> HttpResponse:
```
intgr: Thanks, but we don't add tests covering ordinary stubs (except for particularly complex cases). Covering stubs with tests would require more test code than actual type stub code :)

django-stubs/contrib/messages/views.pyi
```diff
@@ -1,8 +1,12 @@
+from typing import Generic, TypeVar
+
 from django.forms.forms import BaseForm
 from django.http.response import HttpResponse
 from django.utils.functional import _StrOrPromise
 
-class SuccessMessageMixin:
+_F = TypeVar("_F", bound=BaseForm)
+
+class SuccessMessageMixin(Generic[_F]):
```
intgr, line 9: Since you made this generic, it needs to be added to `django-stubs-ext` monkeypatching: `_need_generic` in `ext/django_stubs_ext/patch.py` (otherwise `SuccessMessageMixin[MyForm]` won't be allowed)

intgr (changes requested):
Thanks!

intgr (approved):
:+1: Thanks for your help

PR #1814, opened 2023-11-02
Move `RelatedManager` to `django.db.models.fields.related_descriptors`

### LINKED ISSUE #1454: Add ManyRelatedManager and distinguish from RelatedManager

flaeppe:
Aligns with the module `RelatedManager` is locally declared

## Related issues

Closes #1454


django-stubs/db/models/fields/related_descriptors.pyi
```diff
@@ -86,6 +86,18 @@ class ReverseManyToOneDescriptor:
     def __get__(self, instance: Model, cls: Any = ...) -> type[RelatedManager[Any]]: ...
     def __set__(self, instance: Any, value: Any) -> NoReturn: ...
 
+class RelatedManager(BaseManager[_M], Generic[_M]):
```
intgr, line 90: I think a comment to the fact that this doesn't directly correspond to a runtime class is useful. Feel free to tweak.

```suggestion
# Fake class, Django defines RelatedManager inside a function body
class RelatedManager(BaseManager[_M], Generic[_M]):
```

Would make sense to add the same for `ManyRelatedManager` as well

intgr (approved):
:+1: Makes sense to move it to `related_descriptors` where Django itself defines it.

django-stubs/db/models/fields/related_descriptors.pyi
flaeppe: Am thinking to copy the comment from the `allowlist.txt` where these classes are also present

> 'RelatedManager' does exist and is declared locally, inside a function body

As well as

> 'ManyRelatedManager' does exist and is declared locally, inside a function body

flaeppe: Nvm, that didn't sound so good. Your suggestion reads better

django-stubs/db/models/fields/related_descriptors.pyi
```diff
@@ -86,6 +86,19 @@ class ReverseManyToOneDescriptor:
     def __get__(self, instance: Model, cls: Any = ...) -> type[RelatedManager[Any]]: ...
     def __set__(self, instance: Any, value: Any) -> NoReturn: ...
 
+# Fake class, Django defines 'RelatedManager' inside a function body
+class RelatedManager(BaseManager[_M], Generic[_M]):
```
intgr, line 90: You also changed the base class from `Manager` to `BaseManager`.

I suppose it's more correct (managers *usually* derive from `Manager`, but not always)? But this causes several new inspections in my projects. Seems like something that should be mentioned in release notes.

intgr: Hmm, I think we should export `RelatedManager` and `ManyRelatedManager` classes from `django-stubs-ext` to make their usage more convenient.

Right know if anyone wants to use them in type hints, they have to hide the import behind an `if TYPE_CHECKING:` condition and use quoting for the type, e.g. `blah_set: "RelatedManager[Blah]"`.

flaeppe: I changed base to align the 2 classes. But any of them is probably fine here really.

And yes, I agree, we should export them.

PR #1818, opened 2023-11-05
Simulate `@deconstructible` as a mixin class

flaeppe:
# I have made things!

`@desconstructible` inserts its own `__new__` method and adds a `deconstruct` method to decorated classes. We can simulate that behaviour with a class definition, which decorated classes inherits.

## Related issues

Ref: #1695


sobolevn (approved):
Great work!

UnknownPlatypus:
Nice, I was wondering how to fix this one, this is great :partying_face:

PR #1819, opened 2023-11-05
Improve types for `django.db.models.enums.*`

flaeppe:
# I have made things!

Main thing is to change writable attributes to read only, i.e. properties.


sobolevn (approved):
Thanks!

PR #1822, opened 2023-11-06
Improve types for multiple methods of `QuerySet`

flaeppe:
# I have made things!

Also includes some methods on `Manager` as that propagates due to how they're related in the stubs/plugin.


django-stubs/db/models/query.pyi
```diff
@@ -65,7 +66,7 @@ class _QuerySet(Generic[_T, _Row], Collection[_Row], Reversible[_Row], Sized):
     def __or__(self, other: _QuerySet[_T, _Row]) -> Self: ...
     # IMPORTANT: When updating any of the following methods' signatures, please ALSO modify
     # the corresponding method in BaseManager.
-    def iterator(self, chunk_size: int = ...) -> Iterator[_Row]: ...
+    def iterator(self, chunk_size: int | None = ...) -> Iterator[_Row]: ...
```
intgr, line 69: Nit: This is a departure from how we've ordinarily done things, but IMO it's helpful to include the default value verbatim when it's a primitive type. This is also accepted these days in `typeshed` for example.

```suggestion
    def iterator(self, chunk_size: int | None = None) -> Iterator[_Row]: ...
```

intgr (approved):
Thanks! Needs merge/rebase.

django-stubs/db/models/query.pyi
flaeppe: Hm, I see. I don't have any strong opinions about either alternative really. But surely nice to keep it consistent.

PR #1825, opened 2023-11-07
Fix `get_deleted_objects()` return type from `list[Model]` -> `list[str]`

### LINKED ISSUE #1824: Return from get_deleted_objects-method of ModelAdmin
# Bug report

## What's wrong
When we delete a certain object, we also want to delete some related objects and we want this to be visible in the Delete Confirmation View.

I got some typing issues regarding an override I did for `get_deleted_objects(self, objs, request)` in a subclass to `admin.ModelAdmin`.

Looking at the [Django documentation](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_deleted_objects) for this method it is said to return `tuple[list[str], dict[str, int], set[str], list[str]]`.

I also verified this in a debug session where the `deleted_objects` value was: `'Parameter set: <a href="/admin/onboard/parameterset/341/change/">341</a>'`.

However, in `django_stubs.contrib.admin.options.ModelAdmin.get_deleted_objects` I see the the deleted_objects is typed as `list[Model]`. The full return type is specified as `tuple[list[Model], dict[str, int], set[str], list[str]]`.

I rely on this being a `list[str]` as I append strings to this list using `django.utils.html.format_html`, which fails with the current definition using a `list[Model]`.

Just some example code to easier understand the issue (Not verified if this exact code raises the issue):
```python
class TestAdmin(admin.ModelAdmin):
    def get_deleted_objects(self, objs, request) -> tuple[list[str], dict[str, int], set[str], list[str]]:
        deleted_objects, model_count, perms_needed, protected = super().get_deleted_objects(objs, request)
        for obj in objs:
            deleted_objects.append(
                    format_html(
                        HTML_TEMPLATE,
                        obj.titleid,
                        obj.titleid,
                    )
            )
    return deleted_objects, model_count, perms_needed, protected
```

This would return a mypy-error:
> Return type "tuple[list[str], dict[str, int], set[str], list[str]]" of "get_deleted_objects" incompatible with return type "tuple[list[Model], dict[str, int], set[str], list[str]]" in supertype "ModelAdmin"

Just changing it to list[Model] doesn't work in this case, as the I get:
> Argument 1 to "append" of "list" has incompatible type "SafeString"; expected "Model"

## How is that should be
The return type of `admin.ModelAdmin.get_deleted_objects` should be changed from:
```python
def get_deleted_objects(
    self, objs: Sequence[_ModelT] | QuerySet[_ModelT], request: HttpRequest
) -> tuple[list[Model], dict[str, int], set[str], list[str]]: ...
```
```python
def get_deleted_objects(
    self, objs: Sequence[_ModelT] | QuerySet[_ModelT], request: HttpRequest
) -> tuple[list[str], dict[str, int], set[str], list[str]]: ...
```

## System information

- OS: Fedora Linux 38 (KDE Plasma) @ 6.5.9-200.fc38.x86_64 (64-bit)
- `python` version: 3.11.6
- `django` version: 4.2.7
- `mypy` version: 1.6.1

golgor:
# I have made things!
Changed the return type of admin.ModelAdmin.get_deleted_objects from list[Model] -> list[str] in accordance to the django documentation.

## Related issues

Format is:
- Closes: #1824 
- Refs
  - [Django docs for admin.ModelAdmin.get_deleted_objects](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_deleted_objects).


django-stubs/contrib/admin/options.pyi
```diff
@@ -265,7 +265,7 @@ class ModelAdmin(BaseModelAdmin[_ModelT]):
     def changelist_view(self, request: HttpRequest, extra_context: dict[str, Any] | None = ...) -> HttpResponse: ...
     def get_deleted_objects(
         self, objs: Sequence[_ModelT] | QuerySet[_ModelT], request: HttpRequest
-    ) -> tuple[list[Model], dict[str, int], set[str], list[str]]: ...
+    ) -> tuple[list[str], dict[str, int], set[str], list[str]]: ...
```
intgr, line 268: Thanks! Looks like the related `get_deleted_objects` function in `contrib/admin/utils.pyi` should be updated too.

intgr (changes requested):

django-stubs/contrib/admin/options.pyi
golgor: Fixed in [c99d322](https://github.com/typeddjango/django-stubs/pull/1825/commits/c99d322874b5b74157f681232c4471f8481779e9)

intgr (approved):
Thanks for the PR! In the future, feel free to skip creating an issue and open PR straight away.

golgor:
> Thanks for the PR! In the future, feel free to skip creating an issue and open PR straight away.

Will do, thanks for the quick response and feedback!

PR #1829, opened 2023-11-07
Various improvements in `django.core.management.commands` modules

### LINKED ISSUE #1820: TemplateCommand subclasses `handle` method typing is incorrect
# Bug report

## What's wrong
The `TemplateCommand` class stub (`django-stubs/core/management/templates.pyi`) defines the `handle` command as follows:
```py
def handle(self, app_or_project: str, name: str, target: str | None = ..., **options: Any) -> None: ...
```

We can see that in subclasses (startapp and startproject), Django actually does not follow the super's definition, instead manually parsing kwargs to pass into the super's method:
```py
# django.core.management.commands.startapp
class Command(TemplateCommand):
    help = (
        "Creates a Django app directory structure for the given app name in "
        "the current directory or optionally in the given directory."
    )
    missing_args_message = "You must provide an application name."

    def handle(self, **options):
        app_name = options.pop("name")
        target = options.pop("directory")
        super().handle("app", app_name, target, **options)
```

## How is that should be
This is an issue where Django itself is kinda violating the principles of inheritance by havind a different definition in the child class; however, I believe the stubs should still reflect this behaviour, and should define the class methods for the respective commands in the way they are actually used:
```py
def handle(self, **options: Any) -> Optional[str]: ...
```

## System information

- OS:
- `python` version: 3.11
- `django` version: 4.2.7
- `mypy` version: 1.6.1

UnknownPlatypus:
# I have made things!

Update multiple Django commands to match closely runtime signature (especially for the `handle` hook)

## Related issues

Fixes #1820


UnknownPlatypus:
@intgr for review maybe ? I noticed no-one was assigned and you seem to be following the related issue.

django-stubs/core/management/commands/startapp.pyi
```diff
@@ -1,4 +1,7 @@
+from typing import Any
+
 from django.core.management.templates import TemplateCommand
 
 class Command(TemplateCommand):
     missing_args_message: str
+    def handle(self, **options: Any) -> None: ...  # type: ignore[override]
```
intgr, line 8: Empty line for consistency with other files:

```suggestion

    def handle(self, **options: Any) -> None: ...  # type: ignore[override]
```

django-stubs/core/management/commands/flush.pyi
```diff
@@ -1,6 +1,10 @@
+from typing import Any
+
 from django.core.management.base import BaseCommand
 from django.core.management.color import Style
 
 class Command(BaseCommand):
-    stealth_options: tuple[str]
+    stealth_options: tuple[str, str, str]
```
intgr: Let's remove this. `BaseCommand` already defines `stealth_options: tuple[str, ...]`

```suggestion
```

intgr (commented):
Thanks! A few minor tweaks.

django-stubs/core/management/commands/flush.pyi
UnknownPlatypus: I removed this from `inpsectdb.pyi` too for consistency

intgr (approved):
🥳

PR #1830, opened 2023-11-08
Add types for `MD5` database function

kevinmarsh:
# I have made things!

Added `django.db.models.functions.MD5` (which was added to Django way back in 3.0)

- https://docs.djangoproject.com/en/stable/ref/models/database-functions/#md5
- https://github.com/django/django/blob/3d2370607dafc16f744a1dd824066a2b4c771a52/django/db/models/functions/text.py#L208-L210


sobolevn (approved):
Thanks!

PR #1832, opened 2023-11-09
Constrain multiple `BaseModelAdmin` attributes to be either list or tuple

flaeppe:
Attributes are aligned with Django admin's default checks list found here https://docs.djangoproject.com/en/4.2/ref/checks/#admin

### DIFF
--- a/django-stubs/contrib/admin/options.pyi
+++ b/django-stubs/contrib/admin/options.pyi
@@ -76,19 +76,19 @@ _ListFilterT: TypeAlias = (
 _ModelT = TypeVar("_ModelT", bound=Model)
 
 class BaseModelAdmin(Generic[_ModelT]):
-    autocomplete_fields: Sequence[str]
-    raw_id_fields: Sequence[str]
+    autocomplete_fields: _ListOrTuple[str]
+    raw_id_fields: _ListOrTuple[str]
     fields: _FieldGroups | None
-    exclude: Sequence[str] | None
+    exclude: _ListOrTuple[str] | None
     fieldsets: _FieldsetSpec | None
     form: type[forms.ModelForm[_ModelT]]
-    filter_vertical: Sequence[str]
-    filter_horizontal: Sequence[str]
+    filter_vertical: _ListOrTuple[str]
+    filter_horizontal: _ListOrTuple[str]
     radio_fields: Mapping[str, _Direction]
     prepopulated_fields: dict[str, Sequence[str]]
     formfield_overrides: Mapping[type[Field], Mapping[str, Any]]
-    readonly_fields: Sequence[str]
-    ordering: Sequence[str] | None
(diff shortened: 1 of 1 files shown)


django-stubs/contrib/admin/options.pyi
```diff
@@ -76,19 +76,19 @@ _ListFilterT: TypeAlias = (
...
-    filter_vertical: Sequence[str]
-    filter_horizontal: Sequence[str]
+    filter_vertical: _ListOrTuple[str]
+    filter_horizontal: _ListOrTuple[str]
     radio_fields: Mapping[str, _Direction]
     prepopulated_fields: dict[str, Sequence[str]]
     formfield_overrides: Mapping[type[Field], Mapping[str, Any]]
-    readonly_fields: Sequence[str]
-    ordering: Sequence[str] | None
+    readonly_fields: _ListOrTuple[str] | None
```
intgr, line 90: I don't think `readonly_fields` is allowed to be `None`?

django-stubs/contrib/admin/options.pyi
```diff
@@ -76,19 +76,19 @@ _ListFilterT: TypeAlias = (
 _ModelT = TypeVar("_ModelT", bound=Model)
 
 class BaseModelAdmin(Generic[_ModelT]):
-    autocomplete_fields: Sequence[str]
-    raw_id_fields: Sequence[str]
+    autocomplete_fields: _ListOrTuple[str]
+    raw_id_fields: _ListOrTuple[str]
     fields: _FieldGroups | None
-    exclude: Sequence[str] | None
+    exclude: _ListOrTuple[str] | None
```
intgr, line 82: Should we update the getter methods below as well, e.g. `get_exclude(...) -> Sequence[str]` to `_ListOrTuple`?

sobolevn (approved):
Thank you!

intgr (changes requested):
Thanks

django-stubs/contrib/admin/options.pyi
flaeppe: Oh, nice find. This was a copy paste mistake..

django-stubs/contrib/admin/options.pyi
flaeppe: Hm, I'm not totally sure but it could probably be reasonable to align? I mean, per default these methods just do `return self.<attribute>`

intgr: Yeah my preference would be to update methods as well.

flaeppe: Yeah, I'll push a new PR with fixes very soon. Trying to check their alignment just now.

flaeppe: See #1833

PR #1833, opened 2023-11-09
Constrain multiple `BaseModelAdmin` methods to return either list or tuple

flaeppe:
The get methods serves as extension points for corresponding attributes on the admin class. Thus we align the return type of the the methods with their corresponding attribute's type.

## Related issues

- Ref: #1832


flaeppe:
Let me know if I missed any get method.

django-stubs/contrib/admin/options.pyi
```diff
@@ -107,18 +107,18 @@ class BaseModelAdmin(Generic[_ModelT]):
...
     def get_fieldsets(self, request: HttpRequest, obj: _ModelT | None = ...) -> _FieldsetSpec: ...
     def get_inlines(self, request: HttpRequest, obj: _ModelT | None) -> list[type[InlineModelAdmin]]: ...
-    def get_ordering(self, request: HttpRequest) -> Sequence[str]: ...
-    def get_readonly_fields(self, request: HttpRequest, obj: _ModelT | None = ...) -> Sequence[str]: ...
+    def get_ordering(self, request: HttpRequest) -> _ListOrTuple[str]: ...
+    def get_readonly_fields(self, request: HttpRequest, obj: _ModelT | None = ...) -> _ListOrTuple[str]: ...
     def get_prepopulated_fields(self, request: HttpRequest, obj: _ModelT | None = ...) -> dict[str, Sequence[str]]: ...
     def get_queryset(self, request: HttpRequest) -> QuerySet[_ModelT]: ...
-    def get_sortable_by(self, request: HttpRequest) -> Sequence[str]: ...
+    def get_sortable_by(self, request: HttpRequest) -> _ListOrTuple[str] | _DisplayT: ...
```
flaeppe: This was a little surprise if-statement found that returns `get_list_display` as a default:

https://github.com/django/django/blob/d254a54e7f65e83d8971bd817031bc6af32a7a46/django/contrib/admin/options.py#L424-L430

Not sure if we want to refine this to only be `_DisplayT` or something?

django-stubs/contrib/admin/options.pyi
```diff
@@ -107,18 +107,18 @@ class BaseModelAdmin(Generic[_ModelT]):
     def formfield_for_manytomany(
         self, db_field: ManyToManyField, request: HttpRequest, **kwargs: Any
     ) -> ModelMultipleChoiceField | None: ...
-    def get_autocomplete_fields(self, request: HttpRequest) -> Sequence[str]: ...
+    def get_autocomplete_fields(self, request: HttpRequest) -> _ListOrTuple[str]: ...
     def get_view_on_site_url(self, obj: _ModelT | None = ...) -> str | None: ...
     def get_empty_value_display(self) -> SafeString: ...
-    def get_exclude(self, request: HttpRequest, obj: _ModelT | None = ...) -> Sequence[str] | None: ...
+    def get_exclude(self, request: HttpRequest, obj: _ModelT | None = ...) -> _ListOrTuple[str] | None: ...
```
intgr, line 114: Cool! Previously this wasn't considered an error because `str` satisfies `Sequence[str]` 🙄 

```python
    def get_exclude(self, *args: Any, **kwargs: Any) -> str:
        return "a"
```

django-stubs/contrib/admin/options.pyi
intgr: Looks like any `_ListOrTuple[str]` is compatible with `_DisplayT`, so makes sense.

intgr (approved):
Thanks

django-stubs/contrib/admin/options.pyi
flaeppe: Yes this is precisely what I ran in to: I wrote

```python
ordering = "-date_created"
```

And was surprised when CI passed for mypy but not for system checks..

django-stubs/contrib/admin/options.pyi
flaeppe: I updated to only have `_DisplayT`

PR #1834, opened 2023-11-09
django-stubs-ext: Export `RelatedManager`, `ManyRelatedManager` stub-only classes

intgr:
# I have made things!

As discussed in https://github.com/typeddjango/django-stubs/pull/1814#discussion_r1384844354, `RelatedManager`, `ManyRelatedManager` are sometimes useful for type hinting. But these classes don't exist as is at Django runtime, rather they are defined inside function bodies.

* When `TYPE_CHECKING`, we re-export `django-stubs` fake classes.
* At runtime, we define these as `Protocol[_T]`. This has the advantage that Python prevents them being used with `isinstance()`. It fails with error:

  > TypeError: Instance and class checks can only be used with @runtime_checkable protocols

Usage before:

```python
if TYPE_CHECKING:
    from django.db.models.manager import RelatedManager  # before PR #1814

def get_manager() -> "RelatedManager[MyModel]": ...
```

Usage after:

```python
from django_stubs_ext.db.models.manager import RelatedManager

def get_manager() -> RelatedManager[MyModel]: ...
```

## Related issues

* https://github.com/typeddjango/django-stubs/pull/1814


intgr:
Bikeshedding question: Right now I've defined these as `django_stubs_ext.db.models.RelatedManager`.

But Django itself defines these in `django.db.models.fields.related_descriptors`. Do we want to mirror this structure in django-stubs-ext?

E.g. `django_stubs_ext.db.models.fields.related_descriptors.RelatedManager` (Seems that Django itself considers related_descriptors like an implementation detail)

Or maybe `django_stubs_ext.db.models.managers.RelatedManager` is more future-proof if we add more manager-related things in the future?

I actually slightly prefer the last one: `django_stubs_ext.db.models.managers.RelatedManager`

flaeppe:
Nice!

I don't see it necessary to keep a corresponding import path in this case, as it's typing specific and the classes are not importable from Django anyways.

`django_stubs_ext.db.models.managers.RelatedManager` looks like a relevant import path to me!

intgr:
Moved to `django_stubs_ext.db.models.managers`.

Also had to rename `db/models.py` -> `db/models/__init__.py`

intgr:
Oh wait, Django itself has `django.db.models.manager` (singular), I'll use singular here too for consistency:  `django_stubs_ext.db.models.managers` -> `manager`

flaeppe (approved):

flaeppe:
Side note: adding the `django-stubs-ext` label could probably be automated with https://github.com/actions/labeler (one of my favourite actions)

intgr:
> adding the django-stubs-ext label could probably be automated

Cool, yes. Also `mypy-plugin` label would be useful (so far we've mostly used it for issues, not PRs).

UnknownPlatypus:
> Side note: adding the `django-stubs-ext` label could probably be automated with https://github.com/actions/labeler (one of my favourite actions)

This looks really nice 😯

PR #1835, opened 2023-11-09
Remove incorrect `django.db.models` re-exports

intgr:
# I have made things!

Trying to import these items in Django 4.2.7 fails with `ImportError` at runtime. Not sure if they were removed, or never worked in the first place.

```
>>> from django.db.models import BaseManager
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'BaseManager' from 'django.db.models' (.../site-packages/django/db/models/__init__.py)
```

## Related issues

Discovered when I was testing #1834.


sobolevn (approved):
Thank you!

PR #1841, opened 2023-11-15
Make `StrPromise` not inherit from `Sequence[str]`

### LINKED ISSUE #1626: `capfirst` annotation does not appear usable -- or probably anything using `_StrOrPromiseT`
# Bug report

it appears that the `_StrOrPromiseT` in `capfirst` decays to `Sequence[str]` -- probably because mypy infers the union of `str | _StrPromise` to be `Sequence[str]` (because `str` is also a `Sequence[str]`)

## What's wrong

the original error comes from this line -- though I've isolated the bug to a standalone script: https://github.com/getsentry/sentry/blob/c19f762c70afa9145efb10e8a524498316dcea4e/src/sentry/web/forms/accounts.py#L79

here's the standalone version:

```python
from django.utils.translation import gettext_lazy
from django.utils.text import capfirst

if TYPE_CHECKING:
    from django.utils.functional import _StrPromise

s: _StrPromise | str = gettext_lazy('example')
t = capfirst(s)
reveal_type(t)
```

```console
$ mypy t2.py 
t2.py:10: error: Value of type variable "_StrOrPromiseT" of "capfirst" cannot be "Sequence[str]"  [type-var]
t2.py:11: note: Revealed type is "Union[typing.Sequence[builtins.str], None]"
```

## How is that should be

if I change this instead to an overload it appears to work correctly:

```python
from typing import TYPE_CHECKING, overload
from django.utils.translation import gettext_lazy
from django.utils.text import capfirst

if TYPE_CHECKING:
    from django.utils.functional import _StrPromise


@overload
def capfirst2(x: None) -> None: ...
@overload
def capfirst2(x: str) -> str: ...
@overload
def capfirst2(x: _StrPromise) -> _StrPromise: ...
def capfirst2(x: str | _StrPromise | None) -> str | _StrPromise | None: return x


s: _StrPromise | str = gettext_lazy('example')
t = capfirst2(s)
reveal_type(t)
```

```console
$ mypy t.py 
t.py:21: note: Revealed type is "Union[django.utils.functional._StrPromise, builtins.str]"
Success: no issues found in 1 source file
```

let me know if you'd be interested in a patch for this or if there's some other ideas on how to fix this

## System information

- OS:
- `python` version: 3.8.16
- `django` version: (old, lol)
- `mypy` version: 1.4.1

intgr:
# I have made things!

Even though the problem reported in #1626 is also fixed in mypy 1.7, as discussed in https://github.com/typeddjango/django-stubs/issues/1626#issuecomment-1647527110, we should remove `Sequence[str]` from `StrPromise` base classes since there is no reason why it should subclass it.

## Related issues

* Fixes #1626


django-stubs/utils/functional.pyi
```diff
@@ -20,10 +20,10 @@ class Promise:
     def __radd__(self, other: Any) -> Any: ...
     def __deepcopy__(self, memo: Any) -> Self: ...
 
-class _StrPromise(Promise, Sequence[str]):
+class _StrPromise(Promise):
     def __add__(self, __s: str) -> str: ...
     # Incompatible with Sequence.__contains__
```
intgr:
```suggestion
```

flaeppe (approved):

sobolevn (approved):

PR #1842, opened 2023-11-15
Accurately infer `capfirst()` `None` return

intgr:
# I have made things!

While looking into #1626, I noticed that `capfirst()` is hinted as returning `... | None` even when input is never `None`. This can be fixed with a new `TypeVar`.

```python
reveal_type(capfirst('constant string'))
# Before: Revealed type is "Union[builtins.str, None]"
# After: Revealed type is "builtins.str"
```


sobolevn (approved):
Question: what about `@overload`? Is it better or worse in this case?

intgr:
> what about `@overload`? Is it better or worse in this case?

I think they would work the same. I don't have a preference.

PR #1843, opened 2023-11-15
Restore `RelatedManager`, `ManyRelatedManager` to inherit from `Manager` not `BaseManager`

intgr:
Ping @flaeppe.

The base class of `RelatedManager` was changed from `Manager` to `BaseManager` in #1814. I did not notice the change when reviewing, but only discovered it later when testing `django-stubs` git version out on my projects.

It looks like this will cascade to many users, forcing users to replace `Manager` type hints with `BaseManager`, including `djangorestframework-stubs` and my own projects. So I'd expect some justification for the change, otherwise we should revert it (as this PR does).

I don't want to make a `django-stubs` release with this change, without having discussed and agreed that it's what we really want. Maybe we should just revert (merge this) for the time being and can continue this discussion later?

----

`BaseManager` seems to be intended as an impementation detail in Django -- it's not even mentioned anywhere in their documentation. Inspecting relation objects at Django runtime, they by default inherit from `Manager` also.

What are the use cases for creating custom manager classes from `BaseManager` instead of `Manager`, and are those use cases important enough that we want to move to `BaseManager`?

## Related issues

* #1814
* https://github.com/typeddjango/djangorestframework-stubs/pull/508

### DIFF
--- a/django-stubs/db/models/fields/related_descriptors.pyi
+++ b/django-stubs/db/models/fields/related_descriptors.pyi
@@ -126,7 +126,7 @@ class ManyToManyDescriptor(ReverseManyToOneDescriptor, Generic[_M]):
     def related_manager_cls(self) -> type[ManyRelatedManager[Any]]: ...  # type: ignore[override]
 
 # Fake class, Django defines 'ManyRelatedManager' inside a function body
-class ManyRelatedManager(BaseManager[_M], Generic[_M]):
+class ManyRelatedManager(Manager[_M], Generic[_M]):
     related_val: tuple[int, ...]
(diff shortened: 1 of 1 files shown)


flaeppe:
I changed for 1 of `RelatedManager` and `ManyRelatedManager` to align their bases. That was the sole reason, I took `BaseManager` without digging too deep, since that has all implementations. I'm totally fine with having `Manager` as base instead.

But then we should include a change of base for `ManyRelatedManager` in this change set too

intgr:
OK thanks. Any other opinions @adamchainz @sobolevn?

flaeppe (approved):
LGTM at least

PR #1847, opened 2023-11-18
Updated `TypedModelMeta` `ordering` attribute to allow `OrderBy` objects

### LINKED ISSUE #1846: TypedModelMeta only allows `list[str]` for `ordering`
Hi!

[TypedModelMeta.ordering](https://github.com/typeddjango/django-stubs/blob/master/ext/django_stubs_ext/db/models/__init__.py) currently only allows a Sequence of `str`, but accoding to the [docs,](https://docs.djangoproject.com/en/4.2/ref/models/options/#ordering) expressions like `OrderBy` should also be allowed

image

Example of valid input:
```py
class Entry(models.Model):
  type = models.CharField(max_length=100)
  item = models.CharField(max_length=100, null=True)

  class Meta(TypedModelMeta):
    ordering = ["type", F("item").asc(nulls_first=True)]
```

HansAarneLiblik:
Fixes https://github.com/typeddjango/django-stubs/issues/1846

I ONLY added support for `OrderBy` expression, because this is what is affecting me, and I don't know what else is. suported

### DIFF
--- a/ext/django_stubs_ext/db/models/__init__.py
+++ b/ext/django_stubs_ext/db/models/__init__.py
@@ -3,7 +3,7 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from typing import ClassVar, List, Literal, Sequence, Tuple, Union
 
-    from django.db.models import BaseConstraint, Index
+    from django.db.models import BaseConstraint, Index, OrderBy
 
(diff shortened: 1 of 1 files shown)


intgr (approved):
> I don't know what else is [supported]

I had a look at the doc, I think it's just `str` and `OrderBy`. Good work.

PR #1852, opened 2023-11-24
Fix Self typed custom queryset methods incompatible with base queryse…

### LINKED ISSUE #1840: `Self` typed custom queryset methods incompatible with base queryset type
# Bug report

## What's wrong

Custom queryset methods with a return type of `Self` have a return type incompatible with the base `QuerySet` class.

```python
class LogQuerySet(QuerySet["Log"]):

    def filter_a(self) -> QuerySet["Log"]:
        return self

    def filter_b(self) -> Self:
        return self

LogManager = Manager.from_queryset(LogQuerySet)

class Log(models.Model):
    objects = LogManager()

a: models.QuerySet[Log] = Log.objects.filter_a()  # OK
b: models.QuerySet[Log] = Log.objects.filter_b()  # Incompatible types in assignment
```
mypy reports:
> Incompatible types in assignment (expression has type "ManagerFromLogQuerySet[Log]", variable has type "_QuerySet[Log, Log]")  [assignment]

## How is that should be

It would be great if the return type was inferred to be a sub-type of `QuerySet` with the correct type, eg, `QuerySet[Log]`.

## System information

- OS:
- `python` version: 3.11.6/3.8.18
- `django` version: 3.2.23
- `mypy` version: 1.6.1

moranabadie:
# I have made things!

I made a mistake indeed, methods with -> Self  for as_manager or from_queryset managers returns an instance of the queryset, not the generated manager.

## Related issues

Closes #1840

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -109,19 +109,21 @@ def _process_dynamic_method(
     variables = method_type.variables
     ret_type = method_type.ret_type
 
+    if not is_fallback_queryset:
+        queryset_instance = Instance(queryset_info, manager_instance.args)
+    else:
+        # The fallback queryset inherits _QuerySet, which has two generics
+        # instead of the one exposed on QuerySet. That means that we need
+        # to add the model twice. In real code it's not possible to inherit
+        # from _QuerySet, as it doesn't exist at runtime, so this fix is
+        # only needed for plugin-generated querysets.
+        queryset_instance = Instance(queryset_info, [manager_instance.args[0], manager_instance.args[0]])
+
     # For methods on the manager that return a queryset we need to override the
     # return type to be the actual queryset class, not the base QuerySet that's
     # used by the typing stubs.
     if method_name in MANAGER_METHODS_RETURNING_QUERYSET:
-        if not is_fallback_queryset:
-            ret_type = Instance(queryset_info, manager_instance.args)
-        else:
-            # The fallback queryset inherits _QuerySet, which has two generics
-            # instead of the one exposed on QuerySet. That means that we need
-            # to add the model twice. In real code it's not possible to inherit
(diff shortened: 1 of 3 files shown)


moranabadie:
I have executed the tests without mypy and the types matches with mypy.

Du you agree with the two tests in the yml @MrkGrgsn @flaeppe ?

flaeppe (commented):
I think we would also gain from an assert that a return type of the explicit 'Self' is left untouched.

e.g. that a return type like below is left unchanged

```python
class BaseQuerySet(models.QuerySet["MyModel"]):
    def method(self) -> BaseQuerySet:
        return self

reveal_type(MyModel.objects.method())  # N: Revealed type is "myapp.models.BaseQuerySet"
```

MrkGrgsn:
Looks good. A couple of questions though.
1. The `as_manager` test case is more comprehensive than the `from_queryset` one although both scenarios should have the same result with respect to custom queryset methods returning `Self`. Would it make sense for the queryset set up and assertions to be the same in both?
2. Should we also assert that custom manager methods returning `Self` resolve correctly? I admit I can't think of a use-case for such a method!
```python
class CustomManager(BaseManager):
    def self_return(self) -> Self: ...

NewManager = CustomManager.from_queryset(ModelQuerySet)

reveal_type(MyModel.objects.self_return())  # N: Revealed type is "myapp.models.CustomManagerFromModelQuerySet"
```

moranabadie:
> ```python
> # N: Revealed type is "myapp.models.BaseQuerySet"
> ```

ok I just added the test for this

moranabadie:
> ```python
> # N: Revealed type is "myapp.models.CustomManagerFromModelQuerySet"
> ```

You are right, The two tests are now mostly the same.

I also added the CustomManager test :)

MrkGrgsn:
The updates look good to me 👍

flaeppe (approved):
Thanks

PR #1853, opened 2023-11-27
Refactor: Unify plugin check for model type info

flaeppe:
Replace a couple of repetitions with an auxiliary call

__Side note:__ Not sure if there's any additional gain, but might be fun to introduce some narrowing with a

```python
ModelTypeInfo = NewType("ModelTypeInfo", TypeInfo)
```

Which could be type guarded from the `def is_model_type(info: TypeInfo) -> TypeGuard[ModelTypeInfo]`

Probably possible to do for an `ModelInstance` as well.

But we should probably find a place where it could be utilised first.

## Related issues

Refs: #1805


sobolevn (approved):
Thanks! You can add a `NewType` in the next PR, let's see how it would work.

PR #1854, opened 2023-11-30
Added `Model._do_update()` method signature

pfouque:
# I have made things!

Add typing for a private method (used in django-fsm), which signature has not changed [since 2013](https://github.com/django/django/blame/main/django/db/models/base.py#L1067)

### DIFF
--- a/django-stubs/db/models/base.pyi
+++ b/django-stubs/db/models/base.pyi
@@ -52,9 +52,9 @@ class Model(metaclass=ModelBase):
     def _do_update(
         self,
         base_qs: QuerySet[Self],
-        using: Any,
+        using: str | None,
         pk_val: Any,
-        values: Collection[Any] | None,
+        values: Collection[tuple[Field, type[Model] | None, Any]],
         update_fields: Iterable[str] | None,
         forced_update: bool,
     ) -> bool: ...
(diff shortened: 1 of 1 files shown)


django-stubs/db/models/base.pyi
```diff
@@ -49,6 +49,15 @@ class Model(metaclass=ModelBase):
     def add_to_class(cls, name: str, value: Any) -> Any: ...
     @classmethod
     def from_db(cls, db: str | None, field_names: Collection[str], values: Collection[Any]) -> Self: ...
+    def _do_update(
+        self,
+        base_qs: QuerySet[Self],
+        using: Any,
```
intgr:
```suggestion
        using: str | None,
```

django-stubs/db/models/base.pyi
```diff
@@ -49,6 +49,15 @@ class Model(metaclass=ModelBase):
...
+        pk_val: Any,
+        values: Collection[Any] | None,
```
intgr: The only caller in Django code, `Model._save_table()`, always provides list of a 3-tuple. I think it can't be `None`, unless I'm missing something.

Synthesizing based on the fact that this is passed into `UpdateQuery.add_update_fields()`

```suggestion
        values: Collection[tuple[Field, type[Model] | None, Any]],
```

intgr (approved):
Looks good to me, added a few improvement suggestions.

pfouque:
Thanks for your help @intgr !

adamchainz:
Nice one all 👍

PR #1741, opened 2023-09-28
[5.0] Added many new a-prefixed asynchronous methods

bigfootjon:
# I have made things!

This handles the following commits that landed in the Django 5.0 merge window:
1. https://github.com/django/django/commit/5e98959d9242c57a55c65847758781f82d386fa4
2. https://github.com/django/django/commit/674c23999cb6982a9d447fedec4d72e135201fee
3. https://github.com/django/django/commit/b9473cac65190822e7c94f695f1f7b4d5b49502a
4. https://github.com/django/django/commit/fff14736f1cc594f79ea0e2656b8375d837b7aca

## Related issues

- Refs #1493

PR #1859, opened 2023-12-06
Update and prepare for Django 5.0

intgr:
# I have made thingies!

* Updated Django dependency to 5.0.0 (when using Python 3.10 or newer -- older Pythons continue with Django 4.2.7)
* Created a separate allowlist `allowlist_todo_django50.txt` for new stubtest errors.
* Bumped `setup.py` version to `5.0.0.dev1` for the time being. Final release will be `5.0.0`.
* Django version 4.1 is now officially EOL by Django project, but let's take that separately (there may be some cleanups we can do).

## Related issues

* #1493
* Closes #1855
* Prerequisite for #1741


intgr:
DOH! All pipelines with Python 3.8 and 3.9 are failing, because Django 5.0 is not installable with those Python versions (https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django).

> ERROR: Ignored the following versions that require a different python version: 5.0 Requires-Python >=3.10; 5.0a1 Requires-Python >=3.10; 5.0b1 Requires-Python >=3.10; 5.0rc1 Requires-Python >=3.10
> ERROR: Could not find a version that satisfies the requirement Django==5.0.0 (from versions: [...])
> ERROR: No matching distribution found for Django==5.0.0

requirements.txt
```diff
@@ -7,7 +7,8 @@ pytest-shard==0.1.2
 
 # Django deps:
 psycopg2-binary
-Django==4.2.7
+Django==4.2.7; python_version < '3.10'
+Django==5.0.0; python_version >= '3.10'
```
intgr, line 11: This one weird trick actually works...

intgr: Not sure if this can cause us problems down the road. I'm sure we will be able to solve them in the future as they pop up.

If we need to write Django 5.0-specific tests, we can specifically add `skip: sys.version_info < (3, 10)` to them.

README.md
```diff
@@ -49,6 +49,7 @@ We rely on different `django` and `mypy` versions:
 
 | django-stubs   | Mypy version | Django version | Django partial support | Python version |
 |----------------|--------------|----------------|------------------------|----------------|
+| (next release) | 1.7.x        | 5.0            | 4.2, 4.1, 3.2          | 3.8 - 3.12     |
```
sobolevn, line 52: I think that `3.8 - 3.12` is not correct for `Django@5.0`

sobolevn (commented):
Thank you! Just one question

README.md
intgr: Sure, but that's Django's constraint, not ours. `django-stubs` itself still supports Python 3.8 with Django 4.2.

sobolevn: Ok 👍

intgr: Should I count this as approval to merge?

sobolevn (approved):

intgr:
> Django version 4.1 is now officially EOL by Django project, but let's take that separately (there may be some cleanups we can do).

I had a look at this. We can't really clean up much by removing 4.1, because we support Django 3.2, which still has those symbols.

So given that there's nothing to be gained, we might as well keep 4.1 on the supported list.

PR #1863, opened 2023-12-07
Include ModelBase subclasses in plugin base class hook condition

### LINKED ISSUE #1799: Custom Model metaclass causing TypecheckAssertionError
# Bug report

## What's wrong

We're having some problems using `django-stubs` where we've customised the `ModelBase` metaclass for models. Appreciate this is probably not a super common use case but suspect this is likely not just going to cause issues for us. Whilst investigating I noticed that wagtail for example also utilise this pattern:

https://github.com/wagtail/wagtail/blob/56c62cc29874b98c863792628fa5cdf40db1589b/wagtail/models/__init__.py#L221-L251

I modified an existing test to demonstrate the failure:

```patch
---
 tests/typecheck/managers/querysets/test_filter.yml | 9 +++++++--
 1 file changed, 7 insertions(+), 2 deletions(-)

diff --git a/tests/typecheck/managers/querysets/test_filter.yml b/tests/typecheck/managers/querysets/test_filter.yml
index 83b14bc8..f6e46d6c 100644
--- a/tests/typecheck/managers/querysets/test_filter.yml
+++ b/tests/typecheck/managers/querysets/test_filter.yml
@@ -136,9 +136,14 @@
         -   path: myapp/models.py
             content: |
                 from django.db import models
-                class Publisher(models.Model):
+                from django.db.models.base import ModelBase
+                class MyModelBase(ModelBase):
                     pass
-                class Blog(models.Model):
+                class MyModel(models.Model, metaclass=MyModelBase):
+                    pass
+                class Publisher(MyModel):
+                    pass
+                class Blog(MyModel):
                     publisher = models.ForeignKey(Publisher, on_delete=models.CASCADE, related_name='blogs')
```

which results in this output running the tests:

```python
E   Actual:
E     main:5: error: Cannot resolve keyword 'publisher_id' into field. Choices are: blog, id, publisher  [misc] (diff)
E     main:8: error: Incompatible type for lookup 'publisher': (got "Blog", expected "Union[Publisher, int, None]")  [misc] (diff)
E     main:9: error: Cannot resolve keyword 'publisher_id' into field. Choices are: blog, id, publisher  [misc] (diff)
E     myapp/models:10: error: Incompatible types in assignment (expression has type "ForeignKey[Union[Publisher, Combinable], Publisher]", base class "MyModel" defined the type as "ReverseOneToOneDescriptor[MyModel, Publisher]")  [assignment] (diff)
E   Expected:
E     main:8: error: Incompatible type for lookup 'publisher': (got "Blog", expected "Union[Publisher, int, None]")  [misc] (diff)
E     main:9: error: Incompatible type for lookup 'publisher_id': (got "Blog", expected "Union[str, int]")  [misc] (diff)
E     E: main:8: error: Incompatible type for lookup 'publisher': (got "Blog", ex...
E     A: main:5: error: Cannot resolve keyword 'publisher_id' into field. Choices...
E             ^
```

We were previously running mypy 1.4.1 with django-stubs 4.2.3 and weren't seeing any problems so I think this is a result of some changes that have recently been introduced.

## How is that should be

If there's a simple way for us to hint that our metaclass behaves the same way as `ModelBase` a workaround would be very helpful!

But it would be great if it was possible to allow some way of customising `ModelBase` whilst keeping the stubs from `django-stubs`

## System information

- OS:
- `python` version: 3.11.3
- `django` version:
- `mypy` version: 1.6.0
- `django-stubs-ext` version:

flaeppe:
# I have made things!

We now also trigger the base class hook for subclasses of `ModelBase`.

## Related issues

Refs: https://github.com/typeddjango/django-stubs/pull/1668#issuecomment-1700897353
Supersedes: #1670 
Closes: #1799

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -474,4 +474,9 @@ def resolve_lazy_reference(
 
 
 def is_model_type(info: TypeInfo) -> bool:
-    return info.metaclass_type is not None and info.metaclass_type.type.fullname == fullnames.MODEL_METACLASS_FULLNAME
+    return info.metaclass_type is not None and (
+        # Using the model metaclass shipped by Django
+        info.metaclass_type.type.fullname == fullnames.MODEL_METACLASS_FULLNAME
+        # Using a custom model metaclass that inherits above
+        or info.metaclass_type.type.has_base(fullnames.MODEL_METACLASS_FULLNAME)
+    )
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -474,9 +474,4 @@ def resolve_lazy_reference(
 
 
 def is_model_type(info: TypeInfo) -> bool:
-    return info.metaclass_type is not None and (
-        # Using the model metaclass shipped by Django
-        info.metaclass_type.type.fullname == fullnames.MODEL_METACLASS_FULLNAME
-        # Using a custom model metaclass that inherits above
-        or info.metaclass_type.type.has_base(fullnames.MODEL_METACLASS_FULLNAME)
-    )
+    return info.metaclass_type is not None and info.metaclass_type.type.has_base(fullnames.MODEL_METACLASS_FULLNAME)
--- a/tests/typecheck/models/test_metaclass.yml
+++ b/tests/typecheck/models/test_metaclass.yml
@@ -73,3 +73,25 @@
 
               class Concrete3(Concrete2):
                   ...
+
+-   case: test_custom_model_base_metaclass
+    main: |
+        from myapp.models import This, Other
+
+        this = This(field=Other())
+        reveal_type(this.field)  # N: Revealed type is "myapp.models.Other"
+    installed_apps:
+        - myapp
+    files:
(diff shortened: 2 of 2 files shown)


mypy_django_plugin/lib/helpers.py
```diff
@@ -474,4 +474,9 @@ def resolve_lazy_reference(
 
 
 def is_model_type(info: TypeInfo) -> bool:
-    return info.metaclass_type is not None and info.metaclass_type.type.fullname == fullnames.MODEL_METACLASS_FULLNAME
+    return info.metaclass_type is not None and (
+        # Using the model metaclass shipped by Django
+        info.metaclass_type.type.fullname == fullnames.MODEL_METACLASS_FULLNAME
+        # Using a custom model metaclass that inherits above
+        or info.metaclass_type.type.has_base(fullnames.MODEL_METACLASS_FULLNAME)
```
sobolevn: I think that you can remove `info.metaclass_type.type.fullname == fullnames.MODEL_METACLASS_FULLNAME` part, since `has_base` goes for all `mro` entries, including itself.

sobolevn (approved):

PR #1864, opened 2023-12-08
Ensure mypy plugin processes inherited many to many fields

### LINKED ISSUE #1862: django-stubs 4.2.7 throws type check error for `Group.user_set`
# Bug report

With `django-stubs` 4.2.6, the following simplified code type checked:

```py
from django.contrib.auth.models import Group, User

group = Group()
group.user_set.add(User(username=""))
```

But with with 4.2.7:

```
$ mypy main.py 
main.py:4: error: "Group" has no attribute "user_set"  [attr-defined]
```

## How is that should be

I expect no type check error here. This seems to be a common way to get the `User`s of a `Group`:

- https://stackoverflow.com/a/58453760
- https://snakeycode.wordpress.com/2015/03/11/django-get-all-the-users-in-a-group/
- http://digitaldreamer.net/blog/2010/5/10/get-all-users-group-django/

## System information

- OS: macOS
- `python` version: 3.10.13
- `django` version: 4.2.8
- `mypy` version: 1.7.1

### Configuration

```toml
[tool.mypy]
ignore_missing_imports = true
strict = true
disallow_subclassing_any = false
disallow_untyped_decorators = false
warn_return_any = false

plugins = [
    "pydantic.mypy",
    "mypy_django_plugin.main",
]

[tool.django-stubs]
django_settings_module = "config.settings"
```

flaeppe:
# I have made things!

Ensure the plugin considers `ManyToManyField`s from abstract parent model classes.

## Related issues

Closes: #1862

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -1,7 +1,6 @@
 from collections import deque
-from collections.abc import Iterable
 from functools import cached_property
-from typing import Any, Dict, List, Optional, Type, Union, cast
+from typing import Any, Dict, Iterable, List, Optional, Type, Union, cast
 
 from django.db.models import Manager, Model
 from django.db.models.fields import DateField, DateTimeField, Field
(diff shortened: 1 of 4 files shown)


sobolevn (approved):
Thank you!

Issue #1067, opened 2022-07-21
Mypy can't get custom manager type


maximsht:
I have custom user model that inherits from UserProfile from django-tenant-users package. UserProfile has custom model manager. In my model I see an error from mypy: 
`Mypy: Could not resolve manager type for "Flow_BE.applications.users.models.User.objects"`
and when i try to use this manager I receive error:
`Mypy: "Manager[User]" has no attribute "create_user"`

This is how models specified:
```
class User(UserProfile):
    """This User model replaces default user model."""

# from django-tenant-users
class UserProfile(AbstractBaseUser, PermissionsMixinFacade):

    objects = UserProfileManager()

class UserProfileManager(BaseUserManager):

    def create_user(
            self,
            email=None,
            password=None,
            is_staff=False,
            **extra_fields,
        ):...
```

For User.objects.create_user(...) mypy generates errors. Code works.

## System information

- OS: MacOs 12.3.1
- `python` version: 3.10.4
- `django` version: 4.0.6
- `mypy` version: 0.971
- `django-stubs` version: 1.12.0
- `django-stubs-ext` version: 0.5.0

sobolevn:
Hi @MaximShteygervald 🙂
Nice to see you again!

We are working on better manager types in a new release, so it might be already fixed.
Can you please try `master` version?

flaeppe:
I think this is related to #1023 and the fact that django-tenant-users is an _untyped_ package.

mvaled:
I'm trying to update django-stubs and ext, but I'm being blocked by this issue.   

We usually create a custom queryset and use `as_manager`.  To avoid complications we use an explicit (albeit not 100% accurate, but really useful) annotation:

```python
class AttachmentQS(QuerySet['Attachment']):
     def load(self, pk) -> 'Attachment': ...

class Attachment(Model):
     # ...
     objects: AttachmenQS = AttachmentQS.as_manager()  # type: ignore 
```

This has worked pretty well so far, but trying to update apparently makes our hack being ignored and I get lots of issues like:

```
kaiko/attachments/tasks.py:112:18: error: "Manager[Attachment]" has no attribute "load"
```

when calling `Attachment.objects.load(...)`.  Does `django-stubs-ext` overrides the explicit type hint somehow? (Is it even possible to override it)?

flaeppe:
@mvaled See

- https://github.com/typeddjango/django-stubs#my-queryset-methods-are-returning-any-rather-than-my-model, and
[...]

PR #1865, opened 2023-12-08
Remove section regarding custom queryset methods from README

flaeppe:
Remove section 'My QuerySet methods are returning Any rather than my Model' as it doesn't really communicate anything useful.

## Related issues

Refs: https://github.com/typeddjango/django-stubs/issues/1067#issuecomment-1814753318


intgr (approved):
Agreed, seems very confusing.

PR #1871, opened 2023-12-11
Allow `None` in settings `MIGRATION_MODULES` dict values

asottile:
this is used to disable migrations for a particular module (such as during tests) https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-MIGRATION_MODULES


intgr (approved):
Thanks!

Issue #1264, opened 2022-11-16
Feature: Use PEP 696 defaults for fields


w0rp:
**NOTE:** This feature can't be worked on unless [PEP 696](https://peps.python.org/pep-0696/) is accepted, but given its [inclusion in typing_extensions](https://github.com/python/typing_extensions/blob/main/CHANGELOG.md#release-440-october-6-2022) I'd say it's likely it will be.

`django-stubs` uses set and get value generic types for fields, defined like so:

```py
# __set__ value type
_ST = TypeVar("_ST", contravariant=True)
# __get__ return type
_GT = TypeVar("_GT", covariant=True)

class Field(RegisterLookupMixin, Generic[_ST, _GT]):
```

Specific field types use special attributes that the mypy plugin for django-stubs reads to fill in the get and set types with.

```py
class CharField(Field[_ST, _GT]):
    _pyi_private_set_type: Union[str, int, Combinable]
    _pyi_private_get_type: str
    
    ...
```

This works for `mypy`, but not for other type checkers, where the generic parameters need to be written manually. To better support different type checkers, we could use the default values from PEP 696 instead.

```py
_ST_CHAR = TypeVar("_ST_CHAR", contravariant=True, default=Union[str, int, Combinable])
_GT_CHAR = TypeVar("_GT_CHAR", covariant=True, default=str)

class CharField(Generic[_ST_CHAR, _GT_CHAR], Field[_ST_CHAR, _GT_CHAR]):
    ...
```

The mypy plugin could read the default values instead of the `_pyi_private*` attributes from before. For other type checkers, the instance values for `__get__` and `__set__` will then use the generic type defaults if not specified otherwise.

intgr:
> This feature can't be worked on unless PEP 696 is accepted

I think we don't necessarily have to wait for acceptance of the PEP. As long as it's implemented in `typing_extensions`, there is at least 1 type checker implementation that already supports it, and its adoption does not cause regressions for other type checkers (particularly mypy), I would be in support for merging this.

Once PEP 696 is implemented in mypy itself (https://github.com/python/mypy/issues/14851), maybe we can get rid of this logic in the mypy plugin? Native type constructs tend to be less fragile and better supported in general.

w0rp:
Thanks for the comment. Generally I would like django-stubs to be a bit more useful for tools like Pyright by default without making life harder for people who benefit from the mypy plugin.

christianbundy:
[...]

PR #1873, opened 2023-12-12
Fix/pyright unknown

### LINKED ISSUE #1869: Type of "method_decorator" is partially unknown (pyright)
# Bug report

## What's wrong

Pyright reports error
```
 error: Type of "method_decorator" is partially unknown
    Type of "method_decorator" is "(decorator: ((...) -> Unknown) | Iterable[(...) -> Unknown], name: str = ...) -> ((_T@method_decorator) -> _T@method_decorator)" (reportUnknownVariableType)
278 errors, 0 warnings, 0 informations 
```
on this import 
```python
from django.utils.decorators import method_decorator
```
## How is that should be

Pyright should return no error

Applying this patch seems to fix the error but i am not 100% sure if this is right:
```diff
diff --git a/typings/django/utils/decorators.pyi b/typings/django/utils/decorators.pyi
index 40c1d8252..a9a2ab353 100644
--- a/typings/django/utils/decorators.pyi
+++ b/typings/django/utils/decorators.pyi
@@ -9,7 +9,7 @@ _CallableType = TypeVar("_CallableType", bound=Callable)
 
 classonlymethod = classmethod
 
-def method_decorator(decorator: Callable | Iterable[Callable], name: str = ...) -> Callable[[_T], _T]: ...
+def method_decorator(decorator: Callable[[_T], _T] | Iterable[Callable[[_T], _T]], name: str = ...) -> Callable[[_T], _T]: ...
 def decorator_from_middleware_with_args(middleware_class: type) -> Callable: ...
 def decorator_from_middleware(middleware_class: type) -> Callable: ...
 def make_middleware_decorator(middleware_class: type[MiddlewareMixin]) -> Callable: ...

```

## System information

- OS:
- `python` version: 3.11.6
- `django` version: 4.2.7
- `mypy` version: 1.1.1
- `pyright` version: 1.1.229

dephiros:
# Fix unknown type error in mypy

## Related issues

Format is:
- Closes #1869


django-stubs/apps/registry.pyi
```diff
@@ -24,7 +24,7 @@ class Apps:
     def get_app_config(self, app_label: str) -> AppConfig: ...
     # it's not possible to support it in plugin properly now
     def get_models(self, include_auto_created: bool = ..., include_swapped: bool = ...) -> list[type[Model]]: ...
-    def get_model(self, app_label: str, model_name: str | None = ..., require_ready: bool = ...) -> type[Any]: ...
+    def get_model(self, app_label: str, model_name: str | None = ..., require_ready: bool = ...) -> type[Model]: ...
```
intgr: IIRC it was a deliberate choice to use `Any` instead of `Model` here.

Does this cause issues with pyright?

dephiros: I thought that it caused some issues but cannot replicate now 🤷 
Let me add a comment and undo
@intgr , what was the reason for using `Any` there?

flaeppe: I think the reasoning is here: https://github.com/typeddjango/django-stubs/pull/814#issuecomment-1301957498

django-stubs/utils/decorators.pyi
```diff
@@ -1,15 +1,19 @@
...
 
+from django.http.response import HttpResponseBase
 from django.utils.deprecation import MiddlewareMixin
 from django.views.generic.base import View
+from typing_extensions import TypeAlias
 
 _T = TypeVar("_T", bound=View | Callable)  # Any callable
 _CallableType = TypeVar("_CallableType", bound=Callable)
+_VIEW = TypeVar("_VIEW", bound=Callable[..., HttpResponseBase])
+_DECORATOR: TypeAlias = Callable[..., _VIEW | Callable[..., _VIEW]]
```
flaeppe: As far as I'm aware, regarding mypy, bounds in `TypeAlias` doesn't work as it's outside of class and function scope. Unfortunately `_DECORATOR` needs to be explicitly expanded to not end up with having `_VIEW` as `Any`.

For `mypy` it should probably be visible if you run a `reveal_type(method_decorator)` somewhere.

flaeppe: Here's a playground link that should describe what I'm after: https://mypy-play.net/?mypy=latest&python=3.12&gist=ab4e5d4b4d118bb2e78c146850483952

django-stubs/utils/decorators.pyi
```diff
@@ -1,15 +1,19 @@
 from collections.abc import Callable, Iterable
 from typing import TypeVar
 
+from django.http.response import HttpResponseBase
 from django.utils.deprecation import MiddlewareMixin
 from django.views.generic.base import View
+from typing_extensions import TypeAlias
 
 _T = TypeVar("_T", bound=View | Callable)  # Any callable
```
flaeppe: Perhaps we could define generics for the `Callable` here too, while we're at it?

```suggestion
_T = TypeVar("_T", bound=View | Callable[..., Any])
```

django-stubs/db/models/fields/json.pyi
```diff
@@ -13,7 +13,12 @@ from typing_extensions import Self
 from . import Field
 from .mixins import CheckFieldDefaultMixin
 
-class JSONField(CheckFieldDefaultMixin, Field):
+# __set__ value type
+_ST = TypeVar("_ST", contravariant=True, default=Any)
+# __get__ return type
+_GT = TypeVar("_GT", covariant=True, default=Any)
```
flaeppe, line 20: I don't know how far or if we want to lean in to `PEP 696` (`default=` for `TypeVarLike`) as it's not approved yet

Refs:

- Pyright: https://github.com/microsoft/pyright/issues/4078#issuecomment-1666959148
- Mypy: https://github.com/python/mypy/issues/14851#issue-1613735295 (CC @intgr)
- Us: #1264 
- Steering council: https://github.com/python/steering-council/issues/177#issuecomment-1538925572

Not saying we should or shouldn't, just wanted to highlight for a possible discussion

intgr: As long as it doesn't break any other typecheckers, I'm +1 for adding `default=`.

In this example it shouldn't change anything, as mypy already defaults to `Any` for unbound TypeVars.

(PEP 696 in mypy would be very useful, we could drop a few ugly hacks like `_QuerySet` vs `QuerySet` vs `_QuerySetAny`)

django-stubs/utils/decorators.pyi
sobolevn: Nit: let's please not name it `_T`, since it is bound to `View` it can be `_ViewT` or something

django-stubs/db/models/fields/json.pyi
dephiros: Ah didn't realize that it is still not approved. It seems like it does not change existing behavior since our existing tests are passing?

flaeppe: I can be +1 too, lets try it out. Though we don't really have tests for most annotations, but these 2 are quite involved in the plugin and thus its tests.

No need to add any tests to the suite here I think. And in other cases we shouldn't need any tests for trivial uses of a `TypeVar`

django-stubs/contrib/admin/sites.pyi
```diff
@@ -61,11 +61,11 @@ class AdminSite:
     def is_registered(self, model: type[Model]) -> bool: ...
     def add_action(self, action: _ActionCallback, name: str | None = ...) -> None: ...
     def disable_action(self, name: str) -> None: ...
-    def get_action(self, name: str) -> Callable: ...
+    def get_action(self, name: str) -> _ActionCallback: ...
     @property
     def actions(self) -> Iterable[tuple[str, _ActionCallback]]: ...
     def has_permission(self, request: HttpRequest) -> bool: ...
-    def admin_view(self, view: Callable, cacheable: bool = ...) -> Callable: ...
+    def admin_view(self, view: Callable[..., HttpResponse], cacheable: bool = ...) -> Callable[..., HttpResponse]: ...
```
flaeppe: Hm, `admin_view` looks like a decorator. I think we want to use a `TypeVar` to not ruin any function signature using it. e.g.

```suggestion
    def admin_view(self, view: F, cacheable: bool = ...) -> F: ...
```

Where `F` could be a variable bound to a callable returning a `HttpResponse`

```python
F = TypeVar("F", bound=Callable[..., HttpResponse])
```

flaeppe (approved):
👍

PR #1876, opened 2023-12-14
[5.0] Add `db_default=` parameter to models `Field` classes

Skorpyon:
Django 5 feature update.

Database-computed default values - db_default arg and attr on model Field.

Tests not written, looks like it haven't things to test.

Full [Django 5 updates list](https://github.com/typeddjango/django-stubs/issues/1493).


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -180,6 +181,7 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
         validators: Iterable[validators._ValidatorCallable] = ...,
         error_messages: _ErrorMessagesMapping | None = ...,
         db_comment: str | None = ...,
+        db_default: Any = ...,
```
intgr: We can probably do better than `Any` for this type. According to documentation "This can be a literal value or a database function, such as Now." https://docs.djangoproject.com/en/5.0/ref/models/fields/#db-default

So something like `_ST | Expression | None`?

intgr (changes requested):
Thanks! Let's avoid using `Any` if possible.

intgr:
> Tests not written, looks like it haven't things to test.

Yeah, don't worry about tests for `.pyi` files. Unless it's a particularly complicated case of decorators or generics.

django-stubs/db/models/fields/__init__.pyi
Skorpyon: I walked a lot around it, Any looks too abroad, for sure.
But I feel that I may not narrow user choices, bcause in Django's `__init__` it looks next way:

```python
if db_default is not NOT_PROVIDED and not hasattr(db_default, "resolve_expression"):
    from django.db.models.expressions import Value
    db_default = Value(db_default)
```

So technically it may be anything that follow `.resolve_expression` protocol or a thing, that may be wrapped in `Value`.
Django-stubs define `Value` as 

https://github.com/typeddjango/django-stubs/blob/29c74e6e5f2205a24d42aa2d6a0b596a0ccb9277/django-stubs/db/models/expressions.pyi#L186-L188

so by stub `Value` may be initialized with `value: Any`.

Thats why I some confused. Looks like db_default may be anything. Sure it may be narrowed to `_ST | Expression | None | NOT_PROVIDED`, but I have no idea is it correct or no. Please advice.

Skorpyon: But here as class attr it should be narrowed to:

```python 
    ...
    db_default: NOT_PROVIDED | Expression
```
`__init__` passing expressions or wrap in Value, that is Expression subclass, wdyt?

intgr: > anything that follow .resolve_expression protocol

Don't get too bogged down with "it could be anything", we need to apply a certain amount of pragmatism.

In practice `resolve_expression()` means `Expression` or `BaseExpression`. If users want to define their own DB functions/expressions, they should subclass `Expression`. This is consistent with many other definitions that accept Expression. If this were a problem in practice, we would probably have an issue about it, _then_ we could discuss alternatives (like using `Protocol` instead of concrete class).

```
BaseExpression
├── Expression
│   └── Func
│       ├── Now
│       └── ...
├── Query
└── Subquery
```

Since `Query` and `Subquery` probably aren't allowed or useful in the DEFAULT clause in any database, let's exclude those and go with `Expression` instead of `BaseExpression`.

> so by stub Value may be initialized with value: Any.

Yep, since `Value` doesn't know what type of column it will be used with.

But with the `Field` definition we know the type of field being defined (`CharField` should accept `str`, `IntegerField` should accept `int` etc). That's `_ST`.

intgr: > But here as class attr it should be narrowed to:

Yes, you're right. Good point. `NOT_PROVIDED` -> `type[NOT_PROVIDED]` since it's using the class definition, not a class instance.

Skorpyon:
@intgr may you advice, please?

```shell
Python 3.12.0 (main, Nov 17 2023, 17:15:54) [Clang 15.0.0 (clang-1500.0.40.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> print("\n".join(sys.path))

/Users/anton/.pyenv/versions/3.12.0/lib/python312.zip
/Users/anton/.pyenv/versions/3.12.0/lib/python3.12
/Users/anton/.pyenv/versions/3.12.0/lib/python3.12/lib-dynload
/Users/anton/Documents/projects/github/django-stubs/.venv/lib/python3.12/site-packages
/Users/anton/Documents/projects/github/django-stubs
/Users/anton/Documents/projects/github/django-stubs/ext
```

But `bash ./scripts/stubtest.sh` gives:
```shell
error: django failed to find stubs
error: django.apps failed to find stubs
error: django.apps.config failed to find stubs
...
note: unused allowlist entry django.views.static.DEFAULT_DIRECTORY_INDEX_TEMPLATE
note: unused allowlist entry django.views.static.builtin_template_path
Found 3010 errors (checked 888 modules)
```

I stuck, django-stubs installed:
```shell
> pip freeze
...
distlib==0.3.8
Django==5.0
-e git+ssh://git@github.com/Skorpyon/django-stubs.git@2cf7a510b57c47d2f56ece659cf4796e0f5fb8f3#egg=django_stubs
-e git+ssh://git@github.com/Skorpyon/django-stubs.git@2cf7a510b57c47d2f56ece659cf4796e0f5fb8f3#egg=django_stubs_ext&subdirectory=ext
filelock==3.13.1
...
```

intgr:
> But bash ./scripts/stubtest.sh gives:
> Found 3010 errors (checked 888 modules)

Try removing `.mypy_cache` before running `stubtest`. There seems to be a bug, but I haven't bothered to debug/report it yet.

Skorpyon:
Here we go...
@intgr may you check `files.pyi`, please, not sure that it is correct. `FileField` haven't `_ST` generic in declaration.

Skorpyon:
I remove `db_default` from `ManyToManyField.__init__()` args at all.

Technically it may be passed to init, but latest Django codebase written so ugly, without any forward thinking.
That feature should be realized using some thing like `DbDefaultMixin`, adding `db_default` behaviour to specific fields only, but in Django it realized on top level, added to top parent class `Field`, so `db_default` available for all field subclasses, no matter have it sense or no.

If `db_default` have been set in `ManyToManyField` Django producing pretty fun SQL.

```python
class DBDefaultModel(models.Model):
    created = models.DateField(db_default=Now())
    circumference = models.FloatField(db_default=2 * Pi())
    default_value = models.CharField(max_length=10, db_default=None)
    manyfield = models.ManyToManyField(BlogPost, db_default=123)
```

```ipython
>>> from server.apps.main.models import DBDefaultModel
>>> mm = DBDefaultModel()
>>> mm.save()
```

```sql
INSERT INTO "main_dbdefaultmodel" ("created",  "circumference",  "default_value")
VALUES (DEFAULT, DEFAULT, DEFAULT)
RETURNING 
    "main_dbdefaultmodel"."id",
    "main_dbdefaultmodel"."created",
    "main_dbdefaultmodel"."circumference",
    "main_dbdefaultmodel"."default_value",
    "main_dbdefaultmodel"."manyfield";
```

```shell
django.db.utils.ProgrammingError: column main_dbdefaultmodel.manyfield does not exist
СТРОКА 1: ...mference", "main_dbdefaultmodel"."default_value", "main_dbde...
```

So Django trying retrieve `.manyfield` column value, thats not exists for sure, because it is many-to-many descriptor with intermediate table.

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -182,6 +183,7 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
         validators: Iterable[validators._ValidatorCallable] = ...,
         error_messages: _ErrorMessagesMapping | None = ...,
         db_comment: str | None = ...,
+        db_default: type[NOT_PROVIDED] | Expression | _ST = ...,
```
intgr, line 186: Sorry for sitting on this for so long.

When I was thinking about this later, I realized that `_ST` typevar is usually unbound in model declarations (e.g. `field = CharField(db_default=...)` rather than `field = CharField[str, str](db_default=...)`; the typevars are resolved later by mypy plugin.

In mypy, this doesn't cause errors at usage location because unbound typevars default to `Any` in mypy. But I wasn't sure about pyright.

I tested this now, seems that pyright handles it the same way: https://pyright-play.net/?pyrightVersion=1.1.353&pythonVersion=3.12&pythonPlatform=Linux&strict=true&locale=en&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDiApikSEgMYA0UAKokQGoCGIAUG7VALx0MsgAFACJawgJQcKAG2YBnOVABCsgBYAxJEWkATQcVLkKAbVoBdcQC42UW1B1FgUAPrPUSGK8FztwGjoAjZwdgZgBXaRhLOigAHygAOTBSHigAOgzxKABaAD5E5KJojLSONmAwMFSVZg0tXUFA4MdwyO4ARgAmAGZJAGI7QaG7AD0x8YmJqBQqsnB2IA

I think we can improve this later with typevar defaults on Field classes. There are other reasons as well why that's useful. So I think it's OK to merge this as is.

intgr (approved):
I'll leave this open for a few days in case another maintainer wants to chime in. If not, we can merge this.

PR #1878, opened 2023-12-15
Fix `AsyncClient.defaults` attribute typing

flaeppe:
e.g. overriding client for an ASGI request requires a list

Here's the runtime reference for the available ASGI scope: https://github.com/django/django/blob/617bcf611f3daa796e4054ba041089ece30a32fc/django/test/client.py#L640-L662

### DIFF
--- a/django-stubs/test/client.pyi
+++ b/django-stubs/test/client.pyi
@@ -4,9 +4,9 @@ from io import BytesIO, IOBase
 from json import JSONEncoder
 from re import Pattern
 from types import TracebackType
-from typing import Any, Generic, NoReturn, TypeVar, type_check_only
+from typing import Any, Generic, Literal, NoReturn, TypedDict, TypeVar, type_check_only
 
-from asgiref.typing import HTTPScope
+from asgiref.typing import ASGIVersions
 from django.contrib.auth.base_user import AbstractBaseUser
 from django.contrib.sessions.backends.base import SessionBase
 from django.core.handlers.asgi import ASGIRequest
@@ -154,6 +154,24 @@ class _RequestFactory(Generic[_T]):
     ) -> _T: ...
 
 class RequestFactory(_RequestFactory[WSGIRequest]): ...
+
+# A non total duplication of `asgiref.typing.HTTPScope`
+@type_check_only
+class HTTPScope(TypedDict, total=False):
+    type: Literal["http"]
+    asgi: ASGIVersions
+    http_version: str
+    method: str
+    scheme: str
+    path: str
+    raw_path: bytes
+    query_string: bytes
+    root_path: str
+    headers: Iterable[tuple[bytes, bytes]]
+    client: tuple[str, int] | None
+    server: tuple[str, int | None] | None
+    state: dict[str, Any]
+    extensions: dict[str, dict[object, object]] | None
 
 @type_check_only
 class _AsyncRequestFactory(_RequestFactory[_T]):
--- a/django-stubs/test/client.pyi
+++ b/django-stubs/test/client.pyi
@@ -6,6 +6,7 @@ from re import Pattern
 from types import TracebackType
 from typing import Any, Generic, NoReturn, TypeVar, type_check_only
 
+from asgiref.typing import HTTPScope
 from django.contrib.auth.base_user import AbstractBaseUser
 from django.contrib.sessions.backends.base import SessionBase
 from django.core.handlers.asgi import ASGIRequest
@@ -155,7 +156,8 @@ class _RequestFactory(Generic[_T]):
 class RequestFactory(_RequestFactory[WSGIRequest]): ...
 
 @type_check_only
-class _AsyncRequestFactory(_RequestFactory[_T]): ...
+class _AsyncRequestFactory(_RequestFactory[_T]):
+    defaults: HTTPScope  # type: ignore[assignment]
 
 class AsyncRequestFactory(_AsyncRequestFactory[ASGIRequest]): ...
 
--- a/django-stubs/test/client.pyi
+++ b/django-stubs/test/client.pyi
@@ -157,7 +157,7 @@ class RequestFactory(_RequestFactory[WSGIRequest]): ...
 
 # A non total duplication of `asgiref.typing.HTTPScope`
 @type_check_only
-class HTTPScope(TypedDict, total=False):
+class _HTTPScope(TypedDict, total=False):
     type: Literal["http"]
     asgi: ASGIVersions
     http_version: str
@@ -175,7 +175,7 @@ class HTTPScope(TypedDict, total=False):
 
 @type_check_only
 class _AsyncRequestFactory(_RequestFactory[_T]):
-    defaults: HTTPScope  # type: ignore[assignment]
+    defaults: _HTTPScope  # type: ignore[assignment]
 
 class AsyncRequestFactory(_AsyncRequestFactory[ASGIRequest]): ...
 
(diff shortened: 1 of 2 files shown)


django-stubs/test/client.pyi
```diff
@@ -151,7 +151,10 @@ class _RequestFactory(Generic[_T]):
     ) -> _T: ...
 
 class RequestFactory(_RequestFactory[WSGIRequest]): ...
-class _AsyncRequestFactory(_RequestFactory[_T]): ...
+
+class _AsyncRequestFactory(_RequestFactory[_T]):
+    defaults: dict[str, Any]
```
intgr: Would it make sense to reuse `asgiref.typing.HTTPScope` TypedDict here? Or are there use cases to support other free-form properties here?

Django depends on `asgiref` anyway, we could add it as a `django-stubs` dependency as well.

intgr (approved):
Thanks, makes sense.

django-stubs/test/client.pyi
flaeppe: Nice find with `asgiref.typing.HTTPScope`! I think that's what we _want_ here. But I also see that e.g. `client` is typed as `client: Optional[Tuple[str, int]]` but Django does: `"client": ["127.0.0.1", 0],` 🙂

Not sure what to make of it.. I have nothing against adopting `HTTPScope` anyways though.

I can't think of any free-form properties, but I would suggest starting out strict and relax the type later on to `dict[str, Any]` if anyone runs into troubles.

I wouldn't say we need to add `asgiref`? As you say Django already depend on it, it should follow along. Don't have any strong opinions about it either

intgr: > `client: Optional[Tuple[str, int]]` but Django does: `"client": ["127.0.0.1", 0]`

Interesting. Since you can't typehint a list as `list[str, int]`, it is probably a decent compromise though.

> I wouldn't say we need to add asgiref?

My preference would be to add it to deps and not rely on transitive dependencies. Although granted, the failure mode with missing dependencies in type stubs is very minor.

intgr: According to asgiref docs https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope `client` is `Iterable[str, int]`. So yeah, a compromise.

flaeppe: I've adjusted to `defaults: asgiref.typing.HTTPScope` and added `asgiref` as a first line dependency in `setup.py`

flaeppe: Also, I moved the `defaults` declaration to each class existing runtime: `RequestFactory` and `AsyncRequestFactory` to please the inheritance. I think that's most accurate? Let me know if you see any adjustments

intgr: > Also, I moved the defaults declaration to each class existing runtime: RequestFactory and AsyncRequestFactory to please the inheritance. I think that's most accurate? Let me know if you see any adjustments

I'm not sure. I think the previous approach was slightly closer to Django's implementation, even if Django violates Liskov substitution principle here (as it does in lots of places). But these are internal protocols, so doesn't really matter as long as all classes using them are updated as needed.

You would need to add the `defaults` attribute to all classes that derive from `_RequestFactory` hierarchy, it seems you missed `Client` and `AsyncClient`.

flaeppe: Yeah, I looked again and was about to say just that moving it it instead becomes misplaced. I'll add an ignore instead.

django-stubs/test/client.pyi
```diff
@@ -155,7 +156,8 @@ class _RequestFactory(Generic[_T]):
 class RequestFactory(_RequestFactory[WSGIRequest]): ...
 
 @type_check_only
-class _AsyncRequestFactory(_RequestFactory[_T]): ...
+class _AsyncRequestFactory(_RequestFactory[_T]):
+    defaults: HTTPScope  # type: ignore[assignment]
```
intgr: Wait! I didn't think this `HTTPScope` thing through.

`HTTPScope` isn't using `total=False` or `NotRequired`, so if some user tries to use the `defaults` attribute, they are required to fill **all** fields. 😨 That's not desired here, right?

What are our options? I think we can't change the totality of `HTTPScope TypedDict` on our side. So we could duplicate the whole type on our side, or we could just revert back to `dict[str, Any]`. Sorry about that.

flaeppe: I think that's actually desired?

A full `HTTPScope` is always required I think. As such, if you reassign defaults (e.g. `defaults =`) you need to populate all fields. But you could modify some of them, e.g. `.defaults["client"] = ...`

flaeppe: Haha, now I'm reading this wrong. You're right, since Django updates the scope with `defaults`, all of them aren't required.

flaeppe: `asgiref` only uses `NotRequired` for `state`. But we want `total=False`.

I'm thinking we should duplicate the type.

django-stubs/test/client.pyi
```diff
@@ -154,8 +155,27 @@ class _RequestFactory(Generic[_T]):
 
 class RequestFactory(_RequestFactory[WSGIRequest]): ...
 
+# A non total duplication of `asgiref.typing.HTTPScope`
 @type_check_only
-class _AsyncRequestFactory(_RequestFactory[_T]): ...
+class HTTPScope(TypedDict, total=False):
```
intgr: This should be internal. Also maybe let's add some suffix to more clearly distinguish from `HTTPScope`?

```suggestion
class _HTTPScopeNonTotal(TypedDict, total=False):
```

intgr (changes requested):

django-stubs/test/client.pyi
flaeppe: I'll update it, but I have to admit that I don't follow on why it's considered internal and should thus be prefixed with `_`. Why is it internal?

flaeppe: And to follow up on that. If it's internal I don't see a reason to suffix with `NonTotal`, since it shouldn't be exported to any different module?

intgr: > I don't follow on why it's considered internal and should thus be prefixed with _. Why is it internal?

Because `.pyi` files should describe the corresponding Django `.py` source. But such a symbol doesn't exist in Django itself. You can't do `from django.test.client import HTTPScope` because that would fail at runtime. (It would only work with a `if TYPE_CHECKING` guard, but that's ugly).

If we *want* to export it for users, it would be preferrable to add it to `django_stubs_ext`.

intgr: > And to follow up on that. If it's internal I don't see a reason to suffix with NonTotal, since it shouldn't be exported to any different module?

Fair enough

flaeppe: I was under the impression that `@type_check_only` helped out with unavailable at runtime? Ref: https://docs.python.org/3/library/typing.html#typing.type_check_only

But no, it doesn't change anything in regards to `if TYPE_CHECKING`

intgr: > I was under the impression that @type_check_only helped out with unavailable at runtime?

Oh, I had forgotten what it does and had it confused with some other decorators.

But I tried it and it seems that implementations don't follow it in a helpful manner. PyCharm happily suggests imports for such classes, and mypy does not issue any warnings when using them, e.g.:

```python
from django.conf.global_settings import Router
from django.test.utils import TimeKeeperProtocol

print(Router)
print(TimeKeeperProtocol)
```

So currently the `_` prefix is a more foolproof way to indicate stubs-only symbols.

flaeppe: I see and agree, let's stick with both

intgr (approved):
Looks good now, thanks!

PR #1879, opened 2023-12-15
Add type hints for `JSONField.from_db_value`

### LINKED ISSUE #934: from_db_value not defined in models.JSONField interface
# Bug report

## What's wrong
When attempting to call the superclass `from_db_value` in a class that inherits from `django.db.models.JSONField`, mypy does not believe that `from_db_value` is defined but it is. 
```python3
from django.db import models

class CustomJSONField(models.JSONField):
    def from_db_value(self,value,expression,connection):
        super_value = super().from_db_value(value,expression,connection)
        # Convert value from JSON field further...
        return super_value 
```

gives the mypy error 
```
 error: "from_db_value" undefined in superclass
```

## How is that should be
Django's JSONField implementation explicitly defines a `from_db_value`, to handle the serialisation of a python dictionary to a json string. As such it probably should also be defined in the JSONField interface provided by django-stubs? (it is not defined in the default Field interface as it is not a mandatory method to have defined). 

Please let me know if there is a better way to inherit from this class that avoids this error? 
## System information

- OS:
- `python` version: 3.8.5
- `django` version: 3.2.12
- `mypy` version:  0.942

flaeppe:
## Related issues

Closes: #934

### DIFF
--- a/django-stubs/db/models/fields/json.pyi
+++ b/django-stubs/db/models/fields/json.pyi
@@ -4,6 +4,7 @@ from typing import Any, ClassVar
 
 from django.db.backends.base.base import BaseDatabaseWrapper
 from django.db.models import lookups
+from django.db.models.expressions import Expression
 from django.db.models.fields import TextField
 from django.db.models.lookups import PostgresOperatorLookup, Transform
 from django.db.models.sql.compiler import SQLCompiler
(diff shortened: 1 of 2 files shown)


intgr (approved):
Thanks!

Also noticed this oddity, though not really related to this PR.

```python
class ArrayField(CheckFieldDefaultMixin, Field[_ST, _GT]):
    ...
    from_db_value: Any
```

flaeppe:
> Thanks!
> 
> Also noticed this oddity, though not really related to this PR.
> 
> ```python
> class ArrayField(CheckFieldDefaultMixin, Field[_ST, _GT]):
>     ...
>     from_db_value: Any
> ```

Yeah, I've seen it too. But looked through the runtime and realised it only exists when the child field type of the `ArrayField` has declared it. Which makes it difficult to do anything reasonable about.

PR #1880, opened 2023-12-16
Fix types for UniqueConstraint instantiation

flaeppe:
Gets rid of a couple of allowlist entries


flaeppe:
I'll go ahead and merge this, since it has been idle for a couple of days but is a trivial change.

intgr (approved):

PR #1887, opened 2023-12-22
Fixed error when using async function based view in url path function

### LINKED ISSUE #1874: Type error when using async function based view in url path function
# Bug report

## What's wrong

When using the below snippet ...

```python
from django.http import HttpResponse
from django.urls import path


def sync_view() -> HttpResponse:
    return HttpResponse(status=200)

async def async_view(): 
    return HttpResponse(status=200)

urlpatterns = [
    path("foo/", sync_view),
    path("bar/", async_view)  # <-- type error
]
```

... I got the below type error when calling `path` with the async view:

```
No overloads for "path" match the provided arguments Pyright[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)
conf.pyi(26, 5): Overload 3 is the closest match
Argument of type "() -> Coroutine[Any, Any, HttpResponse]" cannot be assigned to parameter "view" of type "Sequence[URLResolver | str]" in function "path"
  "function" is incompatible with "Sequence[URLResolver | str]"Pyright[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)
def async_view()
Full name: adit.dicom_explorer.urls.async_view
(function) def async_view() -> Coroutine[Any, Any, HttpResponse]
```

I used the latest pyright version (not tested with mypy).

## How it should be

`path` function should accept an async function based view.

## System information

- OS:
- `python` version: 3.11
- `django` version: 4.27
- `pyright` version: 1.1.339

ghost:
Closes #1874


django-stubs/urls/conf.pyi
```diff
@@ -10,6 +10,7 @@ from ..conf.urls import IncludedURLConf
 from ..http.response import HttpResponseBase
 
 _URLConf: TypeAlias = str | ModuleType | Sequence[_AnyURL]
+_ResponseType: TypeAlias = HttpResponseBase | Coroutine[Any, Any, HttpResponseBase] | Coroutine[Any, Any, None]
```
sobolevn: Sorry, I don't think that adding `Coroutine[Any, Any, None]` is correct.

In your case:

```python
async def async_view(): 
    return HttpResponse(status=200)
```

it does not return `None`, it returns `HttpResponse`.

The correct fix is:

```python
async def async_view() -> HttpResponse: 
    return HttpResponse(status=200)
```

sobolevn (changes requested):

django-stubs/urls/conf.pyi
ghost: Yes, I agree with you!)

I changed None to HttpResponse

sobolevn (commented):
I still think that this change is not required, since `Coroutine` is covariant on the return type: https://github.com/python/typeshed/blob/16933b838eef7be92ee02f66b87aa1a7532cee63/stdlib/typing.pyi#L442

ghost:
> I still think that this change is not required, since `Coroutine` is covariant on the return type: https://github.com/python/typeshed/blob/16933b838eef7be92ee02f66b87aa1a7532cee63/stdlib/typing.pyi#L442

We can add it, check how everything will work after if you don't need to delete it after

intgr:

flaeppe:
@ashm-tech do you want to continue the changes here? If yes, I'll make sure to help out with reviewing. I think what we need is a new `@overload` for `path` and `re_path` that covers for async responses.

Have a look at this https://github.com/typeddjango/django-stubs/issues/1874#issuecomment-2077226495 to see what I mean.

We should also add a test for the new overload, anywhere in https://github.com/typeddjango/django-stubs/blob/6f7a19cf26257fb2da1c65c12a72807f9154dfe7/tests/typecheck/urls/test_conf.yml. I believe that will help trying out the changes.

flaeppe:
Oh, sorry, just realised that this PR was superseded by #2085

PR #1888, opened 2023-12-23
Update `django.test.runner` stubs

UnknownPlatypus:
Removes 25 entries form the todolist.

I moved `RemovedInDjangoXX` allowlist entries at the bottom of the `allowlist.txt` so it's easier to find/update them.


django-stubs/test/runner.pyi
```diff
@@ -159,12 +182,23 @@ class DiscoverRunner:
...
-def is_discoverable(label: str) -> bool: ...
+def try_importing(label: str) -> bool: ...
+def find_top_level(top_level: AnyStr) -> AnyStr: ...
+def shuffle_tests(tests: list[TestCase], shuffler: Shuffler) -> Iterator[TestCase]: ...
+def reorder_test_bin(
+    tests: list[TestCase], shuffler: Shuffler | None = ..., reverse: bool = ...
+) -> Iterator[TestCase]: ...
+def reorder_tests(
+    tests: list[TestCase], classes: tuple[type[TestCase], ...], reverse: bool = ..., shuffler: Shuffler | None = ...
+) -> Iterator[TestCase]: ...
```
sobolevn, line 193: Why do you use `Iterator` here and `Generator` in `iter_test_cases`? They should be similar

django-stubs/test/runner.pyi
```diff
@@ -159,12 +182,23 @@ class DiscoverRunner:
     def get_databases(self, suite: TestSuite) -> set[str]: ...
     def run_tests(self, test_labels: list[str], extra_tests: list[Any] | None = ..., **kwargs: Any) -> int: ...
 
-def is_discoverable(label: str) -> bool: ...
+def try_importing(label: str) -> bool: ...
```
sobolevn, line 185: It returns `tuple[bool, bool]`: https://github.com/django/django/blob/14917c9ae272f47d23401100faa6cefa8e1728bf/django/test/runner.py#L1097

django-stubs/test/runner.pyi
```diff
@@ -83,16 +92,30 @@ class ParallelTestSuite(TestSuite):
...
+    def seed_display(self) -> str: ...
+    def shuffle(self, items: list[TestCase], key: Callable[[TestCase], str]) -> list[TestCase]: ...
 
 class DiscoverRunner:
     test_suite: type[TestSuite]
     parallel_test_suite: type[ParallelTestSuite]
     test_runner: type[TextTestRunner]
     test_loader: TestLoader
-    reorder_by: tuple[SimpleTestCase, ...]
+    reorder_by: tuple[type[TestCase], type[SimpleTestCase]]
```
sobolevn, line 118: I think that it should be `tuple[type[X], ...]`, where `X` is a proper type name.

Why? Because it would be easier to subclass and change. Or maybe even `Iterable` / `Sequence`.

django-stubs/test/runner.pyi
```diff
@@ -71,9 +76,13 @@ class RemoteTestRunner:
     def run(self, test: Any) -> Any: ...
 
 def default_test_processes() -> int: ...
+def get_max_test_processes() -> int: ...
+def parallel_type(value: str) -> int | Literal["auto"]: ...
 
 class ParallelTestSuite(TestSuite):
     init_worker: Any
+    process_setup: Callable
```
sobolevn, line 84: Can we type it better?

django-stubs/test/runner.pyi
```diff
@@ -36,14 +36,18 @@ class DebugSQLTextTestResult(TextTestResult):
...
+class DummyList:
+    def append(self, item: Any) -> None: ...
 
 class RemoteTestResult:
     events: list[Any]
     failfast: bool
     shouldStop: bool
     testsRun: int
-    def __init__(self) -> None: ...
+    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
```
sobolevn, line 50: Can you please add a TODO entry to sync this signature with `unittest.TestResult`?

sobolevn (commented):
Thanks!

intgr:

PR #1889, opened 2023-12-23
Generic `forms.ModelChoiceField`

### LINKED ISSUE #698: ModelChoiceField could use a type argument
# Bug report

## What's wrong

When subclassing `django.forms.ModelChoiceField` and overriding `label_from_instance()`, the type of the method argument is `Model`, but the implementation likely needs to know which concrete model type is being passed.

## How is that should be

While it is possible to work around this using `cast()` or `assert isinstance()`, having a type argument for the model type (`_M`) on `ModelChoiceField` would be a more elegant solution.

## System information

- `django` version: 3.2.6

UnknownPlatypus:
# I have made things!

This adds a generic type parameter to `ModelChoiceField` and `ModelMultipleChoiceField`. I've also added a test case to check this handles correctly a few easy-to-make mistakes.

## Related issues
Fixes #698


flaeppe (approved):
LGTM

PR #1891, opened 2023-12-25
Add `ModuleType` as a possible type to `URLResolver.urlconf_name`

Viicos:
According to code comment:
It may also be an object with an urlpatterns attribute (i.e. a module object)

# I have made things!

### DIFF
--- a/django-stubs/urls/resolvers.pyi
+++ b/django-stubs/urls/resolvers.pyi
@@ -103,7 +103,7 @@ class URLPattern:
 
 class URLResolver:
     pattern: _Pattern
-    urlconf_name: str | None | Sequence[_AnyURL] | ModuleType
+    urlconf_name: str | Sequence[_AnyURL] | ModuleType
     callback: None
     default_kwargs: dict[str, Any]
     namespace: str | None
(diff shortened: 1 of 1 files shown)


django-stubs/urls/resolvers.pyi
```diff
@@ -103,7 +103,7 @@ class URLPattern:
 
 class URLResolver:
     pattern: _Pattern
-    urlconf_name: str | None | Sequence[_AnyURL]
+    urlconf_name: str | None | Sequence[_AnyURL] | ModuleType
```
sobolevn: Do we need to update the `__init__` type as well?

Viicos: Yes, it seems to be set in the `__init__` [here](https://github.com/django/django/blob/14917c9ae272f47d23401100faa6cefa8e1728bf/django/urls/conf.py#L74)

sobolevn (approved):
Seems correct to me: https://github.com/django/django/blob/14917c9ae272f47d23401100faa6cefa8e1728bf/django/urls/resolvers.py#L725-L730

Thanks!

sobolevn (changes requested):
Wait a second, can it be `None`? 🤔 
Looks like https://github.com/django/django/blob/14917c9ae272f47d23401100faa6cefa8e1728bf/django/urls/resolvers.py#L483-L486 does not agree with this

Viicos:
> Wait a second, can it be `None`? 🤔
> Looks like https://github.com/django/django/blob/14917c9ae272f47d23401100faa6cefa8e1728bf/django/urls/resolvers.py#L483-L486 does not agree with this

True, [this search](https://github.com/search?q=repo%3Adjango%2Fdjango%20%22URLResolver(%22&type=code) doesn't show any case where it is `None` (at least in the Django codebase)

sobolevn (approved):
Thank you!

PR #1885, opened 2023-12-22
Update mypy to 1.8.0

sobolevn:
Closes https://github.com/typeddjango/django-stubs/pull/1884

All changes in `stubtest`'s allowlist is related to new `@type_check_only` feature:
- https://github.com/typeddjango/django-stubs/blob/125f31a81116f5f149c3e868d950608571e9a5e8/django-stubs/core/files/base.pyi#L31-L32
- https://github.com/typeddjango/django-stubs/blob/125f31a81116f5f149c3e868d950608571e9a5e8/django-stubs/http/response.pyi#L87-L91

PR #1894, opened 2023-12-26
Add `@type_check_only` to all Protocols and known stubs-only classes

intgr:
# I have made things!

Added `@type_check_only` decorators to more relevant classes.

Searched for `Protocol`/`TypedDict` subclasses, class names beginning with `_` and also `stubtest` output.

I didn't check methods. Also it would be great to add this decorator to `TypeAlias`es, but it's not currently possible.

## Related issues

Inspired by https://github.com/typeddjango/django-stubs/pull/1885#issue-2053570416


sobolevn (approved):
Great, thank you!

PR #1895, opened 2023-12-27
Fix type hint of `URLPattern.default_args`

Viicos:
From the implementation:

```python
class URLPattern:
    def __init__(self, pattern, callback, default_args=None, name=None):
        self.pattern = pattern
        self.callback = callback  # the view
        self.default_args = default_args or {}
        self.name = name
```

I'm wondering if it can be `dict[str, Any]`, as defined in `URLResolver` for `default_kwargs`?

### DIFF
--- a/django-stubs/urls/resolvers.pyi
+++ b/django-stubs/urls/resolvers.pyi
@@ -86,7 +86,7 @@ class LocalePrefixPattern:
 class URLPattern:
     pattern: _Pattern
     callback: Callable
-    default_args: dict[str, str] | None
+    default_args: dict[str, str]
     name: str | None
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
I agree these two should be in sync: https://github.com/typeddjango/django-stubs/blob/8118c23c82de2229174071177da3ec89d62e2e3e/django-stubs/urls/resolvers.pyi#L15

Because django passes it to `ResolverMatch`: https://github.com/django/django/blob/751d732a3815a68bdb5b7aceda0e7d5981362c4a/django/urls/resolvers.py#L451

sobolevn (approved):
Thank you and happy New Year! ⛄

PR #1897, opened 2024-01-01
[5.1] Add slicing notation to F expressions

Viicos:
See https://github.com/django/django/pull/17596/

Feel free to merge this whenever is the right time (django 5.1 isn't released yet)

### DIFF
--- a/django-stubs/db/models/expressions.pyi
+++ b/django-stubs/db/models/expressions.pyi
@@ -1,7 +1,7 @@
 import datetime
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from decimal import Decimal
-from typing import Any, ClassVar, Generic, Literal, TypeVar
+from typing import Any, ClassVar, Generic, Literal, NoReturn, TypeVar
 
 from django.db.backends.base.base import BaseDatabaseWrapper
 from django.db.models import Q, fields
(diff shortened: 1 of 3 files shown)


django-stubs/db/models/expressions.pyi
```diff
@@ -152,7 +153,10 @@ class F(_Deconstructible, Combinable):
         nulls_first: bool | None = ...,
         nulls_last: bool | None = ...,
     ) -> OrderBy: ...
-    def copy(self) -> F: ...
+    def copy(self) -> Self: ...
+
+class Sliced(F):
+    def __init__(self, obj: F, subscript: int | slice) -> None: ...
```
Viicos, line 162: Todo I need to add `resolve_expression` here

django-stubs/db/models/expressions.pyi
```diff
@@ -1,7 +1,7 @@
 import datetime
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from decimal import Decimal
-from typing import Any, ClassVar, Generic, Literal, TypeVar
+from typing import Any, ClassVar, Generic, Literal, NoReturn, TypeVar
```
sobolevn, line 4: Please, use `Never`

Viicos: Is there a specific reason to do so? I know they are equivalent but I like sticking to `NoReturn` for return values. What do you think?

sobolevn: OK :)

sobolevn (approved):

PR #1899, opened 2024-01-03
Adding missing `Q` methods: `check()`, `flatten()`

Alexerson:
I noticed that 2 methods on the `Q` object were never typed.
I’m not 100% sure about the signature for `flatten`.


django-stubs/db/models/query_utils.pyi
```diff
@@ -40,6 +40,8 @@ class Q(tree.Node):
         summarize: bool = ...,
         for_save: bool = ...,
     ) -> WhereNode: ...
+    def flatten(self) -> Iterator[Any]: ...
+    def check(self, against: dict[str, Any], using: str) -> bool: ...
```
Viicos:
```suggestion
    def check(self, against: dict[str, Any], using: str = ...) -> bool: ...
```

django-stubs/db/models/query_utils.pyi
```diff
@@ -40,6 +40,8 @@ class Q(tree.Node):
         summarize: bool = ...,
         for_save: bool = ...,
     ) -> WhereNode: ...
+    def flatten(self) -> Iterator[Any]: ...
```
Viicos:
```suggestion
    def flatten(self) -> Iterator[Node | NothingNode | ???]: ...
```
Not sure if anything else could be yielded

Viicos (changes requested):
You can also remove the corresponding entries from `scripts/stubtest/allowlist_todo.txt`

Alexerson:
I fixed `allowlist_todo` and the signature for `check`.
As I said, not quite sure about the signature. I feel your suggestion might be good enough. If noone else suggest something else, I’ll include it.

Viicos:
Yes I'm not sure either, `Any` would be sufficient if we can't really find what can be returned here.

django-stubs/db/models/query_utils.pyi
intgr: Our typing of `Node.children` and `Q.__init__(args, kwargs)` is incomplete also; `_NodeChildren` looks wrong when I look at `flatten()` code.

If you're not interested in un-entangling those, I think best to use `Incomplete` with `from _typeshed import Incomplete`

intgr (approved):
Thanks!

PR #1900, opened 2024-01-04
Add explicit overload to `Field` subclasses

Viicos:
In https://github.com/typeddjango/django-stubs/issues/1264#issuecomment-1859762279, and to try working on bringing `django-stubs` and `django-types` together, I suggested a workaround that would allow type checkers to natively infer `_ST` and `_GT` without the need for a plugin.

Unfortunately, for it to work correctly it would most probably require [higher kinded type vars](https://github.com/python/typing/issues/548) (my current example doesn't play well with subclasses, and I can't use `Self` as it can't be parametrized).

This is an experimental test, and as you can see it adds a lot of verbosity. But this removes the need to a plugin to parametrize the type variables, and it support nullable fields.

Would you accept such a change? I'm ok if this isn't accepted, considering how much it adds


sobolevn (commented):
I like the idea. Several comments:
1. Please, delete the plugin code, so we can be sure that your changes are enough to pass our CI
2. Looks like `stubstest` is not happy with some of the changes. Please, fix or ignore them :)

Viicos:
Great to hear! I'm also experimenting on another solution, but facing some issues (see https://github.com/microsoft/pyright/discussions/6914).

If this end up working, I think it might be preferable as it would avoid redefining all overloads, + people could easily subclass field classes.

Edit:

This alternative solution will require some fixes on both pyright and mypy, but it looks promising. On the pyright side, it is currently ~~being investigated~~ implemented. On the mypy side, it would require https://github.com/python/mypy/issues/3737, and probably https://github.com/python/mypy/pull/11366 (a fix that you created 😄). So I think it's best to wait a bit for this one.

Viicos:
After looking at the related plugin code, I don't know if removing `_pyi_private_set/get_type` support is the best way. The thing is the following function:

https://github.com/typeddjango/django-stubs/blob/1735d9e3ddbece8bbff6c9f166c231350af5e805/mypy_django_plugin/transformers/fields.py#L133-L153

Is being called before mypy actually infers the return type of a call to `FooField`, with the overloads provided in this PR. So I can't really take the "native" mypy solving logic into account. Besides, we can see primary key fields are special cased, so removing the plugin related logic here isn't trivial.

I also saw a [couple hits](https://github.com/search?q=_pyi_private_get_type+path%3A*.py&type=code) of public code making use of these attributes on field subclasses.

The idea here would be to add tests without the plugin added in the mypy config, and assert that they work. I've tried making use of the `parametrized` feature of the pytest plugin, but looking at the CI it doesn't seem to play well.

I might take a second look and see if it's in fact possible to remove the plugin code

django-stubs/contrib/auth/models.pyi
```diff
@@ -23,9 +24,9 @@ class Permission(models.Model):
     content_type_id: int
     objects: ClassVar[PermissionManager]
 
-    name = models.CharField(max_length=255)
-    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
-    codename = models.CharField(max_length=100)
```
Viicos, line 28: Having explicit assignments feels a bit weird in a stub file and was causing issues when testing without the plugin (had error code `var-annotated`)

django-stubs/contrib/auth/models.pyi
```diff
@@ -34,7 +35,7 @@ class GroupManager(models.Manager[Group]):
 class Group(models.Model):
     objects: ClassVar[GroupManager]
 
-    name = models.CharField(max_length=150)
+    name: models.CharField[str | int | Combinable, str]
     permissions = models.ManyToManyField(Permission)
```
Viicos, line 39: ... however I can't explicitly annotate `permissions: models.ManyToManyField[..., ...]` as the second type var is solved by the plugin to a class that doesn't exist in stubs nor at runtime

tests/typecheck/fields/test_base.yml
```diff
@@ -1,4 +1,15 @@
 -   case: test_model_fields_classes_present_as_primitives
+    parametrized:
+        -   mypy_section: |
+                [mypy]
+                plugins =
+                    mypy_django_plugin.main
+        -   mypy_section: |
+                [mypy]
+                disable_error_code = var-annotated
```
Viicos, line 9: ...So I had to disable the error code anyway when testing without the plugin

Viicos:
I rebased everything to have commits easier to review. CI is now passsing, except stubtests, as I had to make arguments following `null` as kw-only. This could be added to the allowlist but doesn't seem like a great solution. Any ideas how this could be handled?

I created a pull request @ mypy to add a `--ignore-positional-only` argument, might be more robust

intgr:
What's the status here, is it waiting for review?

@sobolevn since you commented, do you want to review this?

Viicos:
> What's the status here, is it waiting for review?

I think https://github.com/typeddjango/django-stubs/pull/1900#issuecomment-1923309336 needs to be resolved in some way. Even if not really ideal, the linked mypy open PR seems to be the best option.

A slightly more elegant solution to this PR is described [here](https://github.com/typeddjango/django-stubs/pull/1900#issuecomment-1878363899), but unsupported in mypy currently.

This PR is also imo not a blocker for the next release

Viicos:
I think the additions from https://github.com/typeddjango/django-stubs/pull/2048 is what's breaking the `when_default_for_primary_key_is_specified_allow_none_to_be_set` test case.

django-stubs/contrib/auth/models.pyi
flaeppe: You can have a look here how it can be done: https://github.com/typeddjango/django-stubs/pull/2214/files#diff-0236ca9e909bc4743fa03e6653fc33adbaaef02a34d8698501a63a59ad624f1fR42-R51

flaeppe:
I was playing around with getting rid of `_ST` and `_GT` in fields from `related.pyi` (`ForeignKey` and `OneToOneField`) and got stuck on https://github.com/python/mypy/issues/14764 in regards to overloading on the `null` argument.

Now I'm thinking that https://github.com/python/mypy/issues/14764 must be a problem for these changes too?

PR #1901, opened 2024-01-06
Improve types in `utils.termcolors`

Viicos:
# I have made things!

### DIFF
--- a/django-stubs/utils/termcolors.pyi
+++ b/django-stubs/utils/termcolors.pyi
@@ -1,4 +1,4 @@
-from collections.abc import Callable, Mapping, Sequence
+from collections.abc import Callable, Sequence
 from typing import Any
 
 color_names: tuple[str]
--- a/django-stubs/utils/termcolors.pyi
+++ b/django-stubs/utils/termcolors.pyi
@@ -1,14 +1,14 @@
 from collections.abc import Callable, Mapping, Sequence
 from typing import Any
 
-color_names: Sequence
-foreground: Mapping[str, str]
-background: Mapping[str, str]
+color_names: tuple[str]
+foreground: dict[str, str]
+background: dict[str, str]
 RESET: str
-opt_dict: Mapping[str, str]
+opt_dict: dict[str, str]
 
-def colorize(text: str | None = ..., opts: Sequence[str] = ..., **kwargs: Any) -> str: ...
-def make_style(opts: tuple = ..., **kwargs: Any) -> Callable: ...
+def colorize(text: str | None = ..., opts: Sequence[str] = ..., *, fg: str = ..., bg: str = ...) -> str: ...
+def make_style(opts: Sequence[str] = ..., *, fg: str = ..., bg: str = ...) -> Callable[[str | None], str]: ...
 
 NOCOLOR_PALETTE: str
 DARK_PALETTE: str
--- a/django-stubs/utils/termcolors.pyi
+++ b/django-stubs/utils/termcolors.pyi
@@ -1,10 +1,10 @@
 from collections.abc import Callable, Sequence
-from typing import Any
+from typing import Any, Literal
 
 color_names: tuple[str]
 foreground: dict[str, str]
 background: dict[str, str]
-RESET: str
+RESET: Literal["0"]
 opt_dict: dict[str, str]
 
 def colorize(text: str | None = ..., opts: Sequence[str] = ..., *, fg: str = ..., bg: str = ...) -> str: ...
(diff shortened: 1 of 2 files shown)


django-stubs/utils/termcolors.pyi
```diff
@@ -1,14 +1,14 @@
 from collections.abc import Callable, Mapping, Sequence
 from typing import Any
 
-color_names: Sequence
-foreground: Mapping[str, str]
-background: Mapping[str, str]
+color_names: tuple[str]
+foreground: dict[str, str]
+background: dict[str, str]
```
Viicos: The available colors seem to be fixed, should we switch to:

```suggestion
color_names: Final[tuple["black", "red", "green", "yellow", "blue", "magenta", "cyan", "white"]]
foreground: # Some kind of typed dict
background: # Same
```

django-stubs/utils/termcolors.pyi
```diff
@@ -1,14 +1,14 @@
...
 RESET: str
-opt_dict: Mapping[str, str]
+opt_dict: dict[str, str]
```
Viicos, line 8:
```suggestion
opt_dict: # A typed dict?
```

django-stubs/utils/termcolors.pyi
```diff
@@ -1,14 +1,14 @@
 from collections.abc import Callable, Mapping, Sequence
 from typing import Any
 
-color_names: Sequence
-foreground: Mapping[str, str]
-background: Mapping[str, str]
+color_names: tuple[str]
+foreground: dict[str, str]
+background: dict[str, str]
 RESET: str
```
Viicos:
```suggestion
RESET: Final[Literal["0"]]
```
?

django-stubs/utils/termcolors.pyi
```diff
@@ -1,14 +1,14 @@
...
-opt_dict: Mapping[str, str]
+opt_dict: dict[str, str]
 
-def colorize(text: str | None = ..., opts: Sequence[str] = ..., **kwargs: Any) -> str: ...
-def make_style(opts: tuple = ..., **kwargs: Any) -> Callable: ...
+def colorize(text: str | None = ..., opts: Sequence[str] = ..., *, fg: str = ..., bg: str = ...) -> str: ...
+def make_style(opts: Sequence[str] = ..., *, fg: str = ..., bg: str = ...) -> Callable[[str | None], str]: ...
```
Viicos, line 11:
```suggestion
def colorize(text: str | None = ..., opts: Sequence[str] = ..., **kwargs: Unpack[ColorizeKwargs]) -> str: ...
def make_style(opts: Sequence[str] = ..., **kwargs: Unpack[ColorizeKwargs]) -> Callable[[str | None], str]: ...
```
?

`opts` could also be `Sequence[Literal["bold", "underscore", "blink", "reverse", "conceal", "noreset"]]`

django-stubs/utils/termcolors.pyi
sobolevn: Why `Final`? I think it should be `tuple[Literal[...], ...]`

django-stubs/utils/termcolors.pyi
sobolevn: `Literal["0"]`

django-stubs/utils/termcolors.pyi
sobolevn: `Sequence[str]` is better.

django-stubs/utils/termcolors.pyi
Viicos: So that users don't actually set these variables to something else. But it's fine without it as well, I feel `Final` conveys a stronger meaning (e.g. with `numpy` and `np.PI`).

Viicos: Ended up keeping `tuple[str]`, literals feel a bit overkill

django-stubs/utils/termcolors.pyi
```diff
@@ -1,14 +1,14 @@
...
-from typing import Any
+from collections.abc import Callable, Sequence
+from typing import Any, Literal
 
-color_names: Sequence
-foreground: Mapping[str, str]
-background: Mapping[str, str]
-RESET: str
-opt_dict: Mapping[str, str]
+color_names: tuple[str]
```
intgr: Don't forget, if it's a tuple with multiple elements of the same type, you need `...`. This should fix the allowlist entry too.

```suggestion
color_names: tuple[str, ...]
```

Viicos: Thanks, inattention from my end

intgr (changes requested):
Thanks, apart from the `tuple`, this looks good to me.

intgr:
See stubtest errors.

intgr (approved):
Thanks!

PR #1904, opened 2024-01-14
Improve typing in template registry

### LINKED ISSUE #1903: Support `takes_context` in template registry methods, improve typing of decorators
See https://docs.djangoproject.com/en/5.0/howto/custom-template-tags/

I'll work on this

Viicos:
Fixes #1903


django-stubs/template/defaultfilters.pyi
```diff
@@ -2,13 +2,18 @@ from collections.abc import Callable
...
+from typing_extensions import Concatenate, ParamSpec
 
-register: Any
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
 
-def stringfilter(func: Callable) -> Callable: ...
+register: Library
+
+def stringfilter(func: Callable[Concatenate[str, _P], _R]) -> Callable[Concatenate[object, _P], _R]: ...
```
flaeppe, line 16: How come `object` is in the return type? Is it in place of `Any` or some other reason?

flaeppe (commented):
I'm fine with adding the tests, since e.g. the decorators aren't trivial. But I'd say we should bundle them under the same test, some "decorator test".

Mainly as the suite gets slowed down quite a bit with multiple tests.

Viicos:
> I'm fine with adding the tests, since e.g. the decorators aren't trivial. But I'd say we should bundle them under the same test, some "decorator test".
> 
> Mainly as the suite gets slowed down quite a bit with multiple tests.

Will do. I'll also try to add some comments in the stubs, so that it's easier to reason about which overload matches to which use case

intgr:
Assigned to @flaeppe. If you don't want to lead the review here, feel free to unassign yourself.

django-stubs/template/defaultfilters.pyi
Viicos: Mainly because `stringfilter` applies `str` on the first argument, and `str` expects [an `object` to be passed](https://github.com/python/typeshed/blob/6d19a17f57762815426942c21d0fadc91aa1e817/stdlib/builtins.pyi#L432-L434).

But if there's a good reason to switch to `Any`, I don't think this is a big deal

Viicos:
> I'm fine with adding the tests, since e.g. the decorators aren't trivial. But I'd say we should bundle them under the same test, some "decorator test".
> 
> Mainly as the suite gets slowed down quite a bit with multiple tests.

Thinking again about this, maybe this would things a bit confusing? We would end up with tests of the same module in a separate location.

But if still think we should do it this way, I can work on another PR moving every decorator tests, once this one gets merged (will makes things easier to review)

flaeppe:
> Thinking again about this, maybe this would things a bit confusing? We would end up with tests of the same module in a separate location.

I meant gathering the newly added tests _in this PR_ under the same test case. e.g. `test_library_decorators`. That shouldn't result in multiple locations.

> But if still think we should do it this way, I can work on another PR moving every decorator tests, once this one gets merged (will makes things easier to review)

If you like to refactor existing tests in a separate PR later on, that's fine too

django-stubs/template/defaultfilters.pyi
flaeppe: I see, think we can align and keep `object` then.

PR #1908, opened 2024-01-17
Use PEP 570 syntax

Viicos:
Might be worth waiting a bit on this one


sobolevn (approved):
We don't support 3.7 anymore, so it should be fine! Thank you!

PR #1911, opened 2024-01-18
Fix type hint of `EmailBackend.ssl_keyfile` and `EmailBackend.ssl_certfile`

Viicos:
Only hit in the Django codebase was [here](https://github.com/django/django/blob/10c7c7320baf1c655fcb91202169d77725c9c4bd/django/core/mail/backends/smtp.py#L59-L63), and according to the [typeshed definition](https://github.com/python/typeshed/blob/c51de8ee6859b3b1b411d48e56cd537716893e81/stdlib/ssl.pyi#L405-L407):

```python
    def load_cert_chain(
        self, certfile: StrOrBytesPath, keyfile: StrOrBytesPath | None = None, password: _PasswordType | None = None
    ) -> None: ...
```

### DIFF
--- a/django-stubs/core/mail/backends/smtp.pyi
+++ b/django-stubs/core/mail/backends/smtp.pyi
@@ -1,6 +1,7 @@
 import smtplib
 import threading
 
+from _typeshed import StrOrBytesPath
 from django.core.mail.backends.base import BaseEmailBackend
 
 class EmailBackend(BaseEmailBackend):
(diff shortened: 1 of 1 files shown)


Viicos:
`__init__` should also be added, maybe when I have time in another PR

sobolevn (approved):
Ok, waiting for the next PR :)

Viicos:
I have two open PRs ready for review in the meanwhile :D

PR #1786, opened 2023-10-20
Fix IndexError crash when using `from_queryset()` of custom Manager subclass

moranabadie:
# I have made things!

## Related issues

Fixes https://github.com/typeddjango/django-stubs/issues/1785

```
  File "/.venv/lib/python3.11/site-packages/mypy_django_plugin/transformers/managers.py", line 129, in _process_dynamic_method
    ret_type, base_that_has_method.defn.type_vars[0].fullname, manager_instance.args[0]
                                                               ~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: tuple index out of range
```

Without example I think I can not do more. 

Note : base_that_has_method.defn.type_vars cannot be null. since _has_compatible_type_vars already check that.

Issue #1785, opened 2023-10-19
IndexError: tuple index out of range


christianbundy:
# Bug report

```
version: 1.6.1
Traceback (most recent call last):
  File "/.venv/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/.venv/lib/python3.11/site-packages/mypy/__main__.py", line 15, in console_entry
    main()
  File "/.venv/lib/python3.11/site-packages/mypy/main.py", line 99, in main
    res, messages, blockers = run_build(sources, options, fscache, t0, stdout, stderr)
  File "/.venv/lib/python3.11/site-packages/mypy/main.py", line 178, in run_build
    res = build.build(sources, options, None, flush_errors, fscache, stdout, stderr)
  File "/.venv/lib/python3.11/site-packages/mypy/build.py", line 189, in build
    result = _build(
  File "/.venv/lib/python3.11/site-packages/mypy/build.py", line 262, in _build
    graph = dispatch(sources, manager, stdout)
  File "/.venv/lib/python3.11/site-packages/mypy/build.py", line 2938, in dispatch
    process_graph(graph, manager)
  File "/.venv/lib/python3.11/site-packages/mypy/build.py", line 3336, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/.venv/lib/python3.11/site-packages/mypy/build.py", line 3437, in process_stale_scc
    graph[id].type_check_first_pass()
  File "/.venv/lib/python3.11/site-packages/mypy/build.py", line 2306, in type_check_first_pass
    self.type_checker().check_first_pass()
  File "/.venv/lib/python3.11/site-packages/mypy/checker.py", line 475, in check_first_pass
    self.accept(d)
  File "/.venv/lib/python3.11/site-packages/mypy/checker.py", line 585, in accept
    stmt.accept(self)
  File "/.venv/lib/python3.11/site-packages/mypy/nodes.py", line 1145, in accept
    return visitor.visit_class_def(self)
  File "/.venv/lib/python3.11/site-packages/mypy/checker.py", line 2298, in visit_class_def
    self.accept(defn.defs)
  File "/.venv/lib/python3.11/site-packages/mypy/checker.py", line 585, in accept
    stmt.accept(self)
  File "/.venv/lib/python3.11/site-packages/mypy/nodes.py", line 1226, in accept
    return visitor.visit_block(self)
[...]

PR #1913, opened 2024-01-22
Fix IndexError for custom queryset managers

meshy:
Sometimes Mypy crashes on this line with an IndexError.

This change builds on the change in https://github.com/typeddjango/django-stubs/pull/1786 by making further use of the safer variable `typed_var`.

`typed_var` will equal `manager_instance.args` if `manager_instance.args` is populated, but if it isn't it will have another sensible value we can use instead.

Because that value should be populated, this prevents the crash.

## Related issues

Original issue: https://github.com/typeddjango/django-stubs/issues/1785
Previous fix: https://github.com/typeddjango/django-stubs/pull/1786


intgr (commented):
Thanks. Sorry that this has sat around so long without review.

Would be useful to add a test for this edge case. Possibly based on the `from_queryset_custom_manager_subclass` test that was added to cover a similar case.

intgr (approved):
Since the fix is straightforward this shouldn't cause any issues, I think it's OK to merge without a test.

meshy:
Apologies for not getting back to you on the tests. I've had this on my todo list, but wasn't able to prioritise it. Thank you for merging this!

PR #1925, opened 2024-01-30
Remove incorrect `Collection` base class and `__contains__` method from `QuerySet`

### LINKED ISSUE #1924: Django QuerySet does not implement `__contains__` but Stubs does
## What's wrong
Stubs implement `__contains__` for `QuerySet` but Django does not.

## How is that should be
Django does not implement `__contains__` for `QuerySet`. It relies on the fallback described in the docs at https://docs.python.org/3/reference/expressions.html#membership-test-details. I do not know why this was introduced in https://github.com/typeddjango/django-stubs/pull/33. It is not mentioned there.
```
class Example:
    def __iter__(self):
        return self

    def __next__(self):
        raise StopIteration

print(42 in Example())
```
This Minimal example proves that this is not necessary to type it that way.
`Example` does not implement `__contains__` but `in` is usable. So `42 in myQuerySet` will work but `myQuerySet.__contains__(42)` not, because it is not defined (you can check with `dir`).
`QuerySet` will be type checked correctly by mypy but will not have the same behavior at Runtime because `QuerySet` does not implement `__contains__`. A Django `QuerySet` is not a container because it does not implement `__contains__` as described at
https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes and https://mypy.readthedocs.io/en/stable/protocols.html#container-t. The docs are clear there. No `__contains__` no `Container`. A static type checker will think that a Django Stubs `QuerySet` is a `Container` but a runtime type checker will not recognize a Django `QuerySet` as a Container. Also, using `__contains__(42)` will be allowed by Django Stubs but fail on Django at runtime.

Why `__contains__` is not implemented with `contains` for `QuerySet` in Django, I do not know. But since this project is for typing Django, this is on their site to do. I am planning to open an issue there as well.

## System information
- OS:
- `python`: 3.10
- `django`: 4.2.7
- `mypy`: 1.8
- `django-stubs`: 4.2.6
- `django-stubs-ext`: None

fidoriel:
# I have made things!
## Related issues
Please see fixes https://github.com/typeddjango/django-stubs/issues/1924


fidoriel:
Imho there should be an abstract base type which will make it possible to use this. Currently, Mypy allows passing a QS into a `Collection`. Merging this will break this behavior. 
```
class DjangoContainer(Reversible[_Row], Iterable[_Row], Sized):
    pass
```

should be added. It behaves like a container but isnt one.

```
def myFunc(some_tings: DjangoContainer): ...
```

would allow to pass in a list with Model instances and a queryset.

sobolevn (approved):
I think that we should stay as close to the real source as possible.

@intgr do you have any comments?

intgr:
You are entirely correct, QuerySet is not a Collection. In fact, it's not Reversible either!

```
In [16]: isinstance(qs, Collection)   # ✅
Out[16]: False
In [17]: isinstance(qs, Iterable)
Out[17]: True
In [18]: isinstance(qs, Sized)
Out[18]: True
In [19]: isinstance(qs, Reversible)  # ⚠️
Out[19]: False
```

We keep discovering significant inaccuracies that have been in django-stubs from nearly the beginning. 😅 Both of these date back to 2019.

* https://github.com/typeddjango/django-stubs/commit/f02050911fc5ec4125f9034d79f338329ca7cf75#diff-389fe7d7dcfb77ddaac87c1f3bdc196f3ef0ca33b8047c6638a15f3a3db21b56L124-R125
* https://github.com/typeddjango/django-stubs/commit/324b961d7464d7419e5886da9401a3f35481218d#diff-389fe7d7dcfb77ddaac87c1f3bdc196f3ef0ca33b8047c6638a15f3a3db21b56L49-R53

intgr:
I tested this out. I'm downcasting `QuerySet[x]` to `Collection[x]` in lots of places in my projects. Typically in places where I want to accept both `QuerySet[model]` and `list[model]` as some parameter. So this causes a fair amount of churn.

Most places work fine with `Iterable[x]` as well, but in some places I want to use `len()` on the iterable. AFAIK there's currently no straightforward way to say `Iterable[T] & Sized` in the type sytem. I guess that may have been the motivation in making `QuerySet`s `Collection`.

But there are other solutions:

* Define a custom `class SizedIterable(Iterable[T_co], Sized, Protocol)` for those use cases.
* Instead of `Collection[T]`, use the more explicit `list[T] | QuerySet[T]`

Not a big deal.

intgr (approved):
Thanks, you are correct. Sorry for the long delay in reviewing this.

intgr:
Are you interested in creating another PR to remove `Reversible` too?

PR #1927, opened 2024-02-03
Added `CommandParser` to `commands.__init__`

### LINKED ISSUE #1926: `CommandParser` not in `django.core.management`
# Bug report

I think `CommandParser` should be in `django-stubs/core/management/__init__.pyi`. As of 4.2.7, it's not there: https://github.com/typeddjango/django-stubs/blob/4.2.7/django-stubs/core/management/__init__.pyi

## What's wrong

```python
from django.core.management import BaseCommand, CommandParser
```

## How is that should be

I think we need to add:

```python
from .base import CommandParser as CommandParser
```

## System information

- OS: macOS Ventura 13.5.2
- `python` version: 3.10.13
- `django` version: 3.2.23
- `mypy` version: 1.7.1

jamesbraza:
Closes https://github.com/typeddjango/django-stubs/issues/1926

### DIFF
--- a/django-stubs/core/management/__init__.pyi
+++ b/django-stubs/core/management/__init__.pyi
@@ -2,6 +2,7 @@ from typing import Any
 
 from .base import BaseCommand as BaseCommand
 from .base import CommandError as CommandError
+from .base import CommandParser as CommandParser
 
 def find_commands(management_dir: str) -> list[str]: ...
 def load_command_class(app_name: str, name: str) -> BaseCommand: ...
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
Do we need https://github.com/django/django/blob/0630ca5725ba5b17c61cd1f6a05dce2660c4724e/django/core/management/__init__.py#L23 as well?

jamesbraza:
Yeah thanks for the suggestion! Implemented it just now 👍

jamesbraza:
Ha now I understand why the `as` statements were there... otherwise `pre-commit` will remove them lolz

sobolevn (approved):
Thanks!

PR #1934, opened 2024-02-09
Fix return type of `BaseModelAdmin.formfield_for_dbfield`

Viicos:
Looking at the implementation it falls back to `db_field.formfield`


sobolevn (approved):
I agree: I cannot find any `None` paths. Thanks!

PR #1941, opened 2024-02-13
Fix type hint of `Response.set_cookie.max_age`

Viicos:
# I have made things!

### DIFF
--- a/django-stubs/http/response.pyi
+++ b/django-stubs/http/response.pyi
@@ -57,7 +57,7 @@ class HttpResponseBase:
         self,
         key: str,
         value: str = ...,
-        max_age: int | None = ...,
+        max_age: float | datetime.timedelta | None = ...,
         expires: str | datetime.datetime | None = ...,
         path: str = ...,
         domain: str | None = ...,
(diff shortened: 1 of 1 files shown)


django-stubs/http/response.pyi
```diff
@@ -57,7 +57,7 @@ class HttpResponseBase:
         self,
         key: str,
         value: str = ...,
-        max_age: int | None = ...,
+        max_age: float | datetime.timedelta | None = ...,
```
sobolevn:
```suggestion
        max_age: int | datetime.timedelta | None = ...,
```

https://docs.djangoproject.com/en/5.0/ref/request-response/#django.http.HttpResponse.set_cookie

It says: 

> max_age should be a [timedelta](https://docs.python.org/3/library/datetime.html#datetime.timedelta) object, an integer number of seconds, or None (default) if the cookie should last only as long as the client’s browser session. If expires is not specified, it will be calculated.

Viicos: The docstring says otherwise: https://github.com/django/django/blob/222bf2932b55ebc964ffc5f9a6f47bad083e5ac2/django/http/response.py#L234-L236

But I guess we follow what is publicly documented?

sobolevn: Yes, it would produce unexpected results for `float`: https://github.com/django/django/blob/222bf2932b55ebc964ffc5f9a6f47bad083e5ac2/django/http/response.py#L260

PR #1944, opened 2024-02-13
[5.0] Add django.db.models.GeneratedField

palfrey:
# I have made things!

## Related issues

Part of https://github.com/typeddjango/django-stubs/issues/1493 changes


django-stubs/db/models/fields/generated.pyi
```diff
@@ -0,0 +1,13 @@
+from typing import Any
+
+from django.db import models
+from django.db.backends.base.base import BaseDatabaseWrapper
+from django.db.models.expressions import Expression
+from django.utils.datastructures import DictWrapper
+
+class GeneratedField(models.Field):
+    def __init__(
+        self, *, expression: Expression, output_field: models.Field, db_persist: bool | None = ..., **kwargs: Any
```
Viicos: I think it's better to add all the possible arguments, as it is done for other `Field` subclasses

palfrey: Done

django-stubs/db/models/fields/generated.pyi
```diff
@@ -0,0 +1,47 @@
...
+from django.db import models
+from django.db.backends.base.base import BaseDatabaseWrapper
+from django.db.models.expressions import Expression
+from django.db.models.fields import _ErrorMessagesMapping, _FieldChoices
+from django.db.models.sql import Query
+from django.utils.datastructures import DictWrapper
+from django.utils.functional import _StrOrPromise
+
+class GeneratedField(models.Field):
+    generated: Literal[True] = True
```
sobolevn: We don't need default values here

sobolevn (commented):
Why do we ignore `cached_col`?

palfrey:
> Why do we ignore `cached_col`?

Hit the error at the top of that ignore block `# Ignore @cached_property error "cannot reconcile @property on stub with runtime object"`

django-stubs/db/models/fields/generated.pyi
palfrey: Removed

django-stubs/db/models/fields/generated.pyi
```diff
@@ -0,0 +1,47 @@
...
+        unique_for_year: str | None = ...,
+        choices: _FieldChoices | None = ...,
+        help_text: _StrOrPromise = ...,
+        db_column: str | None = ...,
+        db_comment: str | None = ...,
+        db_tablespace: str | None = ...,
+        validators: Iterable[_ValidatorCallable] = ...,
+        error_messages: _ErrorMessagesMapping | None = ...,
+        **kwargs: Any,
+    ): ...
```
sobolevn:
```suggestion
    ) -> None: ...
```

django-stubs/db/models/fields/generated.pyi
```diff
@@ -0,0 +1,47 @@
...
+from django.db.models.fields import _ErrorMessagesMapping, _FieldChoices
+from django.db.models.sql import Query
+from django.utils.datastructures import DictWrapper
+from django.utils.functional import _StrOrPromise
+
+class GeneratedField(models.Field):
+    generated: Literal[True] = ...
+    db_returning: Literal[True] = ...
+    _query: Query | None = ...
+    output_field: models.Field | None = ...
```
sobolevn:
```suggestion
    generated: Literal[True]
    db_returning: Literal[True]
    _query: Query | None
    output_field: models.Field | None
```

sobolevn (approved):

PR #1945, opened 2024-02-14
Update type hints of `core.signing`

Viicos:
# I have made things!

### DIFF
--- a/django-stubs/core/signing.pyi
+++ b/django-stubs/core/signing.pyi
@@ -1,5 +1,6 @@
 from datetime import timedelta
 from typing import Any, Protocol, type_check_only
+
 from typing_extensions import Unpack
 
 BASE62_ALPHABET: str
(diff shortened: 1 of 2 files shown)


django-stubs/core/signing.pyi
```diff
@@ -20,31 +25,35 @@ class JSONSerializer:
...
-    key: str
+    key: bytes | str
+    fallback_keys: list[bytes | str]
     sep: str
-    salt: str
+    salt: bytes | str
     algorithm: str
     def __init__(
         self,
+        *args: Unpack[tuple[()]],
```
Viicos: This to make stubtest happy.

Two other solutions:
- Add the `__init__` in the ignorelist
- Add an overload, one without `args` (and a empty stararg as it is currently on django's `main`), one with `*args: Any` and `typing_extensions.deprecated` applied to this overload

sobolevn: Let's just use `Any`

sobolevn (approved):
Thanks! Can you please try `deprecated` here?
Just as an experiment.

Viicos:
This is how it looks with Pyright and a LSP implementation:
image

sobolevn (approved):
This is cool! Thanks a lot.

PR #1946, opened 2024-02-14
[5.0] Update `core.paginator`

Viicos:
Part of #1493.

### DIFF
--- a/django-stubs/core/paginator.pyi
+++ b/django-stubs/core/paginator.pyi
@@ -1,7 +1,8 @@
 from collections.abc import Iterable, Iterator, Sequence, Sized
 from typing import ClassVar, Generic, Protocol, TypeVar, overload, type_check_only
 
-from django.utils.functional import _StrOrPromise, _StrPromise, cached_property
+from django.db.fields import _ErrorMessagesDict
+from django.utils.functional import _StrPromise, cached_property
 
 class UnorderedObjectListWarning(RuntimeWarning): ...
 class InvalidPage(Exception): ...
@@ -19,8 +20,8 @@ class _SupportsPagination(Protocol[_T], Sized, Iterable):
 
 class Paginator(Generic[_T]):
     ELLIPSIS: ClassVar[_StrPromise]
-    default_error_messages: ClassVar[dict[str, _StrPromise]]
-    error_messages: dict[str, _StrOrPromise]
+    default_error_messages: ClassVar[_ErrorMessagesDict]
+    error_messages: _ErrorMessagesDict
     object_list: _SupportsPagination[_T]
     per_page: int
     orphans: int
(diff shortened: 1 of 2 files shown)


django-stubs/core/paginator.pyi
```diff
@@ -19,6 +19,8 @@ class _SupportsPagination(Protocol[_T], Sized, Iterable):
 
 class Paginator(Generic[_T]):
     ELLIPSIS: ClassVar[_StrPromise]
+    default_error_messages: ClassVar[dict[str, _StrPromise]]
+    error_messages: dict[str, _StrOrPromise]
```
flaeppe: I think we should declare these the same as model fields does. We can use the alias `django.db.models.fields._ErrorMessagesDict`

```suggestion
    default_error_messages: _ErrorMessagesDict
    error_messages: _ErrorMessagesDict
```

django-stubs/core/paginator.pyi
```diff
@@ -29,6 +31,7 @@ class Paginator(Generic[_T]):
         per_page: int | str,
         orphans: int = ...,
         allow_empty_first_page: bool = ...,
+        error_messages: dict[str, _StrOrPromise] | None = ...,
```
flaeppe: Same here, but with the alias `django.db.models.fields._ErrorMessagesMapping`

```suggestion
        error_messages: _ErrorMessagesMapping | None = ...,
```

flaeppe (commented):
My suggested changes might justify movement of the aliases currently living in `django.db.models.fields`, since they would now be used under a different domain. But that can be done separately, if we'd deem it relevant.

Viicos:
I see `_ErrorMessagesDict` is imported in other files as well, e.g. from `contrib/auth/forms.pyi`, so I think it's fine to leave the type alias where it is currently. Then we might decide to move it at some point

django-stubs/core/paginator.pyi
```diff
@@ -19,6 +20,8 @@ class _SupportsPagination(Protocol[_T], Sized, Iterable):
 
 class Paginator(Generic[_T]):
     ELLIPSIS: ClassVar[_StrPromise]
+    default_error_messages: ClassVar[_ErrorMessagesDict]
```
Viicos, line 23: values are actually guaranteed to be `_StrPromise`, but I think it's fine to leave it this way (`dict[str, _StrOrPromise]`)

flaeppe (approved):

django-stubs/core/paginator.pyi
flaeppe: I suspect you'd then run in to troubles if you want to declare your own `default_error_messages` not bothering about translation strings. e.g.

```python
class X(Paginator):
    default_error_messages = {"key": "not a translated string"}
```

Viicos: Oh yes right

PR #1947, opened 2024-02-14
[5.0] Update `core.validators`

Viicos:
# I have made things!


sobolevn (approved):
`Any | Callable` is still `Any`, but I am fine with that.

Viicos:
> `Any | Callable` is still `Any`, but I am fine with that.

This is true from a typing perspective, but I usually like including it for documentation purposes; I also often use `Any | None` in some cases as well :)

PR #1948, opened 2024-02-14
[5.0] Update `core.cache.backends`, add `RedisCache` and related classes

Viicos:
Not sure why `redis` wasn't spotted before 5.0? It was added in 4.x.

How the redis dependency should be handled?


django-stubs/core/cache/backends/redis.pyi
```diff
@@ -0,0 +1,54 @@
...
+from redis.connection import BaseParser, ConnectionPool
+from typing_extensions import TypeAlias
+
+@type_check_only
+class RedisCacheClientSerializer(Protocol):
+    def dumps(self, obj: Any) -> bytes: ...
+    @overload
+    def loads(self, data: SupportsInt) -> int: ...
+    @overload
+    def loads(self, data: ReadableBuffer) -> Any: ...
```
Viicos, line 18: We might want to define a more simple signature

django-stubs/core/cache/backends/redis.pyi
```diff
@@ -0,0 +1,54 @@
...
+from typing import Any, Callable, Iterable, Protocol, SupportsInt, overload, type_check_only
+
+from _typeshed import ReadableBuffer, Unused
+from django.core.cache.backends.base import BaseCache
+from redis.client import Redis
+from redis.connection import BaseParser, ConnectionPool
+from typing_extensions import TypeAlias
+
+@type_check_only
+class RedisCacheClientSerializer(Protocol):
```
intgr: Let's add underscore to make it clear that this is internal to stubs

```suggestion
class _RedisCacheClientSerializer(Protocol):
```

django-stubs/core/cache/backends/redis.pyi
intgr: What do you mean by "more simple"?

django-stubs/core/cache/backends/redis.pyi
```diff
@@ -0,0 +1,54 @@
...
+class RedisCacheClient:
+    def __init__(
+        self,
+        servers: list[str],
+        serializer: str | Callable[[], RedisCacheClientSerializer] | RedisCacheClientSerializer | None = ...,
+        pool_class: str | type[ConnectionPool] | None = ...,
+        parser_class: str | type[BaseParser] | None = ...,
+        **options: Any,
+    ) -> None: ...
+    def get_client(self, key: Unused = ..., *, write: bool = ...) -> Redis: ...
```
intgr: According to code comment in the method:

> custom client can be implemented which might require the `key` to select the server, e.g. sharding.

So even if Django itself ignores the value, better typehint it as `_Key`

(Also adde default values)

```suggestion
    def get_client(self, key: _Key | None = None, *, write: bool = False) -> Redis: ...
```

django-stubs/core/cache/backends/redis.pyi
```diff
@@ -0,0 +1,54 @@
...
+_Key: TypeAlias = str | bytes | memoryview
+_Expiry: TypeAlias = int | timedelta
+
+class RedisCacheClient:
+    def __init__(
+        self,
+        servers: list[str],
+        serializer: str | Callable[[], RedisCacheClientSerializer] | RedisCacheClientSerializer | None = ...,
+        pool_class: str | type[ConnectionPool] | None = ...,
+        parser_class: str | type[BaseParser] | None = ...,
```
intgr: When the argument default value is a primitive `None/True/False` or short literal, it's helpful to include the default value in type hints. In the past this usually wasn't done, but more recently it has become more established.

```suggestion
        serializer: str | Callable[[], RedisCacheClientSerializer] | RedisCacheClientSerializer | None = None,
        pool_class: str | type[ConnectionPool] | None = None,
        parser_class: str | type[BaseParser] | None = None,
```

django-stubs/core/cache/backends/redis.pyi
```diff
@@ -0,0 +1,54 @@
...
+@type_check_only
+class RedisCacheClientSerializer(Protocol):
+    def dumps(self, obj: Any) -> bytes: ...
+    @overload
+    def loads(self, data: SupportsInt) -> int: ...
+    @overload
+    def loads(self, data: ReadableBuffer) -> Any: ...
+
+class RedisSerializer:
+    def __init__(self, protocol: int | None = ...) -> None: ...
```
intgr:
```suggestion
    def __init__(self, protocol: int | None = None) -> None: ...
```

django-stubs/core/cache/backends/redis.pyi
```diff
@@ -0,0 +1,54 @@
+from collections.abc import Mapping
+from datetime import timedelta
+from typing import Any, Callable, Iterable, Protocol, SupportsInt, overload, type_check_only
+
+from _typeshed import ReadableBuffer, Unused
+from django.core.cache.backends.base import BaseCache
+from redis.client import Redis
+from redis.connection import BaseParser, ConnectionPool
```
intgr: I think handling ImportError should work fine:

```suggestion
try:
    from redis.client import Redis
    from redis.connection import BaseParser, ConnectionPool
except ImportError:
    Redis: TypeAlias = Any
    BaseParser: TypeAlias = Any
    ConnectionPool: TypeAlias = Any
```

Not sure if `Any` or `_typeshed.Incomplete` is preferred. I would be fine with either one.

intgr (changes requested):
Thanks! Some suggestions.

> Not sure why redis wasn't spotted before 5.0? It was added in 4.x.

I think people noticed it, but nobody has put in the effort. Much appreciated. :)

django-stubs/core/cache/backends/redis.pyi
Viicos: Yup, also noticed this in typeshed, and was about to open an issue here to see if we could use https://github.com/JelleZijlstra/stubdefaulter to help us doing so

django-stubs/core/cache/backends/redis.pyi
Viicos: I was just worried it might be painful to implement such a protocol if `RedisCacheClient.serializer` is explicitly specified, but this isn't that big of a concern

django-stubs/core/cache/backends/redis.pyi
Viicos: Hum, I still get `Library stubs not installed for "redis.client"  [import-untyped]`. Maybe we should install the redis types, similar to the ones for pyyaml?

intgr: > Maybe we should install the redis types, similar to the ones for pyyaml?

Interesting. It didn't occur to me that `types-pyyaml` is an unconditional dependency of `django-stubs`, but you are right.

On the other hand: I've seen lots of Django projects over my career and exactly 0 of them used either Redis or Django's PyYAML serialization. :stuck_out_tongue: So it seems a bit of a waste to have these as unconditional dependency, when these dependencies are optional in Django itself.

I think adding it as an optional dependency, and updating `requirements.txt` with `-e .[compatible-mypy,redis]`, should solve this.

Viicos: Yes I think adding an optional dependency to `django-stubs` makes sense. Redis is also a bit different from pyyaml as it is in the process of adding types to the library itself. So depending on `types-redis` from typeshed isn't really reliable as it will [go away](https://github.com/python/typeshed/issues/10592) at some point. I'll add the optional dependency here

django-stubs/core/cache/backends/redis.pyi
intgr: I think most users would just opt to subclass the existing `RedisSerializer` concrete class.

intgr (approved):

PR #1949, opened 2024-02-14
[5.0] Update `core.files`

Viicos:
# I have made things!


django-stubs/core/files/base.pyi
```diff
@@ -26,7 +26,7 @@ class File(FileProxyMixin[AnyStr], IO[AnyStr]):
         exc_value: BaseException | None,
         exc_tb: TracebackType | None,
     ) -> None: ...
-    def open(self, mode: str | None = ...) -> Self: ...
+    def open(self, mode: str | None = ..., *args: Any, **kwargs: Any) -> Self: ...
```
intgr: Since args+kwargs are passed to Python `builtins.open()`, we can copy more precise types from there.

https://github.com/python/typeshed/blob/7acaa912b042e5aa28f679748c94f587da99f846/stdlib/builtins.pyi#L1553-L1564

```suggestion
    def open(
        self,
        mode: str | None = ...,
        buffering: int = -1,
        encoding: str | None = None,
        errors: str | None = None,
        newline: str | None = None,
        closefd: bool = True,
        opener: Callable[[str, int], int] | None = None,
    ) -> Self: ...
```

intgr (changes requested):
Thanks! One suggestion; LGTM otherwise.

django-stubs/core/files/uploadedfile.pyi
```diff
@@ -40,8 +40,9 @@ class InMemoryUploadedFile(UploadedFile):
         content_type: str | None,
         size: int | None,
         charset: str | None,
-        content_type_extra: dict[str, str] = ...,
+        content_type_extra: dict[str, str] | None = ...,
     ) -> None: ...
+    def open(self, mode: str | None = ...) -> Self: ...
```
intgr: Looks like you need to suppress here too

```suggestion
    def open(self, mode: str | None = ...) -> Self: ...  # type: ignore[override]
```

intgr (commented):
Some CI errors.

intgr (approved):
Thanks!

PR #1950, opened 2024-02-15
Add GitHub actions release workflow

### LINKED ISSUE #1922: Use a trusted (GitHub action) publisher when publishing to PyPi
PyPi quite recently introduced "Trusted publishing" (PyPi's term for [OpenID Connect - OIDC](https://openid.net/connect/) standard), which can be used to add a GitHub Actions workflow as trusted publisher in PyPi. That allows for better automation of publishing releases to PyPi that also removes any need for API tokens as GitHub secrets.

Trusted publishing should also integrate into [GitHub's environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) which can be used for deployments and GitHub allows creating protection rules for deployments. As such it could be possible to e.g. require approvals and/or certain reviewers before deploying to PyPi.

Using trusted publishing should probably reduce the current workload of creating releases while also make it simpler to manage who can do it.

A couple of links with more information:

- https://docs.pypi.org/trusted-publishers/
- https://docs.pypi.org/trusted-publishers/adding-a-publisher/
- https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/
- https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment

I could've misinterpreted or got some stuff wrong here, feel free to correct me. I've also set up automatic PyPi releasing for a couple of packages using trusted publishing.

Here's a current `.github/workflows/release.yml` from https://pypi.org/project/virtualenv/ that could be used as reference:

https://github.com/pypa/virtualenv/blob/fa283474fd199e3836f8b2c99510190c6b77e2bc/.github/workflows/release.yml

flaeppe:
The workflow automates pushing new `django-stubs` versions to PyPI triggered from GitHub releases.

It does _not_ release `django-stubs-ext`.

The workflow requires trusted publishing to be set up in PyPI. I'm not sure who has correct access to PyPI, but whoever does needs to follow the few steps lined out here(ping @sobolevn @intgr):

https://docs.pypi.org/trusted-publishers/adding-a-publisher/

I suggest having the GitHub actions environment named `release` (like PyPI's example shows).

I also added a build checking step in the test workflow.

## Related issues

- Closes #1922 

---

Future work could be to upgrade to the more modern packaging tools configuration file `pyproject.toml`. But that deservers its own issue and changeset.


sobolevn:
I am away for a week. I can't review and setup anything, but I like the idea :)

flaeppe:
No worries, whenever you find some motivation and time. Should be quick to get it in place.

intgr:
We already had some release automation way back. Are you familiar with this work? Is there anything useful that could be taken from there?

* Added: #980
* Removed: #1108 (IIRC it was removed because PyPI started requiring 2FA authentication)

intgr:
Anyway the new approach looks more elegant.

But looks like it's difficult to test. Have you figured out a way to test it? In theory we could use `test.pypi.org` & manual `workflow_dispatch` trigger or something similar. But I'm not sure the added complexity would be worth it.

We could just YOLO it, and if release automation fails, restore `release.sh` from git history and do it the old-fashioned way.

flaeppe:
I'm not familiar with the old release workflow and I don't see anything worth including from there. But perhaps someone else sees something I don't?

Other times I've set this up I've just tried to push the workflow through to the default branch, create a release and patch on until I've got all workflow runs green.

Not aware of any good and valuable testing approaches to be honest. All in all I find this setup easy to grasp and patch in case of any problems

sobolevn (approved):
Let's do it!

изображение

YOLO, I guess?

flaeppe:
Let's go!

flaeppe:
Though we might as well rebase and bump the used PyPI action to its latest version before merging

flaeppe:
Seems to work still, all should be ready once trusted publishing is set up on the PyPI side.

PR #1952, opened 2024-02-16
Allowing `set` in `model_to_dict`'s `exclude`

### LINKED ISSUE #1951: `model_to_dict`'s `exclude` allowing `set`
# Bug report

https://github.com/typeddjango/django-stubs/issues/637 was opened about `model_to_dict` not allowing `set`.  This was fixed in https://github.com/typeddjango/django-stubs/pull/640 via `typing.Collection`.

However, https://github.com/typeddjango/django-stubs/pull/909 seems to have clobbered that change :/

So this issue is to bring back https://github.com/typeddjango/django-stubs/pull/640's changes

## What's wrong

https://github.com/typeddjango/django-stubs/blob/4.2.7/django-stubs/forms/models.pyi#L37

Currently, `model_to_dict`'s `exclude` is set to allow `list`s or `tuple`s, but not a `set`.

## How is that should be

Looking at `model_to_dict`, it should allow a `Collection`.

## System information

jamesbraza:
Allowing `set` for `model_to_dict`'s `exclude`

## Related issues

Closes https://github.com/typeddjango/django-stubs/issues/1951

### DIFF
--- a/django-stubs/forms/models.pyi
+++ b/django-stubs/forms/models.pyi
@@ -14,7 +14,7 @@ from django.forms.formsets import BaseFormSet
 from django.forms.renderers import BaseRenderer
 from django.forms.utils import ErrorList, _DataT, _FilesT
 from django.forms.widgets import Widget
-from django.utils.datastructures import _IndexableCollection, _ListOrTuple, _PropertyDescriptor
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks!

PR #1955, opened 2024-02-18
Update type hints of contrib.auth.hashers

yhay81:
# I have made things!


sobolevn (approved):
Thank you, it is correct :)

PR #1956, opened 2024-02-18
Avoid returning None from get_field_related_model_cls

### LINKED ISSUE #1953: AttributeError: 'NoneType' object has no attribute '__module__'. Did you mean: '__reduce__'?
# Bug report

## What's wrong

 mypy . --strict --explicit-package-bases

```Traceback (most recent call last):
  File "/Users/xx/Library/Caches/pypoetry/virtualenvs/xxx-CJszIqT8-py3.11/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
             ^^^^^^^^^^^^^^^
  File "/Users/xx/Library/Caches/pypoetry/virtualenvs/xxx-CJszIqT8-py3.11/lib/python3.11/site-packages/mypy/__main__.py", line 15, in console_entry
    main()
  File "mypy/main.py", line 100, in main
  File "mypy/main.py", line 182, in run_build
  File "mypy/build.py", line 191, in build
  File "mypy/build.py", line 265, in _build
  File "mypy/build.py", line 2890, in dispatch
  File "mypy/build.py", line 3076, in load_graph
  File "mypy/build.py", line 1998, in __init__
  File "mypy/build.py", line 2292, in compute_dependencies
  File "mypy/plugin.py", line 844, in get_additional_deps
  File "/Users/xx/Library/Caches/pypoetry/virtualenvs/xxx-CJszIqT8-py3.11/lib/python3.11/site-packages/mypy_django_plugin/main.py", line 141, in get_additional_deps
    related_model_module = related_model_cls.__module__
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute '__module__'. Did you mean: '__reduce__'?
```

## How is that should be

No errors
## System information

- OS:
- `python` 3.11:
- `django` 5.0.2:
- `mypy` 1.8.0:
- `django-stubs` 4.2.7:
- `django-stubs-ext` 4.2.7:

SingingTree:
https://github.com/typeddjango/django-stubs/pull/1495 updated `get_field_related_model_cls` to raise `UnregisteredModelError` rather than returning `None` for failure paths. However, None can still be returned if the initial retrieval of `related_model_cls` returns None.

This patch adds a check to see if the initial retrieval has got a `None` and then raises the appropriate error rather than letting `None` be returned.

## Related issues

Closes https://github.com/typeddjango/django-stubs/issues/1953

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -369,6 +369,9 @@ class DjangoContext:
         else:
             related_model_cls = field.field.model
 
+        if related_model_cls is None:
+            raise UnregisteredModelError
+
         if isinstance(related_model_cls, str):
             if related_model_cls == "self":  # type: ignore[unreachable]
                 # same model


sobolevn (commented):
Do you think that adding a test case is a good idea here?

UnknownPlatypus:
> Do you think that adding a test case is a good idea here?

yeah I think it would be great to add a test case to `tests/typecheck/fields/test_related.yml` with a minimal reproducible example :pray:

mypy_django_plugin/django/context.py
```diff
@@ -369,6 +369,9 @@ def get_field_related_model_cls(self, field: Union["RelatedField[Any, Any]", For
         else:
             related_model_cls = field.field.model
 
+        if related_model_cls is None:
+            raise UnregisteredModelError
+
```
flaeppe, line 374: I would like it if we cleaned up the if statements here if you wouldn't mind changing a couple more lines?

Essentially raise error in 1 place at the end e.g.

```python
        related_model_cls = None
            related_model_cls = field.remote_field.model
        else:
            related_model_cls = field.field.model

        if isinstance(related_model_cls, str):
            if related_model_cls == "self":  # type: ignore[unreachable]
                # same model
                related_model_cls = field.model
            elif "." not in related_model_cls:
                related_model_fullname = f"{field.model.__module__}.{related_model_cls}"
                related_model_cls = self.get_model_class_by_fullname(related_model_fullname)
            else:
                with contextlib.suppress(LookupError):
                    related_model_cls = self.apps_registry.get_model(related_model_cls)

        if related_model_cls is None:
            raise UnregisteredModelError
        return related_model_cls
```

SingingTree:
Updated to shift location of check.

I'm open to suggestions re tests. The scenario I'm encountering this in is some proprietary code that I'm yet to reduce to something that nicely fits into a test here.

mypy_django_plugin/django/context.py
SingingTree: ✅

flaeppe (approved):

mypy_django_plugin/django/context.py
flaeppe: Oh, my suggestion wouldn't work actually. Since we write `related_model_cls` to a string.

But it's interesting that we have an unreachable ignore too, perhaps that hides something.

Anyways, unless we keep 2 variables I'm not sure we could change to my suggestion

Viicos: I recently went through this code to figure things out. It's a bit hard to understand what Django is doing, but from what I've experienced `field.remote_field.model` can't be a string unless the string reference can't be resolved (which should fail at runtime).

I'm doing things a bit differently from the mypy plugin, but you can find the diff of how I handle it [here](https://github.com/Viicos/django-autotyping/commit/416814bd084a670e2a116b3869dd203914bff018#diff-58b11942d8703e587bdc6b64e799314b2c70974f3fe6f3e3aea3e679073bd13cR138-R143).

(So the unreachable ignore is probably because the type hint says `field.remote_field.model` is always `type[Model]`, not `str`)

flaeppe: Yeah, Django probably sanitizes the field/model attributes. Input is string but never after instantiation.

I suspect the unreachable is easy to get rid of if we remove the type inference, since mypy can't be aware of the above happening.

Something like this:

```python
def get_field_related_model_cls(...) -> Type[Model]:
    related_model_cls: Type[Model] | str | None = None
    ...
    return related_model_cls
```

It doesn't allow for my suggestion above to work or really help out with anything other than getting rid of the unreachable though.

random1st:
@SingingTree

SingingTree:
I've been traveling, and will make some time to get my branch updated over the next couple of days.

SingingTree:
I've reverted the changes back to the first iteration of PR. I'm afraid I haven't had any luck in reducing the faulting code to a test case.

FWIW, the scenario I'm seeing this hit in is using [django taggit](https://github.com/jazzband/django-taggit). I'm not certain what's going on under the hood that's resulting in the unhappiness however.

sobolevn (approved):

flaeppe (approved):

Issue #1931, opened 2024-02-06
Open to cutting a 4.2.8 release?


jamesbraza:
Are you open to cutting a new release (for `mypy==1.8`)?

`django-stubs[compatible-mypy]==4.2.7` is pinned to `mypy==1.7`: https://github.com/typeddjango/django-stubs/blob/4.2.7/setup.py#L35, but the current `master` is not

sobolevn:
It will be `5.0.0`, I guess 🤔

jamesbraza:
Oh if it was a bugfix version bump, I am okay with requesting it 🤓 , but I don't want to rush a major version bump. If it's a major version bump, that would be up to your own choice in timing

meshy:
Is there an appetite for back-fixing other bug fixes? It'd be great to have https://github.com/typeddjango/django-stubs/pull/1913 in an earlier version too!

aleksanb:
looking forward to a release here as well.

intgr:
> but I don't want to rush a major version bump. If it's a major version bump, that would be up to your own choice in timing

Yep it would be 5.0.0. But the major version bump is just to stay in sync with Django version numbers. It shouldn't be significantly more complex than a typical release for us.

It's more that I've been absent from this project lately and nobody else has stepped up to handle the release. I wrote more at https://github.com/typeddjango/django-stubs/issues/1493#issuecomment-1943787982

sobolevn:
I think that `5.0.0` should also drop 3.7 support, since it is EOL.

intgr:
We already dropped Python 3.7 in version 4.2.1 😄 https://github.com/typeddjango/django-stubs?tab=readme-ov-file#version-compatibility

sobolevn:
Sorry, I got confused.

3.8 is not supported by `pre-commit` anymore: https://github.com/typeddjango/django-stubs/actions/runs/7953898472/job/21710530911

But, we cannot drop it yet, since it is still a supported Python version.

delfick:
that could be fixed by making it install pre-commit with a different version if the python version is less than a certain number https://pip.pypa.io/en/stable/reference/requirement-specifiers/

sobolevn:
@delfick PR is welcome :)

delfick:
@sobolevn https://github.com/typeddjango/django-stubs/pull/1961 I believe it needs approval before we can see if CI works against that change

jamesbraza:
[...]

PR #1961, opened 2024-02-20
deps: Upgrade pre-commit for newer versions of python

delfick:
A remake of https://github.com/typeddjango/django-stubs/pull/1957 as suggested by @sobolevn in https://github.com/typeddjango/django-stubs/issues/1931#issuecomment-1952139286

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 # Dev tools:
 black==24.2.0
-pre-commit==3.5.0; python-version < 3.9
-pre-commit==3.6.2; python-version >= 3.9
+pre-commit==3.5.0; python_version < 3.9
+pre-commit==3.6.2; python_version >= 3.9
 pytest==8.0.1
 pytest-mypy-plugins==3.0.0
 pytest-shard==0.1.2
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,6 +1,7 @@
 # Dev tools:
 black==24.2.0
-pre-commit==3.5.0
+pre-commit==3.5.0; python-version < 3.9
+pre-commit==3.6.2
```
jamesbraza: I think you need `; python_version >= 3.9` at the end of this line, otherwise Python 3.8 will get confused as to which to use

delfick: good point, fixed

requirements.txt
```diff
@@ -1,6 +1,7 @@
 # Dev tools:
 black==24.2.0
-pre-commit==3.5.0
+pre-commit==3.5.0; python-version < 3.9
+pre-commit==3.6.2; python-version >= 3.9
```
sobolevn:
```suggestion
pre-commit==3.5.0; python_version < 3.9
pre-commit==3.6.2; python_version >= 3.9
```

sobolevn (approved):
Thanks!

requirements.txt
delfick: so close!

delfick:
you're welcome!

PR #1965, opened 2024-02-24
5.0: Add Signal.asend and Signal.asend_robust

93578237:
# I have made things!
https://github.com/django/django/blob/6e1ece7ed522c904a674966fa985159b7bbf1545/django/dispatch/dispatcher.py#L205
https://github.com/django/django/blob/6e1ece7ed522c904a674966fa985159b7bbf1545/django/dispatch/dispatcher.py#L334

## Related issues

#1493


sobolevn (approved):
It is ok to have `Callable` without type params, because other places in this module do the same thing.

PR #1967, opened 2024-02-24
5.0: Add ModelAdmin.show_facets

93578237:
# I have made things!
https://github.com/django/django/blob/6e1ece7ed522c904a674966fa985159b7bbf1545/django/contrib/admin/options.py#L661

## Related issues

#1493

### DIFF
--- a/django-stubs/contrib/admin/options.pyi
+++ b/django-stubs/contrib/admin/options.pyi
@@ -1,3 +1,4 @@
+import enum
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from typing import Any, Generic, Literal, TypeVar, type_check_only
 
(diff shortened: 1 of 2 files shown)


django-stubs/contrib/admin/options.pyi
```diff
@@ -150,6 +156,7 @@ class ModelAdmin(BaseModelAdmin[_ModelT]):
     save_on_top: bool
     paginator: type
     preserve_filters: bool
+    show_facets: ShowFacets
```
q0w, line 159: Should string literals be accepted?
```suggestion
    show_facets: ShowFacets | Literal["NEVER", "ALLOW", "ALWAYS"]
```

sobolevn (approved):

PR #1969, opened 2024-02-24
5.0: Add BaseConstraint.violation_error_code

93578237:
# I have made things!
https://github.com/django/django/commit/5b3d3e400ab9334ba429ca360c9818c6dfc3a51b

## Related issues

#1493

### DIFF
--- a/django-stubs/db/models/constraints.pyi
+++ b/django-stubs/db/models/constraints.pyi
@@ -15,6 +15,7 @@ class Deferrable(Enum):
 
 class BaseConstraint:
     name: str
+    violation_error_code: str | None
     violation_error_message: _StrOrPromise | None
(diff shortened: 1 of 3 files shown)


django-stubs/db/models/constraints.pyi
```diff
@@ -15,6 +15,7 @@ class Deferrable(Enum):
 
 class BaseConstraint:
     name: str
+    violation_error_code: str | None
```
93578237, line 18: Should it accept Promise as well?
```suggestion
    violation_error_code: _StrOrPromise | None
```

93578237: No, it cant because `ValidationError.code: str`

sobolevn (approved):

PR #1984, opened 2024-02-26
Add URL converter protocol type

adamchainz:
# I have made things!

This PR improves the types for [custom URL converters](https://docs.djangoproject.com/en/5.0/topics/http/urls/#registering-custom-path-converters) with a `Protocol`.

## Related issues

n/a


sobolevn (commented):
Question: calling `converter()` is also a part of the protocol, isn't it? https://github.com/django/django/blob/ef2434f8508551fee183079ab471b1dc325c7acb/django/urls/converters.py#L68

Should we add explicit `__init__(self) -> None` to the protocol?

adamchainz:
I agree, done.

sobolevn (approved):
Thanks!

PR #1991, opened 2024-03-05
Remove some deprecated Django 3.x APIs

sobolevn:
Since we now focus on 5.0, there's no reason to keep these two around.


flaeppe (approved):

intgr:
Django 3.2 technically still has official "security support" until 04 Mar 2024. But the EOL date is close enough that I think its fine to remove it. https://endoflife.date/django

Please remove Django 3.2 from "Django partial support" in README too.

sobolevn:
Thanks everyone!

PR #1993, opened 2024-03-07
Fix BaseModelAdmin.view_on_site annotation

### LINKED ISSUE #1755: Incorrect annotation for ModelAdmin.view_on_site
# Bug report

## What's wrong

I have a ModelAdmin that has a `view_on_site` method similar to [the example in Django docs](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.view_on_site):

```python
class PersonAdmin(admin.ModelAdmin):
    def view_on_site(self, obj):
        url = reverse("person-detail", kwargs={"slug": obj.slug})
        return "https://example.com" + url
```

If I annotate it like so:

```python
class PersonAdmin(admin.ModelAdmin[Person]):
    def view_on_site(self, obj: Person) -> str:
        url = reverse("person-detail", kwargs={"slug": obj.slug})
        return "https://example.com" + url
```

I get a type warning:

```
error: Signature of "view_on_site" incompatible with supertype "BaseModelAdmin"  [override]
note:      Superclass:
note:          bool | Callable[[_ModelT], str]
note:      Subclass:
note:          def view_on_site(self, obj: Person) -> str
```

## How is that should be

It should not generate the warning. I don't know how to fix that though :-/

## System information

- OS:
- `python` version: 3.10.12
- `django` version: 4.2.5
- `mypy` version: 1.5.1

cuu508:
# I have made things!

## Related issues

Closes #1755

### DIFF
--- a/django-stubs/contrib/admin/options.pyi
+++ b/django-stubs/contrib/admin/options.pyi
@@ -99,7 +99,6 @@ class BaseModelAdmin(Generic[_ModelT]):
     readonly_fields: _ListOrTuple[str]
     ordering: _ListOrTuple[str] | None
     sortable_by: _ListOrTuple[str] | None
-    view_on_site: bool | Callable[[_ModelT], str]
     show_full_result_count: bool
     checks_class: Any
     model: type[_ModelT]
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks!

Issue #1992, opened 2024-03-06
Class attributes annotated as mutable types when mutability is not required


samueljsb:
# Bug report

## What's wrong

There are many attributes that are annotated here as mutable types (lists, dicts) that don't actually need to be mutable. It would be perfectly valid to use immutable types (tuples, `MappingProxyType` or frozendicts), but the type checker then complains because the annotation here appear to require mutability.

There are a lot of these, and I don't want to do a half-baked job of listing them exhaustively, so as an example, we can consider `TemplateView`'s `extra_context`:

```python
from types import MappingProxyType

from django.views import generic


class MyView(generic.TemplateView):
    extra_context = MappingProxyType({
        'foo': 1,
        'bar': 42,
    })
```

```console
t.py:7: error: Incompatible types in assignment (expression has type "MappingProxyType[str, int]", base class "ContextMixin" defined the type as "dict[str, Any] | None")  [assignment]
Found 1 error in 1 file (checked 1 source file)
```

## How is that should be

I think this case should be annotated as `Mapping[str, Any] | None`, because a mapping is required but not necessarily a mutable one.

## System information

This example was with the following, but I've experienced this problem with various combinations and versions.

- OS: macOS 14.3.1
- `python` version: v3.11.5
- `django` version: 5.0.3
- `mypy` version: 1.8.0 
- `django-stubs` version: 4.2.7
- `django-stubs-ext` version: 4.2.7

sobolevn:
PR is welcome.

Generally we annotate attributes with their concrete types. Because Django does not provide any promises about how they would be used.

So, if you have any speicific examples, we can decide on a per-case basis.

sobolevn:
Please, open new issues for any other places you find this problem 👍

PR #1994, opened 2024-03-07
Allow immutable `extra_context` on `TemplateView`s

samueljsb:
Prior to this change, the `extra_context` attribute of a template view was annotated as a `dict` (or `None`). This means that if a value is provided, it *must* be a mutable object. Since this is an attribute on the class, this creates mutable shared state, which is a source of bugs. It was not possible to use an immutable mapping (`MappingProxyType or the 3rd-party `frozendict`) without getting type-checking errors.

However, there is no need for this attribute to be mutable. This change allows *any* mapping type, which allows authors to write template views with immutable `extra_context` attributes.

Refs #1992

### DIFF
--- a/django-stubs/views/generic/base.pyi
+++ b/django-stubs/views/generic/base.pyi
@@ -1,6 +1,6 @@
 import logging
 from collections.abc import Callable, Sequence
-from typing import Any
+from typing import Any, Mapping
 
 from django.http.request import HttpRequest
 from django.http.response import HttpResponse, HttpResponseBase
@@ -9,7 +9,7 @@ from django.utils.functional import _Getter
 logger: logging.Logger
 
 class ContextMixin:
-    extra_context: dict[str, Any] | None
+    extra_context: Mapping[str, Any] | None
     def get_context_data(self, **kwargs: Any) -> dict[str, Any]: ...
 
 class View:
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
This case is explicitly documented to be `dict`: https://docs.djangoproject.com/en/5.0/ref/class-based-views/mixins-simple/#django.views.generic.base.ContextMixin.extra_context

samueljsb:
I don't think that documentation is intended to be an exhaustive definition of the types that can be passed in, though. For example, it doesn't mention that the extra context can be `None`.

Moreover, the way `extra_context` is being used does not require mutability (it is being passed to `dict.update`, which will accept any mapping), so it is acceptable to set an immutable mapping on that attribute. Django will never mutate the `extra_context` on a view class because doing so would leak state between requests.

Is that sufficient for django-stubs to annotate it to allow immutable objects here?

adamchainz (approved):
IMO yes, let's do it.

sobolevn:
Thank you all for the feedback! 👍

samueljsb:
It has just been pointed out to me that the commit message is incorrect - it should have been "allow *im*mutable `extra_context`". I've fixed the PR title, but whoever writes the changelog will want to make sure they get that the right way around. Sorry!

sobolevn:
No worries, it is fine :)

PR #1999, opened 2024-03-11
Add BoundField.__html__()

### LINKED ISSUE #1998: Add type for BoundField.__html__()
# Bug report

## What's wrong

BoundField can be rendered as html by calling `__html__()`, via RenderableFieldMixin:
https://github.com/django/django/blob/3592e9fcb13b1353cfa78f03c2574fe9282ce5ac/django/forms/utils.py#L81

`__html__()` is currently not defined on the type definition for BoundField.

## How is that should be
`__html__()` should be defined (returning SafeString)

pelme:
Not sure if the test case is needed? Let me know and I will drop it!

Closes #1998


tests/typecheck/test_forms.yml
```diff
@@ -112,3 +112,12 @@
 
                 class Article(models.Model):
                     name = models.CharField(max_length=128)
+-   case: boundfield_html
```
sobolevn: No. It is not needed :)

pelme: Done, it is now gone!

sobolevn (approved):
Thank you!

PR #2000, opened 2024-03-11
Remove `class Meta` from `Model` and `Form` class stubs

jorenham:
This fixes several abstract models and model-forms with incorrect annotations for a non-existant `Meta` type, e.g.

```pycon
In [1]: from django import forms

In [2]: forms.BaseForm.Meta
Cell In[2], line 1
----> 1 forms.BaseForm.Meta

AttributeError: type object 'BaseForm' has no attribute 'Meta'
```

Additionally, the `django_stubs_ext.db.models.ModelMeta` type alias was added, which can be used as a less-incorrect variant of `django_stubs_ext.db.models.TypedModelMeta` when annotating the `Meta` class attribute of a `Model`, e.g.:

```python
class SupportsModelMetaAndSpam(typing.Protocol):
    Meta: type[django_stubs_ext.db.models.ModelMeta]
    spam: str
```


jorenham:
Congratz on the 2000th issue+PR btw 🎉

sobolevn (commented):
I think the whole reason for this annotation is that all `BaseForm` subclasses do have `Meta` attribute. Including user-defined classes.

@flaeppe do you agree with this change?

flaeppe:
> I think the whole reason for this annotation is that all `BaseForm` subclasses do have `Meta` attribute. Including user-defined classes.
> 
> @flaeppe do you agree with this change?

Yes, I do. But arguably incomplete, since `class Meta` exists on `BaseModelForm` and behaves the same way as for `Model.Meta`: it only exists for subclasses of `BaseModelForm`.

Ref: https://github.com/django/django/blob/80fe2f439102dc748ff8ddd661d94935915bd3e7/django/forms/models.py#L270-L277

In any case it should be removed from `BaseForm`.

jorenham:
There is actually a much larger issue with the way that `Meta` is currently typed:

```python
from typing import Protocol

class SupportsMeta(Protocol):
    class Meta:
        ...

class MyModel:
    class Meta:
        verbose_name = 'my model'

obj_with_meta: SupportsMeta = MyModel()
"""
Pyright:

Expression of type "MyModel" cannot be assigned to declared type "SupportsMeta"
  "MyModel" is incompatible with protocol "SupportsMeta"
    "Meta" is an incompatible type
      "Untitled.MyModel.Meta" is incompatible with "Untitled.SupportsMeta.Meta"
      Type "type[Untitled.MyModel.Meta]" cannot be assigned to type "type[Untitled.SupportsMeta.Meta]"  (reportAssignmentType)
"""
```

Solving this appeared to be harder than expected, though... The best I could come up with was this:

```python

class _MetaVerboseName(Protocol):
    verbose_name: str

class _MetaVerboseNamePlural(Protocol):
    verbose_name_plural: str

# ... [one protocol for each allowed Meta attribute]

# emulates type with >=1 of the allowed Meta class attributes
type AnyMeta = _MetaVerboseName | _MetaVerboseNamePlural


# must be made generic, otherwise the meta type won't be covariant
class SupportsMeta[M: AnyMeta](Protocol):
    Meta: type[M]

type SupportsAnyMeta = SupportsMeta[Any]


class MyModel:
    class Meta:
        verbose_name = 'my model'
        verbose_name_plural = 'our models'

obj_with_meta: SupportsAnyMeta = MyModel()
```

Even in pyright strict mode, this is valid.

sobolevn:
Yes, using a protocol is a good idea in this case! Thanks for working on this issue :)

jorenham:
@sobolevn  Would you like me to introduce this fix in this PR, or keep it as-is and open up a new one for it?

sobolevn:
I think we can continue working here, but it is up to you: what feels easier.

flaeppe:
In the README there's currently a different (but more relaxed?) approach mentioned for model `Meta`:

https://github.com/typeddjango/django-stubs/?tab=readme-ov-file#type-checking-of-model-meta-attributes

Also, not sure how much it could impact. Not too familiar with nested class definitions and how type checkers like them. But _I think_ `Meta` could be declared a `@property` i.e. read only attribute on the protocol?

Quite ingenious idea with the union approach to get rid of the issue supporting partial protocols (wanted to but couldn't find any issue link right now)

jorenham:
@flaeppe the `Protocol` solution I proposed would make the `TypedModelMeta` redundant

flaeppe:
Yes, I understood that. Just wanted to mention it in case you hadn't seen it.

jorenham:
> I think we can continue working here, but it is up to you: what feels easier.

I'll make the 2000th PR a big one then :)

jorenham:
> Yes, I understood that. Just wanted to mention it in case you hadn't seen it.

Yea thanks for that; it saves me from having to figure out the attribute annotations myself :)

jorenham:
BTW, here's the pyright-playground link with the Protocol-based Meta solution:

https://pyright-play.net/?pyrightVersion=1.1.353&pythonVersion=3.12&pythonPlatform=Linux&strict=true&locale=en&code=MQAgZgTg9gtiAuBPADgSwHYHMSpsqE8IACtPFAMZQA2AULaBdQIYDOrIAygK7L6GsAsgFN4zABSko5KtQCUALgYgVKpmw4ixS0Kr0gAdEfqMW7EIMSCoAE2HUd%2B9ea3NH%2B1QDdhEAEZRWYQB9dGYYYRAAXhAAchhEEBhbexiTEChfACsggHdUeAALIPDtLl5%2BeCFRZiiLK2TqcTllACI25WJECFRMAvglZQBRAA9kCGF2VCh0dLAEFAiWy2s7ahaQCmZ0dGkQXwiNHvRhGwQoEDt1cdOkZEWePgJK1xblFSX61fXUDgwqPGY8FQvmoETyhRAY2klBoIBaDwqVTEr10qiW1W%2BHC2OHQ-2QgOBoPmdze%2BhaAFV0ED4KCbAZlg16RicL9cbB8UCQWD8gU4ZTqbSDAinkjmEzkaS9AAVBZw27CADa-PygoZq3FzAAuutNttdvsQIdMMcbud5XKFkqqSqTkLyiLXBrtSpxOMKgBBSbG8JUmV3ZqgNqvWiQWDEjDYXAVEDu9CIAA0JDIMLotGcHCCrgAaj5-IEAHJhYSSZOyRS0PTePwBYKhcIKECseAQejpkCZ6o56sFovEajcCDMRpSGQ0cuV3M1kJFoLIfuDhyN5tpIwGEAK6YRKGj6jgAggYTMCi8ofUKA5E4WaqG%2BDN4HceDCTVpYQwbgsR8cc3g3kAPkiACMswIAUBzUGeF6nK4GxmFit7dL4D4TLQ5qxlY17RB2YhdnmwiFuEIAAD7ttmk49uEfYDkO9DKG%2BTZ7BEMDMHYICYMIxzdBQibSKBEB5IEIGMde37TDERAGlQnjMN0WzwGmsFlI8AiuAqggNmhriaiW0JlkoeiuA28qqc%2BKGysKAgaRhimIipaEmfJGh1Cs9h6aobYGRWHggFWuHTgR0RxAkSSrKkXneWRtYznOVG7gFUADokDSsKktAZNkP7FNUDbmZUlliLUar2E0tBAA

sobolevn:
https://pyright-play.net/?pyrightVersion=1.1.353&pythonVersion=3.12&pythonPlatform=Linux&strict=true&locale=en&code=MQAgZgTg9gtiAuBPADgSwHYHMSpsqE8IACtPFAMZQA2AULaBdQIYDOrIAygK7L6GsAsgFN4zABSko5KtQCUALgYgVKpmw4ixS0Kr0gAdEfqMW7EIMSCoAE2HUd%2B9ea3NH%2B1QDdhEAEZRWYQB9dGYYYRAAXhAAchhEEBhbexiTEChfACsggHdUeAALIPDtLl5%2BeCFRZiiLK2TqcTllACI25WJECFRMAvglZQBRAA9kCGF2VCh0dLAEFAiWy2s7ahaQCmZ0dGkQXwiNHvRhGwQoEDt1cdOkZEWePgJK1xblFSX61fXUDgwqPGY8FQvmoETyhRAY2klBoIBaDwqVTEr10qiW1W%2BHC2OHQ-2QgOBoPmdze%2BhaAFV0ED4KCbAZlg16RicL9cbB8UCQWD8gU4ZTqbSDAinkjmEzkaS9AAVBZw27CADa-PygoZq3FzAAuutNttdvsQIdMMcbud5XKFkqqSqTkLyiLXBrtSpxOMKgBBSbG8JUmV3ZqgNqvWiQWDEjDYXAVEDu9CIAA0JDIMLotGcHCCrgAaj5-IEAHJhYSSZOyRS0PTePwBYKhcIKECseAQejpkCZ6o56sFovEajcCDMRpSGQ0cuV3M1kJFoLIfuDhyN5tpIwGEAK6YRKGj6jgAggYTMCi8ofUKA5E4WaqG%2BDN4HceDCTVpYQwbgsR8cc3g3kAPkiACMswIAUBzUGeF6nK4GxmFit7dL4D4TLQ5qxlY17RB2YhdnmwiFuEIAAD7ttmk49uEfYDkO9DKG%2BTZ7BEMDMHYICYMIxzdBQibSKBEB5IEIGMde37TDERAGlQnjMN0WzwGmsFlI8AiuAqggNmhriaiW0JlkoeiuA28qqc%2BKGysKAgaRhimIipaEmfJGh1Cs9h6aobYGRWHggFWuHTgR0RxAkSSrKkXneWRtYznOVG7tEAAcCWJYl9AZNkP7FNUDbmZUlliLUar2E0tBAA

this is interesting

jorenham:
@sobolevn I don't see pyright allows `int <: str` in this case. Adding `ClassVar` everywhere doesn't help either. Perhaps a bug?

jorenham:
@sobolevn Ah I understand now. If you remove the correct `verbose_name: str`, leaving only the `verbose_name_plural: int`, it'll give an error as expected. 

Because of the union; pyright only requires one of the attributes to be correct. I.e. pyright uses short-circuit evaluation on the union. Unfortunately, that's technically correct: Since a hypothetical subtype of `_MetaVerboseName` could have a `verbose_name_plural: int`. 

I can think of only way around this: Make each individual protocols `@final`, and additionally create separate `@final` Protocols for each of the possible combinations. 
However, since we have 24 attributes, that would require `math.factorial(24) == 620_448_401_733_239_439_360_000` protocols...

django-stubs/contrib/auth/models.pyi
```diff
@@ -61,8 +63,7 @@ class PermissionsMixin(models.Model):
     groups = models.ManyToManyField(Group)
     user_permissions = models.ManyToManyField(Permission)
 
-    class Meta:
-        abstract: Literal[True]
+    Meta: ClassVar[type[_ModelMeta]]
```
sobolevn: I think that this now misses that fact that `abstact = True` is set here.

jorenham: Yes; and that's ok. 

This model's `Meta` attribute is a type, and nothing but a type. So the specific attributes that it has (i.e. its subtype), should be considered implementation. 

Otherwise, all of the subtypes of the model must also must have that particular `Meta` type. So in this case, that would require all models that extend `PermissionsMixin` to have a `Meta` with `abstract = True`.

jorenham:
I don't understand the reason behind the failing tests 🤔

sobolevn:
it is something related to our plugin 🤔 
Refs https://github.com/typeddjango/django-stubs/commit/a1b1f51f432c4f4ec136c071931c4f07bb4f34fb and https://github.com/typeddjango/django-stubs/commit/53cdbe4c9e31a491378e027f08c6b5421a843958 CC @flaeppe

django-stubs/contrib/auth/models.pyi
flaeppe: The plugin tests currently expects to find a type definition for the `abstract` attribute here.

A different approach would be to e.g. hard code any abstract models that the plugin needs to know about. That's totally fine to do for the models shipped with Django

jorenham: @flaeppe Alternatively, is it an option to use a comment, e.g. `# django-stubs: abstract`, to make the plugin aware of it?

jorenham: admittedly, I am not familiar with the internals of mypy, or mypy plugins for that matter

flaeppe: Perhaps, but I think I would suggest instead keeping it internally in the plugin

sobolevn: Comment won't work, it needs to be a part of the ast. 

@flaeppe what about other django-plugins that would have annotated abstract models?
Right now it would work for them to just add `abstract: Literal[True]`, but if we list all abstract models manually, it won't work.

Maybe adding something like `_pyi_model_abstract: Literal[True]`? We already have several custom attributes like that. We just can add one more.

flaeppe: We could still keep the code for `abstract: Literal[True]`, should still work. It's just that we, to cover the Django internals, extend with a curated list of abstract models shipped with Django.

sobolevn: It also sounds fine! 👍

flaeppe: I've opened #2017 just now that should do what I've been rambling about

flaeppe:
> @flaeppe the `Protocol` solution I proposed would make the `TypedModelMeta` redundant

In regards to the above comment, should we remove `TypedModelMeta` and the [README section](https://github.com/typeddjango/django-stubs/?tab=readme-ov-file#type-checking-of-model-meta-attributes) in this PR?

flaeppe:
Also, I'm wondering if `Meta: ClassVar[type[_ModelMeta]]` should be declared on e.g. `django.db.models.Model`? 

I think that would remove each declaration introduced in this PR?

intgr:
> > the Protocol solution I proposed would make the TypedModelMeta redundant
>
> In regards to the above comment, should we remove TypedModelMeta and the [README section](https://github.com/typeddjango/django-stubs/?tab=readme-ov-file#type-checking-of-model-meta-attributes) in this PR?

Hold your horses. First it needs to be established that `TypedModelMeta` truly is redundant. I think the `Union[]` solution is at best a partial solution (as mentioned in https://github.com/typeddjango/django-stubs/pull/2000#issuecomment-1992547122)

More problematic... I think mypy doesn't support this approach at all. I tested this MR and I don't think mypy is capable of type-checking `class Meta` when `Model` defines it is as `Meta: ClassVar[type[_ModelMeta]]`. I see the focus here was on Pyright, but let's not forget mypy.

I submitted a feature request for this last year:

* https://github.com/python/mypy/issues/14767

flaeppe:
I see, so let's not remove `TypedModelMeta`

jorenham:
@intgr Since the tests are all mypy-based, it certainly won't be forgotton :). 

The reason I tend to use pyright in my examples, is that in my experience it is a lot stricter than mypy, which your mypy issue is a good example of. 

This is unrelated to this PR, but if I run `pyright django-stubs` (with the default config, so not even in `strict` mode), it outputs `423 errors, 19 warnings, 0 informations` 😅. So I can't say that there's any focus on pyright at all, although perhaps there should be 🤷🏻‍♂️ .

jorenham:
> Also, I'm wondering if `Meta: ClassVar[type[_ModelMeta]]` should be declared on e.g. `django.db.models.Model`?
> 
> I think that would remove each declaration introduced in this PR?

I see your point @flaeppe , but it would be incorrect:

```pycon
>>> import django.db.models
>>> django.db.models.Model.Meta
  File "<stdin>", line 1, in <module>
AttributeError: type object 'Model' has no attribute 'Meta'
```

sobolevn:
@jorenham do you want to contribute a pyright CI job? I can be just hidden under `|| true` or `allow-errors: true` or whatever.

intgr:
> @intgr Since the tests are all mypy-based, it certainly won't be forgotton :).

Just to be clear, the discussion that I opposed was about removing documentation. Mypy hasn't yet grown LLM powers that it can typecheck documentation. 😄

I'm very much in favor of improving Pyright support, as long as we don't degrade support for mypy users.

### Union of protocols is problematic

In response to https://github.com/typeddjango/django-stubs/pull/2000#issuecomment-1992547122

> I can think of only way around this: Make each individual protocols `@final`, and additionally create separate `@final` Protocols for each of the possible combinations. 
> However, since we have 24 attributes, that would require `math.factorial(24) == 620_448_401_733_239_439_360_000` protocols...

Yep, that won't work. :(

As it stands currently, this PR only typechecks that `class Meta` contains 1 attribute with valid type, remaining attributes are allowed to be incorrect. IMO that's worse than no typechecking, since it can give users a false sense of security.

Can anyone think of any alternatives or work-arounds to achieve what we want? If not, we're at an impasse.

#### Optional protocol members

What we really need is what PEP 544 calls "[optional protocol members](https://peps.python.org/pep-0544/#support-optional-protocol-members)"

> it would be handy to be able to say that a method or data attribute does not need to be present in a class implementing a protocol, but if it is present, it must conform to a specific signature or type.

I believe no type checkers currently implement such a thing? There are old open issues in python typing project and mypy, with no recent progress:

* https://github.com/python/typing/issues/601
* https://github.com/python/mypy/issues/5504

#### Reuse TypedModelMeta?

If we want to enforce strict checking of `Meta`, we could possibly hint it as `Meta: ClassVar[type[TypedModelMeta]]`, which would *require* users to always derive Meta classes from `TypedModelMeta`.

I'm not a fan of this approach, I think it would be a bit too intrusive for people who try to adopt `django-stubs` gradually.

jorenham:
@intgr I don't think it's a good idea to require changes to the implementation for the purposes of typing, of which `TypedModelMeta` is an example.

I tend to agree with your "all or nothing" mentality regarding typing. But since "all" is practically impossible here, we should look at the remaining $1 + 0.99 \approx 2$ options:

1. Remove `Meta` altogether (no, not that one; please calm down mr. Zuckerberg...)
2. Scrape Django projects on Github to find the 99% most frequently used `Meta` attribute combinations, and create union'ed Protocols for those.

My preference would be to go for option 1 first, and place option 1.99 under the "future work" section.

jorenham:
> @jorenham do you want to contribute a pyright CI job? I can be just hidden under `|| true` or `allow-errors: true` or whatever.

Yea sure, I'll have a look in the coming days

intgr:
> Remove `Meta` altogether

Indeed. But if you want to propose that, please open a separate PR. As nice as number 2000 is, most of the discussion here would be irrelevant for that proposal.

#### Model.Meta

As we've established here, having `Meta` there does nothing for typechecking inner attributes of Meta.

And it seems Model metaclass magic even tries to hide the `Meta` attribute of models. For example, [despite `Permission` class having explicit `class Meta:`](https://github.com/django/django/blob/71368b6f00b44595115aac8e26811f42b5db77b7/django/contrib/auth/models.py#L72-L76), it's not actually accessible "from the outside":

```python
>>> from django.contrib.auth.models import Permission
>>> Permission.Meta
AttributeError: type object 'Permission' has no attribute 'Meta'
>>> Permission().Meta
AttributeError: 'Permission' object has no attribute 'Meta'
```

In almost all situations, the "correct" approach would be to access `Model._meta` anyway, not `.Meta`. (Despite being named with underscore, `_meta` is stable public API in Django).

So 👍 from me. Does anyone see any downsides or issues that could cause?

#### BaseForm.Meta

I haven't used Django forms for years, so I don't have an informed position on this. Is `class Meta` required or optional in forms? Is there an equivalent of `_meta` in form classes? Is the `form.Meta` attribute accessible from the outside?

intgr:
> Scrape Django projects on Github to find the 99% most frequently used Meta attribute combinations, and create union'ed Protocols for those.

👎 I don't think what will work. We don't want to flag the remaining 1% as errors if they're actually correct. And if we add a "loose protocol" to the union for those use cases, then it also spoils typechecking for the remaining 99%. Besides that, it would be an extremely ugly hack.

sobolevn:
Yes, I also agree. Hacking something should be justified: right now using `Protocol`s here does not seem like a proper solution. We need something else.

jorenham:
@intgr model forms also cast the same black magic spell as models on `Meta` to turn it into `_meta`. Here's the incantation:  https://github.com/django/django/blob/5.0.3/django/forms/models.py#L277

jorenham:
But you make fair points @intgr and @sobolevn ; I'll remove the messy `Meta` magic. 

But for the sake of backwards compatibility; I assume that we're gonna want to keep `TypedModelMeta` for now, right? And perhaps mark it as deprecated, e.g. with `@typing_extensions.deprecated`?

ext/django_stubs_ext/db/models/__init__.py
```diff
@@ -1,6 +1,7 @@
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
+
```
intgr: Unnecessary unrelated change.

```suggestion
```

scripts/stubtest/allowlist_todo.txt
```diff
@@ -100,22 +100,20 @@ django.contrib.auth.admin.UserAdmin.form
...
-django.contrib.auth.forms.BaseUserCreationForm.Meta.field_classes
-django.contrib.auth.forms.BaseUserCreationForm.Meta.model
+django.contrib.auth.forms.BaseUserCreationForm.Meta
 django.contrib.auth.forms.BaseUserCreationForm.declared_fields
 django.contrib.auth.forms.PasswordChangeForm
 django.contrib.auth.forms.PasswordResetForm
 django.contrib.auth.forms.SetPasswordForm
-django.contrib.auth.forms.UserChangeForm.Meta.field_classes
-django.contrib.auth.forms.UserChangeForm.Meta.model
+django.contrib.auth.forms.UserChangeForm.Meta
```
intgr: Since it's our intentional choice now: please move all these `.Meta` ignores to `allowlist.txt`, add a comment header to them and link to this PR.

django-stubs/contrib/flatpages/forms.pyi
```diff
@@ -1,7 +1,8 @@
 from typing import Any
 
 from django import forms
+from django.contrib.flatpages.models import FlatPage
 
-class FlatpageForm(forms.ModelForm):
+class FlatpageForm(forms.ModelForm[FlatPage]):
```
intgr: Unrelated to the PR topic. But if you don't feel like spliting this out, I'm fine with merging it together too.

intgr (changes requested):
A few suggestions.

I didn't notice before that you had updated this. I was expecting a separate PR ("if you want to propose that, please open a separate PR"). But no big deal; since you already pushed here, no need for additional bureaucracy.

intgr:
> I assume that we're gonna want to keep `TypedModelMeta` for now, right?

Yes, I haven't seen any reasons to remove it.

django-stubs/contrib/flatpages/forms.pyi
jorenham: I must've been on auto-pilot; I'll remove it

jorenham:
> A few suggestions.
> 
> I didn't notice before that you had updated this. I was expecting a separate PR ("if you want to propose that, please open a separate PR"). But no big deal; since you already pushed here, no need for additional bureaucracy.

Oh woops I didn't notice. But either way, it would've been sad if the 2000th PR wouldn't have made it :)

intgr (approved):
Awesome, thanks!

PR #2001, opened 2024-03-12
Allow timedelta type for session.set_expiry() argument

mlazar-endear:
# I have made things!

Fixes the type hint for `BaseSession.set_expiry` to allow `timedelta`.

The allowed types for this method are enumerated in the docstring:

https://github.com/django/django/blob/80fe2f439102dc748ff8ddd661d94935915bd3e7/django/contrib/sessions/backends/base.py#L249-L275

## Related issues

This was such a minor fix that I didn't make an issue for it.


sobolevn (approved):
Thanks!

PR #2004, opened 2024-03-13
5.0: Update `django.contrib.admin`

ngnpope:
# I have made things!

Updates stubs for `django.contrib.admin` for Django 5.0.

## Related issues

Refs #1493


django-stubs/contrib/admin/filters.pyi
```diff
@@ -17,18 +20,22 @@ class ListFilter:
...
+    def get_facet_counts(self, pk_attname: str, filtered_qs: QuerySet[Model]) -> dict[str, Count]: ...
+    def get_facet_queryset(self, changelist: ChangeList) -> dict[str, int]: ...
+
+class SimpleListFilter(FacetsMixin, ListFilter):
     parameter_name: str | None
     lookup_choices: Any
     def value(self) -> str | None: ...
     def lookups(self, request: HttpRequest, model_admin: ModelAdmin) -> Iterable[tuple[Any, str]] | None: ...
-    def choices(self, changelist: Any) -> Iterator[dict[str, Any]]: ...
+    def choices(self, changelist: ChangeList) -> Iterator[dict[str, Any]]: ...
```
sobolevn: We can remove all `choices`, because they are defined on `ListFilter`

django-stubs/contrib/admin/views/main.pyi
```diff
@@ -77,7 +77,7 @@ class ChangeList:
     def get_ordering_field(self, field_name: Callable | str) -> Expression | str | None: ...
     def get_ordering(self, request: HttpRequest, queryset: QuerySet) -> list[Expression | str]: ...
     def get_ordering_field_columns(self) -> dict[int, Literal["desc", "asc"]]: ...
-    def get_queryset(self, request: HttpRequest) -> QuerySet: ...
+    def get_queryset(self, request: HttpRequest, exclude_parameters: list[str | None] | None = ...) -> QuerySet: ...
```
sobolevn, line 80: Is it really a `list`? Or can we use other `Collection`s?

sobolevn (commented):
Thanks!

django-stubs/contrib/admin/filters.pyi
ngnpope: Will have a look at this later. I think there are some other bits that are unnecessarily duplicated too.

intgr:

django-stubs/contrib/admin/views/main.pyi
ngnpope: In this case I think it makes sense to leave this as a `list`.

On current `stable/5.0.x`, a grep yields the following:

```console
$ git grep -n exclude_parameters
django/contrib/admin/filters.py:85:            self.request, exclude_parameters=self.expected_parameters()
django/contrib/admin/views/main.py:531:    def get_queryset(self, request, exclude_parameters=None):
django/contrib/admin/views/main.py:544:                exclude_parameters is None
django/contrib/admin/views/main.py:545:                or filter_spec.expected_parameters() != exclude_parameters
```

It's doing a direct comparison against the return type of `ListFilter.expected_parameters()` which is typed as `list[str | None]`. Based on the implementation, accepting `Collection[str | None]` would break things.

sobolevn (approved):
Thank you!

Issue #2012, opened 2024-03-15
Django Google Summer of Code 2024


adamchainz:
Like last year (#1337), Django is doing GSoC again and django-stubs is on the list of potential projects.

It actually remained there because it was just copied from last year’s wiki page, with two projects that I think are now irrelevant:

1. Mypy strict-mode support - the stubs should be mostly(?) strict-mode compatible since #1556.
2. Merging django-types back in - somewhat covered by supporting other checkers in #1628, and hard to merge further due to continued drift.

So I just deleted those projects from the wiki page and added a generic one, [Expand django-stubs coverage](https://code.djangoproject.com/wiki/SummerOfCode2024#Expanddjango-stubscoverage), which would cover any proposal for clearing some of the entries from the stubtest TODO files.

Other maintainers and regular contributors, do you think this is okay? Would you be willing to mentor, and do you have any other possible projects?

nebiyuelias1:
@adamchainz thank you very much for creating this issue (I was the one who got interested in those project ideas from last year). I was wondering how you produced the allowlist_todo.txt and allowlist_todo_django50.txt files.

adamchainz:
They are produced by the stubtest tool. Read the relevant part of the contributing guide and try run it yourself locally.

sobolevn:
I won't participate in GSoC ever again: https://developers.google.com/open-source/gsoc/faq?hl=en#are_participants_from_ukraine_russia_or_belarus_allowed_to_participate_in_gsoc_2024

But, I have no problem with other contributors / maintainers participating :)

nebiyuelias1:
Can anyone please explain to me why the stubtest reports a lot of errors when I run it locally whereas the github action seems to be showing green(all went well I suppose)? I was also wondering why the the number of errors decreases when I remove an entry from the allowlist_todo.txt file? I was expecting the opposite.

sobolevn:
It depends on how you install your stub package (see https://github.com/typeddjango/django-stubs/blob/f92821a3f55f95854a1b1802d5077ced9b5d205d/.github/workflows/test.yml#L34) and how you set your `MYPYPATH` (see https://github.com/typeddjango/django-stubs/blob/f92821a3f55f95854a1b1802d5077ced9b5d205d/scripts/stubtest.sh#L7)

Does it help? Feel free to send a PR with fixes to `CONTRIBUTING.md` :)

nebiyuelias1:
[...]

PR #2015, opened 2024-03-18
Clean the cache on each run of `stubtest`

sobolevn:
This should solve the issue in https://github.com/typeddjango/django-stubs/issues/2012
CC @nebiyuelias1 @intgr 

Please, try this change, report whether it works for you or not.


intgr (approved):
Thanks! Tested and this solves the issue.

I should also open a bug report against stubtest so the bug can be fixed upstream.

PR #2017, opened 2024-03-19
Keep abstract Django models internally in the plugin

flaeppe:
Renders `class Meta` declarations unnecessary in any .pyi files, for database models.

- Refs: #2000

CC @jorenham @sobolevn


flaeppe:
The changes here should resolve the failing test in #2000

sobolevn (approved):
Thanks!

PR #2019, opened 2024-03-21
Pyright joins the workflow in an advisory capacity

jorenham:
as suggested in https://github.com/typeddjango/django-stubs/pull/2000#issuecomment-2009326310

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,6 +95,12 @@ jobs:
       - name: Run stubtest
         run: bash ./scripts/stubtest.sh
 
+      - name: Run pyright
+        uses: jakebailey/pyright-action@v2
+        with:
+          pylance-version: latest-release
+        continue-on-error: true
+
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks! Here's my plan for the future:
- We should check our own source code as strict as possible, errors can be fixed and checked rather soon
- stubs will be not-so-strict, because they are harder to type, we can start with the similar approach typeshed has for different kind of types and different settings files

Viicos:
Would be happy to contribute more with pyright compatibility as this is one of my main goals!

sobolevn:
@Viicos please, go ahead :)

See:
- https://github.com/python/typeshed/blob/main/pyrightconfig.json
- https://github.com/python/typeshed/blob/main/pyrightconfig.scripts_and_tests.json
- https://github.com/python/typeshed/blob/main/pyrightconfig.stricter.json

PR #2021, opened 2024-03-21
feat: Allow setting django_settings_module from env

### LINKED ISSUE #2018: Allow setting `django_settings_module` from the `DJANGO_SETTINGS_MODULE` env variable
Hi there!

Hope everyone is doing alright! I have a use-case where I deploy my Django application with different `INSTALLED_APPS` in two different deployments. The least invasive way I found to do this was having two separate settings files and switching between them based on whichever one I am deploying. Things like `pytest` allow the setting of `django_settings_module` from both the environment and the `pyproject.toml`. It seems like `django_stubs` wants the settings to be static which doesn't really make sense to me considering Django has deemed it as an env variable for that flexibility. 

Let me know if I am missing something, or if there is a workaround/fix that would be easier.

Thanks!

armanckeser:
## What changed
* `django_settings_module` is now not required as part of the config
* When not given, `django_settings_module` will be populated from `DJANGO_SETTINGS_MODULE` env variable
* Error messages now mention setting either of these values
* Tests added to see correct population of the config

## Why
* Some projects use `DJANGO_SETTINGS_MODULE` to deploy multiple variants of an app, this will allow the config to be dynamic to accommodate
* `DJANGO_SETTINGS_MODULE` acts as the single source of truth when it comes to the project, so allowing population from that variable will reduce duplication.

## Related issues

- Closes #2018 
- An updated version of #153

### DIFF
--- a/mypy_django_plugin/config.py
+++ b/mypy_django_plugin/config.py
@@ -83,9 +83,11 @@ class DjangoPluginConfig:
         except KeyError:
             toml_exit(MISSING_SECTION.format(section="tool.django-stubs"))
 
-        self.django_settings_module = config.get("django_settings_module") or os.getenv(DJANGO_SETTINGS_ENV_VAR)
-        if not self.django_settings_module:
+        django_settings_module = config.get("django_settings_module") or os.getenv(DJANGO_SETTINGS_ENV_VAR)
+        if not django_settings_module:
             toml_exit(MISSING_DJANGO_SETTINGS)
+
+        self.django_settings_module = django_settings_module
 
         if not isinstance(self.django_settings_module, str):
             toml_exit("invalid 'django_settings_module': the setting must be a string")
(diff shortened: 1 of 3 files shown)


mypy_django_plugin/config.py
```diff
@@ -80,10 +83,10 @@ def parse_toml_file(self, filepath: Path) -> None:
         except KeyError:
             toml_exit(MISSING_SECTION.format(section="tool.django-stubs"))
 
-        if "django_settings_module" not in config:
+        self.django_settings_module = config.get("django_settings_module") or os.getenv(DJANGO_SETTINGS_ENV_VAR)
```
armanckeser: For the sake of backwards compatibility, I elected to default to config instead of the env var

mypy_django_plugin/config.py
```diff
@@ -14,22 +15,24 @@
 (config)
 ...
 [mypy.plugins.django-stubs]
-django_settings_module = str (required)
+django_settings_module = str (default: DJANGO_SETTINGS_MODULE env var)
```
sobolevn:
```suggestion
django_settings_module = str (default: `os.getenv(DJANGO_SETTINGS_MODULE)`)
```

let's be consistent.

mypy_django_plugin/config.py
```diff
@@ -14,22 +15,24 @@
...
 strict_settings = bool (default: true)
 ...
 """
 TOML_USAGE = """
 (config)
 ...
 [tool.django-stubs]
-django_settings_module = str (required)
+django_settings_module = str (default: DJANGO_SETTINGS_MODULE env var)
```
sobolevn: and here

mypy_django_plugin/config.py
sobolevn: yes, that's correct 👍

sobolevn (commented):
Nice, thanks!

sobolevn (approved):

sobolevn:
Thank you!

PR #2022, opened 2024-03-22
Update `_default_manager` and `_base_manager` to be `Manager`

flaeppe:
Default behaviour can be verified with the following snippet:

```python
from django.db import models
class MyModel(models.Model):
    ...
print(type(MyModel._default_manager))  # <class 'django.db.models.manager.Manager'>
print(type(MyModel._base_manager))  # <class 'django.db.models.manager.Manager'>
```

Check e.g. [Django `_base_manager` implementation](https://github.com/django/django/blob/8dbfef469582128c9d8487bf3f45d861b2ecfcb9/django/db/models/options.py#L477-L481)


sobolevn:
Thanks! I disabled these annoying annotations in #2023

sobolevn (approved):
Question: are we sure that there are no other cases where this can behave in a different way and return `BaseManager`? (I don't think so, just double checking)

flaeppe:
`_default_manager` is completely dynamic and you can change it by setting `default_manager_name = ` under `class Meta`. But the default behaviour is to return `Manager` (the implicit `objects` manager Django adds per default). I think the reason we've declared `_default_manager` in the stubs is for convenience

sobolevn:
Ok 👍

intgr (approved):
There is some precedent to this as well:

* #1843
* BaseManager rejected in DRF: https://github.com/typeddjango/djangorestframework-stubs/pull/508

It's true that users can override manager with a class that doesn't derive from `Manager`. I'm not sure how prevalent that is.

* In most cases managers do derive from `Manager`.
* The existence of `BaseManager` isn't even documented in Django.
* Currently many APIs (e.g. `get_object_or_404()`) expect `QuerySet[T] | Manager[T]` as input, which isn't satisfied by `BaseManager`.

flaeppe:
Thanks for digging up those links and info @intgr I'll go ahead and merge this

PR #2023, opened 2024-03-22
Do not annotate PRs with pyright

sobolevn:
Looking at https://github.com/typeddjango/django-stubs/pull/2022/files I've noticed that pyright annotates our code, which is not good. Disabled.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -116,6 +116,7 @@ jobs:
         uses: jakebailey/pyright-action@v2
         with:
           pylance-version: latest-release
+          annotate: false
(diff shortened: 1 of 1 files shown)


sobolevn:
No annotations 🎉

intgr:
Agreed, right now there are too many annotations that it becomes noise.

But we should restore the annotations when the pyright situation gets better.

sobolevn:
+++

We would still need to use this hack: https://github.com/python/typeshed/blob/9dcadd2d57d14dd93b84552362656da284e413e7/.github/workflows/tests.yml#L144

PR #2024, opened 2024-03-22
Use `PRI_MYPY` in `get_additional_deps` hook

sobolevn:
This value comes from here: https://github.com/python/mypy/blob/394d17b758bae6c95cbe91b84c5cccf0f4d73c28/mypy/build.py#L382

It is more correct than `PRI_MED` that we used, because it is a typing-only import.

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -3,6 +3,7 @@ import sys
 from functools import partial
 from typing import Any, Callable, Dict, List, Optional, Tuple, Type
 
+from mypy.build import PRI_MYPY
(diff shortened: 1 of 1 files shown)


intgr:
What effect does mypy import priority actually have?

sobolevn:
It mostly works to solve import cycles, in our case - it should have no effect.

intgr:
Looks like this change causes additional "Import cycle from Django settings module prevents type inference" mypy errors to show up with django-stubs 5.0.0.

See discussion: https://github.com/typeddjango/django-stubs/discussions/2097#discussioncomment-9321035

PR #2025, opened 2024-03-22
Set the calculated metaclass when creating type info in the plugin

flaeppe:
Simply calling `<TypeInfo>.calculate_metaclass_type()` doesn't modify any instance but rather returns the metaclass type.

Setting the metaclass revealed `objects` and `_default_manager` attributes not being `ClassVar` on plugin generated through models. Which is also fixed in this PR.


sobolevn (approved):
Great catch!

PR #2026, opened 2024-03-22
Add `ManyRelatedManager.through` attribute and generic type parameter

### LINKED ISSUE #1940: `through` attribute not defined within model definition
# Bug report

## What's wrong

Mypy cannot detect the `through` attribute of a `ManyToManyField` within the definition of the enclosing model. To reproduce, add the following test method to `MyModel` in `tests/typecheck/fields/test_related.yml::test_many_to_many`:

```python
class MyModel(models.Model):
    auto_through = models.ManyToManyField(Other, related_name="autos")
    # Have multiple M2Ms with implicit through
    other_again = models.ManyToManyField(Other, related_name="others_again")
    custom_through = models.ManyToManyField(Other, through=CustomThrough, related_name="customs")

    objects = MyModelManager()

    def test(self) -> None:
        reveal_type(self.custom_through.through)
```
Results in:
```
myapp/models/__init__:31: error: "Other_ManyRelatedManager" has no attribute "through"  [attr-defined]
myapp/models/__init__:31: note: Revealed type is "Any"
```

## How is that should be

Revealed type should be `"CustomThrough"`

## System information

- OS: macOS 14.2.1 
- `python` version: 3.12.1
- `django` version: 4.2.9
- `mypy` version: 1.7.1

flaeppe:
`ManyRelatedManager` has a `.through` attribute which is typed with a generic type parameter.

The plugin is also changed to populate and consider the new generic type parameter.

## Related issues

- Closes #1940


ext/django_stubs_ext/db/models/manager.py
```diff
@@ -13,11 +13,12 @@
     from typing import Protocol, TypeVar
 
     _T = TypeVar("_T")
+    _M = TypeVar("_M")
```
intgr: Let's call this `_Through` for consistency IMO

mypy_django_plugin/transformers/manytomany.py
```diff
@@ -150,7 +150,7 @@ def get_model_from_expression(
     return None
 
 
-def get_related_manager_and_model(ctx: MethodContext) -> Optional[Tuple[Instance, Instance]]:
+def get_related_manager_and_model(ctx: MethodContext) -> Optional[Tuple[Instance, Instance, Instance]]:
```
intgr, line 153: Please add docstring explaining what each of the returned values means.

mypy_django_plugin/transformers/manytomany.py
```diff
@@ -159,13 +159,15 @@ def get_related_manager_and_model(ctx: MethodContext) -> Optional[Tuple[Instance
         # Returning a 'ManyRelatedManager'. Which we want to, just like Django, build from the
         # default manager of the related model.
         many_related_manager = ctx.default_return_type
-        # Require first type argument of 'ManyRelatedManager' to be a model
+        # Require first and second type argument of 'ManyRelatedManager' to be models
         if (
-            many_related_manager.args
+            len(many_related_manager.args) >= 2
```
intgr, line 175: Can this regress users who only annotated 1 typevar out of 2? Or is this function only dealing with plugin-generated ManyRelatedManager?

django-stubs/db/models/fields/related_descriptors.pyi
```diff
@@ -125,40 +125,41 @@ class ManyToManyDescriptor(ReverseManyToOneDescriptor, Generic[_To, _Through]):
...
     @overload  # type: ignore[override]
     def __get__(self, instance: None, cls: Any = ...) -> Self: ...
     @overload
-    def __get__(self, instance: Model, cls: Any = ...) -> ManyRelatedManager[_To]: ...
+    def __get__(self, instance: Model, cls: Any = ...) -> ManyRelatedManager[_To, _Through]: ...
 
 # Fake class, Django defines 'ManyRelatedManager' inside a function body
 @type_check_only
-class ManyRelatedManager(Manager[_M], Generic[_M]):
+class ManyRelatedManager(Manager[_To], Generic[_To, _Through]):
```
intgr, line 136: Just a thought: `ManyRelatedManager` gained a new type argument, but users may have usages that only contain 1 typevar.

Mypy 1.9 gained support for PEP 696 `TypeVar(default=...)` argument (using `typing_extensions.TypeVar`). We could make `_Through` default to `Model`, which would be a bit safer than mypy's implicit fallback `Any`.

intgr (commented):
Some initial thoughts. Haven't looked at the plugin code in depth yet.

ext/django_stubs_ext/db/models/manager.py
flaeppe: Sure, I'll update

mypy_django_plugin/transformers/manytomany.py
flaeppe: No, it's correct that it's only plugin related but this code is only reached via one of the overloads of `ManyToManyField.__get__`, namely the one when there's a model instance and a `ManyRelatedManager` is returned.

Here's the condition:

https://github.com/typeddjango/django-stubs/blob/0a006f2dda989679b3d09fc876677bd90e1e5cb4/mypy_django_plugin/main.py#L193-L197

django-stubs/db/models/fields/related_descriptors.pyi
flaeppe: Sure, we can set a `default=`

mypy_django_plugin/transformers/manytomany.py
```diff
@@ -179,26 +192,48 @@ def refine_many_to_many_related_manager(ctx: MethodContext) -> MypyType:
...
-            bases=[many_related_manager, default_manager_node.type],
+            bases=[
+                many_related_manager.copy_modified(
+                    args=[
+                        # Keep the same '_To' as the (parent) `ManyRelatedManager` instance
+                        many_related_manager.args[0],
+                        # But reset the '_Through' `TypeVar` declared for `ManyRelatedManager`
+                        _through_type_var.copy_modified(),
+                    ]
+                ),
```
intgr: This is trying to do too much in one statement. I'd suggest creating a temporary variable for `many_related_manager.copy_modified()`.

mypy_django_plugin/transformers/manytomany.py
```diff
@@ -150,7 +150,18 @@ def get_model_from_expression(
     return None
 
 
-def get_related_manager_and_model(ctx: MethodContext) -> Optional[Tuple[Instance, Instance]]:
+def get_related_manager_and_model(ctx: MethodContext) -> Optional[Tuple[Instance, Instance, Instance]]:
+    """
+    Returns a 3-tuple consisting of:
+      1. A `ManyRelatedManager` instance
+      2. The first type parameter instance of 1. when it's a model
+      3. The second type parameter instance of 1. when it's a model
```
intgr:
```suggestion
      2. The first type parameter (_To) instance of 1. when it's a model
      3. The second type parameter (_Through) instance of 1. when it's a model
```

tests/typecheck/fields/test_related.yml
```diff
@@ -1114,6 +1115,13 @@
 
         reveal_type(Other.autos.through.objects)
         reveal_type(Other.autos.through._default_manager)
+
+        reveal_type(MyModel().auto_through)
+        reveal_type(MyModel().auto_through.through)
+        reveal_type(MyModel().other_again)
+        reveal_type(MyModel().other_again.through)
+        reveal_type(MyModel().custom_through)
+        reveal_type(MyModel().custom_through.through)
```
intgr: Maybe also check that the expected fields exist on through model?

```suggestion
        reveal_type(MyModel().auto_through.through)
        reveal_type(MyModel().auto_through.through.mymodel_id)
        reveal_type(MyModel().other_again)
        reveal_type(MyModel().other_again.through)
        reveal_type(MyModel().other_again.through)
        reveal_type(MyModel().other_again.through.mymodel_id)
        reveal_type(MyModel().custom_through)
        reveal_type(MyModel().custom_through.through)
        reveal_type(MyModel().custom_through.my_model_id)
```

intgr (changes requested):
A few minor tweaks; other than that, looks good to me.

flaeppe:
I've rebased and updated with your suggestions. Let me know if there's anything more I should fix

intgr (approved):
Looks good to me, thanks!

PR #2027, opened 2024-03-23
Determine the type of queryset methods on unions

### LINKED ISSUE #1532: Cannot resolve shared manager/queryset method for union
# Bug report

## What's wrong

I have two models with matching managers and querysets, and sometimes use them as a union.

```python3
model: type[Foo] | type[Bar]

result = model.objects.shared_method()
```

```
Unable to resolve return type of queryset/manager method "shared_method" 
```

## How is that should be

`result` is inferred as the union of return types of `Foo.objects.shared_method()` and `Bar.objects.shared_method()`.

## System information

- OS: Docker `Linux b5ca85c5b236 5.15.49-linuxkit #1 SMP PREEMPT Tue Sep 13 07:51:32 UTC 2022 aarch64 GNU/Linux`
- `python` version: 3.11
- `django` version: 4.2
- `mypy` version: 1.3

delfick:
Closes: #1532

Prior to this if a type was a union of multiple django model managers then it wouldn't be able to determine the type of the queryset methods on them.

This change will mean that in the case of a union, the type of the queryset method will be a union of the types of all the queryset methods on each item in the union

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -19,7 +19,18 @@ from mypy.nodes import (
 from mypy.plugin import AttributeContext, ClassDefContext, DynamicClassDefContext
 from mypy.semanal import SemanticAnalyzer
 from mypy.semanal_shared import has_placeholder
-from mypy.types import AnyType, CallableType, FunctionLike, Instance, Overloaded, ProperType, TypeOfAny, TypeVarType
+from mypy.types import (
+    AnyType,
+    CallableType,
+    FunctionLike,
+    Instance,
+    Overloaded,
+    ProperType,
+    TypeOfAny,
+    TypeVarType,
+    UnionType,
+    get_proper_type,
+)
 from mypy.types import Type as MypyType
 from mypy.typevars import fill_typevars
 
@@ -274,6 +285,13 @@ def resolve_manager_method(ctx: AttributeContext) -> MypyType:
 
     if isinstance(ctx.type, Instance):
         return resolve_manager_method_from_instance(instance=ctx.type, method_name=method_name, ctx=ctx)
+    elif isinstance(ctx.type, UnionType) and all(isinstance(instance, Instance) for instance in ctx.type.items):
+        resolved = tuple(
+            resolve_manager_method_from_instance(instance=instance, method_name=method_name, ctx=ctx)
+            for instance in ctx.type.items
+            if isinstance(instance, Instance)
+        )
+        return get_proper_type(UnionType(resolved))
     else:
         ctx.api.fail(f'Unable to resolve return type of queryset/manager method "{method_name}"', ctx.context)
         return AnyType(TypeOfAny.from_error)
--- a/tests/typecheck/managers/querysets/test_as_manager.yml
+++ b/tests/typecheck/managers/querysets/test_as_manager.yml
@@ -140,6 +140,44 @@
                 class MyModel(models.Model):
                     objects = ModelQuerySet.as_manager()
 
+-   case: includes_custom_queryset_methods_on_unions
+    main: |
+        from myapp.models import MyModel1, MyModel2
+        import typing
+        kls: typing.Type[typing.Union[MyModel1, MyModel2]] = MyModel1
+        reveal_type(kls.objects.custom_queryset_method())  # N: Revealed type is "Union[myapp.models.ModelQuerySet1, myapp.models.ModelQuerySet2]"
+        reveal_type(kls.objects.all().custom_queryset_method())  # N: Revealed type is "Union[myapp.models.ModelQuerySet1, myapp.models.ModelQuerySet2]"
+        reveal_type(kls.objects.returns_thing())  # N: Revealed type is "Union[builtins.int, builtins.str]"
+        reveal_type(kls.objects.get())  # N: Revealed type is "Union[myapp.models.MyModel1, myapp.models.MyModel2]"
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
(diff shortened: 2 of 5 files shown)


mypy_django_plugin/transformers/managers.py
```diff
@@ -274,6 +285,19 @@ def resolve_manager_method(ctx: AttributeContext) -> MypyType:
...
+                f'Unable to resolve return type of queryset/manager method "{method_name}" on a union involving Any',
+                ctx.context,
+            )
+            return AnyType(TypeOfAny.from_error)
+        else:
+            resolved = [
+                resolve_manager_method_from_instance(instance=instance, method_name=method_name, ctx=ctx)
+                for instance in ctx.type.items
+            ]
+            return get_proper_type(UnionType(tuple(resolved)))
```
```suggestion
            resolved = tuple(
                resolve_manager_method_from_instance(instance=instance, method_name=method_name, ctx=ctx)
                for instance in ctx.type.items
            )
            return get_proper_type(UnionType(resolved))
```

We can build a tuple directly instead of casting

flaeppe (commented):
Just 1 tiny change, otherwise I thinks it looks good

mypy_django_plugin/transformers/managers.py
delfick: ah yeah, reasonable, also I fixed the broken tests

delfick: fixed

flaeppe (approved):
👍

mypy_django_plugin/transformers/managers.py
```diff
@@ -274,6 +285,26 @@ def resolve_manager_method(ctx: AttributeContext) -> MypyType:
...
+                ctx.context,
+            )
+            return AnyType(TypeOfAny.from_error)
+        else:
+            resolved = tuple(
+                resolve_manager_method_from_instance(instance=instance, method_name=method_name, ctx=ctx)
+                for instance in ctx.type.items
+                if isinstance(instance, Instance)
+            )
+            return get_proper_type(UnionType(resolved))
```
flaeppe: Since this now only requires `Instance`(?) I think we can switch to `if all(...)` and drop the outermost `else`

delfick: I kept the difference because when there is an Any, the error message is more precise. I don't have a strong opinion though, I can collapse that

tests/typecheck/managers/querysets/test_as_manager.yml
```diff
@@ -140,6 +140,48 @@
                 class MyModel(models.Model):
                     objects = ModelQuerySet.as_manager()
 
+-   case: includes_custom_queryset_methods_on_unions
+    main: |
+        from myapp.models import MyModel1, MyModel2
+        import sys
+        if sys.version_info <= (3, 9):
```
delfick: you can't do `type[MyModel]` in python 3.8 and you can't do `MyModel1 | MyModel2` in python3.9

mypy_django_plugin/transformers/managers.py
```diff
@@ -274,6 +285,20 @@ def resolve_manager_method(ctx: AttributeContext) -> MypyType:
...
+            ctx.api.fail(
+                f'Unable to resolve return type of queryset/manager method "{method_name}"',
+                ctx.context,
+            )
+            return AnyType(TypeOfAny.from_error)
+
+        resolved = tuple(
+            resolve_manager_method_from_instance(instance=instance, method_name=method_name, ctx=ctx)
+            for instance in ctx.type.items
+            if isinstance(instance, Instance)
```
delfick, line 292: mypy doesn't know about the all() check above and we have to reassure it these are indeed Instance objects

mypy_django_plugin/transformers/managers.py
delfick: fixed, and I got all tests to pass. For real this time!

mypy_django_plugin/transformers/managers.py
flaeppe: What if we apply this diff?

```diff
-    elif isinstance(ctx.type, UnionType):
-        if not all(isinstance(instance, Instance) for instance in ctx.type.items):
-            ctx.api.fail(
-                f'Unable to resolve return type of queryset/manager method "{method_name}"',
-                ctx.context,
-            )
-            return AnyType(TypeOfAny.from_error)
-
+    elif isinstance(ctx.type, UnionType) and all(isinstance(instance, Instance) for instance in ctx.type.items):
         resolved = tuple(
             resolve_manager_method_from_instance(instance=instance, method_name=method_name, ctx=ctx)
             for instance in ctx.type.items
-            if isinstance(instance, Instance)
         )
         return get_proper_type(UnionType(resolved))
-    else:
-        ctx.api.fail(f'Unable to resolve return type of queryset/manager method "{method_name}"', ctx.context)
-        return AnyType(TypeOfAny.from_error)
+    ctx.api.fail(f'Unable to resolve return type of queryset/manager method "{method_name}"', ctx.context)
+    return AnyType(TypeOfAny.from_error)
```

tests/typecheck/managers/querysets/test_as_manager.yml
intgr: Don't bother with the conditional. Use this style in tests:

```python
from typing import Type, Union
kls: Type[Union[MyModel1, MyModel2]] = MyModel1
```

delfick: ok, done

mypy_django_plugin/transformers/managers.py
delfick: it still needs that extra isinstance in that case, but is definitely still an improvement :)

flaeppe (approved):

delfick:
thanks!

PR #2029, opened 2024-03-25
Add first stub for get_model_admin

nebiyuelias1:
# I have made things!

I'm making this PR to fix some stub inconsistencies for django version 5, specifically we're missing a stub for get_model_admin method. I'm hoping to participate for GSOC 2024 in clearing up the allowed lists.
## Related issues
- Refs #2012


sobolevn (approved):
Thanks!

https://github.com/django/django/blob/ae10146793dca0e0594c7acdee20ca3810983f39/django/contrib/admin/sites.py#L96

nebiyuelias1:
@sobolevn Thank you! Can you elaborate a little further about the link you provided please?

sobolevn:
This is a note to self: that the type is indeed `type[Model]` and not `Model`

nebiyuelias1:
Got it! Thanks for the clarification.

Viicos:
Could this be made generic?

```python
def get_model_admin(self, model: type[_ModelT]) -> ModelAdmin[_ModelT]: ...
```

flaeppe:
> Could this be made generic?
> 
> ```python
> def get_model_admin(self, model: type[_ModelT]) -> ModelAdmin[_ModelT]: ...
> ```

Yes, I think so. It looks more correct

Viicos:
I'll open a PR then :)

PR #2034, opened 2024-03-29
[5.0] Add missing stubs for geos

nebiyuelias1:
Added missing stubs for the `django.contrib.gis.geos` package. 

## Related issues
Refs #1493

I also wanted to add a missing stub for `django.contrib.gis.geos.prototypes.io.DEFAULT_TRIM_VALUE` for but I was getting a rather cryptic error from mypy. This is the error I was getting: `TypeError: union[int, str, bool, float] object expected; got django.utils.functional.SimpleLazyObject`. But I tried to add a stub like the following:
```
DEFAULT_TRIM_VALUE: Any
```
I even tried to use SimpleLazyObject as type but the error didn't change.


django-stubs/contrib/gis/geos/prototypes/predicates.pyi
```diff
@@ -22,6 +22,7 @@ geos_crosses: Any
 geos_disjoint: Any
 geos_equals: Any
 geos_equalsexact: Any
+geos_equalsidentical: Any
```
sobolevn: `BinaryPredicate`? https://github.com/django/django/blob/5f3cdf219de71021cc5b965695dbe0c881c0a4f1/django/contrib/gis/geos/prototypes/predicates.py#L42

You can type other values here as well.

sobolevn (commented):
`DEFAULT_TRIM_VALUE` is a perfect example of a thing we would add to `stubtest/allowlist.txt`. We know that it will behave as `bool`, but it is just wrapped into `SimpleLazyObject`. https://github.com/django/django/blob/5f3cdf219de71021cc5b965695dbe0c881c0a4f1/django/contrib/gis/geos/prototypes/io.py#L190

Thanks!

nebiyuelias1:
> `DEFAULT_TRIM_VALUE` is a perfect example of a thing we would add to `stubtest/allowlist.txt`. We know that it will behave as `bool`, but it is just wrapped into `SimpleLazyObject`. https://github.com/django/django/blob/5f3cdf219de71021cc5b965695dbe0c881c0a4f1/django/contrib/gis/geos/prototypes/io.py#L190
> 
> Thanks!

Does the type checker think `SimpleLazyObject` is a `union[int, str, bool, float]` object? Also, should I add `DEFAULT_TRIM_VALUE ` to the to the `stubtest/allowlist.txt` file?

sobolevn:
No, you just type `DEFAULT_TRIM_VALUE` as `bool` and add `DEFAULT_TRIM_VALUE` to `allowlist.txt`

nebiyuelias1:
> No, you just type `DEFAULT_TRIM_VALUE` as `bool` and add `DEFAULT_TRIM_VALUE` to `allowlist.txt`

After doing this, actually `stubtest.sh` is producing errors. It doesn't seem to be happy about the inclusion of the type annotation to `DEFAULT_TRIM_VALUE` even if it's explicitly added to `allowlist.txt`. For reference, the error is still the same as before: `TypeError: union[int, str, bool, float] object expected; got django.utils.functional.SimpleLazyObject`

sobolevn (approved):
Ok, we can do that later. Thanks!

PR #2035, opened 2024-03-29
Relax type for `fields` argument of `Model.refresh_from_db()`

mthuurne:
According to the [docstring](https://github.com/django/django/blob/5f3cdf219de71021cc5b965695dbe0c881c0a4f1/django/db/models/base.py#L683), `Model.refresh_from_db()` accepts any iterable or `None` for its `fields` argument. Currently the stubs require `Sequence[str] | None `, which excludes `set[str]`.

There is [a bug](https://code.djangoproject.com/ticket/34925) that is fixed in Django's `main` branch but not in any release as of the time writing, where `remove()` can be called on the `fields` argument. Since this is a bug, I think it should not affect the annotation. In any case, `remove()` is only part of `MutableSequence`, not `Sequence`, so the old annotation matched neither the documentation nor the implementation.


sobolevn (approved):
Thanks!

PR #2038, opened 2024-03-31
Make `AdminSite.get_model_admin` generic

Viicos:
Follow up on https://github.com/typeddjango/django-stubs/pull/2029


sobolevn (approved):
Seems correct to me! Thank you!

PR #2044, opened 2024-04-05
Fallback to checking kwargs of function call when looking for arguments of a call

### LINKED ISSUE #2043: Custom fields with overwritten passthrough constructors don't change types based on attributes
# Bug report

## What's wrong

When creating a custom model field with a constructor that uses `*args, **kwargs` to pass arguments to the parent constructor, attributes that change the typing of the field are not considered.

Below is a simple example where `my_custom_field` has `null=True` kwarg so the type of the field on the model instance should be `Union[builtins.int, None]` but is incorrectly `builtins.int`.

```
-   case: test_custom_model_fields
    main: |
        from myapp.models import User
        user = User()
        reveal_type(user.id)  # N: Revealed type is "builtins.int"
        reveal_type(user.my_custom_field)  # N: Revealed type is "Union[builtins.int, None]"
    monkeypatch: true
    installed_apps:
        - myapp
    files:
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                from django.db import models

                class MyIntegerField(fields.IntegerField[_ST, _GT]):

                class User(models.Model):
                    my_custom_field = MyIntegerField(null=True)
```

## How is that should be

`reveal_type(user.my_custom_field)  # N: Revealed type is "Union[builtins.int, None]"`

## System information

- OS:
- `python` version: 3.11
- `django` version: 4.2.5
- `mypy` version: 1.7.1

md384:
# I have made things!

## Related issues

Closes #2043

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -173,6 +173,19 @@ def get_class_fullname(klass: type) -> str:
 def get_call_argument_by_name(ctx: Union[FunctionContext, MethodContext], name: str) -> Optional[Expression]:
     """
     Return the expression for the specific argument.
+
+    It first checks the function definition for the presence of the argument in its parameters, returning the associated
+    expression for the argument if found. This handles the case of getting the expression of `null` in
+
+    >>> def first_case(null: bool=False) -> Any: ...
+    >>> first_case(null=True)
+
+    If not found, it is searched for in the names of actual arguments in the call expression, returning the expression
+    for the argument if found. This handles the case of getting the expression of `null` in
+
+    >>> def second_case(*args, **kwargs) -> Any: ...
+    >>> second_case(null=True)
+
     This helper should only be used with non-star arguments.
     """
     # first check for named arg on function definition
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -173,16 +173,22 @@ def get_class_fullname(klass: type) -> str:
 def get_call_argument_by_name(ctx: Union[FunctionContext, MethodContext], name: str) -> Optional[Expression]:
     """
     Return the expression for the specific argument.
-    This helper should only be used with non-star arguments.
     """
-    if name not in ctx.callee_arg_names:
-        return None
-    idx = ctx.callee_arg_names.index(name)
-    args = ctx.args[idx]
-    if len(args) != 1:
-        # Either an error or no value passed.
-        return None
-    return args[0]
+    # first check for named arg on function definition
+    if name in ctx.callee_arg_names:
+        idx = ctx.callee_arg_names.index(name)
+        args = ctx.args[idx]
+        if len(args) != 1:
+            # Either an error or no value passed.
+            return None
+        return args[0]
+
+    # check for named arg in function call keyword args
+    if name in ctx.arg_names[1]:
+        idx = ctx.arg_names[1].index(name)
+        return ctx.args[1][idx]
+
+    return None
 
 
 def get_call_argument_type_by_name(ctx: Union[FunctionContext, MethodContext], name: str) -> Optional[MypyType]:
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -173,6 +173,7 @@ def get_class_fullname(klass: type) -> str:
 def get_call_argument_by_name(ctx: Union[FunctionContext, MethodContext], name: str) -> Optional[Expression]:
     """
     Return the expression for the specific argument.
+    This helper should only be used with non-star arguments.
     """
     # first check for named arg on function definition
     if name in ctx.callee_arg_names:
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -185,9 +185,10 @@ def get_call_argument_by_name(ctx: Union[FunctionContext, MethodContext], name: 
         return args[0]
 
     # check for named arg in function call keyword args
-    if name in ctx.arg_names[1]:
-        idx = ctx.arg_names[1].index(name)
-        return ctx.args[1][idx]
+    for arg_group_idx, arg_group in enumerate(ctx.arg_names):
+        if name in arg_group:
+            arg_name_idx = arg_group.index(name)
+            return ctx.args[arg_group_idx][arg_name_idx]
 
     return None
 
--- a/tests/typecheck/managers/querysets/test_custom_queryset.yml
+++ b/tests/typecheck/managers/querysets/test_custom_queryset.yml
@@ -1,0 +1,42 @@
+- case: test_custom_queryset_with_passthrough_values_list
+  main: |
+    from typing import Any, TypeVar, Self
+    from django.db.models.base import Model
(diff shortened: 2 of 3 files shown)


mypy_django_plugin/lib/helpers.py
```diff
@@ -173,16 +173,22 @@ def get_class_fullname(klass: type) -> str:
 def get_call_argument_by_name(ctx: Union[FunctionContext, MethodContext], name: str) -> Optional[Expression]:
     """
     Return the expression for the specific argument.
-    This helper should only be used with non-star arguments.
```
md384, line 176: This suggests we shouldn't do this - not sure if that is still applicable (or means something else)

md384: I guess maybe it means *args rather than **kwargs, so this comment should go back in

sobolevn (commented):
We have several usages of this function: https://github.com/search?q=repo%3Atypeddjango%2Fdjango-stubs%20get_call_argument_by_name&type=code

Can you please also test the `QuerySet` one?

mypy_django_plugin/lib/helpers.py
md384: I added the comment back

tests/typecheck/managers/querysets/test_custom_queryset.yml
```diff
@@ -0,0 +1,42 @@
...
+    from typing import Any, TypeVar, Self
+    from django.db.models.base import Model
+    from django.db.models.query import QuerySet
+    from myapp.models import MyUser
+
+    _Model = TypeVar("_Model", bound=Model, covariant=True)
+
+    class CustomQuerySet(QuerySet[_Model]):
+        def values_list(self, *args: Any, **kwargs: Any) -> QuerySet[_Model]:
+            return super().values_list(*args, **kwargs)
```
sobolevn, line 12: stupid question: what will happen if we do `kwargs.pop('flat'); super().values_list(*args, flat=True, **kwargs)`?

md384: The `flat=True` doesn't change the return type of the `get` call in the tests. This seems like a separate bug and is beyond my knowledge of mypy or mypy plugins -there doesn't seem to be anything usable in the context.

mypy_django_plugin/lib/helpers.py
```diff
@@ -175,14 +175,21 @@ def get_call_argument_by_name(ctx: Union[FunctionContext, MethodContext], name:
...
+    if name in ctx.callee_arg_names:
+        idx = ctx.callee_arg_names.index(name)
+        args = ctx.args[idx]
+        if len(args) != 1:
+            # Either an error or no value passed.
+            return None
+        return args[0]
+
+    # check for named arg in function call keyword args
+    if name in ctx.arg_names[1]:
```
flaeppe: I don't think usage of `ctx.arg_names[1]` will work like this here.

For example if I instead declare my field like:

```python
class SecondField(fields.IntegerField[_ST, _GT]):
    def __init__(self, *args: Any, blank: bool = True, **kwargs: Any) -> None:
        super().__init__(*args, blank=blank, **kwargs)
```

And call it with: `SecondField(blank=True, null=True)`

`ctx.arg_names` could look like:

```python
[[], ['blank'], ['null']]
```

Which results in a mismatch, resulting in the issue we're trying to solve.

With the above in mind. If I instead then do:

```python
class ThirdField(fields.IntegerField[_ST, _GT]):
    def __init__(self) -> None:
        super().__init__()
```

And call it like: `ThirdField()` I get an `IndexError` crash on `ctx.arg_names[1]`

flaeppe (commented):
I'm suspecting there's more cases to cover for here.

mypy_django_plugin/lib/helpers.py
md384: I think we can do something along the lines of 

```py
    # check for named arg in function call keyword args
    for idx, arg_group in enumerate(ctx.arg_names):
        if name in arg_group:
            sub_idx = arg_group.index(name)
            return ctx.args[idx][sub_idx]
```
to handle these cases.

flaeppe:
Did you see my comment: https://github.com/typeddjango/django-stubs/issues/2043#issuecomment-2041361782?

md384:
> Did you see my comment: [#2043 (comment)](https://github.com/typeddjango/django-stubs/issues/2043#issuecomment-2041361782)?

Yes, but that means the developer has to know about these "magic" kwargs in a number of functions. 

IMO, the plugin should be able to handle the `*args, **kwargs` case since it is a common pattern and the presence/absence of some kwargs unexpectedly (at least for the typical django-stubs user) changes the return type.

flaeppe:
I disagree with that the plugin should try to resolve `*args: Any` and `**kwargs: Any` here. In general the plugin should help express things that can't be statically declared, but this isn't such a case. There's a way to statically declare this and, in my eyes, introducing plugin code here just increases the risk of interference.

I can agree with that it might be tedious to declare any or all overridden arguments, especially since there are quite a few..

But we want the plugin to do as little as possible and fallback on mypy to do the rest.

At least that's my opinion. @intgr or @sobolevn might have different thoughts here.

mypy_django_plugin/lib/helpers.py
```diff
@@ -175,14 +175,22 @@ def get_call_argument_by_name(ctx: Union[FunctionContext, MethodContext], name:
...
+    # first check for named arg on function definition
+    if name in ctx.callee_arg_names:
+        idx = ctx.callee_arg_names.index(name)
+        args = ctx.args[idx]
+        if len(args) != 1:
+            # Either an error or no value passed.
+            return None
+        return args[0]
+
+    # check for named arg in function call keyword args
```
intgr, line 200: Would be helpful to have more detailed docstring or code comments explaining what's going on here, why there are 2 ways to look up an argument by name and when each is triggered.

If I didn't already know the context explained in this the PR description, I probably would be confused reading this function.

intgr (commented):
I'm torn on this.

From a typing perspective, `*args: Any, **kwargs: Any` is problematic, since it suppresses all typechecking of the call. You can write `MyIntegerField(1, 2, 3, null=Ellipsis, no_such_arg=123)` and mypy won't bat an eyelid. From this perspective it seems consistent to also reject inspecting `null=` if the function definition is vague.

On the other hand, this PR does seem like an improvement in making django-stubs less finnicky. The implementation makes sense and is self-contained in the helper function; other callers of this function may benefit from the changes as well.

If we can't think of any downsides of this approach, seems unfair to block it. So a weak vote for merging from me.

sshishov:
Hi guys, just want to highlight as the ordinary user of the plugin and Django itself. The point of this change to help create "custom" fields based on the standard fields with some overrides.

- If we are creating completely custom field (derived from `Field`) then probably you are right that we should express everything ourselves.
- If we create the field from another field and just override some "kwargs" value, like defaults or adding validators or anything else then ask the user to re-define all the kwargs to make mypy work properly - is overkill.

Why this issue is noticable? A lot of people creating custom IntegerField, DecimalField where `null=False` by default and then in the model they create `CustomField(null=True)` and mypy do not validate properly this `null=True` part anymore. And 100% for everyone this is a surprise as we never override how `null` keyword should work.

Thank you for you work, folks!

Issue #336, opened 2020-03-02
Resources for annotating third party Django libraries


coiax:
![image](https://user-images.githubusercontent.com/609465/75683882-cb5a6880-5c8f-11ea-994f-1a6211f6f236.png)

In our project, we're using [django-enumfield](https://github.com/5monkeys/django-enumfield), and their custom field.

Currently, the custom field is typed as an `Any`. The get type should be the specific Enum that you associate with the field, while the set type should be `Union[int, MyEnum]`.

Some resources on how to write stubs for third party libraries would be really helpful!

sobolevn:
You can try to use class-level attribute annotation with `Field`: https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/fields/__init__.pyi#L45

It is a generic type with two parameters: 
- `_ST` for types that can be used when settings a value
- `_GT` as a type that will be returned when getting a value

Something like:

```python
my_attr: Field[SupportsInt, str] = MyCustomIntField()
```

Can you please report if this works for you? If so, I will add this example to FAQ.

coiax:
I tried to make a test setup, with the snippet you suggested, my results are in this gist.

https://gist.github.com/coiax/e4a26eb224c97b33cfb303f1df2f9fe0

Typing the field with `Field[Union[int, Colour], Colour]` only provides typing coverage when using an instance of the object, but looks like there's trouble with other ways of engaging with the object, unless I set up my test environment wrong!

sobolevn:
Thanks! Can you please also attach `mypy` output to your gist?

coiax:
Done.

flaeppe:
With all of its quirks and limitations. Just as a reference, this is how I annotated a package with a model field: https://github.com/flaeppe/django-choicefield/

PR #2048, opened 2024-04-08
Use field generic types for descriptors

md384:
# I have made things!

## Related issues

Refs #336

This PR tries to address typing for descriptors on fields that use generic types, e.g.

On current master, you get the following
```py
from django.db import models

class User(models.Model):
    my_custom_field1 = fields.Field[int, int]()
    
  user = User()
  reveal_type(user.my_custom_field1)  # N: Revealed type is "Any"

  user = User(my_custom_field1="hey")  # No error on the set type being str instead of int
  reveal_type(user.my_custom_field1)  # N: Revealed type is "Any"
```
  
with the changes in this PR
```py
from django.db import models

class User(models.Model):
    my_custom_field1 = fields.Field[int, int]()
    
  user = User()
  reveal_type(user.my_custom_field1)  # N: Revealed type is "builtins.int"

  user = User(my_custom_field1="hey")  # error: Incompatible type for "my_custom_field1" of "User" (got "str", expected "int")  [misc]
  reveal_type(user.my_custom_field1)  # N: Revealed type is "builtins.int"
```

### DIFF
--- a/mypy_django_plugin/transformers/fields.py
+++ b/mypy_django_plugin/transformers/fields.py
@@ -7,7 +7,7 @@ from django.db.models.fields.reverse_related import ForeignObjectRel
 from mypy.maptype import map_instance_to_supertype
 from mypy.nodes import AssignmentStmt, NameExpr, TypeInfo
 from mypy.plugin import FunctionContext
-from mypy.types import AnyType, Instance, ProperType, TypeOfAny, UninhabitedType, UnionType
+from mypy.types import AnyType, Instance, NoneType, ProperType, TypeOfAny, UninhabitedType, UnionType
 from mypy.types import Type as MypyType
 
 from mypy_django_plugin.django.context import DjangoContext
@@ -160,13 +160,13 @@ def set_descriptor_types_for_field(
 
     # bail if either mapped_set_type or mapped_get_type have type Never
     if not (isinstance(mapped_set_type, UninhabitedType) or isinstance(mapped_get_type, UninhabitedType)):
-        # only replace set_type and get_type with mapped types if their original value is Any
-        set_type = helpers.convert_any_to_type(
-            set_type, helpers.make_optional(mapped_set_type) if is_set_nullable or is_nullable else mapped_set_type
-        )
-        get_type = helpers.convert_any_to_type(
-            get_type, helpers.make_optional(mapped_get_type) if is_get_nullable or is_nullable else mapped_get_type
-        )
+        # always replace set_type and get_type with (non-Any) mapped types
+        set_type = helpers.convert_any_to_type(mapped_set_type, set_type)
+        get_type = helpers.convert_any_to_type(mapped_get_type, get_type)
+
+        # the get_type must be optional if the field is nullable
+        if (is_get_nullable or is_nullable) and not (isinstance(get_type, NoneType) or helpers.is_optional(get_type)):
+            ctx.api.fail("Field is nullable but generic get type parameter is not optional", ctx.context)
 
     return helpers.reparametrize_instance(default_return_type, [set_type, get_type])
 
--- a/mypy_django_plugin/transformers/fields.py
+++ b/mypy_django_plugin/transformers/fields.py
@@ -13,7 +13,6 @@ from mypy.types import Type as MypyType
 from mypy_django_plugin.django.context import DjangoContext
 from mypy_django_plugin.exceptions import UnregisteredModelError
 from mypy_django_plugin.lib import fullnames, helpers
-from mypy_django_plugin.lib.fullnames import FIELD_FULLNAME
 from mypy_django_plugin.lib.helpers import parse_bool
 from mypy_django_plugin.transformers import manytomany
 
@@ -154,7 +153,7 @@ def set_descriptor_types_for_field(
     )
 
     # reconcile set and get types with the base field class
-    base_field_type = next(base for base in default_return_type.type.mro if base.fullname == FIELD_FULLNAME)
+    base_field_type = next(base for base in default_return_type.type.mro if base.fullname == fullnames.FIELD_FULLNAME)
     mapped_instance = map_instance_to_supertype(default_return_type, base_field_type)
     mapped_set_type, mapped_get_type = mapped_instance.args
 
@@ -166,7 +165,10 @@ def set_descriptor_types_for_field(
 
         # the get_type must be optional if the field is nullable
         if (is_get_nullable or is_nullable) and not (isinstance(get_type, NoneType) or helpers.is_optional(get_type)):
-            ctx.api.fail("Field is nullable but generic get type parameter is not optional", ctx.context)
+            ctx.api.fail(
+                f"{default_return_type.type.name} is nullable but its generic get type parameter is not optional",
+                ctx.context,
+            )
 
     return helpers.reparametrize_instance(default_return_type, [set_type, get_type])
 
--- a/mypy_django_plugin/transformers/fields.py
+++ b/mypy_django_plugin/transformers/fields.py
@@ -4,14 +4,16 @@ from django.core.exceptions import FieldDoesNotExist
 from django.db.models.fields import AutoField, Field
 from django.db.models.fields.related import RelatedField
 from django.db.models.fields.reverse_related import ForeignObjectRel
+from mypy.maptype import map_instance_to_supertype
 from mypy.nodes import AssignmentStmt, NameExpr, TypeInfo
 from mypy.plugin import FunctionContext
-from mypy.types import AnyType, Instance, ProperType, TypeOfAny, UnionType
+from mypy.types import AnyType, Instance, ProperType, TypeOfAny, UninhabitedType, UnionType
 from mypy.types import Type as MypyType
 
 from mypy_django_plugin.django.context import DjangoContext
 from mypy_django_plugin.exceptions import UnregisteredModelError
 from mypy_django_plugin.lib import fullnames, helpers
+from mypy_django_plugin.lib.fullnames import FIELD_FULLNAME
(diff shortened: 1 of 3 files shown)


tests/typecheck/fields/test_custom_fields.yml
```diff
@@ -0,0 +1,62 @@
...
+        from myapp.models import User, CustomFieldValue
+        user = User()
+        reveal_type(user.id)  # N: Revealed type is "builtins.int"
+        reveal_type(user.my_custom_field1)  # N: Revealed type is "myapp.models.CustomFieldValue"
+        reveal_type(user.my_custom_field2)  # N: Revealed type is "myapp.models.CustomFieldValue"
+        reveal_type(user.my_custom_field3)  # N: Revealed type is "builtins.bool"
+        reveal_type(user.my_custom_field4)  # N: Revealed type is "myapp.models.CustomFieldValue"
+        reveal_type(user.my_custom_field5)  # N: Revealed type is "Union[myapp.models.CustomFieldValue, None]"
+        # user.my_custom_field6 is incorrectly typed as non-optional
+        # reveal_type(user.my_custom_field6) ## N: Revealed type is "Union[myapp.models.CustomFieldValue, None]"
```
md384: This is incorrect on master too and does not pick up on null=True. When debugging, the instance seemed correct and the args were optional so I think this issue is coming from another piece of code somewhere.

mypy_django_plugin/transformers/fields.py
```diff
@@ -150,6 +152,22 @@ def set_descriptor_types_for_field(
...
     )
+
+    # reconcile set and get types with the base field class
+    base_field_type = next(base for base in default_return_type.type.mro if base.fullname == FIELD_FULLNAME)
+    mapped_instance = map_instance_to_supertype(default_return_type, base_field_type)
+    mapped_set_type, mapped_get_type = mapped_instance.args
+
+    # bail if either mapped_set_type or mapped_get_type have type Never
+    if not (isinstance(mapped_set_type, UninhabitedType) or isinstance(mapped_get_type, UninhabitedType)):
+        # only replace set_type and get_type with mapped types if their original value is Any
```
md384: Perhaps we just want to always replace if the mapped types are not Any.

This is actually a use-case I need where I use a `SmallInteger` field to represent a custom enum type so I'll have to set `_pyi_private_set_type` and `_pyi_private_get_type` to `Any` in my code.

tests/typecheck/fields/test_custom_fields.yml
flaeppe: Slightly off topic but still related:

I think #1900 is involved and/or need to be considered regarding `null=True` and optional.

IMO we should finish explore that approach primarily, before moving on to a plugin approach.

tests/typecheck/fields/test_custom_fields.yml
```diff
@@ -0,0 +1,62 @@
...
+        reveal_type(user.my_custom_field7)  # N: Revealed type is "Union[builtins.bool, None]"
+        reveal_type(user.my_custom_field8)  # N: Revealed type is "Union[myapp.models.CustomFieldValue, None]"
+        reveal_type(user.my_custom_field9)  # N: Revealed type is "myapp.models.CustomFieldValue"
+        reveal_type(user.my_custom_field10)  # N: Revealed type is "Union[myapp.models.CustomFieldValue, None]"
```
flaeppe: I would say the behaviour for fields: `my_custom_field5`, `my_custom_field8` and `my_custom_field10` is ambiguous. Since e.g. `my_custom_field5` is explicitly annotated like:

`GenericField[Union[CustomFieldValue, int], CustomFieldValue](null=True)`

But the get type is then rewritten from `CustomFieldValue` to `Union[myapp.models.CustomFieldValue, None]` due to `null=True`. I get the convenient part of this, but when passing generic types explicitly I'd argue we shouldn't rewrite. Especially not without any note or message.

In some sense I think the explicitly passed get type(`CustomFieldValue`) and the underlying genericness of a field, triggered by `null=True`, could raise an incompatibility error. I wouldn't find it strange if the declaration of `my_custom_field5`, `my_custom_field8` or `my_custom_field10` yielded an error regarding the mentioned incompatibility, saying that their _get_ type needs include `None` as an additional type.

I have no idea if the above is related to this PR or not though.

mypy_django_plugin/transformers/fields.py
flaeppe: I don't see any harm in supporting what you mention, don't think it'll affect anything else

tests/typecheck/fields/test_custom_fields.yml
Viicos: Would really like to see https://github.com/typeddjango/django-stubs/pull/1900 making it as well, although in its current form it doesn't really simplify creating custom field subclasses (and I believe addressing https://github.com/typeddjango/django-stubs/pull/1900#issuecomment-1878363899 would help in this case but requires significant work on mypy)

tests/typecheck/fields/test_custom_fields.yml
md384: Okay, I will change to throw an error. I would think that any field that maps to non-optional and has null=True should also throw this error, for example, 

```py

class CustomValueField(fields.Field[Union[CustomFieldValue, int], CustomFieldValue]): ...

CustomValueField(null=True)
```
would also be incompatible.

md384: I will also remove the code that rewrites fields with null=True to be optional and we'll rely on developers being explicit

mypy_django_plugin/transformers/fields.py
md384: Okay, I have changed this so explicit type params will override the `_pyi_private_set_type` and `_pyi_private_get_type` variables

mypy_django_plugin/transformers/fields.py
```diff
@@ -150,6 +152,22 @@ def set_descriptor_types_for_field(
...
+
+    # bail if either mapped_set_type or mapped_get_type have type Never
+    if not (isinstance(mapped_set_type, UninhabitedType) or isinstance(mapped_get_type, UninhabitedType)):
+        # always replace set_type and get_type with (non-Any) mapped types
+        set_type = helpers.convert_any_to_type(mapped_set_type, set_type)
+        get_type = helpers.convert_any_to_type(mapped_get_type, get_type)
+
+        # the get_type must be optional if the field is nullable
+        if (is_get_nullable or is_nullable) and not (isinstance(get_type, NoneType) or helpers.is_optional(get_type)):
+            ctx.api.fail("Field is nullable but generic get type parameter is not optional", ctx.context)
```
flaeppe: What do you think about extracting the name of the initiated field from `ctx` (`FunctionContext`) and include it in the error message?

e.g.

```
CustomValueField is nullable but its generic get type parameter is not optional  [misc]
```

mypy_django_plugin/transformers/fields.py
```diff
@@ -4,14 +4,16 @@
...
 from mypy.nodes import AssignmentStmt, NameExpr, TypeInfo
 from mypy.plugin import FunctionContext
-from mypy.types import AnyType, Instance, ProperType, TypeOfAny, UnionType
+from mypy.types import AnyType, Instance, NoneType, ProperType, TypeOfAny, UninhabitedType, UnionType
 from mypy.types import Type as MypyType
 
 from mypy_django_plugin.django.context import DjangoContext
 from mypy_django_plugin.exceptions import UnregisteredModelError
 from mypy_django_plugin.lib import fullnames, helpers
+from mypy_django_plugin.lib.fullnames import FIELD_FULLNAME
```
```suggestion
```

mypy_django_plugin/transformers/fields.py
```diff
@@ -150,6 +152,22 @@ def set_descriptor_types_for_field(
         is_set_nullable=is_set_nullable or is_nullable,
         is_get_nullable=is_get_nullable or is_nullable,
     )
+
+    # reconcile set and get types with the base field class
+    base_field_type = next(base for base in default_return_type.type.mro if base.fullname == FIELD_FULLNAME)
```
```suggestion
    base_field_type = next(base for base in default_return_type.type.mro if base.fullname == fullnames.FIELD_FULLNAME)
```

Since we already have `fullnames` imported

tests/typecheck/fields/test_custom_fields.yml
```diff
@@ -0,0 +1,78 @@
...
+        reveal_type(user.my_custom_field13)  # N: Revealed type is "Union[myapp.models.CustomFieldValue, None]"
+        reveal_type(user.my_custom_field14)  # N: Revealed type is "Union[builtins.bool, None]"
+        reveal_type(user.my_custom_field15)  # N: Revealed type is "None"
+    monkeypatch: true
+    out: |
+      myapp/models:31: error: Field is nullable but generic get type parameter is not optional  [misc]
+      myapp/models:32: error: Field is nullable but generic get type parameter is not optional  [misc]
+      myapp/models:33: error: Field is nullable but generic get type parameter is not optional  [misc]
+      myapp/models:34: error: Field is nullable but generic get type parameter is not optional  [misc]
+      myapp/models:35: error: Field is nullable but generic get type parameter is not optional  [misc]
```
flaeppe:
```suggestion
      myapp/models:35: error: Field is nullable but its generic get type parameter is not optional  [misc]
```

flaeppe (commented):
Looks great. I just have 1 minor thought about the message

tests/typecheck/fields/test_custom_fields.yml
flaeppe: Sounds good 👍

mypy_django_plugin/transformers/fields.py
md384: Yes, good idea

flaeppe (approved):
Nice, thank you!

PR #2049, opened 2024-04-09
CI: Update Django 4.2 version used for test suite

intgr:
# I have made things!

Django 4.2.7 was used when testing with Python versions 3.9 and 3.8. Updated to 4.2.11.

* Solves security advisory https://github.com/typeddjango/django-stubs/security/dependabot/6
  (though the vulnerability has 0 impact for us)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ pytest-shard==0.1.2
 
 # Django deps:
 psycopg2-binary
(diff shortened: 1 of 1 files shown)


intgr:
Merging myself as trivial change.

flaeppe:
👍 I find it interesting that @dependabot didn't pick up on that version. But it seems to do on the `5.x.x` row, ref: #2042

PR #2050, opened 2024-04-09
Add `DeferredAttribute.__get__()`

mthuurne:
This PR adds the `DeferredAttribute.__get__()` method that was previously missing from the stubs.

I assumed that the `DeferredAttribute` descriptor will always be used on `Model` classes.

I left the returned data type as `Any`. In theory it should match the field's get type, but that doesn't seem worth the hassle: as the descriptor stubs are relevant for custom descriptors that inherit it, not for application code that accesses the fields, the data type is not that important.


sobolevn (approved):
Thanks!

PR #2051, opened 2024-04-09
Add missing methods and superclass to `FieldFile`

mthuurne:
This adds the pickle methods `__getstate__()` and `__setstate__()` to `django.db.models.fields.files.FieldFile`.


django-stubs/db/models/fields/files.pyi
```diff
@@ -31,6 +31,8 @@ class FieldFile(File):
     def delete(self, save: bool = ...) -> None: ...
     @property
     def closed(self) -> bool: ...
+    def __getstate__(self) -> dict[str, Any]: ...
+    def __setstate__(self, state: dict[str, Any]) -> None: ...
```
sobolevn, line 36: You can also add `__eq__` and `__hash__` as methods from https://github.com/django/django/blob/617bcf611f3daa796e4054ba041089ece30a32fc/django/db/models/fields/files.py#L25-L33

and you can also add `AltersData` base class: https://github.com/django/django/blob/617bcf611f3daa796e4054ba041089ece30a32fc/django/db/models/fields/files.py#L17

While we are at it.

sobolevn (commented):
Thanks!

django-stubs/db/models/fields/files.pyi
mthuurne: Done.

Please check whether you agree with how I annotated `__eq__()`:
- I assumed `hasattr(other, "name")` was a lazy way to check for `FieldFile`, but if it is intended to accept other classes with a `name` attribute, we should define a `Protocol` instead.
- It accepts `str` for backwards compatibility, but it doesn't state that compatibility is deprecated, so I added it as part of the signature.

django-stubs/db/models/fields/files.pyi
```diff
@@ -33,6 +34,8 @@ class FieldFile(File):
     def closed(self) -> bool: ...
     def __getstate__(self) -> dict[str, Any]: ...
     def __setstate__(self, state: dict[str, Any]) -> None: ...
+    def __eq__(self, other: FieldFile | str) -> bool: ...
```
sobolevn: It can be `object`, because any value (with correct `__eq__`) can be compared with `self.name`

django-stubs/db/models/utils.pyi
```diff
@@ -7,3 +7,5 @@ def make_model_tuple(model: type[Model] | str | tuple[str, str]) -> tuple[str, s
 def resolve_callables(mapping: MutableMapping[str, Any]) -> Iterator[tuple[str, Any]]: ...
 def unpickle_named_row(names: Iterable[str], values: Iterable[Any]) -> NamedTuple: ...
 def create_namedtuple_class(*names: str) -> type[NamedTuple]: ...
+
+class AltersData: ...
```
sobolevn:
```suggestion
class AltersData:
    def __init_subclass__(cls, **kwargs: Any) -> None: ...
```

django-stubs/db/models/fields/files.pyi
```diff
@@ -31,6 +32,10 @@ class FieldFile(File):
     def delete(self, save: bool = ...) -> None: ...
     @property
     def closed(self) -> bool: ...
+    def __getstate__(self) -> dict[str, Any]: ...
+    def __setstate__(self, state: dict[str, Any]) -> None: ...
+    def __eq__(self, other: FieldFile | str) -> bool: ...  # type: ignore[override]
```
sobolevn:
```suggestion
    def __eq__(self, other: object) -> bool: ...
```

sobolevn (approved):
Thanks!

PR #2052, opened 2024-04-09
Refine return type for `ManyToOneRel.get_accessor_name()`

mthuurne:
Only for the special case of a symmetrical m2m relation on self, `ForeignObjectRel.get_accessor_name()` will return `None`. However, `ManyToOneRel` and its subclasses will never be symmetrical, so for those classes `get_accessor_name()` does not return `None`.

Link to [relevant Django code](https://github.com/django/django/blob/ca5cd3e3e8e53f15e68ccd727ec8fe719cc48099/django/db/models/fields/reverse_related.py#L235).


sobolevn (approved):
Thanks!

Issue #2053, opened 2024-04-12
Postgre and GIS model fields do not support the `db_comment` argument.


JaeHyuckSa:
# Bug report
Mypy error occurs when using db_comment argumment of model fields in postgre and gis.

## What's wrong

``` python
from django.contrib.gis.db.models import PointField
from django.contrib.postgres.fields import ArrayField
from django.db import models


class Place(models.Model):
    point = PointField(db_comment="Position coordinates")
    visible_phone_number = ArrayField(
        models.CharField(max_length=16), 
        db_comment="Exposed phone numbers to show customers", 
        size=2
    )
```

You can check the error with the terminal below.
```python
models.py:23: error: Unexpected keyword argument "db_comment" for "ArrayField"  [call-arg]
/Users/sajh/.pyenv/versions/3.11.5/envs/mung/lib/python3.11/site-packages/django-stubs/contrib/postgres/fields/array.pyi:26: note: "ArrayField" defined here

/models.py:27: error: Unexpected keyword argument "db_comment" for "PointField"  [call-arg]
/Users/sajh/.pyenv/versions/3.11.5/envs/mung/lib/python3.11/site-packages/django-stubs/contrib/gis/db/models/fields.pyi:81: note: "PointField" defined here

```

## How is that should be

#1492  The contributor to the issue seems to have forgotten to process the postgres and gis fields while adding the value of the db_comment argument. In order to solve the error, add db_comment argument to GeometryField and BaseSpatialField, which are higher classes in ArrayField and PointField.

## System information

- OS:
- `python` version: 3.11.5
- `django` version: 4.2.11
- `mypy` version: 1.9.0
- `django-stubs` version: 4.2.7
- `django-stubs-ext` version: 4.2.7

PR #2054, opened 2024-04-12
Add `db_comment=` parameter to Postgres and GIS model fields

JaeHyuckSa:
# I have made things!
Add the value of the db_comments argument for the postgres and gis fields.

## Related issues
#2053


JaeHyuckSa:
I didn't add the test separately because I think it's a similar PR to #1492. If necessary, I'll fill out the test and put it in.

sobolevn (approved):
Thanks!

PR #2055, opened 2024-04-12
Correct type for `db.models.sql.query.Query.join()` argument

mthuurne:
The `join` argument of `db.models.sql.query.Query.join()` should be an instance, not a class.

I think the confusion comes from the docstring. In Django 3.2 it reads:

> 'join' is either a sql.datastructures.BaseTable or Join.

The word "a" is ambiguous here: from the `join()` implementation it's clear that an instance is required, but looking at only the docstring it could be misinterpreted as requiring a subclass.

Note that in the latest Django code, the docstring reads:

> 'join' is either a base_table_class or join_class.

This is because new `base_table_class` and `join_class` attributes were added to `db.models.sql.query.Query`. I don't know if it's worth adding type arguments to the `Query` class for this. Either way, I think it would be better to first correct the annotations to work with the non-customized `Query` use case.


sobolevn (approved):

sobolevn:
Thanks!

PR #2058, opened 2024-04-14
Disable mypy `ignore_missing_imports` option

sudosubin:
# I have made things!

Enabled mypy `ignore_missing_imports` option, and updated `test_import_all.yml`.

- Remove removed module imports.
- Ignored mypy error report from `psycopg2.extras`, `redis.*`, and `xml.sax.expatreader`


flaeppe (approved):

intgr:
Why is it not an error if non-existing modules are in this test? Can we enable some mypy option here to discover more such examples?

flaeppe (approved):

flaeppe:
I think it's due to having `ignore_missing_imports = true` in the conf

https://github.com/typeddjango/django-stubs/blob/c25653e9cccccb93466e066082444779ed77fb65/mypy.ini#L6

flaeppe:
Could we just add the following to the `import_all_modules` test case?

```diff
   case: import_all_modules
+    mypy_config: |
+        [mypy]
+        ignore_missing_imports = false
```

sudosubin:
There were some errors caused by `psycopg2.extras`, `redis.*`, and `xml.sax.expatreader`, so I excluded it from the error report.

django-stubs/core/paginator.pyi
```diff
@@ -1,7 +1,7 @@
 from collections.abc import Iterable, Iterator, Sequence, Sized
 from typing import ClassVar, Generic, Protocol, TypeVar, overload, type_check_only
 
-from django.db.fields import _ErrorMessagesDict
+from django.db.models.fields import _ErrorMessagesDict
```
intgr, line 4: Cool, looks like this already found some bugs.

intgr (commented):
The test suite isn't happy about this change. We could disable `ignore_missing_imports` globally, there aren't that many suppressions we need to add.

But this quickly ballooned into a bigger project. I'm OK with reducing the scope of this, if you don't feel like you want to solve all that @sudosubin.

flaeppe:
I agree with that we probably should enable `ignore_missing_imports` globally, better to be strict and relax when necessary. No need to do it all in the same place or at once though, as previously mentioned

sudosubin:
> 
I agree with that we probably should enable `ignore_missing_imports` globally, better to be strict and relax when necessary. No need to do it all in the same place or at once though, as previously mentioned

~It looks like disable `ignore_missing_imports` makes errors like below.~

~It cannot be suppressed by `mypy.ini`'s `[mypy-django.contrib.gis.*] ignore_missing_imports = true` settings, sadly 😢~

~It looks like installing the `gdal` package in the CI environment will fix this issue, but I'm not sure if that's a good idea. (Should it also be added to the `requirements.txt`?)~

~This issue was discovered while I was working on the 5.0 issue resolution, and I'd be happy to take it slow or have other maintainers do it. :)~

The above issue was a problem in my local environment. When I cleared the cache and ran it inside a debian container, there were no problem. Sorry :)

django-stubs/contrib/postgres/fields/ranges.pyi
```diff
@@ -2,7 +2,7 @@ from typing import Any, ClassVar, Literal
 
 from django.db import models
 from django.db.models.lookups import PostgresOperatorLookup
-from psycopg2.extras import DateRange, DateTimeTZRange, NumericRange, Range
+from psycopg2.extras import DateRange, DateTimeTZRange, NumericRange, Range  # type: ignore [import-untyped]
```
intgr, line 5: This should fix the remaining CI errors:

```suggestion
from psycopg2.extras import DateRange, DateTimeTZRange, NumericRange, Range  # type: ignore [import-untyped,unused-ignore]
```

sudosubin:
```
django-stubs/utils/timezone.pyi:6: error: Cannot find implementation or library stub for module named "zoneinfo"  [import-not-found]
django-stubs/utils/timezone.pyi:6: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

Django 5.0 supports Python 3.10, 3.11, and 3.12, and the type error raised from 3.8 matrix. (which does not have `zoneinfo` builtin module yet.)

Does anyone have any ideas to solve this problem? (Remove 3.8, 3.9 support, or add `# type: ignore` comment to the `zoneinfo` import)

intgr:
> Django 5.0 supports Python 3.10, 3.11, and 3.12, and the type error raised from 3.8 matrix. (which does not have zoneinfo builtin module yet.)

Suppressing sounds fine to me. Or maybe a `try / except ImportError`, though not sure if that avoids the `import-not-found` error.

sudosubin:
Thanks! I hope my commit 8e00810a764b38ce8c8249a1d752792473762a98 solve the error :D

andersk:
The PR should be retitled from “Enable …” to “Disable …”, as the value is being changed from `true` to `false`.

sudosubin:
> The PR should be retitled from “Enable …” to “Disable …”, as the value is being changed from true to false.

Oh, that's right. I updated the PR's title. Thanks :)

flaeppe:
Thank you for this

PR #2070, opened 2024-04-16
5.0: Update `django.core.servers.basehttp`

sudosubin:
# I have made things!

Update stubs for `django.core.servers.basehttp` for Django 5.0.

- [x] `django.core.servers.basehttp`
  - [x] `django.core.servers.basehttp.run`'s `on_bind` argument was added
  - [x] `django.core.servers.basehttp.is_broken_pipe_error` was missed

## Related issues

Refs #1493

### DIFF
--- a/django-stubs/core/servers/basehttp.pyi
+++ b/django-stubs/core/servers/basehttp.pyi
@@ -37,7 +37,7 @@ class WSGIRequestHandler(simple_server.WSGIRequestHandler):
 
 def run(
     addr: str,
-    port: str,
+    port: int,
     wsgi_handler: WSGIHandler,
     ipv6: bool = ...,
     threading: bool = ...,
--- a/django-stubs/core/servers/basehttp.pyi
+++ b/django-stubs/core/servers/basehttp.pyi
@@ -1,9 +1,12 @@
 import socketserver
 from io import BytesIO
-from typing import Any
+from typing import Any, Callable
 from wsgiref import simple_server
 
 from django.core.handlers.wsgi import WSGIHandler, WSGIRequest
+
+def get_internal_wsgi_application() -> WSGIHandler: ...
+def is_broken_pipe_error() -> bool: ...
 
 class WSGIServer(simple_server.WSGIServer):
     request_queue_size: int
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
This is strange, why no allowlist entries are modified?

sudosubin:
> This is strange, why no allowlist entries are modified?

I had a question about this too: are the functions that are only used within a module and not imported from other modules automatically excluded? 👀

sobolevn:
> are the functions that are only used within a module and not imported from other modules automatically excluded?

no, 100%

django-stubs/core/servers/basehttp.pyi
```diff
@@ -32,4 +35,12 @@ class WSGIRequestHandler(simple_server.WSGIRequestHandler):
     request_version: str
     def handle(self) -> None: ...
 
-def get_internal_wsgi_application() -> WSGIHandler: ...
+def run(
+    addr: str,
+    port: str,
```
flaeppe: I think the port here can be passed all the way back to initialising a `socketserver.TCPServer` and that has `port` typed as `int`.

Ref: https://github.com/python/typeshed/blob/ce049d505f1960e95a47f24cff29d15ce3cadb0b/stdlib/socketserver.pyi#L78

```suggestion
    port: int,
```

django-stubs/core/servers/basehttp.pyi
```diff
@@ -32,4 +35,12 @@ class WSGIRequestHandler(simple_server.WSGIRequestHandler):
     request_version: str
     def handle(self) -> None: ...
 
-def get_internal_wsgi_application() -> WSGIHandler: ...
+def run(
+    addr: str,
```
flaeppe: I think `addr` here can be passed all the way back to initialising `socketserver.TCPServer` and that has `addr` typed as `str | bytes | bytearray`

Ref: https://github.com/python/typeshed/blob/ce049d505f1960e95a47f24cff29d15ce3cadb0b/stdlib/socketserver.pyi#L78

```suggestion
    addr: str | bytes | bytearray,
```

django-stubs/core/servers/basehttp.pyi
sudosubin: Thank you for confirming what I missed! :)

django-stubs/core/servers/basehttp.pyi
sudosubin: Thanks :)

flaeppe (approved):
Thanks

PR #2071, opened 2024-04-16
Update django app related types

sudosubin:
# I have made things!

- From Django 4.1, `default_app_config` was removed.
- Fix some django app related type hints.


sobolevn (approved):
Thanks!

PR #2074, opened 2024-04-18
fix typing for URL validator.__call__

asottile:
while it will only be successful for `str`, the first step of the validator is to check the type.  this matches the RegexValidator above


sobolevn (approved):
Thanks! 👍

PR #2075, opened 2024-04-19
5.0: Add `django.utils.choices`

sudosubin:
# I have made things!

Update stubs for `django.utils.choices` for Django 5.0.

- [x] `django.forms.fields`
  - [x] `django.forms.fields.CallableChoiceIterator` was removed
- [x] `django.forms.models`
  - [x] `django.forms.models.ModelChoiceIterator` was changed
- [x] `django.utils.choices` was added
  - [x] `django.utils.choices.BaseChoiceIterator` was added
  - [x] `django.utils.choices.BlankChoiceIterator` was added
  - [x] `django.utils.choices.CallableChoiceIterator` was added
  - [x] `django.utils.choices.flatten_choices` was added
  - [x] `django.utils.choices.normalize_choices` was added

I'm not sure if this change follows the existing conventions of this project well, and if I need to define the types more precisely than current.

Please leave any feedback :)

## Related issues

Refs #1493

### DIFF
--- a/django-stubs/utils/choices.pyi
+++ b/django-stubs/utils/choices.pyi
@@ -1,5 +1,7 @@
 from collections.abc import Iterable, Iterator
-from typing import Any, Protocol, TypeAlias, TypeVar, type_check_only
+from typing import Any, Protocol, TypeVar, type_check_only
+
+from typing_extensions import TypeAlias
 
 _Choice: TypeAlias = tuple[Any, Any]
 _ChoiceNamedGroup: TypeAlias = tuple[str, Iterable[_Choice]]
--- a/django-stubs/utils/choices.pyi
+++ b/django-stubs/utils/choices.pyi
@@ -1,0 +1,29 @@
+from collections.abc import Iterable, Iterator
+from typing import Any, Protocol, TypeAlias, TypeVar, type_check_only
+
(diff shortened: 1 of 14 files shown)


django-stubs/utils/choices.pyi
```diff
@@ -0,0 +1,31 @@
+from collections.abc import Iterable, Iterator
+from typing import Any, Protocol, TypeVar, type_check_only
+
+from typing_extensions import TypeAlias
+
+_Choice: TypeAlias = tuple[Any, Any]
+_ChoiceNamedGroup: TypeAlias = tuple[str, Iterable[_Choice]]
+_Choices: TypeAlias = Iterable[_Choice | _ChoiceNamedGroup]
```
sobolevn, line 8: Is the any reason to change the name from `_FieldChoices` to `_Choices`? I think that keeping the old name can safe a lot of changed lines.

sudosubin: I wanted to indicate that `_FieldChoices` is no longer just a type for `Field`s, as it has been moved from `django.db.models.fields` to `django.utils.choices`.

But I'm not sure if this approach is correct 😂

sobolevn: I am fine with both, just double checking.

sobolevn (approved):

PR #2077, opened 2024-04-20
Update pyright report options (`reportMissingTypeArgument`, `reportPrivateUsage`)

sudosubin:
# I have made things!

I'm using pyright and pylance in VSCode, and I've found that the below error occurs most often.

If we do not plan to make any special improvements on this, I suggest changing the options as shown below.

- Changed `reportMissingTypeArgument` from `error` to `warning`, since this projects is already using many generics without type arguments.
- Turned off `reportPrivateUsage`, since this project is already using a lot of private type imports.

```toml
[tool.pyright]
reportMissingTypeArgument = "warning"
reportPrivateUsage = "none"
```

| before | after |
| ------ | ----- |
| CleanShot 2024-04-20 at 16 20 55@2x | CleanShot 2024-04-20 at 16 20 33@2x |

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,8 @@ exclude = [
     ".mypy_cache",
     "build",
 ]
+reportMissingTypeArgument = "warning"
+reportPrivateUsage = "none"
 stubPath = "."
 typeCheckingMode = "strict"
 


sobolevn (approved):
Thanks, I agree

sobolevn:
Before: 1754 errors, 0 warnings, 0 informations 
After: 886 errors, 671 warnings, 0 informations

PR #2078, opened 2024-04-20
5.0: Update `django.contrib.postgres`

sudosubin:
# I have made things!

Update stubs for `django.contrib.postgres` for Django 5.0.

- [x] `django.contrib.postgres.constraints`
  - [x] `django.contrib.postgres.constraints.ExclusionConstraint.__init__`'s `opclasses` argument was removed

Update missing todos for `django.contrib.postgres`.

- Some `__init__`, `formfield` methods did not work. These methods receive `*args` and `**kwarg`s and then pass the arguments to the parent method as is (with some `setdefaults`), but from the perspective of type support, I did not want to overwrite them with `Any`s.

## Related issues

Refs #1493


django-stubs/contrib/postgres/fields/ranges.pyi
```diff
@@ -21,32 +25,41 @@ class RangeOperators:
     NOT_GT: Literal["&<"]
     ADJACENT_TO: Literal["-|-"]
 
-class RangeField(models.Field):
+_Range = TypeVar("_Range", bound=Range[Any])
```
sobolevn: nitpick: when using a name that is really close to an existing one `Range` vs `_Range`, I would prefer to have `Range` and `_RangeT`

sobolevn (commented):
The scope of this PR is a little bit too much for a careful review :)

Next time, please, create smaller ones for small-brained people like me 😆

django-stubs/contrib/postgres/fields/ranges.pyi
sudosubin: Oh, that seems more reasonable. Thanks! https://github.com/typeddjango/django-stubs/pull/2078/commits/b4c339cc37d8e1a5255005e1a3c92ab7460edc42

sobolevn (approved):

django-stubs/contrib/postgres/fields/ranges.pyi
```diff
@@ -21,32 +25,41 @@ class RangeOperators:
...
+
+class RangeField(models.Field[Any, _Range]):
     empty_strings_allowed: bool
-    base_field: models.Field
-    range_type: type[Range]
+    base_field: type[models.Field]
```
delfick, line 32: @sobolevn @sudosubin this is still an instance though isn't it?

https://github.com/django/django/blob/40a60d589e1d0d290c3b79c7e97d9cd0c94e52e3/django/contrib/postgres/fields/ranges.py#L65

PR #2085, opened 2024-04-25
Add `path` signature for async views

jlost:
# I have made things!

Creates an extra overload that permits async views inside of `path`.

## Related issues

#1874

### DIFF
--- a/django-stubs/urls/conf.pyi
+++ b/django-stubs/urls/conf.pyi
@@ -1,6 +1,6 @@
 from collections.abc import Callable, Sequence
 from types import ModuleType
-from typing import Any, overload
+from typing import Any, Coroutine, overload
 
 from django.urls import URLPattern, URLResolver, _AnyURL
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thank you

django-stubs/urls/conf.pyi
```diff
@@ -33,6 +40,13 @@ def re_path(
     route: _StrOrPromise, view: Callable[..., HttpResponseBase], kwargs: dict[str, Any] = ..., name: str = ...
 ) -> URLPattern: ...
 @overload
+def re_path(
+    route: _StrOrPromise,
+    view: Callable[..., Coroutine[Any, Any, HttpResponseBase]],
```
intgr, line 45: Should this be `Coroutine[None, None, HttpResponseBase]`? I don't think view coroutines should yield any values?

flaeppe: Sounds fair enough

PR #2089, opened 2024-04-27
Add missing py.typed File

max-muoto:
Currently the lack of a `py.typed` file means that in VSCode the default bundled Django stubs (derived from [djagno-types](https://github.com/sbdchd/django-types)) are prioritized). This can make it quite confusing (for Pyright users) why you're seeing different results when running the Pyright CLI versus Pyright's type-checking in VSCode. 

For reference, one is present in Djagno-Types as well: https://github.com/sbdchd/django-types


sobolevn:
Sorry, I am not a pyright user. I cannot review :(

max-muoto:
> Sorry, I am not a pyright user. I cannot review :(

Any other reviewers I could possibly ping?

flaeppe:
I don't use pyright either. Does this change affect any other type checker, if not, could we merge and try it out?

intgr:
> in VSCode the default bundled Django stubs (derived from [djagno-types](https://github.com/sbdchd/django-types)) are prioritized).

[According to PEP 561](https://peps.python.org/pep-0561/#stub-only-packages), stubs-only packages don't need to include `py.typed`.

> Note that for stub-only packages adding a py.typed marker is not needed since the name *-stubs is enough to indicate it is a source of typing information.

Did you try that this actually fixes the issue?

I wasn't aware that VSCode or pyright bundle the django-types package. I'm willing to consider work-around hacks, but this is a problem caused by VSCode/pyright first and foremost. I think manually installed stubs should override the defaults. Please open an issue for them and see what they think. In any case, they are in a better position to suggest what the appropriate solution is.

Viicos:
This is an existing issue on the Pylance side: https://github.com/microsoft/pylance-release/issues/5031, so indeed not related to the missing `py.typed` marker which is not required for stubs.

max-muoto:
From what I can tell, while it isn't required, it is a way of at least tricking Pylance into prioritizing it.

intgr:
OK, per issue linked by Viicos, this has been reported upstream and there hasn't been any progress for months.

But did you test that this solves the issue with django-stubs? Usually it's also needed to change `setup.py` to ensure `py.typed` is included in the package.

heejaechang:
some of stub packages bundeld in pylance is a partial stub. py.typed file in them indicate it is partial stub not full stub

See https://peps.python.org/pep-0561/#partial-stub-packages for more detail

That said it sounds like a bug in import resolution order around partial stubs. U should open an issue in pylance

intgr:
Given the evidence that `py.typed` won't actually fix the issue, I'll close this PR.

Follow upstream issue here:
* https://github.com/microsoft/pylance-release/issues/5031

PR #2104, opened 2024-04-30
Remove QuerySet alias hacks via PEP 696 TypeVar defaults

### LINKED ISSUE #681: Error messages refer to private _QuerySet class

intgr:
# I have made things!

The `QuerySet` class was previously named `_QuerySet` and had three aliases: `QuerySet`, `QuerySetAny` and `ValuesQuerySet`. These hacks were mainly needed to for the ergonomic single-parameter `QuerySet[Model]`, which expanded into `_QuerySet[Model, Model]`

Now that mypy 1.10 implements [PEP 696 Type Defaults for Type Parameters](https://peps.python.org/pep-0696/) to a fuller extent (Pyright also supports it), the 2nd type parameter can be a simple TypeVar that defaults to 1st type parameter. All usages of `QuerySetAny` and `ValuesQuerySet` can now be replaced with simple `QuerySet`.

* Bumps `typing-extensions` requrement to `>=4.11.0`, since that's the one that adds `TypeVar(default=)` parameter. https://github.com/python/typing_extensions/blob/main/CHANGELOG.md#release-4110-april-5-2024

TODO:

- [x] Test behavior on mypy 1.9 and earlier
- [x] Update readme

## Related issues

* Fixes #681


ext/django_stubs_ext/aliases.py
```diff
@@ -6,7 +6,7 @@
     from django.utils.functional import _StrPromise as StrPromise
 
     QuerySetAny = _QuerySetAny
-    ValuesQuerySet = _QuerySet[_T, _Row]
+    ValuesQuerySet = _QuerySet[_T, _Row]  # TODO think about this!
```
intgr: Reminder for self.

sobolevn (commented):
Great! I think we can now (in the next PRs) get rid of `WithAnnotations` and provide it as the third type var to `QuerySet`.

ext/django_stubs_ext/aliases.py
intgr: Tested and this works fine, even though `query.pyi` now uses `_Model` TypeVar instead of `_T` in the `QuerySet` class (`_T` still exists).

It's best to keep these aliases as is for now. Since the dependency constraint between `django-stubs` and `django-stubs-ext` only goes one way, users may be running e.g. `django-stubs-ext 5.0.1` and `django-stubs 5.0.0` together.

We can simplify these aliases more a few releases down the line.

intgr:
After this change, we strictly require mypy >=1.10.0. Fortunately, older mypy versions will fail quite loudly.

intgr:
> Great! I think we can now (in the next PRs) get rid of WithAnnotations and provide it as the third type var to QuerySet.

A 3rd typevar is not a good long-term solution for that. What we really need is `Intersection` types.

* Tracking issue: https://github.com/python/typing/issues/213

----

The type of `Publisher.objects.annotate(num_books=Count("book"))` should boil down to:

```python
class NumBooksAnnotation(Protocol):
    num_books: int

QuerySet[Publisher, Publisher & NumBooksAnnotation]
```

We currently emulate the `&` that with our own hand-grown `Annotations` type and `TypedDict` instead of `Protocol`.

We can't get rid of `WithAnnotations` with a `QuerySet` typevar anyway. It's not needed for just `QuerySet`s, it's also needed when passing around `Model` instances containing annotations, e.g.

```python
class NumBooksTypedDict(TypedDict):
    num_books: int

def handle_publisher_with_numbooks(
    publisher: WithAnnotations[Publisher, NumBooksTypedDict]  # <-- (!) not a QuerySet
):
    print(publisher.num_books)


for publisher in Publisher.objects.annotate(num_books=Count("book")):
    handle_publisher_with_numbooks(publisher)
```

tests/typecheck/db/models/test_query.yml
```diff
@@ -54,3 +54,19 @@
...
+
+        a: QuerySet[User]
+        reveal_type(a)  # N: Revealed type is "django.db.models.query.QuerySet[django.contrib.auth.models.User, django.contrib.auth.models.User]"
+        b: QuerySet[User, int]
+        reveal_type(b)  # N: Revealed type is "django.db.models.query.QuerySet[django.contrib.auth.models.User, builtins.int]"
+        c: ValuesQuerySet[User, int]
+        reveal_type(c)  # N: Revealed type is "django.db.models.query.QuerySet[django.contrib.auth.models.User, builtins.int]"
+
+        d: QuerySet[int]        # E: Type argument "int" of "QuerySet" must be a subtype of "Model"  [type-var]
+        e: ValuesQuerySet[int]  # E: Type argument "int" of "QuerySet" must be a subtype of "Model"  [type-var]
```
intgr, line 72: Seemed helpful to add a test for these behaviors. But not entirely sure it's worth it.

intgr:
Anyway this is ready for review/merging from my point of view.

django-stubs/db/models/query.pyi
```diff
@@ -205,28 +208,28 @@ class RawQuerySet(Iterable[_T], Sized):
...
-    def prefetch_related(self, *lookups: Any) -> RawQuerySet[_T]: ...
+    def prefetch_related(self, *lookups: Any) -> RawQuerySet[_Model]: ...
     def resolve_model_init_order(self) -> tuple[list[str], list[int], list[tuple[str, int]]]: ...
-    def using(self, alias: str | None) -> RawQuerySet[_T]: ...
-
-_QuerySetAny: TypeAlias = _QuerySet  # noqa: PYI047
+    def using(self, alias: str | None) -> RawQuerySet[_Model]: ...
 
-QuerySet: TypeAlias = _QuerySet[_T, _T]
+# Obsolete aliases of QuerySet, for compatibility only.
```
sobolevn: Do we really need this compat alias?

ext/django_stubs_ext/aliases.py
```diff
@@ -5,14 +5,16 @@
...
 
+    # Obsolete type aliases. Use the QuerySet class directly instead.
     QuerySetAny = _QuerySetAny
     ValuesQuerySet = _QuerySet[_T, _Row]
 else:
     from django.db.models.query import QuerySet
     from django.utils.functional import Promise as StrPromise
 
+    StrOrPromise = typing.Union[str, StrPromise]
+    # Obsolete type aliases. Use the QuerySet class directly instead.
```
sobolevn:
```suggestion
    # Deprecated type aliases. Use the QuerySet class directly instead.
```

I think that we should deprecate them in the future with a real warning.

sobolevn (commented):
Thanks a lot, really great feature!

django-stubs/db/models/query.pyi
intgr: Currently, yes. `ext/django_stubs_ext/aliases.py` still expects to import `from django.db.models.query import _QuerySet, _QuerySetAny`. I'll need to think if we change `aliases.py` without breaking compatibility.

It's really difficult to think about these, because the dependency constraint between django-stubs and django-stubs-ext only goes one way, users may be running e.g. `django-stubs-ext 5.0.1` and `django-stubs 5.0.0` together.

I wonder if we can make the version constraint strict `django-stubs-ext==5.0.0`, or will that cause issues for package managers?

intgr: btw when adding review comments, please add it on the *last* relevant line of a change, rather than the first line, this way more of the diff context is visible.

Right now the message is missing the context that it's about `_QuerySet` and `_QuerySetAny`.

sobolevn (approved):
Fair enough, thanks again!

django-stubs/db/models/query.pyi
intgr: Here's an idea:

*  #2117

We can get rid of `_QuerySetAny` after the above is merged. But the `_QuerySet` alias must remain for the time being.

intgr:
Created an "announcement" discussion for this deprecation as well:

* https://github.com/typeddjango/django-stubs/discussions/2194

PR #2108, opened 2024-05-01
Fix annotation for `request.auser`

bshelkhonov:
# I have made things!

Previously type tests failed on function like this because `request.auser` was annotated incorrectly. I fixed this.
```python
async def get_user_async(request: HttpRequest) -> AbstractBaseUser | AnonymousUser:
    return await request.auser()
```

### DIFF
--- a/tests/typecheck/test_request.yml
+++ b/tests/typecheck/test_request.yml
@@ -106,3 +106,15 @@
         reveal_type(req.GET)  # N: Revealed type is "django.http.request._ImmutableQueryDict"
         req.GET['foo'] = 'bar'  # E: This QueryDict is immutable.  [misc]
         x = 1  # E: Statement is unreachable  [unreachable]
+
+-   case: request_user_auser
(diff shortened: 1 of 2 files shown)


tests/typecheck/test_request.yml
```diff
@@ -106,3 +106,15 @@
         reveal_type(req.GET)  # N: Revealed type is "django.http.request._ImmutableQueryDict"
         req.GET['foo'] = 'bar'  # E: This QueryDict is immutable.  [misc]
         x = 1  # E: Statement is unreachable  [unreachable]
+
+-   case: request_user_auser
```
sobolevn: Thanks for it, but I don't think we need this test. The change is simple enough.

bshelkhonov: Yeah, i see my test broke the pipeline. I removed test

sobolevn (approved):
Thanks!

PR #2117, opened 2024-05-03
Prepare django-stubs-ext for QuerySet typevar refactor

intgr:
* Preparation for: #2104

If we merge this first, and don't change `aliases.py` later, then we can be sure that next `django-stubs-ext 5.0.1` remains compatible with `django-stubs 5.0.0` (since CI will have checked both combinations)


sobolevn (approved):
Let's try this! This feature has multiple tests, so if they pass - I am quite confident.

PR #2121, opened 2024-05-03
`BaseBackend.get_user`: Relax `user_id` param type

### LINKED ISSUE #2120: `django.contrib.auth.BaseBackend.get_user()`: `user_id` is too strict
# Bug report

## What's wrong

If a [custom django `User` model](https://docs.djangoproject.com/en/5.0/topics/auth/customizing/#auth-custom-user) is used via [`AUTH_USER_MODEL`](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-AUTH_USER_MODEL).

In our case, using a `UUID` primary key runs into issues, as when we subclass, it will try to force `user_id` to be an `int`:

```
.../app_name/backends.py:25: error: Argument 1 of "get_user" is incompatible with supertype "BaseBackend"; supertype defines the argument type as "int"  [override]
.../app_name/backends.py:25: note: This violates the Liskov substitution principle
.../app_name/backends.py:25: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

## How is that should be

[The typing for `user_id` in `django.contrib.auth.BaseBackend.get_user()`](https://github.com/typeddjango/django-stubs/blob/5.0.0/django-stubs/contrib/auth/backends.pyi#L16) should be relaxed  - as `str` is too strict and breaks in cases where a `UUID` primary key is used.

Possibilities:

- A narrow typing (risky): `user_id: int | str | uuid.UUID`, `int` is a possibility for incrementing primary keys - but other types of primary keys, like `UUID`s and `str` are possible.
- A broad typing (safest): `user_id: Any`, [`typing.Any`](https://docs.python.org/3/library/typing.html#the-any-type) would give us the best result here, as it's not predictable what the underlying user model's primary key will be.
- Any other opinions?

Also, some django links:
- Source of [`django.contrib.auth.BaseBackend.get_user()`](https://github.com/django/django/blob/5.0.4/django/contrib/auth/backends.py#L12) as of Django 5.0.4
- [Writing an authentication backend](https://docs.djangoproject.com/en/5.0/topics/auth/customizing/#writing-an-authentication-backend) in Django project docs

## System information

- OS: macOS
- `python` version: 3.11.3
- `django` version: 4.2
- `mypy` version: 1.9.0
- `django-stubs` version: Various versions, but this is verified up to django-stubs 5.0
- `django-stubs-ext` version: Various versions

tony:
# Loosen the `user_id` typing in `BaseBackend.get_user()`

In theory, this pk could be anything, given that `AUTH_USER_MODEL` can resolved to any custom model (often a `AbstractBaseUser` subclass.)

See also:
- https://docs.djangoproject.com/en/5.0/topics/auth/customizing/#substituting-a-custom-user-model
- https://github.com/django/django/blob/5.0.4/django/contrib/auth/backends.py

## Related issues

- Closes #2120


sobolevn (approved):
Makes sense for `BaseBackend`: `uuid`, `str`, `ObjectID`, etc

django-stubs/contrib/auth/backends.pyi
```diff
@@ -13,7 +13,7 @@ UserModel: Any
 
 class BaseBackend:
     def authenticate(self, request: HttpRequest | None, **kwargs: Any) -> AbstractBaseUser | None: ...
-    def get_user(self, user_id: int) -> AbstractBaseUser | None: ...
+    def get_user(self, user_id: Any) -> AbstractBaseUser | None: ...
```
intgr, line 16: Personally I would have preferred `int | str | UUID`. But OK, let's keep this for now

PR #2125, opened 2024-05-05
Improve typing of `django.db.backends.mysql.base.DatabaseWrapper` attributes

JaeHyuckSa:
# I have made things!
I modified the `Any` value of the `django.db.backends.mysql.base.DatabaseWrapper`'s arguments .

- [x] `django.db.backends.mysql.base.DatabaseWrapper.data_types`
- [x] `django.db.backends.mysql.base.DatabaseWrapper.operators`
- [x] `django.db.backends.mysql.base.DatabaseWrapper.pattern_ops`
- [x] `django.db.backends.mysql.base.DatabaseWrapper.isolation_levels`
- [x] `django.db.backends.mysql.base.DatabaseWrapper.SchemaEditorClass`
- [x] `django.db.backends.mysql.base.DatabaseWrapper.isolation_level`
- [x] `django.db.backends.mysql.base.DatabaseWrapper.needs_rollback`

### DIFF
--- a/django-stubs/db/backends/mysql/base.pyi
+++ b/django-stubs/db/backends/mysql/base.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Container, Iterator
-from typing import Any, Literal
+from typing import Any, Literal, Protocol
 
 from django.db.backends.base.base import BaseDatabaseWrapper
 
@@ -14,6 +14,8 @@ from .validation import DatabaseValidation
 version: Any
 django_conversions: Any
 server_version_re: Any
+
+class MySQLDatabase(Protocol): ...
 
 class CursorWrapper:
     codes_for_integrityerror: Any
@@ -40,12 +42,11 @@ class DatabaseWrapper(BaseDatabaseWrapper):
     validation_class: type[DatabaseValidation]
 
     vendor: str
-    data_types: dict[str, str]
     operators: dict[str, str]
     pattern_esc: str
     pattern_ops: dict[str, str]
     isolation_levels: set[str]
-    Database: Any
+    Database: MySQLDatabase
     SchemaEditorClass: type[DatabaseSchemaEditor]
     isolation_level: set[str]
     def get_connection_params(self) -> dict[str, Any]: ...
@@ -53,7 +54,6 @@ class DatabaseWrapper(BaseDatabaseWrapper):
     def init_connection_state(self) -> None: ...
     def create_cursor(self, name: Any | None = ...) -> CursorWrapper: ...
     def disable_constraint_checking(self) -> Literal[True]: ...
-    needs_rollback: bool
     def enable_constraint_checking(self) -> None: ...
     def check_constraints(self, table_names: Any | None = ...) -> None: ...
     def is_usable(self) -> bool: ...
--- a/django-stubs/db/backends/mysql/base.pyi
+++ b/django-stubs/db/backends/mysql/base.pyi
@@ -8,6 +8,7 @@ from .creation import DatabaseCreation
 from .features import DatabaseFeatures
 from .introspection import DatabaseIntrospection
 from .operations import DatabaseOperations
+from .schema import DatabaseSchemaEditor
 from .validation import DatabaseValidation
 
 version: Any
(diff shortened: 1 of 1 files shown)


django-stubs/db/backends/mysql/base.pyi
```diff
@@ -39,20 +40,20 @@ class DatabaseWrapper(BaseDatabaseWrapper):
     validation_class: type[DatabaseValidation]
 
     vendor: str
-    data_types: Any
-    operators: Any
+    data_types: dict[str, str]
```
sobolevn: This is a cached property. Maybe we should reflect that? https://github.com/django/django/blob/9a27c76021f934201cccf12215514a3091325ec8/django/db/backends/mysql/base.py#L139

django-stubs/db/backends/mysql/base.pyi
```diff
@@ -39,20 +40,20 @@ class DatabaseWrapper(BaseDatabaseWrapper):
...
+    operators: dict[str, str]
     pattern_esc: str
-    pattern_ops: Any
-    isolation_levels: Any
+    pattern_ops: dict[str, str]
+    isolation_levels: set[str]
     Database: Any
```
sobolevn: It is `Database = Database` https://github.com/django/django/blob/9a27c76021f934201cccf12215514a3091325ec8/django/db/backends/mysql/base.py#L202

django-stubs/db/backends/mysql/base.pyi
```diff
@@ -39,20 +40,20 @@ class DatabaseWrapper(BaseDatabaseWrapper):
...
-    isolation_level: Any
+    SchemaEditorClass: type[DatabaseSchemaEditor]
+    isolation_level: set[str]
     def get_connection_params(self) -> dict[str, Any]: ...
     def get_new_connection(self, conn_params: Any) -> Any: ...
     def init_connection_state(self) -> None: ...
     def create_cursor(self, name: Any | None = ...) -> CursorWrapper: ...
     def disable_constraint_checking(self) -> Literal[True]: ...
-    needs_rollback: Any
+    needs_rollback: bool
```
sobolevn: please, add a comment that this is conditionally defined in https://github.com/django/django/blob/9a27c76021f934201cccf12215514a3091325ec8/django/db/backends/mysql/base.py#L310

saJaeHyukc: I will add comment for that part :)

django-stubs/db/backends/mysql/base.pyi
saJaeHyukc: Thank you for the good point :) I will fix that part

django-stubs/db/backends/mysql/base.pyi
saJaeHyukc: For external libraries, not `django`, could you tell me how to check the type?

intgr: You can also define a custom `Protocol` class that lists all members expected to exist in the database module. Mypy at least supports assigning modules to a Protocol.

django-stubs/db/backends/mysql/base.pyi
intgr: Where possible, please just remove attributes that are present in the base class and have the same definition.

`BaseDatabaseWrapper` already defines `needs_rollback: bool`, so this can be removed here.

django-stubs/db/backends/mysql/base.pyi
saJaeHyukc: Can you tell me what is the best file to place a `Protocol` custom class?

sobolevn: near the place where they are used :)

django-stubs/db/backends/mysql/base.pyi
```diff
@@ -8,12 +8,15 @@ from .creation import DatabaseCreation
...
 from .introspection import DatabaseIntrospection
 from .operations import DatabaseOperations
+from .schema import DatabaseSchemaEditor
 from .validation import DatabaseValidation
 
 version: Any
 django_conversions: Any
 server_version_re: Any
 
+class MySQLDatabase(Protocol): ...
```
sobolevn, line 18: This protocol should have at least some attributes / methods

saJaeHyukc: I'm having a hard time figuring it out. I think I just need to add methods or attributes related to the connection, but is there anything else I can include?

PR #2126, opened 2024-05-05
5.0: Update `django.contrib.messages.test`

JaeHyuckSa:
# I have made things!
Update stubs for `django.contrib.messages.test` for Django 5.0.

- [x]  `django.contrib.messages.test` was added

## Related issues

Refs

Upstream PR
- https://github.com/django/django/pull/17101

### DIFF
--- a/django-stubs/contrib/messages/test.pyi
+++ b/django-stubs/contrib/messages/test.pyi
@@ -1,0 +1,9 @@
+from typing import Any
+
+from django.contrib.messages.storage.base import Message
+from django.http.response import HttpResponse
+
+class MessagesTestMixin:
+    def assertMessages(
+        self, response: HttpResponse, expected_messages: list[dict[str, Any] | type[Message]], *, ordered: bool = ...
+    ) -> None: ...
--- a/django-stubs/contrib/messages/test.pyi
+++ b/django-stubs/contrib/messages/test.pyi
@@ -1,9 +1,9 @@
 from typing import Any
 
-from django.contrib.messages.storage.base import Message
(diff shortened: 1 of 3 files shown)


django-stubs/contrib/messages/test.pyi
```diff
@@ -0,0 +1,9 @@
+from typing import Any
+
+from django.contrib.messages.storage.base import Message
+from django.http.response import HttpResponse
+
+class MessagesTestMixin:
+    def assertMessages(
+        self, response: HttpResponse, expected_messages: list[dict[str, Any] | type[Message]], *, ordered: bool = ...
```
sobolevn: I am not sure that this is correct. `expected_messages` should be equal to `list(get_messages(...))` result: https://github.com/django/django/blob/9a27c76021f934201cccf12215514a3091325ec8/django/contrib/messages/test.py#L6

`get_messages()` definition: https://github.com/typeddjango/django-stubs/blob/b0858a799b2cc7e9abec1e05ec5c0ce2c1875320/django-stubs/contrib/messages/api.pyi#L16

saJaeHyukc: I see. I'll fix it and upload it :)

sobolevn (approved):
You can transform `BaseStorage` into `list`. See https://github.com/django/django/blob/34a503162fe222033a1cd3249bccad014fcd1d20/django/contrib/messages/storage/base.py#L64-L72

And it will work with https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual

So, I would prefer to keep it here 👍 

Thanks!

PR #2131, opened 2024-05-07
Update django to 5.0.5

sobolevn:
Closes #2130


requirements.txt
```diff
@@ -9,9 +9,9 @@ pytest-shard==0.1.2
 # Django deps:
 psycopg2-binary
 Django==4.2.11; python_version < '3.10'
```
intgr:
```suggestion
Django==4.2.12; python_version < '3.10'
```

intgr:
The new `djang-stubs[oracle]` extra should be mentioned in release notes.

scripts/stubtest/allowlist.txt
```diff
@@ -393,6 +393,10 @@ django.urls.resolvers.URLPattern.lookup_str
 django.urls.resolvers.URLResolver.url_patterns
 django.urls.resolvers.URLResolver.urlconf_module
 django.utils.connection.BaseConnectionHandler.settings
+django.db.models.Q.referenced_base_fields
```
sobolevn, line 396: We also have to figure out a way to work with `cached_property` without this hack

intgr: I think Django was planning to replace their homegrown `cached_property` implementation with `functools.cached_property` in 5.1 or 5.2.

If we have the patience to wait until then, this will resolve itself 😄

flaeppe: I think it might take quite some time? According to this comment and ticket https://code.djangoproject.com/ticket/30949#comment:14

Or is it mentioned to be replaced someplace else?

intgr: No, I think you're right, it will take years.

PR #2133, opened 2024-05-07
Simplify `matrix-test`, drop old django versions

sobolevn:


.github/workflows/test.yml
```diff
@@ -125,19 +125,15 @@ jobs:
...
-          - python-version: '3.8'
-            django-version: '3.2'
           - python-version: '3.8'
             django-version: '4.2'
           - python-version: '3.9'
-            django-version: '3.2'
-          - python-version: '3.9'
             django-version: '4.2'
-          - python-version: '3.12'
-            django-version: '4.1'
```
intgr, line 140: Oh, I didn't realize Django 4.1 has been EOL for so long already.

We should remove Django 4.1 "partial support" from README as well.

intgr:
Release notes should also mention removal of 4.1 support.

PR #2134, opened 2024-05-07
Use a `TypedDict` for `SimpleUploadedFile.from_dict`

Viicos:
# I have made things!

### DIFF
--- a/django-stubs/core/files/uploadedfile.pyi
+++ b/django-stubs/core/files/uploadedfile.pyi
@@ -1,7 +1,7 @@
-from typing import IO
+from typing import IO, TypedDict
 
 from django.core.files.base import File
-from typing_extensions import Self
+from typing_extensions import NotRequired, Self
 
 class UploadedFile(File):
     content_type: str | None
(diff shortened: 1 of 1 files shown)


django-stubs/core/files/uploadedfile.pyi
```diff
@@ -44,7 +44,9 @@ class InMemoryUploadedFile(UploadedFile):
     ) -> None: ...
     def open(self, mode: str | None = ...) -> Self: ...  # type: ignore[override]
 
+_FileDict = TypedDict("_FileDict", {"filename": str, "content": bytes | None, "content-type": NotRequired[str])
```
sobolevn, line 47: Please, use class syntax.

django-stubs/core/files/uploadedfile.pyi
```diff
@@ -44,7 +44,9 @@ class InMemoryUploadedFile(UploadedFile):
...
+
 class SimpleUploadedFile(InMemoryUploadedFile):
     def __init__(self, name: str, content: bytes | None, content_type: str = ...) -> None: ...
     @classmethod
-    def from_dict(cls, file_dict: dict[str, str | bytes]) -> Self: ...
+    def from_dict(cls, file_dict: _FileDict) -> Self: ...
```
sobolevn, line 52: I vote -1 on this change, because of this: https://mypy-play.net/?mypy=latest&python=3.12&gist=53713000a8df5755e88ebc7a0e9261d0

```
main.py:10: error: Argument 1 to "accepts_my_dict" has incompatible type "dict[str, Any]"; expected "MyDict"  [arg-type]
main.py:16: error: Argument 1 to "accepts_my_dict" has incompatible type "dict[str, object]"; expected "MyDict"  [arg-type]
```

django-stubs/core/files/uploadedfile.pyi
Viicos: inline syntax used because of the `content-type` key.

django-stubs/core/files/uploadedfile.pyi
Viicos: Understandable, when doing `from_dict({...})`, it's a nice improvement, but your example with variable `c` looks too restrictive. Let's close this for now

sobolevn:
Thanks a lot for the suggestion! Let's get back to it when mypy will change.

PR #2136, opened 2024-05-09
4.2.8: Update `Query.resolve_lookup_value()` parameters

JaeHyuckSa:
# I have made things!
I checked and updated that 4.2.8 was missing while I was doing a version up to fit django5.0.

- [x]  `django.db.models.sql.query`
    - [x] `django.db.models.sql.query.Query.resolve_lookup_value`'s `summarize` argument was added

[Django 4.2.8 Commit history](https://github.com/django/django/compare/4.2.7...4.2.8#:~:text=%5B4.2.x%5D%20Fixed%20%2334975%20%2D%2D%20Fixed%20crash%20of%20conditional%20aggregate()%20over%20a%E2%80%A6)
This is the highlighted 4.28 release commit history.

## Related issues

Refs

Upstream PR
- https://github.com/django/django/pull/17487


django-stubs/db/models/sql/query.pyi
```diff
@@ -111,7 +111,9 @@ class Query(BaseExpression):
     def get_initial_alias(self) -> str: ...
     def count_active_tables(self) -> int: ...
     def resolve_expression(self, query: Query, *args: Any, **kwargs: Any) -> Query: ...  # type: ignore[override]
-    def resolve_lookup_value(self, value: Any, can_reuse: set[str] | None, allow_joins: bool) -> Any: ...
+    def resolve_lookup_value(
+        self, value: Any, can_reuse: set[str] | None, allow_joins: bool, summarize: bool = ...
```
intgr: Let's use literals in default vaues if they are simple.

```suggestion
        self, value: Any, can_reuse: set[str] | None, allow_joins: bool, summarize: bool = False
```

intgr (changes requested):
Thanks, 1 trivial quibble.

django-stubs/db/models/sql/query.pyi
saJaeHyukc: I have a question. Is `Literal[False]` okay?

intgr: No, the default value after `=` expects a *value*, not _the type of_ a literal value. You'd get a type error from mypy.

saJaeHyukc: Thank you for letting me know :)

PR #2139, opened 2024-05-09
fix: UserAdmin accepts AbstractUser

vadimshvetsov:
# I have made things!
Currently when CustomUser is created Admin doesn't accept it.

There was initial PR wih this fix #1783, but comments weren't addressed yet but I also need this merged.

Let me know if we can do it better.

### DIFF
--- a/django-stubs/contrib/auth/admin.pyi
+++ b/django-stubs/contrib/auth/admin.pyi
@@ -5,9 +5,10 @@ from django.contrib.auth.models import AbstractUser, Group
 from django.http.request import HttpRequest
 from django.http.response import HttpResponse
 
+_AbstractUserT = TypeVar("_AbstractUserT", bound=AbstractUser)
+
 csrf_protect_m: Any
 sensitive_post_parameters_m: Any
-_AbstractUserT = TypeVar("_AbstractUserT", bound=AbstractUser)
 
 class GroupAdmin(admin.ModelAdmin[Group]): ...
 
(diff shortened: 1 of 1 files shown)


django-stubs/contrib/auth/admin.pyi
```diff
@@ -1,16 +1,17 @@
...
 
 from django.contrib import admin
-from django.contrib.auth.models import Group, User
+from django.contrib.auth.models import AbstractUser, Group
 from django.http.request import HttpRequest
 from django.http.response import HttpResponse
 
 csrf_protect_m: Any
 sensitive_post_parameters_m: Any
+_M = TypeVar("_M", bound=AbstractUser)
```
sobolevn:
```suggestion
_AbstractUserT = TypeVar("_AbstractUserT", bound=AbstractUser)
```

django-stubs/contrib/auth/admin.pyi
```diff
@@ -1,16 +1,17 @@
...
 
 class GroupAdmin(admin.ModelAdmin[Group]): ...
 
-class UserAdmin(admin.ModelAdmin[User]):
+class UserAdmin(admin.ModelAdmin[_M]):
```
sobolevn:
```suggestion
class UserAdmin(admin.ModelAdmin[_AbstractUserT]):
```

sobolevn (approved):
Thanks a lot!

PR #2144, opened 2024-05-10
FIx test looking for env options

delfick:
Originally this PR was about how I wanted to remove the `from db.models` imports so I could remove the option that made mypy ignore missing imports (because https://github.com/python/mypy/issues/7777)

Turns out that's all been fixed in latest master. So this PR only removes the option ignoring mypy types and fixes the test for testing adding `MYPYPATH` to the yml `env`.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -3,7 +3,6 @@
 [mypy]
 allow_redefinition = true
 check_untyped_defs = true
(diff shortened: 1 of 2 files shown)


delfick:
oh, that's already there in latest master too. I will close this, my aplogies for the nose.

PR #2150, opened 2024-05-12
Add GitHub actions release workflow for `django-stubs-ext`

### LINKED ISSUE #2114: Configure PyPI for release automation
PR #1950 introduced automation for making releases. But when I tried this out on Tuesday, it did not work: https://github.com/typeddjango/django-stubs/actions/runs/8892716988/job/24417384679

> Token request failed: the server refused the request for the following reasons:
>
> * `invalid-publisher`: valid token, but no corresponding publisher (All lookup strategies exhausted)

It requires some setup to be done on PyPI side. I believe only @sobolevn can do it. See below:

----

https://github.com/typeddjango/django-stubs/pull/2087#issuecomment-2084867311

> @intgr Was the trusted publisher set up on the PyPI side? Do you have access to see if it's configured?
> 
> It should just be to follow this if it's not: https://docs.pypi.org/trusted-publishers/adding-a-publisher/
> 
> Should just be to fill in and post a form

https://github.com/typeddjango/django-stubs/pull/2087#issuecomment-2084870025

> Owner: typeddjango
> Repository name: django-stubs
> Workflow name: release.yml
> Environment name: release

flaeppe:
Similar to #1950 we need to add a trusted publisher for `django-stubs-ext` to PyPI (CC @sobolevn, @intgr)

This time the trusted publisher form at [PyPI](https://pypi.org/) should be filled in with

```
Owner: typeddjango
Repository name: django-stubs
Workflow name: release.yml
Environment name: release-ext
```

## Related issues

- Closes #2114

---

Worth noting is that this workflow releases `django-stubs-ext` every time that `django-stubs` is released


sobolevn (approved):
Trusted publisher is set 👍

flaeppe:
Great! 👍  Now we should be ready for a new round of an automated release

PR #2152, opened 2024-05-12
5.0: Update `django.db.models.field`: add `generated` attribute to model fields

JaeHyuckSa:
# I have made things!
Update stubs for `django.db.models.field.Field` for Django 5.0.

- [x] `django.db.models.fields.__init__`
  - [x] `django.db.models.fields.Field.generated` was added

## Related issues

Refs

Upstream PR
- https://github.com/django/django/pull/16860


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -120,6 +120,7 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
     remote_field: ForeignObjectRel | None
     is_relation: bool
     related_model: type[Model] | Literal["self"] | None
+    generated: bool
```
```suggestion
    generated: ClassVar[bool]
```

See https://github.com/django/django/blob/1a36dce9c5e0627d46582829d7abd47ed872e3aa/django/db/models/fields/__init__.py#L174

saJaeHyukc: I'm having difficulty understanding this section. Could you please clarify why you opted to use `ClassVar`?

sobolevn: Yes, most values defined on a class level are class var (unless their are overriden on an instance level). We need to keep this in mind :)

saJaeHyukc: Thank you for the clarification! I'll be sure to incorporate that into my future work as well!

django-stubs/db/models/fields/generated.pyi
```diff
@@ -11,7 +11,6 @@ from django.utils.datastructures import DictWrapper
 from django.utils.functional import _StrOrPromise
 
 class GeneratedField(models.Field):
-    generated: Literal[True]
```
sobolevn, line 14: Why not `generated: ClassVar[Literal[True]]`?

saJaeHyukc: I thought the value was already defined in the parent class so I didn't need it.
If the values are not the same, do I need to specify them separately?

sobolevn: Yes, when you can narrow the type - we can specify extra values.

sobolevn (approved):

PR #2151, opened 2024-05-12
5.0: Update `django.db.models.field.related`

JaeHyuckSa:
# I have made things!
Update stubs for `django.db.models.field.related` for Django 5.0.

- [x]  `django.db.models.fields.related`
  - [x]  `django.db.models.fields.related.ForeignObject.get_joining_fields` was added
  - [x]  `django.db.models.fields.related.ForeignObject.get_reverse_joining_fields` was added

## Related issues

Refs
- https://github.com/typeddjango/django-stubs/issues/1493

Upstream PR
- https://github.com/django/django/pull/16632

PR #2154, opened 2024-05-12
Add `django.db.models.fields.related.ForeignObject` missing methods and attributes

JaeHyuckSa:
# I have made things!

Some of the Django 5.0 updates were missing from the related classes, so I added them referring to the allowlist_todo.

If you need any additional work, please let me know and I'll work on it right away. Thank you :)

<details>

<summary>  

`django.db.models.ForeignObject`

</summary>

- [x] `django.db.models.ForeignObject.__copy__`
- [x] `django.db.models.ForeignObject.contribute_to_related_class`
- [x] `django.db.models.ForeignObject.forward_related_accessor_class`
- [x] `django.db.models.ForeignObject.get_class_lookups`
- [x] `django.db.models.ForeignObject.get_extra_descriptor_filter`
- [x] `django.db.models.ForeignObject.get_foreign_related_value`
- [x] `django.db.models.ForeignObject.get_instance_value_for_fields`
- [x] `django.db.models.ForeignObject.get_joining_columns`
- [x] `django.db.models.ForeignObject.get_local_related_value`
- [x] `django.db.models.ForeignObject.get_path_info`
- [x] `django.db.models.ForeignObject.get_reverse_joining_columns`
- [x] `django.db.models.ForeignObject.get_reverse_path_info`
- [x] `django.db.models.ForeignObject.related_accessor_class`
- [x] `django.db.models.ForeignObject.requires_unique_target`

</details>

## Related issues

- https://github.com/typeddjango/django-stubs/pull/2151


django-stubs/db/models/fields/related.pyi
```diff
@@ -87,6 +94,9 @@ class RelatedField(FieldCacheMixin, Field[_ST, _GT]):
 
 class ForeignObject(RelatedField[_ST, _GT]):
```
sobolevn, line 88: Looks like this class has lots of unsolved problems. Let's solve them!

1. Please, use `ClassVar` where needed: https://github.com/django/django/blob/a09082a9bec18f8e3ee8c10d473013ec67ffe93b/django/db/models/fields/related.py#L522-L530

scripts/stubtest/allowlist_todo.txt
```diff
@@ -969,28 +935,13 @@ django.db.models.fields.json.KeyTransform.as_sqlite
...
-django.db.models.fields.related.ForeignObject.get_instance_value_for_fields
-django.db.models.fields.related.ForeignObject.get_joining_columns
-django.db.models.fields.related.ForeignObject.get_local_related_value
-django.db.models.fields.related.ForeignObject.get_path_info
-django.db.models.fields.related.ForeignObject.get_reverse_joining_columns
-django.db.models.fields.related.ForeignObject.get_reverse_path_info
 django.db.models.fields.related.ForeignObject.path_infos
-django.db.models.fields.related.ForeignObject.related_accessor_class
-django.db.models.fields.related.ForeignObject.requires_unique_target
+django.db.models.fields.related.ForeignObject.get_extra_restriction
```
sobolevn, line 939: 2. Let's add other missing attributes to `ForeignObject`

django-stubs/db/models/fields/related.pyi
```diff
@@ -124,6 +134,7 @@ class ForeignObject(RelatedField[_ST, _GT]):
         error_messages: _ErrorMessagesMapping | None = ...,
         db_comment: str | None = ...,
     ) -> None: ...
+    def __copy__(self) -> Empty: ...
```
intgr: I think this should actually be `Self`

The `Empty` class is just an artifact of a weird hack that Django uses, but it modifies the `__class__` attribute so that it's no longer actually the `Empty` class

```python
    def __copy__(self):
        # We need to avoid hitting __reduce__, so define this
        # slightly weird copy construct.
        obj = Empty()
        obj.__class__ = self.__class__
        obj.__dict__ = self.__dict__.copy()
        return obj
```

django-stubs/db/models/fields/related.pyi
saJaeHyukc: Thank you. As you suggested, I have handled all the relevant parts as class variables.

scripts/stubtest/allowlist_todo.txt
saJaeHyukc: I have two questions:

1. I tried to define the `get_extra_restriction` method, but I encountered an error due to the parameter mismatch with the subclass. I attempted to use `type: ignore` on the method in the subclass, but it was not allowed. What steps can I take to resolve this issue?

2. Should I also add the cached_property method?

django-stubs/db/models/fields/related.pyi
saJaeHyukc: Based on your explanation, that seems correct. Thank you for pointing that out!

PR #2157, opened 2024-05-13
5.0: Update `django.db.models.field.related_descriptors`

JaeHyuckSa:
# I have made things!
Update stubs for `django.db.models.field.related_descriptors` for Django 5.0.

- [x] `django.db.models.fields.related_descriptors`
  - [x]  `django.db.models.fields.related_descriptors.ForwardManyToOneDescriptor.get_prefetch_querysets` was added
  - [x]  `django.db.models.fields.related_descriptors.ReverseOneToOneDescriptor.get_prefetch_querysets` was added

## Related issues

Refs

Upstream PR
- https://github.com/django/django/pull/17136


django-stubs/db/models/fields/related_descriptors.pyi
```diff
@@ -31,6 +31,9 @@ class ForwardManyToOneDescriptor(Generic[_F]):
     def get_prefetch_queryset(
         self, instances: list[Model], queryset: QuerySet | None = ...
     ) -> tuple[QuerySet, Callable, Callable, bool, str, bool]: ...
+    def get_prefetch_querysets(
+        self, instances: list[Model], querysets: list[QuerySet] | None = ...
+    ) -> tuple[QuerySet, Callable, Callable, bool, str, bool]: ...
```
sobolevn: Let's always use generics with type params even if they are just `Any`.

django-stubs/db/models/fields/related_descriptors.pyi
```diff
@@ -60,6 +63,9 @@ class ReverseOneToOneDescriptor(Generic[_From, _To]):
     def get_prefetch_queryset(
         self, instances: list[_From], queryset: QuerySet[_To] | None = ...
     ) -> tuple[QuerySet[_To], Callable[..., Any], Callable[..., Any], bool, str, bool]: ...
+    def get_prefetch_querysets(
```
sobolevn, line 66: Why these two signatures are different? `ForwardManyToOneDescriptor(Generic[_F])` looks generic too

django-stubs/db/models/fields/related_descriptors.pyi
saJaeHyukc: Based on my understanding, are you suggesting that, in addition to the code I have added, it is also acceptable to modify the existing code as you have described?

django-stubs/db/models/fields/related_descriptors.pyi
saJaeHyukc: I simply missed it. I thought that the existing code was written that way for some reason. If you could provide an answer above, I will go ahead and modify all the relevant parts of the class accordingly.

django-stubs/db/models/fields/related_descriptors.pyi
sobolevn: I mean when adding new types instead of `QuerySet` and `Callable` use `QuerySet[Any]`, `Callable[..., Any]`, etc

sobolevn:
@saJaeHyukc please, rebase :)

JaeHyuckSa:
> @saJaeHyukc please, rebase :)

Applied as you mentioned :)

sobolevn (approved):
We can work on annotations later to remove `Any`, it requires quite a lot of effort from both contributor and reviewer.

JaeHyuckSa:
> We can work on annotations later to remove `Any`, it requires quite a lot of effort from both contributor and reviewer.

I agree with the handling of 'Any'. I will continue to work on it as much as possible, but even if I am busy, I will actively help whenever I can!

PR #2164, opened 2024-05-15
Improve typing of view decorators

Viicos:
A new type variable is defined, matching (async) callables taking a request and returning a response.

Some decorators are defined as views to comply with pyright (see https://github.com/microsoft/pyright/issues/5821#issuecomment-1694422066 for an explanation).

Note that some decorators allow `*args, **kwargs` to exist on the decorated views, e.g. `never_cache`:

```python
def never_cache(view_func):
    """
    Decorator that adds headers to a response so that it will never be cached.
    """

    if iscoroutinefunction(view_func):

        async def _view_wrapper(request, *args, **kwargs):
            _check_request(request, "never_cache")
            response = await view_func(request, *args, **kwargs)
            add_never_cache_headers(response)
            return response

    else:

        def _view_wrapper(request, *args, **kwargs):
            _check_request(request, "never_cache")
            response = view_func(request, *args, **kwargs)
            add_never_cache_headers(response)
            return response

    return wraps(view_func)(_view_wrapper)
``` 

However, I think it's good to actually enforce views to take a single `request` argument. What do you think?


tests/typecheck/views/decorators/test_csrf.yml
```diff
@@ -0,0 +1,15 @@
+-   case: view_decorator
+    main: |
+      from django.http.request import HttpRequest
+      from django.http.response import HttpResponse
+      from django.views.decorators.csrf import csrf_protect
+
+      @csrf_protect
+      def good_view(request: HttpRequest) -> HttpResponse:
+          return HttpResponse()
```
sobolevn: pls, add `reveal_type(good__view)`

flaeppe:
> However, I think it's good to actually enforce views to take a single request argument. What do you think?

What about path parameters?

flaeppe:
It's probably fine to keep the first positional argument a request, but it needs to allow an arbitrary amount of args and kwargs after that

django-stubs/views/decorators/__init__.pyi
```diff
@@ -0,0 +1,5 @@
+from typing import TypeVar
+
+from django.utils.deprecation import _AsyncGetResponseCallable, _GetResponseCallable
+
+_ViewFuncT = TypeVar("_ViewFuncT", bound=_AsyncGetResponseCallable | _GetResponseCallable)  # noqa: PYI018
```
intgr: ~~Would make sense to keep `_AsyncGetResponseCallable`, `_GetResponseCallable` definitions together in the same file with `_ViewFuncT`.~~

~~Maybe move those 2 into this file? Or put all of them in `django-stubs/views/__init__.pyi`?~~

django-stubs/views/decorators/debug.pyi
```diff
@@ -1,9 +1,9 @@
...
 
-_F = TypeVar("_F", bound=Callable[..., Any])
+from . import _ViewFuncT
 
 coroutine_functions_to_sensitive_variables: dict[int, Literal["__ALL__"] | tuple[str, ...]]
 
-def sensitive_variables(*variables: str) -> Callable[[_F], _F]: ...
-def sensitive_post_parameters(*parameters: str) -> Callable[[_F], _F]: ...
+def sensitive_variables(*variables: str) -> Callable[[_ViewFuncT], _ViewFuncT]: ...
+def sensitive_post_parameters(*parameters: str) -> Callable[[_ViewFuncT], _ViewFuncT]: ...
```
intgr: `@sensitive_*` decorators can be applied to any regular function, not just view functions.

tests/typecheck/views/decorators/test_csrf.yml
```diff
@@ -0,0 +1,15 @@
+-   case: view_decorator
+    main: |
+      from django.http.request import HttpRequest
+      from django.http.response import HttpResponse
+      from django.views.decorators.csrf import csrf_protect
+
+      @csrf_protect
+      def good_view(request: HttpRequest) -> HttpResponse:
```
intgr: Add a test case with async view as well, please.

django-stubs/views/decorators/__init__.pyi
intgr: Waitaminute. Views aren't necessarily compatible with `_GetResponseCallable`.

Views can have extra arguments from URL parameters, e.g. `def good_view2(request: HttpRequest, url_param: str) -> HttpResponse`

This causes error

> error: Value of type variable "_ViewFuncT" of "csrf_protect" cannot be "Callable[[HttpRequest, str], HttpResponse]"  [type-var]

intgr (changes requested):
Thanks, I like creating a common shared definition for `_ViewFuncT`.

But some changes are needed.

tests/typecheck/views/decorators/test_csrf.yml
intgr: And add a test case with additional parameters, e.g. `def good_view2(request: HttpRequest, url_param: str) -> HttpResponse`

intgr:
> However, I think it's good to actually enforce views to take a single request argument. What do you think?

I believe it would break the hack that our FAQ suggests to constrain `request.user` type for authenticated views? https://github.com/typeddjango/django-stubs?tab=readme-ov-file#how-can-i-create-a-httprequest-thats-guaranteed-to-have-an-authenticated-user

I never liked this hack to begin with (https://github.com/typeddjango/django-stubs/discussions/2046#discussioncomment-9048731), but I think lots of users are relying on it, we shouldn't break it, unless there's strong consensus to do so.

django-stubs/views/decorators/__init__.pyi
Viicos: Yep as suggested by `flaeppe`. I got mislead because some of these decorators are defined using `decorator_from_middleware`, which will call `get_response` (i.e. the view function) only by passing the `request`. So probably we need another view func type for those

Viicos: Ok my bad, `decorator_from_middleware` also supports extra arguments specified in a view function. So in the end I only need to update the definition of `_ViewFuncT`.

tests/assert_type/views/decorators/test_csrf.py
Viicos, line 1: I went for the assert types. Unfortunately I can't do:

```python
assert_type(good_view, Callable[[HttpRequest], HttpResponse])
```

As `Callable` does not have information about the name of the arguments, while `good_view` is inferred as `"(request: HttpRequest) -> HttpResponse"` (which is a good thing).

Considering the decorators are defined with the type vars in a straightforward way I don't think it is that much of an issue to not be able to test this

django-stubs/views/decorators/__init__.pyi
```diff
@@ -0,0 +1,15 @@
...
+from django.http.response import HttpResponseBase
+from typing_extensions import Concatenate
+
+# Examples:
+# def (request: HttpRequest, path_param: str) -> HttpResponseBase
+# async def (request: HttpRequest) -> HttpResponseBase
+_ViewFuncT = TypeVar(  # noqa: PYI018
+    "_ViewFuncT",
+    bound=Callable[Concatenate[HttpRequest, ...], HttpResponseBase]
+    | Callable[Concatenate[HttpRequest, ...], Awaitable[HttpResponseBase]],
```
intgr: Unfortunately this still breaks the `AuthenticatedHttpRequest` hack, https://github.com/typeddjango/django-stubs/pull/2164#issuecomment-2115383139

tests/assert_type/views/decorators/test_csrf.py
flaeppe: I say that we should prefer the type over the name of an argument and thus add an `assert_type` to ensure the return type of the decorator.

You should be able to use `assert_type(good_view, Callable[[HttpRequest], HttpResponse])` if your `good_view` forces the request argument to be positional e.g.

```diff
@csrf_protect
-def good_view(request: HttpRequest) -> HttpResponse:
+def good_view(request: HttpRequest, /) -> HttpResponse:
    return HttpResponse()
+assert_type(good_view, Callable[[HttpRequest], HttpResponse])
```

django-stubs/views/decorators/__init__.pyi
flaeppe: Have we tried/do you think we can use callable protocols here? IIRC `Concatenate` only works in function scope(?)

What if we add a callable protocol for each type? e.g. `View` and `AsyncView`

```python
class View(Protocol):
    def __call__(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponseBase: ...

class AsyncView(Protocol):
    async def __call__(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponseBase: ...

_ViewFuncT = TypeVar("_ViewFuncT", bound=View | AsyncView)
```

tests/assert_type/views/decorators/test_csrf.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@csrf_protect  # type: ignore  # pyright: ignore
+def bad_view(request: int) -> str:
+    return ""
+
+
+@csrf_protect  # type: ignore  # pyright: ignore
+def bad_view_no_arguments() -> HttpResponse:
+    return HttpResponse()
```
flaeppe: Could we add the ignore code(s) too?

django-stubs/views/decorators/__init__.pyi
flaeppe: Also, can we incorporate the `AuthenticatedHttpRequest` hack in the test suite? Unless we haven't done so already

Viicos: `__call__(self, request: HttpRequest, *args: Any, **kwargs: Any)`  will require every view function to have `*args, **kwargs` int their signature. `Callable[Concatenate[HttpRequest, ...], HttpResponseBase]` is currently the only way to express what we want here iirc.

I'll take a look at the `AuthenticatedHttpRequest` hack

flaeppe: > `__call__(self, request: HttpRequest, *args: Any, **kwargs: Any)` will require every view function to have `*args, **kwargs` int their signature. `Callable[Concatenate[HttpRequest, ...], HttpResponseBase]` is currently the only way to express what we want here iirc.

From experimenting with mypy playground, I don't think that `*args` and `**kwargs` need to be included in the signature. Check this playground link: https://mypy-play.net/?mypy=latest&python=3.12&flags=strict&gist=e5d6cc2baf2095c4eab2863ea645faf3

Let me know if I missed something

Edit: Here's a different playground link that includes a `TypeVar` and decorator, much similar to what we have in this PR; https://mypy-play.net/?mypy=latest&python=3.12&flags=strict&gist=277b229672c75316b1b9473c07d4712a

Viicos: Interesting, this seems to have changed really recently in the typing spec:
- spec: https://github.com/python/typing/pull/1723
- mypy: This behavior was implemented from the beginning apparently.
- pyright: https://github.com/microsoft/pyright/issues/7848.

Viicos:
The `AuthenticatedHttpRequest` think is annoying :/ Maybe it could be explicitly exported in `django-stubs-ext` or in the stubs itself, and added as an overload?

tests/assert_type/views/decorators/test_csrf.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+# `assert_type` can only be used when `request` is pos. only.
+assert_type(good_view_positional, Callable[[HttpRequest], HttpResponse])
+
+
+# The decorator works too if `request` is not explicitly pos. only.
+@csrf_protect
+def good_view(request: HttpRequest) -> HttpResponse:
```
intgr, line 20: Ignoring `AuthenticatedHttpRequest` for the moment, the added `test_csrf.py` is failing for me with mypy... Do you know what's up with that?

> tests/assert_type/views/decorators/test_csrf.py:19: error: Value of type variable "_ViewFuncT" of "csrf_protect" cannot be "Callable[[HttpRequest], HttpResponse]"  [type-var]
> tests/assert_type/views/decorators/test_csrf.py:24: error: Value of type variable "_ViewFuncT" of "csrf_protect" cannot be "Callable[[HttpRequest], Coroutine[Any, Any, HttpResponse]]"  [type-var]
> tests/assert_type/views/decorators/test_csrf.py:29: error: Value of type variable "_ViewFuncT" of "csrf_protect" cannot be "Callable[[HttpRequest, int, str], HttpResponse]"  [type-var]
> tests/assert_type/views/decorators/test_csrf.py:34: error: Value of type variable "_ViewFuncT" of "csrf_protect" cannot be "Callable[[HttpRequest, int, str], Coroutine[Any, Any, HttpResponse]]"  [type-var]

intgr: I wanted to look into what could be done about `AuthenticatedHttpRequest`, but this is blocking that.

django-stubs/views/decorators/__init__.pyi
```diff
@@ -0,0 +1,13 @@
...
+from django.http.response import HttpResponseBase
+
+# `*args: Any, **kwargs: Any` means any extra argument(s) can be provided, or none.
+class _View(Protocol):
+    def __call__(self, request: HttpRequest, /, *args: Any, **kwargs: Any) -> HttpResponseBase: ...
+
+class _AsyncView(Protocol):
+    async def __call__(self, request: HttpRequest, /, *args: Any, **kwargs: Any) -> HttpResponseBase: ...
+
+_ViewFuncT = TypeVar("_ViewFuncT", bound=_View | _AsyncView)  # noqa: PYI018
```
intgr, line 13: btw `django.db.transaction.non_atomic_requests` decorator expects views as well, and could re-use this.

tests/assert_type/views/decorators/test_csrf.py
flaeppe: Problem is probably the forced positional of `request` in the protocols

intgr: Indeed. Removing the `/,` from `_View` signature fixes that, but then makes line 9 fail.

We can fix this by duplicating both View protocols for both positional/non-positional cases. I think that would be fine. Unless there are better solutions?

```python
class _View(Protocol):
    def __call__(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponseBase: ...

class _ViewPositionalRequest(Protocol):
    def __call__(self, request: HttpRequest, /, *args: Any, **kwargs: Any) -> HttpResponseBase: ...

class _AsyncView(Protocol):
    async def __call__(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponseBase: ...

class _AsyncViewPositionalRequest(Protocol):
    async def __call__(self, request: HttpRequest, /, *args: Any, **kwargs: Any) -> HttpResponseBase: ...

_ViewFuncT = TypeVar(
    "_ViewFuncT", bound=_View | _ViewPositionalRequest | _AsyncView | _AsyncViewPositionalRequest
)  # noqa: PYI018
```

intgr: EDIT: Sorry, this comment was a mistake. Will need to think about this more.

django-stubs/views/decorators/__init__.pyi
```diff
@@ -0,0 +1,13 @@
+from typing import Any, Protocol, TypeVar
+
+from django.http.request import HttpRequest
+from django.http.response import HttpResponseBase
+
+# `*args: Any, **kwargs: Any` means any extra argument(s) can be provided, or none.
+class _View(Protocol):
+    def __call__(self, request: HttpRequest, /, *args: Any, **kwargs: Any) -> HttpResponseBase: ...
```
intgr, line 8: Looks like defining a Protocol for `request` wouldn't work either.

I'm open to just hinting as `request: Any` to solve the `AuthenticatedHttpRequest` hack. That's basically the approach we had until now.

tests/assert_type/views/decorators/test_csrf.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@csrf_protect  # type: ignore[type-var]  # pyright: ignore[reportArgumentType, reportUntypedFunctionDecorator]
+def bad_view(request: int) -> str:
+    return ""
+
+
+@csrf_protect  # type: ignore[type-var]  # pyright: ignore[reportArgumentType, reportUntypedFunctionDecorator]
+def bad_view_no_arguments() -> HttpResponse:
+    return HttpResponse()
```
intgr, line 46: Let's add a test for the hack as well:

```python


class AuthenticatedHttpRequest(HttpRequest):
    user: User


@csrf_protect
def view_hack_authenticated_request(request: AuthenticatedHttpRequest) -> HttpResponse:
    return HttpResponse()
```

django-stubs/views/decorators/__init__.pyi
```diff
@@ -0,0 +1,13 @@
+from typing import Any, Protocol, TypeVar
+
+from django.http.request import HttpRequest
+from django.http.response import HttpResponseBase
+
+# `*args: Any, **kwargs: Any` means any extra argument(s) can be provided, or none.
+class _View(Protocol):
```
intgr, line 7: Add `@type_check_only` to all protocols.

tests/assert_type/views/decorators/test_csrf.py
flaeppe: I don't know what is best to do here really.. Django runs all this with a first, "forced" positional, `HttpRequest` argument. But it's annoying that _everyone_ then _have_ to add the forced positional marker(`/`) to avoid violating the protocol.

An additional thing to keep in mind with the protocols here is that if the `request` argument isn't a forced positional it dictates the name of the argument to be "request".

i.e. trying something like below will yield an error:

```python
@csrf_protect
def someview(req: HttpRequest) -> HttpResponse: ...
```

---

An additional approach here could be to dig deeper into what attributes the decorator expects. For instance, I think, `@csrf_protect` "just" requires an object of something has the attributes `session` and `COOKIES`(went looking again and found there was a bunch of more attributes/methods, but you get the point) instead of the `HttpRequest` type. But it requires it as the first argument.

I'm not sure how much or if that can help, I think we still would have to decide on forced positional or not, but just wanted to mention it as an alternate approach to the `HttpRequest` type

PR #2168, opened 2024-05-16
Restore inner `class Meta` in models where it is accessible

### LINKED ISSUE #2112: AbstractUser lacks Meta-attribute
# Bug report

## What's wrong

Since version 5.0.0 (probably since #2000), the following code fails:

```python
from django.contrib.auth.models import AbstractUser

class User(AbstractUser):
    class Meta(AbstractUser.Meta):  # error: Name "AbstractUser.Meta" is not defined
        pass
```

obviously this example is shortened, but I think you get the Idea. While it is true that a basic `Model` does not have a `Meta`-attribute, this is not valid for basically any Subclass for Meta, and using this inheritance is actually the recommended way by django docs:

* https://docs.djangoproject.com/en/dev/topics/db/models/#meta-inheritance
* https://docs.djangoproject.com/en/dev/topics/auth/customizing/#using-a-custom-user-model-when-starting-a-project

## How is that should be

`AbstractUser.Meta` should exist for inheritance.

## System information

- OS: Linux (Alpine and Arch tested, but the problem should be valid regardless)
- `python` version: 3.12
- `django` version: 5.0.4
- `mypy` version: 1.10.0
- `django-stubs-ext` version: none

intgr:
# I have made things!

Turns out that the inner `Meta` class is accessible for some models, but not others. Added `class Meta` only to stub model classes where it's accessible from outside (e.g. `AbstractBaseUser.Meta` does not crash with AttributeError). Used a hacky script at https://gist.github.com/intgr/352cac61b4f95e921bb33a0d4b76d324. See https://github.com/typeddjango/django-stubs/issues/2112#issuecomment-2106201549 for details.

## Related issues

* Closes #2112
* Reverts parts of #2000

### DIFF
--- a/django-stubs/contrib/sessions/base_session.pyi
+++ b/django-stubs/contrib/sessions/base_session.pyi
@@ -16,6 +16,11 @@ class AbstractBaseSession(models.Model):
     session_key: str
     objects: Any
 
+    class Meta:
+        verbose_name: str
+        verbose_name_plural: str
+        abstract: bool
+
     @classmethod
     def get_session_store_class(cls) -> type[SessionBase] | None: ...
     def get_decoded(self) -> dict[str, Any]: ...
--- a/django-stubs/contrib/admin/models.pyi
+++ b/django-stubs/contrib/admin/models.pyi
@@ -21,6 +21,7 @@ class LogEntryManager(models.Manager[LogEntry]):
     ) -> LogEntry: ...
 
 class LogEntry(models.Model):
+    # Note: do not create `class Meta` here.
     action_time: models.DateTimeField
     user: models.ForeignKey
     content_type: models.ForeignKey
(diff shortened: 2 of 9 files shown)


django-stubs/contrib/sessions/base_session.pyi
```diff
@@ -16,6 +16,11 @@ class AbstractBaseSession(models.Model):
     session_key: str
     objects: Any
 
+    class Meta:
+        verbose_name: str
+        verbose_name_plural: str
+        abstract: bool
```
sobolevn, line 22: What about `Literal[True]`?

sobolevn (commented):
Oh, I am lost! :)

So, what will happen now if we inherit some model with existing `Meta` and create our own `Meta`?
Can we:
1. Inherit `Meta` as well as in `class MyModel(SomeModel):  class Meta(SomeModel.Meta)`?
2. Or do we have to use `# type: ignore`?
3. Can we use `.Meta` on these types?

Do we need to test any of these?

django-stubs/contrib/sessions/base_session.pyi
intgr: I think using `Literal[True]` here won't help anyone, to the contrary.

If the user subclasses this Meta and sets `abstract = False`, mypy will consider that a type error since it's not  True. But it would be perfectly legal usage.

emmanuel-ferdman:
Superseded by PR #3200, closing :)

PR #2175, opened 2024-05-21
5.0: Update django.db.models.constraints.UniqueConstraint

steve148:
# I have made things!
    
Adds in nulls_distinct field to the UniqueConstraint class. Field can be None or a boolean.

https://docs.djangoproject.com/en/5.0/ref/models/constraints/#nulls-distinct

## Related issues

Refs #1493

## Feedback

None from me. Let me know if I missed anything.

### DIFF
--- a/django-stubs/db/models/constraints.pyi
+++ b/django-stubs/db/models/constraints.pyi
@@ -53,6 +53,7 @@ class UniqueConstraint(BaseConstraint):
     fields: Sequence[str]
     condition: Q | None
     deferrable: Deferrable | None
+    nulls_distinct: bool | None
 
     @overload
     def __init__(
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
Please, reflect this line as well: https://github.com/django/django/blob/15fff62d5d8f28b6b61d63341535c9aafc64c865/django/db/models/constraints.py#L356

steve148:
@sobolevn I amended the change, let me know if this is what you meant. I added the attribute to the class so it shows up as an arg on both the init and as an instance attribute. I also added a quick test to make sure it works.

tests/typecheck/db/models/test_constraints.yml
```diff
@@ -35,3 +35,13 @@
         main:4: note: Possible overload variants:
         main:4: note:     .*
         main:4: note:     .*
+
+-   case: unique_constraint_nulls_distinct
```
sobolevn: Test is not required, can you please remove it? They are really slow :(

steve148: Works for me :+1:

sobolevn (approved):

LaurierMantel:
This is big - thank you @steve148

PR #2180, opened 2024-05-24
`AbstractBaseSession`: Use model fields for subclassed cases

### LINKED ISSUE #2179: `AbstractBaseSession` subclassing: Fields not overridable
# Bug report

Django allows [Extending database-backed session engines](https://docs.djangoproject.com/en/5.0/topics/http/sessions/#extending-database-backed-session-engines) ([example](https://docs.djangoproject.com/en/5.0/topics/http/sessions/#example)). Some of these fields need typings

## What's wrong

Fields of `AbstractBaseSession` (Django source: [5.0.6](https://github.com/django/django/blob/5.0.6/django/contrib/sessions/base_session.py), django-stubs: [5.0.0](https://github.com/typeddjango/django-stubs/blob/5.0.0/django-stubs/contrib/sessions/base_session.pyi#L13-L21)) are set to their return values instead of fields.

Consider the following Custom Session Backend, extending `session_key`'s `max_length`:

```python
from django.contrib.sessions.base_session import AbstractBaseSession
from django.db import models


class CustomServiceSession(AbstractBaseSession):
    session_key = models.CharField(
        _("session key"),
        max_length=512,
        primary_key=True,
    )
```

This raises:

```
example.py:7: error: Incompatible types in assignment (expression has type "CharField[str | int | Combinable, str]", base class "AbstractBaseSession" defined the type as "str")  [assignment]
```

## How is that should be

`session_key` and other `AbstractBaseSession` fields should be overridable as model fields.

Perhaps something similar to `AbstractBaseUser` (django: [5.0.6](https://github.com/django/django/blob/5.0.6/django/contrib/auth/base_user.py#L60-L61), django-stubs: [5.0.0](https://github.com/typeddjango/django-stubs/blob/5.0.0/django-stubs/contrib/auth/base_user.pyi#L16-L21))

## System information

- OS: mac / linux
- `python` version: 3.10.1
- `django` version: 4.2
- `mypy` version: 1.9.0
- `django-stubs-ext` version: 0.7.0+

tony:
# Changes

## `AbstractBaseSession` fields

These currently raise on mypy when overridden, such as by overriding `max_length`, see example in #2179.

In situations where these fields are overridden in custom models, for instance extending 'session_key`'s `max_length`. Follow a similar style to `AbstractBaseUser` (django: [5.0.6](https://github.com/django/django/blob/5.0.6/django/contrib/auth/base_user.py#L60-L61), django-stubs: [5.0.0](https://github.com/typeddjango/django-stubs/blob/5.0.0/django-stubs/contrib/auth/base_user.pyi#L16-L21)).

See also:
- https://docs.djangoproject.com/en/5.0/topics/http/sessions/#extending-database-backed-session-engines
- https://github.com/django/django/blob/5.0.6/django/contrib/sessions/base_session.py
- https://github.com/typeddjango/django-stubs/blob/5.0.0/django-stubs/contrib/sessions/base_session.pyi#L13-L21

## Related issues

- Fixes #2179

### DIFF
--- a/django-stubs/contrib/sessions/base_session.pyi
+++ b/django-stubs/contrib/sessions/base_session.pyi
@@ -11,9 +11,9 @@ class BaseSessionManager(models.Manager[_T]):
     def save(self, session_key: str, session_dict: dict[str, Any], expire_date: datetime) -> _T: ...
 
 class AbstractBaseSession(models.Model):
-    expire_date: datetime
-    session_data: str
-    session_key: str
+    session_key = models.CharField(primary_key=True)
+    session_data = models.TextField()
+    expire_date = models.DateTimeField()
     objects: Any
 
     @classmethod
--- a/django-stubs/contrib/sessions/base_session.pyi
+++ b/django-stubs/contrib/sessions/base_session.pyi
@@ -1,8 +1,9 @@
 from datetime import datetime
-from typing import Any, TypeVar
+from typing import Any, ClassVar, TypeVar
 
 from django.contrib.sessions.backends.base import SessionBase
 from django.db import models
+from typing_extensions import Self
 
 _T = TypeVar("_T", bound=AbstractBaseSession)
 
@@ -14,6 +15,7 @@ class AbstractBaseSession(models.Model):
     session_key = models.CharField(primary_key=True)
     session_data = models.TextField()
     expire_date = models.DateTimeField()
+    objects: ClassVar[BaseSessionManager[Self]]
 
     @classmethod
     def get_session_store_class(cls) -> type[SessionBase] | None: ...
--- a/django-stubs/contrib/sessions/base_session.pyi
+++ b/django-stubs/contrib/sessions/base_session.pyi
@@ -14,7 +14,6 @@ class AbstractBaseSession(models.Model):
     session_key = models.CharField(primary_key=True)
     session_data = models.TextField()
     expire_date = models.DateTimeField()
-    objects: Any
 
     @classmethod
     def get_session_store_class(cls) -> type[SessionBase] | None: ...
--- a/scripts/stubtest/allowlist_todo.txt
+++ b/scripts/stubtest/allowlist_todo.txt
@@ -509,6 +509,7 @@ django.contrib.sessions.backends.signed_cookies.SessionStore.exists
 django.contrib.sessions.base_session.AbstractBaseSession.expire_date
 django.contrib.sessions.base_session.AbstractBaseSession.get_next_by_expire_date
 django.contrib.sessions.base_session.AbstractBaseSession.get_previous_by_expire_date
+django.contrib.sessions.base_session.AbstractBaseSession.objects
 django.contrib.sessions.base_session.AbstractBaseSession.session_data
 django.contrib.sessions.base_session.AbstractBaseSession.session_key
 django.contrib.sessions.base_session.BaseSessionManager.__slotnames__


django-stubs/contrib/sessions/base_session.pyi
```diff
@@ -11,9 +11,9 @@ class BaseSessionManager(models.Manager[_T]):
     def save(self, session_key: str, session_dict: dict[str, Any], expire_date: datetime) -> _T: ...
 
 class AbstractBaseSession(models.Model):
-    expire_date: datetime
-    session_data: str
-    session_key: str
+    session_key: models.CharField
+    session_data: models.TextField
+    expire_date: models.DateTimeField
     objects: Any
```
flaeppe: We should add type parameters to the fields and while we're at it also set a type for `objects`.

e.g. `session_key: models.CharField[str, str]`

Looks like `objects` should have: `objects: ClassVar[BaseSessionManager[Self]]`

Though it looks like fields on other models in the stubs have been declared as runtime expressions. So let's do that here too for the sake of consistency

```suggestion
    session_key = models.CharField(primary_key=True)
    session_data = models.TextField()
    expire_date = models.DateTimeField()
    objects: ClassVar[BaseSessionManager[Self]]
```

django-stubs/contrib/sessions/base_session.pyi
```diff
@@ -11,9 +11,9 @@ class BaseSessionManager(models.Manager[_T]):
     def save(self, session_key: str, session_dict: dict[str, Any], expire_date: datetime) -> _T: ...
 
 class AbstractBaseSession(models.Model):
-    expire_date: datetime
-    session_data: str
-    session_key: str
+    session_key = models.CharField(primary_key=True)
+    session_data = models.TextField()
+    expire_date = models.DateTimeField()
     objects: Any
```
intgr: Also let's remove the `objects` member, it's defined in the `Model` class.

```suggestion
```

flaeppe: It shouldn't be removed, I think, but rather set to:

```suggestion
    objects: ClassVar[BaseSessionManager[Self]]
```

django-stubs/contrib/sessions/base_session.pyi
tony: With the exception of `objects`, my original commit at https://github.com/typeddjango/django-stubs/commit/43caa11fb7e617f90a7d383cdb0e5e36bc0c45d3 was like this, except I made a mistake.

<details>

image

</details>

tony:
@intgr @flaeppe When applying suggested changes, do you prefer I keep your commits atomic, or just squash them into mine? (Up to you)

flaeppe:
> @intgr @flaeppe When applying suggested changes, do you prefer I keep your commits atomic, or just squash them into mine? (Up to you)

I can't say that I have any preference. From my end you can do whatever you feel works best

django-stubs/contrib/sessions/base_session.pyi
tony: I was going to ask to have `objects` deferred to a subsequent PR - but it looks like this may be removed in favor of `Model`s: https://github.com/typeddjango/django-stubs/pull/2180#discussion_r1613639182.

flaeppe: While looking at the runtime, I don't think `objects` should be removed: https://github.com/django/django/blob/bcbc4b9b8a4a47c8e045b060a9860a5c038192de/django/contrib/sessions/base_session.py#L32

intgr:
Yeah, do whatever is quickest. The commits will be squashed on merge anyway.

django-stubs/contrib/sessions/base_session.pyi
tony: In re [here](https://github.com/typeddjango/django-stubs/actions/runs/9226729993/job/25387161956?pr=2180#step:6:14):

> error: django.contrib.sessions.base_session.AbstractBaseSession.objects variable differs from runtime type django.db.models.manager.ManagerDescriptor
> Stub: in file /home/runner/work/django-stubs/django-stubs/django-stubs/contrib/sessions/base_session.pyi:18
> django.contrib.sessions.base_session.BaseSessionManager[Self`0]
> Runtime:
> <django.db.models.manager.ManagerDescriptor object at 0x7f56a53328a0>

Does it want it to be `ClassVar[BaseSessionManager[_T]]` maybe?

tony: @flaeppe @intgr Any objection to keeping `objects: Any` for now?

flaeppe: @tony Lets keep `objects: ClassVar[BaseSessionManager[Self]]`

And then add `django.contrib.sessions.base_session.AbstractBaseSession.objects` to the `allowlist_todo.txt`. It's missing from there since it's been typed as `objects: Any`. But adding it to `allowlist_todo.txt` in this case is fine, it aligns with `django.contrib.auth.models.AbstractUser.objects`, which is an identical case.

Add it here:

https://github.com/typeddjango/django-stubs/blob/e19698594e69f3dc925533044d96f6d84ea9e063/scripts/stubtest/allowlist_todo.txt#L510-L512

```diff
django.contrib.sessions.base_session.AbstractBaseSession.get_next_by_expire_date
django.contrib.sessions.base_session.AbstractBaseSession.get_previous_by_expire_date
+django.contrib.sessions.base_session.AbstractBaseSession.objects
django.contrib.sessions.base_session.AbstractBaseSession.session_data
```

flaeppe: Then it's possible to come back to this, at a later time, and handle the descriptor case in whatever way is deemed proper. But the descriptor is "pass through" anyways([runtime ref](https://github.com/django/django/blob/bcbc4b9b8a4a47c8e045b060a9860a5c038192de/django/db/models/manager.py#L184-L204)), so having it as `objects: ClassVar[BaseSessionManager[Self]]` is totally fine to begin with.

flaeppe (approved):
Nice work. Thank you!

tony:
@flaeppe @intgr Vice versa! We made it!

tony:
> We made it!

Maybe I spoke too soon. @intgr Is this solution fine with you?

flaeppe:
I'll merge this in the meantime. Lets circle back if we figure out that there's anything more that needs to be done.

django-stubs/contrib/sessions/base_session.pyi
```diff
@@ -11,10 +12,10 @@ class BaseSessionManager(models.Manager[_T]):
...
 
 class AbstractBaseSession(models.Model):
-    expire_date: datetime
-    session_data: str
-    session_key: str
-    objects: Any
+    session_key = models.CharField(primary_key=True)
+    session_data = models.TextField()
+    expire_date = models.DateTimeField()
+    objects: ClassVar[BaseSessionManager[Self]]
```
intgr, line 18: Yep this is better because it's more precise, thanks.

PR #2182, opened 2024-05-24
Migrate 'test_conf.yml' to 'assert_type' tests

flaeppe:
I tried to get started on moving `.yml` tests to `assert_type` style tests. Also tried to adjust the cases to fit the `assert_type` style a little bit better.

Let me know what you think.


django-stubs/urls/conf.pyi
```diff
@@ -11,9 +11,7 @@ from ..http.response import HttpResponseBase
 
 _URLConf: TypeAlias = str | ModuleType | Sequence[_AnyURL]
 
-def include(
-    arg: _URLConf | tuple[_URLConf, str], namespace: str | None = ...
-) -> tuple[Sequence[URLResolver | URLPattern], str | None, str | None]: ...
+def include(arg: _URLConf | tuple[_URLConf, str], namespace: str | None = ...) -> IncludedURLConf: ...
```
sobolevn: Maybe we can:
- rename it to `_IncludedURLConf`
or
- add it to `ext` module and import from there?

sobolevn (approved):
Test changes look good! Thanks! One question

django-stubs/urls/conf.pyi
flaeppe: Perhaps prefix with an underscore to `_IncludedURLConf`? Thinking since it's only an alias

flaeppe: Done

sobolevn (approved):

Issue #2020, opened 2024-03-21
Next release planning (5.0.0)


intgr:
After a few months hiatus, I'm back and planning a release. But don't hold your breath quite yet.

This will be the first release to be officially based on Django 5.0. This **does not mean 100% type coverage for Django 5.0 features**. We are dependent on volunteers to contribute and review updates. Many improvements are already completed and should not be delayed further because we're not perfect yet.

* See #1493 (NB! Not up to date currently, but we're working to update it)

I'm in the process of working through the backlog of unhandled pull requests, there are quite a few. Any additional help in reviewing them would be appreciated.

**Reviewers & maintainers:** To clarify who's handling what, I'd like to ask reviewers to please add themselves as assignee on PRs where they feel comfortable leading the review. (Suggestions welcome for alternative ways of handling PRs)

> ![image](https://github.com/typeddjango/django-stubs/assets/137616/31e57889-cc71-43d0-b190-1651a7aebfe5)

intgr:
5.0.0 is out!

* Release notes: https://github.com/typeddjango/django-stubs/releases/tag/5.0.0
* django-stubs: https://pypi.org/project/django-stubs/5.0.0/#history
* django-stubs-ext: https://pypi.org/project/django-stubs-ext/5.0.0/#history

PR #2183, opened 2024-05-25
Prepare for a release `5.0.1`

flaeppe:
There's already a bunch of contributions since `5.0.0`, I was thinking it might be time to release a `5.0.1`.

I've updated some numbers in preparation here, let me know if you know anything more that we should prepare before creating a new release.

## Related issues

- Ref: https://github.com/typeddjango/django-stubs/pull/2133#issuecomment-2104855760

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ We rely on different `django` and `mypy` versions:
 
 | django-stubs   | Mypy version | Django version | Django partial support | Python version |
 |----------------|--------------|----------------|------------------------|----------------|
+| 5.0.1          | 1.10.x       | 5.0            | 4.2                    | 3.8 - 3.12     |
 | 5.0.0          | 1.10.x       | 5.0            | 4.2, 4.1               | 3.8 - 3.12     |
 | 4.2.7          | 1.7.x        | 4.2            | 4.1, 3.2               | 3.8 - 3.12     |
 | 4.2.6          | 1.6.x        | 4.2            | 4.1, 3.2               | 3.8 - 3.12     |
--- a/ext/setup.py
+++ b/ext/setup.py
@@ -13,7 +13,7 @@ dependencies = [
 # It's fine to skip django-stubs-ext releases, but when doing a release, update this to newest django-stubs version.
 setup(
     name="django-stubs-ext",
-    version="5.0.0",
+    version="5.0.1",
     description="Monkey-patching and extensions for django-stubs",
     long_description=readme,
     long_description_content_type="text/markdown",
@@ -38,7 +38,6 @@ setup(
         "Programming Language :: Python :: 3.12",
         "Typing :: Typed",
         "Framework :: Django",
-        "Framework :: Django :: 4.1",
         "Framework :: Django :: 4.2",
         "Framework :: Django :: 5.0",
     ],
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ extras_require = {
 
 setup(
     name="django-stubs",
-    version="5.0.0",
+    version="5.0.1",
     description="Mypy stubs for Django",
     long_description=readme,
     long_description_content_type="text/markdown",
@@ -69,7 +69,6 @@ setup(
         "Programming Language :: Python :: 3.12",
         "Typing :: Typed",
         "Framework :: Django",
-        "Framework :: Django :: 4.1",
         "Framework :: Django :: 4.2",
         "Framework :: Django :: 5.0",
     ],


sobolevn (commented):
I would like to include these two PRs:
- https://github.com/typeddjango/django-stubs/pull/2182
- https://github.com/typeddjango/django-stubs/pull/2174

flaeppe:
@sobolevn the 2 PR:s you mentioned have been merged. What do you think, should we proceed with doing a release?

sobolevn (approved):
:rocket:

flaeppe:
Haha @sobolevn the `django-stubs` release worked this time (yay!) but it seems that `django-stubs-ext` publisher didn't work.

Action run: https://github.com/typeddjango/django-stubs/actions/runs/9253970455/job/25454746769

Fortunately we can just retry it if we update the publisher.

Does the publisher you set here: https://github.com/typeddjango/django-stubs/pull/2150#pullrequestreview-2051425119 look like below?

```
Owner: typeddjango
Repository name: django-stubs
Workflow name: release.yml
Environment name: release-ext
```

intgr:
Sorry, I didn't have time to review, I was away all weekend and had a busy Monday. But everything looks great to me 👍

I can also release `django-stubs-ext` manually, @flaeppe should I do it?

flaeppe:
@intgr no problem and yes, go ahead and do a manual one. I saw that there was some `ext/` changes that we want released. Would be great if you could push that out, thank you!

intgr:
Actually, you missed one thing. The `django-stubs-ext>=5.0.0` requirement should be updated every time. We don't want to support arbitrary combinations of `django-stubs` and `django-stubs-ext`. https://github.com/typeddjango/django-stubs/pull/2104 is a sensitive change in this regard.

Let's make a quick 5.0.2 release which bumps the dependency?

flaeppe:
Sounds good! But I'm wondering, can that requirement be updated before we've done a release of `django-stubs-ext`?

intgr:
Yes, the `django-stubs` upload to PyPI will pass even if `django-stubs-ext` isn't released yet.

intgr:
But it's a good point, we should tweak the `release.yml` workflow, such that `django-stubs-ext` release is done first. Otherwise users may get surprising dependency errors between the releases.

intgr:
> Actually, you missed one thing. The django-stubs-ext>=5.0.0 requirement should be updated every time.

And it's not at all your fault. We should have a checklist of what needs to be done for each release, written down somewhere.

flaeppe:
I'm not sure if it's a good idea to create a dependency in the release workflow? A problem that then shows up is that if e.g. `django-stubs-ext` succeeds but `django-stubs` fails, it's not possible to rerun the release action.

I suppose it then becomes a measurement of which problem is worse, since I don't think we can support both, at least not in a simple way.

> And it's not at all your fault. We should have a checklist of what needs to be done for each release, written down somewhere.

I agree that we should try to write it down, but I should've figured it out. I was looking through the PR:s marked with a "release notes reminder".

I think we should continue with an open issue for the next release (e.g. like #2020). I liked that approach, and then we can try to update that with a checklist and other stuff relevant for the next release.

flaeppe:
@intgr I've opened #2186

intgr:
> I'm not sure if it's a good idea to create a dependency in the release workflow? A problem that then shows up is that if e.g. django-stubs-ext succeeds but django-stubs fails, it's not possible to rerun the release action.
>
> I suppose it then becomes a measurement of which problem is worse, since I don't think we can support both, at least not in a simple way.

I think we can control this with [the workflow `concurrency` key](https://docs.github.com/en/actions/using-jobs/using-concurrency)?

```yaml
concurrency:
  group: release
```

Then both release jobs will be run in the same concurrency group and are ran in sequence. And we need to re-order them such that `django-stubs-ext` starts first.

flaeppe:
I don't think we should add any concurrency key, because; if a _different_ workflow run with the same key is started, any already running is shut down. That means we could possibly be halfway through releasing and if a new one is started, GitHub actions would cancel the first one halfway through.

I think we rather want to declare [needs](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#defining-prerequisite-jobs) for `django-stubs`:

```diff
jobs:
  build-and-publish:
    runs-on: ubuntu-latest
    environment:
      name: release
      url: https://pypi.org/p/django-stubs
+   needs: [build-and-publish-ext]
    permissions:
      id-token: write
    steps:
      - name: Setup python to build package
        uses: actions/setup-python@v5
        with:
          python-version: '3.12'
      - name: Install build
        run: python -m pip install build
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
      - name: Build package
        run: python -m build
      - name: Publish to PyPI
        uses: pypa/gh-action-pypi-publish@v1.8.14
```

That should force sequential runs while allowing us to retry the `build-and-publish` job _only_ if e.g. it fails

intgr:
> if a different workflow run with the same key is started, any already running is shut down.

That's only if using the `cancel-in-progress: true` key, which defaults to false AFAICT.

> I think we rather want to declare [needs](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#defining-prerequisite-jobs) for django-stubs:

Sure, that sounds better.

flaeppe:
> That's only if using the cancel-in-progress: true key, which defaults to false AFAICT.

Ah, I missed that section in the docs 🙂 I was also completely oblivious to the meaning of the `cancel-in-progress` name.

We should do both the above mentioned things then! That would mean any _new_ workflow will wait for a currently running one and we will release `django-stubs-ext` before `django-stubs`.

Last addition is that we'll also need to resolve #2187 when we've put `django-release-ext` as a blocking job, to get everything working.

intgr:
> We should do both the above mentioned things then! That would mean any new workflow will wait for a currently running one

I think that's not necessary, we don't make releases so often that the workflows can run in parallel. 😄

flaeppe:
> I think that's not necessary, we don't make releases so often that the workflows can run in parallel. 😄

That's true! But it's very easy for 2 persons to press e.g. "Rerun failing jobs" for a failed release workflow.

PR #2185, opened 2024-05-26
Don't crash when inspecting classes loaded from cache

### LINKED ISSUE #2184: attribute 'bases' of 'TypeInfo' undefined when running mypy after adding a model
# Bug report

Hi, I'm creating a [mypy plugin that wraps django-stubs](https://extended-mypy-django-plugin.readthedocs.io/en/latest/) so I can work with Abstract models in a way that lets us upgrade past mypy 1.4. I'm writing tests in there that run mypy after modifications and I came across an interesting corner case.

## What's wrong

It seems under some special circumstance, django-stubs can end up with this error

```
  File "mypy/semanal.py", line 1825, in apply_class_plugin_hooks
  File "/Users/stephen.moore/Projects/testing/django-stubs-bug/example-django-stubs-bug-20240526/.venv/lib/python3.10/site-packages/mypy_django_plugin/transformers/models.py", line 954, in process_model_class
    initializer_cls(ctx, django_context).run()
  File "/Users/stephen.moore/Projects/testing/django-stubs-bug/example-django-stubs-bug-20240526/.venv/lib/python3.10/site-packages/mypy_django_plugin/transformers/models.py", line 667, in run
    for statement in self.statements():
  File "/Users/stephen.moore/Projects/testing/django-stubs-bug/example-django-stubs-bug-20240526/.venv/lib/python3.10/site-packages/mypy_django_plugin/transformers/models.py", line 643, in statements
    for base in model.info.bases:
AttributeError: attribute 'bases' of 'TypeInfo' undefined
```

## System information

- OS: OSX/Linux
- `python` version: 3.10
- `django` version: 5.0.0 
- `mypy` version: 1.10.0
- `django-stubs-ext` version: 5.0.0

My project also tests against mypy-1.4/django-stubs-4.2.3 and it appears to only be in the newer version of django-stubs

## Reproducing

I've created a script here to reproduce the bug https://gist.github.com/delfick/46bac4f1609e866f01385f77f639d39d. Gist also includes the output it creates. Essentially bootstraps a virtualenv and creates a very basic django project (where one app creates concrete models using an abstract model defined in a seperate app), runs mypy, adds a model and runs mypy again. The second run produces the above exception. if mypy is not run before the additional model is added (or if run with `--no-incremental`) then it works fine.

I found it when I wrote [this test](https://github.com/delfick/extended-mypy-django-plugin/blob/4c0e780efc465852898854d3321e8dac96be8e51/tests/test_concrete_annotations.py#L224-L352). My plugin wraps django-stubs and so I'm able to [put in this "fix"](https://github.com/delfick/extended-mypy-django-plugin/blob/bdc8cbde2963e40a845705726ae73e52f9ce5687/extended_mypy_django_plugin/plugin/_plugin.py#L146-L203) that appears to work in the meantime where I trigger the exception and catch it so I can find and populate the missing information before passing control back to django-stubs itself.

I'm wondering if you were able to find it easier than me to understand why this happens and if there was a better fix?

Thanks!

flaeppe:
`ClassDef`s are populated with faked `TypeInfo` when loaded from cache. A faked `TypeInfo` doesn't have all attributes of a `TypeInfo` set. So we'll avoid inspecting the faked `TypeInfo`

## Related issues

Closes #2184

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -14,6 +14,7 @@ from mypy.nodes import (
     CallExpr,
     Context,
     Expression,
+    FakeInfo,
     NameExpr,
     RefExpr,
     Statement,
@@ -640,6 +641,10 @@ class ProcessManyToManyFields(ModelClassInitializer):
         while model_bases:
             model = model_bases.popleft()
             yield from model.defs.body
+            if isinstance(model.info, FakeInfo):
+                # While loading from cache ClassDef infos are faked and 'FakeInfo' doesn't have
+                # all attributes of a 'TypeInfo' set. See #2184
+                continue
             for base in model.info.bases:
                 # Only produce any additional statements from abstract model bases, as they
                 # simulate regular python inheritance. Avoid concrete models, and any of their


sobolevn (approved):
Thanks!

Issue #2187, opened 2024-05-27
Configure PyPI trusted publisher for `django-stubs-ext` 


flaeppe:
PR #2150 introduced automated releases of `django-stubs-ext` https://github.com/typeddjango/django-stubs/actions/runs/9254742886 but PyPI complained that the trusted publisher is invalid. Perhaps it was misconfigured?

Before any next release we should check and adjust the trusted publishing for `django-stubs-ext` at PyPI to correspond with our release workflow.

Below is a print of the output from the action run:

<img width="671" alt="Screenshot 2024-05-27 at 14 30 25" src="https://github.com/typeddjango/django-stubs/assets/3703560/857a9fc8-02d8-43b3-8505-b646c653305d">

---

Ref: #2183

sobolevn:
Done! 
<img width="444" alt="Снимок экрана 2024-05-28 в 09 15 49" src="https://github.com/typeddjango/django-stubs/assets/4660275/377efa25-0927-4eb3-9943-67f8629bb667">

flaeppe:
Superb :ok_hand:, thank you!

PR #2188, opened 2024-05-27
Release workflow: publish `django-stubs-ext` before `django-stubs`

flaeppe:
Also set up a concurrency group, per release name, so that we don't run releases in parallel.

Refs:
- https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=published#release
- https://docs.github.com/en/actions/using-jobs/using-concurrency
- #2183 
- From now on we're depending on solving #2187 to get _any_ automatic releasing

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,29 +4,10 @@ on:
   release:
     types: [published]
 
-jobs:
-  build-and-publish:
-    runs-on: ubuntu-latest
-    environment:
-      name: release
-      url: https://pypi.org/p/django-stubs
-    permissions:
-      id-token: write
-    steps:
-      - name: Setup python to build package
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.12'
-      - name: Install build
-        run: python -m pip install build
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -4,12 +4,15 @@ on:
   release:
     types: [published]
 
+concurrency:
+  group: release-${{ github.event.release.tag_name }}
```
flaeppe, line 8: I added concurrency per tag name, since that's what's unique towards PyPI.

.github/workflows/release.yml
```diff
@@ -44,9 +50,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Build ext package
-        run: python -m build ext/
-      - name: Publish ext to PyPI
+      - name: Build package
+        run: python -m build
+      - name: Publish to PyPI
```
flaeppe, line 55: Diff looks like it does because the jobs switched place. Felt more natural to swap place due to the new dependency

(I highlighted all lines but GitHub didn't want to. In case someone wonders why the context is really off for this comment)

intgr (approved):
Looks good to me, thanks!

PR #2191, opened 2024-05-28
Expect a `QuerySet` for multiple defaults in `MultipleObjectMixin`

### LINKED ISSUE #1928: `ListView.get_queryset()` cannot be treated as a queryset
# Bug report

## What's wrong

The return type of `get_queryset` on a `ListView` seems to be `_SupportsPagination[Any]`, which is throwing an error when I try and add a `select_related` to the queryset. 

Apologies if I'm missing something here (entirely possible!), but I've been going back and forth. I can see it's coming from the `MultipleObjectMixin`. 

https://github.com/typeddjango/django-stubs/blob/d283970c1f816994f54e1e90c9a7c0752c79668e/django-stubs/views/generic/list.pyi#L21

```
class MyModelList(ListView): 
    ...
    
    def get_queryset(self): -> QuerySet[Any]:
        qs = super().get_queryset().select_related("city")
```

There's nothing overly complicated about the model setup here, I just want to add the select related to the base queryset.

## How is that should be

`super().get_queryset()` should (probably?) not throw a type error when I treat it as a queryset in a `ListView`

## System information

- OS: Ubuntu 20.04.6 LTS (5.15.133.1-microsoft-standard-WSL2 under Windows 11)
- `python` version: 3.11.5
- `django` version: 5.0.1
- `mypy` version: 1.8.0

flaeppe:
When `MultipleObjectMixin` adds a more specific type than "something that can be sliced" (i.e. `_SupportsPagination`) for default types of queryset attributes/arguments a couple of subclass views gets a better default behaviour.

The changes here go under the same argument as in https://github.com/typeddjango/django-stubs/pull/2174#issuecomment-2124394572

## Related issues

- Closes #1928 
- Refs #2174

### DIFF
--- a/django-stubs/views/generic/list.pyi
+++ b/django-stubs/views/generic/list.pyi
@@ -2,7 +2,7 @@ from collections.abc import Sequence
 from typing import Any, Generic, TypeVar
 
 from django.core.paginator import Page, Paginator, _SupportsPagination
-from django.db.models import Model
+from django.db.models import Model, QuerySet
 from django.http import HttpRequest, HttpResponse
 from django.views.generic.base import ContextMixin, TemplateResponseMixin, View
 
@@ -10,7 +10,7 @@ _M = TypeVar("_M", bound=Model, covariant=True)
 
 class MultipleObjectMixin(Generic[_M], ContextMixin):
     allow_empty: bool
-    queryset: _SupportsPagination[_M] | None
+    queryset: QuerySet[_M] | None
     model: type[_M] | None
     paginate_by: int | None
     paginate_orphans: int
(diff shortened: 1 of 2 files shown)


flaeppe:
@sobolevn and @intgr does any of you want to check this out?

sobolevn (approved):

sobolevn:
I think that the main reason for a protocol is that it can also paginate lists, tuples, etc: https://github.com/typeddjango/django-stubs/blob/590f033abb9f0b3d10360a975c17b851a8222b88/django-stubs/core/paginator.pyi#L14-L19

But, anyway

flaeppe:
Yes, definitely. But it conflicts with having access to builtin queryset methods e.g. `.filter`

PR #2195, opened 2024-05-29
Update `django.db.models.functions`: allow `Combinable` as parameter, add `Round` function

noamkush:
# I have made things!

This fixes the annotation for database functions which can take `F` objects and field names, not just an `Expression`. Also added annotation for the specialized `Round` constructor.

* Closes #2196


intgr:
Per stubtest output, please also remove these entries from allowlist:

```
note: unused allowlist entry django.db.models.functions.Round.__init__
note: unused allowlist entry django.db.models.functions.math.Round.__init__
```

noamkush:
Done

django-stubs/db/models/functions/datetime.pyi
```diff
@@ -41,14 +41,14 @@ class TruncBase(TimezoneMixin, Transform):
     tzinfo: Any
 
     def __init__(
-        self, expression: Expression, output_field: Field | None = ..., tzinfo: tzinfo | None = ..., **extra: Any
+        self, expression: Combinable | str, output_field: Field | None = ..., tzinfo: tzinfo | None = ..., **extra: Any
```
intgr, line 44: I'm not very familiar QuerySet operators/functions.

Most `Combinable`s seem to inherit from `Expression`. ~~Can you give an example of something that is `Combinable` but not `Expression`?~~

intgr: Also some PostgreSQL functions are typehinted as `BaseExpression | Combinable | str`. I wonder if there's a reason for that, or in practice all useful expressions already inherit from `Combinable`.

intgr: Sry, you already answered the first question in PR description. `F()` objects inherit from `Combinable` but not `Expression`.

intgr: > I wonder if there's a reason for that, or in practice all useful expressions already inherit from `Combinable`.

But unless someone can think of a use case for `BaseExpression`, I'm inclined to go with `Combinable | str`. Let's not complicate things unless we know it's necessary.

intgr (approved):
Thanks, looks good to me.

Leaving open for a bit, in case someone has thoughts on `BaseExpression`: https://github.com/typeddjango/django-stubs/pull/2195#discussion_r1618916679

mkurnikov:
Looks good to me, I checked, all possible `BaseExpression` children also inherit `Combinable`, either directly, or via `Expression` base class. 
I'd advise to add one test where you could pass every possible `Expression` child class to one of the places you've annotated as `Combinable | str`, so we won't miss this in the future. 
But it's up to you, could be too much to add every single case. Maybe not every single case, but just one per "unique inheritance chain", like one of the `LessThan` / `GreaterThan` and so on.

intgr:
Thanks for confirming the class hierarchy wrt `BaseExpression`.

I'm not against adding simple tests, but our usual policy is to not require tests for simple stubs, I'd just merge this as is. Once these hints are fixed, it's very unlikely they will regress again. (The bigger challenge is remembering to use the correct types when new classes are added, but that can't really be covered with current test infra)

PR #2167, opened 2024-05-16
5.0: Update `django.db.models.functions.datetime`

JaeHyuckSa:
# I have made things!

Update stubs for `django.db.models.functions.datetime` for Django 5.0.

- [x] `django.db.models.functions.datetime`
  - [x]  `django.db.models.functions.datetime.TruncBase.__init__`'s `is_dst` argument was removed
  - [x]  `django.db.models.functions.datetime.TruncBase.is_dst` was removed
  - [x]  `django.db.models.functions.datetime.Trunc.__init__`'s `is_dst` argument was removed

I have added the necessary `__init__ `and `as_sql` methods to accommodate the updates in Django 5.0.

## Related issues

Refs
- https://github.com/typeddjango/django-stubs/issues/1493

Upstream PR
- https://github.com/django/django/pull/16432

PR #2196, opened 2024-05-29
accept str for Trunc expression

asottile:
fixes regression in #2167


intgr:
This is also changed in another MR: https://github.com/typeddjango/django-stubs/pull/2195/files which also expands `Expression` to `Combinable`.

I think we can close this and proceed with #2195.

PR #2198, opened 2024-05-30
remove `as_<dbname>()` methods from stubs

mkurnikov:
I want to clean up `Lookup` inheritance chain to remove some false-positives. Most of the annotations there came from the automatic generation of stubs in the early days of this project. They're not user-facing, and could be removed. 

There's more PRs to come once this one is merged, I just want to make sure I understand the current process. 

@intgr @sobolevn


sobolevn (commented):
> They're not user-facing, and could be removed.

But, is there any real reason to do so?
They are public methods of public types.

They don't cause any trouble as far as I know.

mkurnikov:
You're probably right. Let me then to address the issue I wanted to fix in a separate PR.

PR #2199, opened 2024-05-30
`Lookup` should be a subtype of `Expression`

mkurnikov:
# I have made things!

Added test to trigger the type failure. 
Had to change signature for `set_source_expressions` to `Combinable | Expression`, and remove some inherited methods. 

@sobolevn


sobolevn (approved):

tests/typecheck/db/models/test_constraints.yml
```diff
@@ -35,3 +35,13 @@
...
         main:4: note:     .*
+
+-   case: model_level_check_contraint_should_allow_lookups
+    main: |
+        from django.db.models import CheckConstraint, F
+        from django.db.models.lookups import LessThan
+
+        CheckConstraint(
+            name="less_than_constraint",
+            check=LessThan(F("months"), 1)
```
intgr, line 46: Since mypy typecheck tests are slow, and we also want to test with Pyright, such tests are better added under `tests/assert_type`

mkurnikov: Thanks, I should've waited for your review before merging.
I'll fix it in a follow-up.

intgr: No problem, this is more of a minor quibble. Follow-up PR is welcome. :)

Welcome back to django-stubs development @mkurnikov! :tada:

intgr: FWIW @flaeppe already changed this https://github.com/typeddjango/django-stubs/pull/2204

mkurnikov: Great, thanks @flaeppe

PR #2204, opened 2024-06-04
Tests: Migrate `db/models/test_constraints.yml` to 'assert_type' tests

flaeppe:
Ref: https://github.com/typeddjango/django-stubs/pull/2199#discussion_r1621550006


intgr (approved):
Thanks!

PR #2205, opened 2024-06-04
Remove generic argument of `django.db.models.lookups.Lookup`

flaeppe:
I might be missing something obvious here but I don't see what purpose the generic argument has or what it intends to control.


flaeppe:
Aha, I think I found it under `mypy_django_plugin.django.context.resolve_lookup_expected_type` in the Mypy plugin:

https://github.com/typeddjango/django-stubs/blob/afbbf37a87b9ff0e35d4893d17765475133fedd1/mypy_django_plugin/django/context.py#L494-L505

But I currently think it looks a bit off with a generic argument on the `Lookup` type. I'm guessing it instead should always prefer to be based on the "get type" of the field.

UnknownPlatypus:
Superseded by #3235

PR #2209, opened 2024-06-06
Tests: Migrate `db/test_transaction.yml` to 'assert_type' tests

flaeppe:


tests/assert_type/db/test_transaction.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+assert_type(func1, Callable[[int], list[int]])
+
+
+@atomic(using="bla", savepoint=False)
+def func2(x: int, /) -> list[int]: ...
+
+
+assert_type(func2, Callable[[int], list[int]])
```
sobolevn: this does not differentiate `(x: int)` and `(x: int, /)`. A Protocol, maybe?

flaeppe: `assert_type` doesn't work without forced positional, as it doesn't seem to be possible to do something like:

```python
assert_type(func2, Callable[[Arg(int, "x")], list[int]]
```

flaeppe: Ah, alright, lets just avoid `assert_type` for the Callables and it should be fine I think. Could update to something like:

```python
x: Callable[[int], list[int]] = func2
```

Ref: https://github.com/python/mypy/issues/13935

tests/assert_type/db/test_transaction.py
```diff
@@ -0,0 +1,33 @@
...
+
+from django.db.transaction import atomic, non_atomic_requests
+from django.http import HttpRequest, HttpResponse
+
+
+@atomic
+def func1(x: int) -> List[int]: ...
+
+
+x1: Callable[[int], List[int]] = func1
```
intgr, line 12: If due to some stubs issue, mypy thinks `@atomic` turns the function into `Any`, then this check will ignore and won't catch the issue. :(

flaeppe: Crap, that's true

intgr: It's an ugly hack, but we can fix it if we add both a positive and negative "assignment" test

```python
x1: Callable[[int], List[int]] = func1
y1: Callable[[], None] = func1  # type: ignore[...]     # <-- Expects an error to occur, but error will disappear with Any
```

flaeppe: @intgr hm, couldn't the [disallow_any_decorated](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_any_decorated) setting save us from all this happening?

flaeppe: Never mind, I read that setting as if it disallowed the return type of the decorator to be `Any`. But it's even stricter; it disallows `Any` from being anywhere in the returned signature

flaeppe: Instead of assigning to a different type I called the callable with wrong types e.g.

```python
x1: Callable[[int], List[int]] = func1
func1("str")
```

Which yields an error to suppress too

sobolevn: (that's why I don't really like `assert_type` tests 😄)

flaeppe: Yeah it's tedious, but it felt fair enough to invoke with invalid types. Quite small addition as a workaround

UnknownPlatypus:
Superseded by #3231

PR #2210, opened 2024-06-06
Allow any function for admin's `display` and `action` decorators

flaeppe:
I realised that:
- The `display` decorator isn't limited to any specific type of callable, it accepts anything really.
- The `action` decorator was trying to dictate a callable type that the `ModelAdmin.actions` expects. But that's the responsibility of `ModelAdmin`. And `action` also accepts anything.

Refs:
- [The runtime call site for admin actions]( https://github.com/django/django/blob/adae619426b6f50046b3daaa744db52989c9d6db/django/contrib/admin/options.py#L1664) (which displays what is passed to an action function)
- Future is today; when `Self` is possible to use for actions -- ref: https://github.com/typeddjango/django-stubs/pull/1292#discussion_r1048313390

### DIFF
--- a/django-stubs/contrib/admin/decorators.pyi
+++ b/django-stubs/contrib/admin/decorators.pyi
@@ -3,45 +3,33 @@ from typing import Any, TypeVar, overload
 
 from django.contrib.admin import ModelAdmin
 from django.contrib.admin.sites import AdminSite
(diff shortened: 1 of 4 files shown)


sobolevn (approved):
Thanks!

PR #2218, opened 2024-06-13
Model._meta.pk is non-None

asottile:
django internals frequently reference (for example [this](https://github.com/django/django/blob/e99187e5c94516ee35f37cc41a36d906b395808d/django/db/models/base.py#L655)) this as if it is always a field and looking at the source it is always set to some field during _prepare:

- https://github.com/django/django/blob/e99187e5c94516ee35f37cc41a36d906b395808d/django/db/models/options.py#L302-L320
    - first branch through `setup_pk`: https://github.com/django/django/blob/e99187e5c94516ee35f37cc41a36d906b395808d/django/db/models/options.py#L361-L362
    - second branch through:
        - add_to_class: https://github.com/django/django/blob/e99187e5c94516ee35f37cc41a36d906b395808d/django/db/models/base.py#L371
        - contribute_to_class: https://github.com/django/django/blob/e99187e5c94516ee35f37cc41a36d906b395808d/django/db/models/fields/__init__.py#L951
        - add_field: https://github.com/django/django/blob/e99187e5c94516ee35f37cc41a36d906b395808d/django/db/models/options.py#L337
        - (and then again setup_pk)

### DIFF
--- a/django-stubs/db/models/options.pyi
+++ b/django-stubs/db/models/options.pyi
@@ -66,7 +66,7 @@ class Options(Generic[_M]):
     required_db_features: list[str]
     required_db_vendor: Literal["sqlite", "postgresql", "mysql", "oracle"] | None
     meta: type | None
-    pk: Field | None
+    pk: Field
     auto_field: AutoField | None
     abstract: bool
     managed: bool


flaeppe (approved):

intgr:
This is incorrect. Unsaved models have `pk` set to `None`.

```python
In [1]: from django.contrib.auth.models import User

In [2]: new_user = User()

In [3]: print(new_user.pk)
None
```

And I have used this in several places in my code to distinguish between saved and unsaved models. I think this change should be reverted. But it does seem helpful to add a code comment here, to explain this situation.

intgr:
Oh, sorry for jumping the gun, I did not notice that the change was involving the `Options` class, not `Model`. Never mind then 😅  🤦

PR #2219, opened 2024-06-13
cache's version can be Any

asottile:
matches the annotation with default_key_func

in practice, we use a hexdigest for cache versioning which is forbidden by int but works fine at runtime


django-stubs/core/cache/backends/base.pyi
```diff
@@ -20,50 +20,50 @@ class BaseCache:
...
     def clear(self) -> None: ...
     async def aclear(self) -> None: ...
-    def incr_version(self, key: Any, delta: int = ..., version: int | None = ...) -> int: ...
-    async def aincr_version(self, key: Any, delta: int = ..., version: int | None = ...) -> int: ...
-    def decr_version(self, key: Any, delta: int = ..., version: int | None = ...) -> int: ...
-    async def adecr_version(self, key: Any, delta: int = ..., version: int | None = ...) -> int: ...
+    def incr_version(self, key: Any, delta: int = ..., version: Any | None = ...) -> int: ...
+    async def aincr_version(self, key: Any, delta: int = ..., version: Any | None = ...) -> int: ...
+    def decr_version(self, key: Any, delta: int = ..., version: Any | None = ...) -> int: ...
+    async def adecr_version(self, key: Any, delta: int = ..., version: Any | None = ...) -> int: ...
```
flaeppe, line 66: If `version` argument is set to `Any`, I think increment/decrement should return `Any`(?) and `delta` should perhaps be the same type as `version` (except `None`).

It sure looks like there could be a `TypeVar` here?

flaeppe (commented):
In a way it looks like `BaseCache` could be generic on `version` with a default as `int`. But I suspect having that might be an inconvenience too

django-stubs/core/cache/backends/base.pyi
flaeppe: I'm not totally sure about this but since Django e.g. adds `delta` to `version` at the end of `incr_version` like:

```python
return version + delta
```

Correct me if I'm wrong, but I suppose there could be a case of declaring something like

```python
def incr_version(self, key: Any, delta: SupportsAdd[T, T] = ..., version: SupportsAdd[T, T] | None = ...) -> T: ...
```

`SupportsAdd` is declared in the `_typeshed` module and can be imported under `if TYPE_CHECKING:`

asottile: I guess with decr_version existing this can't work as written

so this patch is probably dead

PR #2227, opened 2024-06-21
Make `ReverseManyToOneDescriptor` generic over a model

flaeppe:
This allows us to remove some plugin code, as we now can fall back to the default related manager on the descriptor instead of having to implement the fallback behaviour manually.

## Related issues

Refs: https://github.com/typeddjango/django-stubs/pull/2214#discussion_r1632216834


sobolevn (commented):
Should we add this to `ext` to be monkeypatched? Or is it an internal tool only?

flaeppe:
Good point, we should patch `ReverseManyToOneDescriptor` with `__class_getitem__`

sobolevn (approved):

Issue #2224, opened 2024-06-18
`(ManagerClass).from_queryset` is not creating a valid base class


asottile-sentry:
# Bug report

so I realize this is not the best of reports -- all attempts I've made to make a minimal case do not reproduce the bug so I suspect it is something to do with how our repository is set up somehow.

I have thrown together a reproducible docker image to show the problem:

```dockerfile
FROM python:3.11-slim

RUN : \
    && apt-get update \
    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends dumb-init git \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/*

WORKDIR /src

RUN : \
    && git clone https://github.com/getsentry/sentry --branch master . \
    && git checkout 0bcc06037ce710ff3a97346f8d956766c4cb4787

ENV PATH=/src/.venv/bin:$PATH PIP_NO_CACHE_DIR=1
RUN : \
    && python3 -m venv .venv \
    && pip install -r requirements-dev-frozen.txt \
    # ensure we reproduce the problem without our fork
    && pip uninstall -y sentry-forked-django-stubs \
    && pip install django-stubs==5.0.2 \
    && python3 -m tools.fast_editable --path .

RUN sentry init

# apply the "patch" which should allow it to have queryset methods
RUN : \
    && sed  -i \
        's/class BaseManager.*$/_base = DjangoBaseManager.from_queryset(BaseQuerySet)\nclass BaseManager(_base[M]):/g' \
        src/sentry/db/models/manager/base.py \
    && echo 'from typing import TYPE_CHECKING' | tee -a src/sentry/db/models/manager/base.py \
    && echo 'if TYPE_CHECKING: reveal_type(BaseManager.using_replica)' | tee -a src/sentry/db/models/manager/base.py

ENTRYPOINT ["dumb-init", "--"]
CMD ["mypy", "src/sentry/db/models/manager/base.py"]
```

after the patch is applied [this line here](https://github.com/getsentry/sentry/blob/9ebd80d38e5c2a905090ad63623d7bee65a746ec/src/sentry/db/models/manager/base.py#L47) should look like this (the dockerfile does this automatically):

```python
_base = DjangoBaseManager.from_queryset(BaseQuerySet)
class BaseManager(_base[M]):
```

and the end of the file should have:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING: reveal_type(BaseManager.using_replica)
``` 

[...]

PR #2228, opened 2024-06-21
add PlaceholderNode when deferring manager class hook

asottile:
see #2224

### DIFF
--- a/tests/typecheck/managers/querysets/test_from_queryset.yml
+++ b/tests/typecheck/managers/querysets/test_from_queryset.yml
@@ -847,6 +847,41 @@
                     self.ttl = ttl
                     super().__init__(*a, **k)
 
+-   case: test_from_queryset_with_deferral
+    main: |
+        from myapp.models import Concrete
+        reveal_type(Concrete.objects.qs_meth())  # N: Revealed type is "builtins.int"
+        reveal_type(Concrete.objects.manager_meth())  # N: Revealed type is "builtins.str"
+    mypy_config: |
+        [mypy.plugins.django-stubs]
+        django_settings_module = myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from typing import TypeVar, ClassVar
+                from typing_extensions import Self
+                from django.db import models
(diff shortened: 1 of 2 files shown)


asottile:
ok so this passes the current tests -- but how do I demonstrate the improvement is tricky.  I need to somehow make a circular Manager <-> QuerySet ?  would appreciate some help on writing a test for this :S

tests/typecheck/managers/querysets/test_from_queryset.yml
```diff
@@ -847,6 +847,41 @@
...
                     super().__init__(*a, **k)
 
+-   case: test_from_queryset_with_deferral
+    main: |
+        from myapp.models import Concrete
+        reveal_type(Concrete.objects.qs_meth())  # N: Revealed type is "builtins.int"
+        reveal_type(Concrete.objects.manager_meth())  # N: Revealed type is "builtins.str"
+    mypy_config: |
+        [mypy.plugins.django-stubs]
+        django_settings_module = myapp
```
asottile, line 857: this is needed to skip the importing of models -- since the forced-deferral trick `models.py` triggers a `NameError` when actually imported

asottile (commented):
test originally failing with:

```
_______________________ test_from_queryset_with_deferral _______________________
/home/asottile/workspace/django-stubs/tests/typecheck/managers/querysets/test_from_queryset.yml:853: 
E   Actual:
E     main:2: note: Revealed type is "Any"          (diff)
E    main:3: note: Revealed type is "builtins.str"
E     myapp/models:15: error: Variable "myapp.models._base" is not valid as a type  [valid-type] (diff)
E     myapp/models:15: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases (diff)
E     myapp/models:15: error: Invalid base class "_base"  [misc] (diff)
E   Expected:
E     main:2: note: Revealed type is "builtins.int" (diff)
E    main:3: note: Revealed type is "builtins.str"
E     E: ...te: Revealed type is "builtins.int"
E     A: ...te: Revealed type is "Any"
```

flaeppe (approved):
Nice work. Great that you figured out a test case

PR #2230, opened 2024-06-22
Make plugin handle explicitly declared reverse descriptors for FKs

flaeppe:
If a reverse descriptor was explicitly declared on a model, the plugin skipped to create a more specific subclass for its related manager.

Refs: #2227

### DIFF
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -1035,6 +1035,36 @@
                 class Book(PrintedGood):
                     name = models.CharField()
 
+-   case: test_explicit_reverse_many_to_one_descriptor
+    main: |
+        from myapp.models import Other
+        reveal_type(Other.explicit_descriptor)  # N: Revealed type is "django.db.models.fields.related_descriptors.ReverseManyToOneDescriptor[myapp.models.MyModel]"
+        reveal_type(Other().explicit_descriptor)  # N: Revealed type is "myapp.models.MyModel_RelatedManager"
+        reveal_type(Other().explicit_descriptor.custom_method())  # N: Revealed type is "builtins.int"
+    installed_apps:
+        -   myapp
+    monkeypatch: true
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models/__init__.py
+            content: |
+                from typing import ClassVar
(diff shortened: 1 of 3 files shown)


tests/typecheck/fields/test_related.yml
```diff
@@ -1035,6 +1035,36 @@
...
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models/__init__.py
+            content: |
+                from typing import ClassVar
+                from django.db import models
+                from django.db.models.fields.related_descriptors import ReverseManyToOneDescriptor
+
+                class Other(models.Model):
+                    explicit_descriptor: ClassVar[ReverseManyToOneDescriptor["MyModel"]]
```
sobolevn, line 1056: When is this useful? Isn't it better to just write `= ReverseManyToOneDescriptor(...)`?

flaeppe: When you don't use mypy and also don't want to override anything that Django does during runtime.

flaeppe: Or rather: when you don't use our mypy plugin

sobolevn: Maybe we then need a `assert_type` test for this instead?

flaeppe: I suppose we could, but we'd have to make it assert to a different type depending on if we're running with mypy or pyright. Since the plugin creates a more specific type. It didn't feel very useful to be honest. But I can add it if you find that it could be useful

The test here was merely for the plugin, it failed initially so it was some form of regression. I wanted to ensure that the more specific type was set even though an explicit descriptor was declared. I discovered the plugin had an early exit that stopped the descriptor refinement

flaeppe: Actually, thinking twice about it. I don't think that mypy would output anything different, since it currently requires the runtime to output a more specialised type

flaeppe: FWIW I added an `assert_type` test

flaeppe:
What do you think? Does this look fine?

sobolevn (approved):

PR #2231, opened 2024-06-22
Do many related manager creation during semantic analysis

flaeppe:
When finding a `ManyToManyField` we can directly create 2 `ManyRelatedManager`s, one for each side of the relation. This makes the model declaring the `ManyToManyField` be the one who contributes the `ManyRelatedManager`s.

Trying to place some related code closer to each other. This time we move some code out of the type checking phase to semantic analysis.

Refs:
- #1805 
- #2026

### DIFF
--- a/scripts/stubtest/allowlist.txt
+++ b/scripts/stubtest/allowlist.txt
@@ -27,6 +27,13 @@ django.core.files.storage.default_storage
 django.contrib.admin.models.LogEntry_RelatedManager
 django.contrib.auth.models.Permission_RelatedManager
 
+# '<Model>_ManyRelatedManager' entries are plugin generated and these subclasses only exist
+# _locally/dynamically_ runtime -- Created via
+# 'django.db.models.fields.related_descriptors.create_forward_many_to_many_manager'
+django.contrib.auth.models.Group_ManyRelatedManager
+django.contrib.auth.models.Permission_ManyRelatedManager
+django.contrib.auth.models.User_ManyRelatedManager
+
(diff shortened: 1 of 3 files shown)


scripts/stubtest/allowlist.txt
```diff
@@ -27,6 +27,13 @@ django.core.files.storage.default_storage
 django.contrib.admin.models.LogEntry_RelatedManager
 django.contrib.auth.models.Permission_RelatedManager
 
+# '<Model>_ManyRelatedManager' entries are plugin generated and these subclasses only exist
+# _locally/dynamically_ runtime -- Created via
+# 'django.db.models.fields.related_descriptors.create_forward_many_to_many_manager'
+django.contrib.auth.models.Group_ManyRelatedManager
+django.contrib.auth.models.Permission_ManyRelatedManager
+django.contrib.auth.models.User_ManyRelatedManager
+
```
flaeppe, line 36: These are only here since the stubs includes `= ManyToManyField(...)` in the following places:

https://github.com/typeddjango/django-stubs/blob/ec37d06ef630f7bc78a9df3a3cfd9030566655ad/django-stubs/contrib/auth/models.pyi#L34-L38

https://github.com/typeddjango/django-stubs/blob/ec37d06ef630f7bc78a9df3a3cfd9030566655ad/django-stubs/contrib/auth/models.pyi#L59-L62

Once they would move to a type hint form like: `field: ManyToManyField[..., ...]` they will disappear from the allowlist.

sobolevn (approved):

PR #2233, opened 2024-06-24
Expand CheckRegistry.register as function (#2232)

### LINKED ISSUE #2232: `CheckRegistry().register()` should allow `*tags` when called as function
# Bug report

The type hints for `CheckRegistry.register` do not allow passing `tags` when calling as a function:

https://github.com/typeddjango/django-stubs/blob/edfebe45c1626683b09ab2dffe62b4e131f46b48/django-stubs/core/checks/registry.pyi#L40-L47

## What's wrong

According to the [System check framework docs](https://docs.djangoproject.com/en/4.2/topics/checks/), these (annotated) variants should be equivalent:

```python
from typing import Any, Sequence, Optional, Union, List
from django.core.checks import register, Tags, messages


@register(Tags.security)
def my_check_decorated(
    app_configs: Union[Sequence[AppConfig], None], **kwargs: Any
) -> List[messages.Warning]:
    return []


def my_check(
    app_configs: Union[Sequence[AppConfig], None], **kwargs: Any
) -> List[messages.Warning]:
    return []


register(my_check, Tags.security)
```

However, the second version does not pass `mypy`:

```
example.py:20: error: No overload variant matches argument types "Callable[[Sequence[AppConfig] | None, KwArg(Any)], list[Warning]]", "str"  [call-overload]
example.py:20: note: Possible overload variants:
example.py:20: note:     def [_C <: _CheckCallable] register(self, _C, /) -> _ProcessedCheckCallable[_C]
example.py:20: note:     def register(self, *tags: str, **kwargs: Any) -> Callable[[_C], _ProcessedCheckCallable[_C]]
```

## How is that should be

The second version should be accepted. This way, `register()` can be called in `AppConfig.ready()`, such as in https://github.com/mozilla/django-csp/blob/main/csp/apps.py.

## System information

- OS:
- `python` version: 3.12.3
- `django` version: 5.0.6
- `mypy` version: 1.8.0 (compiled: yes)
- `django-stubs` version: 5.0.2
- `django-stubs-ext` version: 5.0.2

jwhitlock:
Allow setting tags and "deploy" flag when using register as a function. Previously, only the decorator overload allowed these. Closes #2232.


tests/typecheck/core/test_checks.yml
```diff
@@ -27,3 +27,45 @@
 
     @register()  # E: Value of type variable "_C" of function cannot be "Callable[[int], Sequence[CheckMessage]]"  [type-var]
     def wrong_args(bla: int) -> Sequence[CheckMessage]: ...
+- case: test_checks_register_as_function
```
sobolevn: Thanks! But tests for this feature is not required :)

jwhitlock: Does that mean I should remove the new tests before this PR is accepted?

sobolevn: Yes, please :)

sobolevn (approved):
Thanks!

PR #2238, opened 2024-06-26
Improve types of `BaseDatabaseOperations`

Viicos:
# I have made things!


intgr (approved):
Thanks! I found some more related inaccurate stubs too.

image

Possibly these methods could be removed from subclasses, if they have the same signature as the `BaseDatabaseOperations` base class.

Viicos:
Made some other improvements to remove some unneeded subclass overrides. There's more work needed here, but I don't want to process everything at once

django-stubs/db/backends/base/operations.pyi
```diff
@@ -66,16 +69,16 @@ class BaseDatabaseOperations:
...
     def sql_flush(
-        self, style: Any, tables: Sequence[str], *, reset_sequences: bool = ..., allow_cascade: bool = ...
+        self, style: Style, tables: Sequence[str], *, reset_sequences: bool = ..., allow_cascade: bool = ...
     ) -> list[str]: ...
-    def execute_sql_flush(self, sql_list: Iterable[str]) -> None: ...
-    def sequence_reset_by_name_sql(self, style: Style | None, sequences: list[Any]) -> list[Any]: ...
-    def sequence_reset_sql(self, style: Style, model_list: Sequence[type[Model]]) -> list[Any]: ...
+    def execute_sql_flush(self, sql_list: list[str]) -> None: ...
+    def sequence_reset_by_name_sql(self, style: Style, sequences: list[dict[str, str | None]]) -> list[str]: ...
+    def sequence_reset_sql(self, style: Style, model_list: list[type[Model]]) -> list[str]: ...
```
intgr, line 76: Since `sequence_reset_sql` only iterates over the `model_list`, I typing this as `list` seems too strict. `Sequence[]` or `Iterable[]` would be better.

Same for `execute_sql_flush`

django-stubs/db/backends/base/operations.pyi
```diff
@@ -66,16 +69,16 @@ class BaseDatabaseOperations:
...
     def start_transaction_sql(self) -> str: ...
     def end_transaction_sql(self, success: bool = ...) -> str: ...
-    def tablespace_sql(self, tablespace: str | None, inline: bool = ...) -> str: ...
-    def prep_for_like_query(self, x: str) -> str: ...
-    prep_for_iexact_query: Any
+    def tablespace_sql(self, tablespace: str, inline: bool = ...) -> str: ...
+    def prep_for_like_query(self, x: object) -> str: ...
+    def prep_for_iexact_query(self, x: object) -> str: ...
```
intgr, line 81: Nit: I think `Any` has been preferred over `object` recently.

intgr (changes requested):
Thanks! Few suggestions.

django-stubs/db/backends/base/operations.pyi
Viicos: I went for `list` as these methods receive an actual `list` at runtime (e.g. `sequence_reset_sql` receives the result of `sql_flush`), so I assumed it is safe to allow `BaseDatabaseOperations` subclasses to rely on the available attributes/methods of `list` in these methods. On the other hand, keeping `Sequence` seems fine as well

PR #2239, opened 2024-06-26
Add attributes set by application/middleware to `HttpRequest`

bigfootjon:
# I have made things!

Ref the docs: https://docs.djangoproject.com/en/5.0/topics/http/urls/#how-django-processes-a-request

> Django determines the root URLconf module to use. Ordinarily, this is the value of the ROOT_URLCONF setting, but if the incoming HttpRequest object has a urlconf attribute (set by middleware), its value will be used in place of the ROOT_URLCONF setting.

## Related issues

N/A

### DIFF
--- a/django-stubs/http/request.pyi
+++ b/django-stubs/http/request.pyi
@@ -49,6 +49,7 @@ class HttpRequest(BytesIO):
     resolver_match: ResolverMatch | None
     content_type: str | None
     content_params: dict[str, str] | None
+    urlconf: str | None
     _body: bytes
     _stream: BinaryIO
     # Attributes added by optional parts of Django
--- a/django-stubs/http/request.pyi
+++ b/django-stubs/http/request.pyi
@@ -54,6 +54,7 @@ class HttpRequest:
     resolver_match: ResolverMatch | None
     content_type: str | None
     content_params: dict[str, str] | None
+    urlconf: str | None
     _body: bytes
     _stream: BinaryIO
     # Attributes added by optional parts of Django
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
I am not sure what to do here.

1. It is indeed assigned in a middleware
2. and used in several places: https://github.com/django/django/blob/e56a32b89bb7fadffdfaa2cdf12b4863ccd5af9b/django/middleware/locale.py#L20 and https://github.com/django/django/blob/e56a32b89bb7fadffdfaa2cdf12b4863ccd5af9b/django/middleware/common.py#L68
3. But, it is not a part of `HttpRequest`'s public API: https://github.com/django/django/blob/e56a32b89bb7fadffdfaa2cdf12b4863ccd5af9b/django/http/request.py

CC @adamchainz

ngnpope:
This is documented here:

https://docs.djangoproject.com/en/stable/ref/request-response/#attributes-set-by-application-code

I would suggest adding all of these and grouping them together with a link to the documentation.

Similarly, it's probably worth checking these and grouping them together also:

https://docs.djangoproject.com/en/stable/ref/request-response/#attributes-set-by-middleware

sobolevn:
what about adding these properties as `self.urlconf = None` to `HttpRequest.__init__`?

adamchainz:
> what about adding these properties as `self.urlconf = None` to `HttpRequest.__init__`?

`None` isn't a valid value - Django uses `hasattr()` to check for them...

UnknownPlatypus:
I've implemented the suggestion by  @ngnpope https://github.com/typeddjango/django-stubs/pull/2239#issuecomment-2194388277. 

I think this is a good middleground and makes our definition more consistant.
These attribute could not be there if the associated middleware/code is not used but we cannot statically know that so it's better to have a false negative than a false positive.

Also I expect people trying to interact with them to be working a project using the associated middleware

ngnpope (approved):
This seems like a good compromise.

Ideally we'd have something like `MaybeUndefined[...]` to flag up cases like this where we know we're abusing the safety. Alternatively, with `Intersection`, a `Protocol` could be used to mixin for cases where we know these are set.

PR #2240, opened 2024-06-27
don't copy magic or private methods in from_queryset managers

asottile:
failing before this change with:

```
_______________ queryset_inheritable_does_not_clobber_super_init _______________
/home/asottile/workspace/django-stubs/tests/typecheck/managers/querysets/test_from_queryset.yml:795: 
E   Actual:
E     myapp/models:19: error: Unable to resolve return type of queryset/manager method  [misc] (diff)
E   Expected:
E     (empty)
```

the error came from inspecting `super().__init__(...)` in the manager, trying to look up queryset's `__init__`


flaeppe (approved):
Ah, great

PR #2241, opened 2024-06-28
fix handling of explicit `objects` annotation

asottile:
test originally failing with:

```
_______________________ from_queryset_model_inheritance ________________________
/home/asottile/workspace/django-stubs/tests/typecheck/managers/querysets/test_from_queryset.yml:47: 
E   Actual:
E     myapp/models:10: error: Incompatible types in assignment (expression has type "MyManager[Base]", variable has type "MyManager[Self]")  [assignment] (diff)
E   Expected:
E     (empty)
```

we use this pattern at sentry to ensure the correct types for model classes which inherit from our base model

### DIFF
--- a/tests/typecheck/managers/querysets/test_from_queryset.yml
+++ b/tests/typecheck/managers/querysets/test_from_queryset.yml
@@ -43,6 +43,31 @@
                 class MyModelWithoutSelf(models.Model):
                     objects = ManagerWithoutSelf()
 
+-   case: from_queryset_model_inheritance
+    main: |
+        from myapp.models import Base, Sub
+        reveal_type(Base.objects)  # N: Revealed type is "myapp.models.MyManager[myapp.models.Base]"
+        reveal_type(Sub.objects)  # N: Revealed type is "myapp.models.MyManager[myapp.models.Sub]"
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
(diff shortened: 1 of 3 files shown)


tests/typecheck/managers/querysets/test_from_queryset.yml
```diff
@@ -43,6 +43,31 @@
...
+                from typing import ClassVar
+                from typing_extensions import Self
+                from django.db.models import Model
+                from django.db.models.manager import BaseManager
+                from django.db.models.query import QuerySet
+
+                MyManager = BaseManager.from_queryset(QuerySet, "MyManager")
+
+                class Base(Model):
+                    objects: ClassVar[MyManager[Self]] = MyManager()
```
flaeppe, line 66: Do you think we also need to check that the plugin is able to infer the manager type in a scenario like this?

e.g. also test

```python
class Base(Model):
    objects = MyManager()
```

I suppose there might be a bunch of other tests doing that, so it might just be a duplication

flaeppe (approved):

tests/typecheck/managers/querysets/test_from_queryset.yml
asottile: most of the tests do that already fortunately!

flaeppe: Alright, lets merge

PR #2242, opened 2024-06-30
Add atomic_blocks to base DB wrapper

palfrey:
# I have made things!

Add `atomic_blocks` as per https://github.com/django/django/blob/main/django/db/backends/base/base.py#L77 and https://github.com/django/django/blob/main/django/db/transaction.py#L222


sobolevn (approved):
Thanks!

PR #2244, opened 2024-07-01
walk mro for Field set type annotations

asottile:
test failed prior to this with:

```
_____________________ mro_is_walked_for_field_annotations ______________________
/home/asottile/workspace/django-stubs/tests/typecheck/models/test_create.yml:64: 
E   Actual:
E     main:2: error: Incompatible type for "dct" of "Child" (got "Dict[Any, Any]", expected "Union[str, Combinable]")  [misc] (diff)
E   Expected:
E     (empty)
```

test is a simplified example from sentry code

### DIFF
--- a/tests/typecheck/models/test_create.yml
+++ b/tests/typecheck/models/test_create.yml
@@ -60,6 +60,27 @@
                 class Child4(Child1):
                     value4 = models.IntegerField()
 
+-   case: mro_is_walked_for_field_annotations
+    main: |
+        from myapp.models import Child
+        Child.objects.create(dct={"hello": "world"})  # no errors
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from __future__ import annotations
+
+                from django.db import models
+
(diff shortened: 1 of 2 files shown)


tests/typecheck/models/test_create.yml
```diff
@@ -60,6 +60,27 @@
                 class Child4(Child1):
                     value4 = models.IntegerField()
 
+-   case: mro_is_walked_for_field_annotations
+    main: |
+        from myapp.models import Child
+        Child.objects.create(dct={"hello": "world"})  # no errors
```
sobolevn, line 66: Can you please add:

```yml
c: Child
reveal_type(c.dct)
```

?

This is needed to prevent a possible regression that `dct` is `Any`.

sobolevn (commented):
Good catch, thanks!

tests/typecheck/models/test_create.yml
asottile-sentry: I don't think that's relevant / useful -- the type for `dct` is explicitly annotated.  the bug here is strictly in the plugin parts of `create`

sobolevn (approved):

PR #2248, opened 2024-07-03
use field annotations for values_list types

asottile:
test originally failing with: 

```
 N: Revealed type is "builtins.list[Tuple[builtins.int, builtins.str]]
```

I reorganized the similar setter type determination as well as they had symmetry


flaeppe (approved):
Nice work

PR #2252, opened 2024-07-05
Remove some unnecessary defer rounds

flaeppe:
Change so that we don't trigger unnecessary defer rounds during manager creation via the `from_queryset` method call.

Refs:
- #1044 
- #1023

### DIFF
--- a/tests/typecheck/managers/test_managers.yml
+++ b/tests/typecheck/managers/test_managers.yml
@@ -519,7 +519,7 @@
                     reveal_type(user.bookingowner_set)
                     reveal_type(user.booking_set)
 
-                    # Check QuerySet methods on UnknownManager
+                    # Check QuerySet methods on unknown manager
                     reveal_type(Booking.objects.all)
                     reveal_type(Booking.objects.custom)
                     reveal_type(Booking.objects.all().filter)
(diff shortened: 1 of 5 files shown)


tests/typecheck/managers/test_managers.yml
```diff
@@ -539,30 +539,27 @@
...
+        myapp/models:56: error: "QuerySet[Any, Any]" has no attribute "custom"  [attr-defined]
         myapp/models:56: note: Revealed type is "Any"
-        myapp/models:57: note: Revealed type is "Union[myapp.models.Booking, None]"
-        myapp/models:58: note: Revealed type is "myapp.models.Booking"
-        myapp/models:59: note: Revealed type is "builtins.list[myapp.models.Booking]"
-        myapp/models:60: note: Revealed type is "builtins.list[myapp.models.Booking]"
+        myapp/models:57: note: Revealed type is "Union[Any, None]"
+        myapp/models:58: note: Revealed type is "Any"
+        myapp/models:59: note: Revealed type is "builtins.list[Any]"
+        myapp/models:60: note: Revealed type is "builtins.list[Any]"
```
flaeppe, line 562: I'm going to argue that the changes here improves correctness. Since the manager in the test comes from a function that specifies a manager as return type:

```python
def LocalManager() -> models.Manager:
    """
    Returns a manager instance of an inlined manager type that can't
    be resolved.
    """
    class InnerManager(models.Manager): ...
    return InnerManager()
```

This (now) assigns a type `models.Manager[Any]`, the return type of the function. I would say that the previous behaviour did too much, since it tried to express that assigned an invalid manager. But to know that it had to draw conclusions from the function body and thus "override" the return type.

The plugin should now only consider calls of the format: `objects = <Manager>.from_queryset(<QuerySet>)()` or `objects = <QuerySet>.as_manager()` in this case

flaeppe:
To clarify a little bit more; previously the plugin _could_ tell mypy to defer on e.g. a case like this:

```python
def not_a_manager() -> Any: ...

class SomeModel(models.Model):
    objects = not_a_manager()
```

Deferring isn't gonna make it a more of a manager. The changed code looks for a more narrow/specific case that we want and can handle that comes from `from_queryset`. For any other situation we now just continue without deferring.

flaeppe:
@sobolevn, @intgr would any of you guys mind checking this out? I'm suspecting it _could_ improve a couple of errors in #1023 (though I have no proof that it will)

PR #2253, opened 2024-07-08
Update function signatures for `django.db.models.sql.Query` class

DongwookKim0823:
# I have made things!

This PR updates the function signatures for the `Query` class in `django.db.models.sql.query`. Additionally, it removes unnecessary functions.


django-stubs/db/models/sql/query.pyi
```diff
@@ -81,51 +81,71 @@ class Query(BaseExpression):
...
-        branch_negated: bool = ...,
-        current_negated: bool = ...,
-        can_reuse: set[str] | None = ...,
-        allow_joins: bool = ...,
-        split_subq: bool = ...,
-        check_filterable: bool = ...,
-        update_join_types: bool = ...,
-    ) -> tuple[WhereNode, Iterable[str]]: ...
-    def add_filter(self, filter_clause: tuple[str, Any]) -> None: ...
+        filter_expr: Any,
```
sobolevn: Why did you change `filter_expr`'s type?

django-stubs/db/models/sql/query.pyi
```diff
@@ -81,51 +81,71 @@ class Query(BaseExpression):
...
+        branch_negated: bool = False,
+        current_negated: bool = False,
+        can_reuse: Any | None = None,
```
sobolevn: And `can_reuse`?

django-stubs/db/models/sql/query.pyi
DongwookKim0823: The change was a mistake. I've reverted the type of filter_expr to its original state.

DongwookKim0823: The change was a mistake. I've reverted the type to its original state.

django-stubs/db/models/sql/query.pyi
```diff
@@ -81,51 +81,73 @@ class Query(BaseExpression):
...
     @cached_property
     def base_table(self) -> str: ...
+    def add_annotation(self, annotation: Any, alias: str, select: bool = True) -> None: ...
     def sql_with_params(self) -> tuple[str, tuple]: ...
     def __deepcopy__(self, memo: dict[int, Any]) -> Query: ...
-    def get_compiler(self, using: str | None = ..., connection: BaseDatabaseWrapper | None = ...) -> SQLCompiler: ...
+    def get_compiler(
+        self, using: str | None = None, connection: Any | None = None, elide_empty: bool = True
+    ) -> Any: ...
+    def join_parent_model(self, opts: Any, model: Any | None, alias: str, seen: dict[Any | None, str]) -> str: ...
```
sobolevn:
```suggestion
    def join_parent_model(self, opts: Any, model: Any | None, alias: str, seen: dict[Any, str]) -> str: ...
```

sobolevn (approved):
Thank you!

PR #2254, opened 2024-07-08
Add `Options._get_fields`

Viicos:
Although private, this method is pretty useful when inspecting models

### DIFF
--- a/django-stubs/db/models/options.pyi
+++ b/django-stubs/db/models/options.pyi
@@ -122,6 +122,14 @@ class Options(Generic[_M]):
     def get_fields(
         self, include_parents: bool = ..., include_hidden: bool = ...
     ) -> list[Field[Any, Any] | ForeignObjectRel | GenericForeignKey]: ...
+    def _get_fields(
+        self,
+        forward: bool = ...,
+        reverse: bool = ...,
+        include_parents: bool | object = ...,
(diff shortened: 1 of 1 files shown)


django-stubs/db/models/options.pyi
```diff
@@ -122,6 +122,14 @@ class Options(Generic[_M]):
     def get_fields(
         self, include_parents: bool = ..., include_hidden: bool = ...
     ) -> list[Field[Any, Any] | ForeignObjectRel | GenericForeignKey]: ...
+    def _get_fields(
+        self,
+        forward: bool = ...,
+        reverse: bool = ...,
```
Viicos, line 128: We could probably have an overload on `reverse`, I think if `reverse=False`, `ForeignObjectRel` can't be present in the return type

sobolevn (commented):
Do you have any real-life / opensource examples of using this method?

Viicos:
A github search shows some examples: https://github.com/search?q=%22_meta._get_fields%22&type=code

Otherwise, I'm mainly using it when I want to exclude/include forward/reverse relationships, as they are all included by default. However, filtering can also be done on the result of `get_fields`, so this is really up to you!

sobolevn (approved):
Thank you!

Issue #579, opened 2021-03-20
Use django-stubs with type checkers other than mypy


kylebebak:
Hey all,

I know the README says it doesn't make sense to use `django-stubs` without mypy, because mypy uses a plugin for cases where type hints alone don't work. Sorry if this question is a waste of time, and feel free to close this issue. 

That said, this library ships with some very thorough type stubs for just about everything in Django. Is it possible to get a subset of the benefits from these stubs using a different type checker, like `pyright`?

I tried using the [stubs from the latest commit](https://github.com/typeddjango/django-stubs/commit/124d90794c6a303ff7bba2ebcb84b80d9c234618), without the mypy plugin, with pyright, but have found that `pyright` can't infer the types of model fields, which would be **very** useful. Nor can it infer types returned by `Manager` methods.

I'm sure pyright is sourcing the stubs just fine; I can manually edit the stubs for something like `timezone.now()`, and pyright picks it up immediately.

@sobolevn @mkurnikov 

Is it expected that most things having to do with the ORM don't work without running the mypy plugin? Is there manual monkey-patching that can be done, without running the mypy plugin, to help type checkers infer the correct types from model fields and Manager methods?

If this were possible it would be really awesome, and would help more users of type hints get lots of value out of `django-stubs` =)

I'm using Django 3.1.6, and [stubs from the latest commit](https://github.com/typeddjango/django-stubs/commit/124d90794c6a303ff7bba2ebcb84b80d9c234618).

## Examples

Here are some images of `pyright` doing type inference in my text editor:

<img width="355" alt="Screen Shot 2021-03-20 at 2 01 14 PM" src="https://user-images.githubusercontent.com/1524088/111884122-0ad02180-8985-11eb-81d6-d6302e55c7b0.png">

<img width="349" alt="Screen Shot 2021-03-20 at 2 01 26 PM" src="https://user-images.githubusercontent.com/1524088/111884130-2cc9a400-8985-11eb-83ef-510e708e7657.png">

I instantiate a `prod` from a `Product` models, which has a `name` field that's a `CharField`.

`pyright` knows `prod` is a `Product`, but it thinks `prod.name` has a type of `Unknown`. It thinks `Product.name` has a type of `CharField[Unknown, Unknown]`

<img width="532" alt="Screen Shot 2021-03-20 at 2 35 42 PM" src="https://user-images.githubusercontent.com/1524088/111884819-9ea3ec80-8989-11eb-8187-14e3a6fcb0a2.png">

[...]

PR #2261, opened 2024-07-15
Allow single generic param for Field in ForeignKey

rafonseca:
# Allow a less verbose specification of foreign keys

This set he default type of getter so that the user can choose to specify the foreign key using one or two generic params.

Recently, I've moved from django-types to django-stubs, and this modification allowed for a smooth transition.

## Related issues

- Refs #579
- Refs  #1264

### DIFF
--- a/tests/typecheck/models/test_related_fields.yml
+++ b/tests/typecheck/models/test_related_fields.yml
@@ -142,3 +142,58 @@
 
         class Other(models.Model):
             field = models.ForeignKey(MyModel, related_name="others", on_delete=models.CASCADE)
+
+
+- case: test_related_fields_with_two_generic_parameters
+  main: |
+        from myapp.models import Address, School, Student
+        reveal_type(Student().school)  # N: Revealed type is "myapp.models.School"
+        reveal_type(Student().address)  # N: Revealed type is "myapp.models.Address"
+        s = Student()
+        s.school = School()
+        s.address = Address()
+  installed_apps:
+    - myapp
+  files:
+    - path: myapp/__init__.py
+    - path: myapp/models.py
+      content: |
+        from django.db import models
+
+        class School(models.Model):
+            pass
+
+        class Address(models.Model):
+            pass
+
+        class Student(models.Model):
+            school = models.ForeignKey["School","School"](to="School", on_delete=models.CASCADE)
+            address = models.OneToOneField["Address","Address"](to="Address", on_delete=models.CASCADE)
+
+
+- case: test_related_fields_with_one_generic_parameter
+  expect_fail: True
+  main: |
+        from myapp.models import Address, School, Student
+        reveal_type(Student().school)  # N: Revealed type is "myapp.models.School"
+        reveal_type(Student().address)  # N: Revealed type is "myapp.models.Address"
+        s = Student()
+        s.school = School()
+        s.address = Address()
+  installed_apps:
+    - myapp
+  files:
+    - path: myapp/__init__.py
+    - path: myapp/models.py
+      content: |
+        from django.db import models
+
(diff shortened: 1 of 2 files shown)


flaeppe (commented):
This looks fine to me but a bunch of plugin code is about/around these `_ST` and `_GT` type variables so for that reason and any future changes, I would like that a test is included here that uses a single generic argument. Would be great if both `OneToOneField` and `ForeignKey` are included in that test

sobolevn (commented):
Yes, please, add a test for single-arg `Field`

rafonseca:
I've forgot to mention I used this with pyright... I'm not acquainted with mypy but what I understand here is that either mypy does not propagates the type defined in the first parameter to the second parameter or it is a printing issue. In the latter case, maybe there is another way to write the test.
@sobolevn , @flaeppe , can you give me some advice here?

flaeppe:
I think it's the former, mypy not handling type vars as default, while recognizing what you say from https://github.com/python/mypy/issues/14851#issuecomment-1988786756

Our own related issue is #1264

tests/typecheck/models/test_related_fields.yml
```diff
@@ -142,3 +142,51 @@
 
         class Other(models.Model):
             field = models.ForeignKey(MyModel, related_name="others", on_delete=models.CASCADE)
+
+
+- case: test_related_fields_with_two_generic_parameters
+  main: |
+        from myapp.models import Student
+        reveal_type(Student().school)  # N: Revealed type is "myapp.models.School"
+        reveal_type(Student().address)  # N: Revealed type is "myapp.models.Address"
```
flaeppe: Let's test the set type(`_ST`) argument results in no error(s) as well:

```suggestion
        from myapp.models import Address, School, Student
        reveal_type(Student().school)  # N: Revealed type is "myapp.models.School"
        reveal_type(Student().address)  # N: Revealed type is "myapp.models.Address"
        s = Student()
        s.school = School()
        s.address = Address()
```

tests/typecheck/models/test_related_fields.yml
```diff
@@ -142,3 +142,51 @@
...
+        class Student(models.Model):
+            school = models.ForeignKey["School","School"](to="School", on_delete=models.CASCADE)
+            address = models.OneToOneField["Address","Address"](to="Address", on_delete=models.CASCADE)
+
+
+- case: test_related_fields_with_one_generic_parameter
+  main: |
+        from myapp.models import Student
+        reveal_type(Student().school)  # N: Revealed type is "myapp.models.School"
+        reveal_type(Student().address)  # N: Revealed type is "myapp.models.Address"
```
flaeppe: Same here:

```suggestion
        from myapp.models import Address, School, Student
        reveal_type(Student().school)  # N: Revealed type is "myapp.models.School"
        reveal_type(Student().address)  # N: Revealed type is "myapp.models.Address"
        s = Student()
        s.school = School()
        s.address = Address()
```

flaeppe (commented):
Since Mypy doesn't seem to break by adding the `default=` argument I would be fine adding it already for the sakes of pyright.

Which hopefully allows us to just fill in for Mypy later on when https://github.com/python/mypy/issues/14851#issuecomment-1988786756 is resolved

We could add `expect_fail: true` (`xfail`) for the 2 new tests to start with

rafonseca:
Thanks for the quick answer @flaeppe . I've updated with your suggestions. Should we add any note about the failing test or the link to this PR in the final commit message is enough?

sobolevn (approved):

flaeppe (approved):
No need to add a note, I feel like the git history and link to this PR will be enough.

Issue #2272, opened 2024-07-24
reverse relations are not recognized by mypy


AlexandrVasilchuk:
# Bug report

# Description
I am facing an issue with django-stubs where reverse relations are not recognized by mypy. Despite trying multiple threads and solutions, including those in the following discussions and issues:

[Discussion 2192](https://github.com/typeddjango/django-stubs/discussions/2192)
[Issue 150](https://github.com/typeddjango/django-stubs/issues/150)

The problem persists.

## Actual Behavior
mypy raises errors indicating that the reverse relation field is not recognized. Specifically, it cannot resolve the keyword 'proposals' into the field.

## Expected Behavior
mypy should recognize the reverse relations between models and not raise errors when these relations are used.

## Steps to Reproduce
Define models with ManyToManyField and reverse relations.
Run mypy to check for typing errors.

![Снимок экрана 2024-07-24 в 15 43 03](https://github.com/user-attachments/assets/180ca64f-5f20-4598-93b0-a3492e9717a1)
![Снимок экрана 2024-07-24 в 15 43 28](https://github.com/user-attachments/assets/ca2dcc4f-1407-49cf-a122-5444be80bf32)
![Снимок экрана 2024-07-24 в 15 44 04](https://github.com/user-attachments/assets/04600e4d-9eed-4dcc-a412-ec08f705d23a)

## System information

- OS:
- `python` 3.11.4:
- `django` 4.2.7,:
- `mypy` "1.10.1":
- `django-stubs` "5.0.2":
- `django-stubs-ext` "5.0.2":

-------------------------

How can I make mypy recognize the reverse relation between these models? It is also important to mention that the models implement standard objects.

Thank you for your assistance.

flaeppe:
I've tried to investigate this further, but I'm having trouble trying to figure out if your provided snippets are a repro case or not. Could you please fill in the missing pieces so that we can check if we have a repro case here?

From your images:

- What type is `document` in `if document.pcrs.filter(proposals__project__isnull=True).exists():`?
- What does the related `PCR` model look like?

I'm not saying that you necessarily need to include all model fields, a runnable minimal breaking case is fine. If possible, of course.

AlexandrVasilchuk:
@flaeppe 

# Thank You and Apologies

First of all, thank you for the quick response to my previous issue. I apologize for not providing the complete code necessary to reproduce the problem in my initial post. Below, I have included a more comprehensive example.

## Reproduction Code

[...]

PR #2275, opened 2024-07-25
Check calls to filtering manager methods involving `ManyToManyField`

flaeppe:
Ensure that related managers and auto created models/managers (through table) are checked when calling methods `filter`, `get` or `exclude`

## Related issues

Found when further investigating #2272


sobolevn (approved):
Thanks!

PR #2276, opened 2024-07-25
lookup manager type via mro

asottile:
there's a lot going on in the test but this specifically breaks lookup through `TypeVar` in mypy 1.11 (mypy 1.10 also failed in this way as well -- but was fixed in python/mypy#17381)

test originally failing with:

```
/tmp/django-stubs/tests/typecheck/models/test_inheritance.yml:114:
E   Actual:
E     myapp/models:23: error: Incompatible return value type (got "Bound", expected "T")  [return-value] (diff)
E   Expected:
E     (empty)
```

### DIFF
--- a/tests/typecheck/managers/test_managers.yml
+++ b/tests/typecheck/managers/test_managers.yml
@@ -543,7 +543,6 @@
         myapp/models:23: error: Could not resolve manager type for "myapp.models.TwoUnresolvable.objects"  [django-manager-missing]
         myapp/models:24: error: Could not resolve manager type for "myapp.models.TwoUnresolvable.second_objects"  [django-manager-missing]
         myapp/models:27: error: Could not resolve manager type for "myapp.models.AbstractUnresolvable.objects"  [django-manager-missing]
-        myapp/models:32: error: Could not resolve manager type for "myapp.models.InvisibleUnresolvable.objects"  [django-manager-missing]
         myapp/models:36: note: Revealed type is "django.db.models.manager.Manager[myapp.models.User]"
         myapp/models:37: note: Revealed type is "django.db.models.manager.Manager[myapp.models.User]"
         myapp/models:39: note: Revealed type is "myapp.models.UnknownManager[myapp.models.Booking]"
--- a/tests/typecheck/models/test_inheritance.yml
+++ b/tests/typecheck/models/test_inheritance.yml
@@ -109,3 +109,44 @@
                         abstract = True
                 class User(Mixin1, Mixin2):
                     name = models.TextField()
+
+-   case: test_manager_typevar_through_bounds
+    main: |
+        from myapp.models import ResultProcessorConcrete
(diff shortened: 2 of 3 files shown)


sobolevn (approved):
Thank you! It will be in `5.0.3`

tests/typecheck/models/test_inheritance.yml
```diff
@@ -109,3 +109,42 @@
...
+        reveal_type(ResultProcessorConcrete().f())  # N: Revealed type is "myapp.models.Concrete"
+    installed_apps:
+        -   myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db.models import Model
+                from django.db.models.manager import Manager
+                from typing import TypeVar, Generic, ClassVar
```
sobolevn, line 127:
```suggestion
                from typing import TypeVar, Generic, ClassVar, Type
```

tests/typecheck/models/test_inheritance.yml
```diff
@@ -109,3 +109,42 @@
...
+                class Base(Model):
+                    custom_objects: ClassVar[BaseManager[Self]] = BaseManager()
+
+                class Bound(Base): pass
+
+                T = TypeVar("T", bound=Bound)
+
+                class ResultProcessorBase(Generic[T]):
+                    @property
+                    def model_cls(self) -> type[T]:
```
sobolevn, line 143:
```suggestion
                    def model_cls(self) -> Type[T]:
```

asottile:
looks like I have a few things to fix still as well -- I think I need more `Self` probably :)

tests/typecheck/managers/test_managers.yml
```diff
@@ -543,7 +543,6 @@
         myapp/models:23: error: Could not resolve manager type for "myapp.models.TwoUnresolvable.objects"  [django-manager-missing]
         myapp/models:24: error: Could not resolve manager type for "myapp.models.TwoUnresolvable.second_objects"  [django-manager-missing]
         myapp/models:27: error: Could not resolve manager type for "myapp.models.AbstractUnresolvable.objects"  [django-manager-missing]
-        myapp/models:32: error: Could not resolve manager type for "myapp.models.InvisibleUnresolvable.objects"  [django-manager-missing]
```
asottile, line 546: this makes sense to me at least -- we're no longer redefining an `objects` here because the base class already has an `UnknownManager[Self]` definition

asottile:
@sobolevn should be ready for another round of review now \o/

tests/typecheck/managers/test_managers.yml
sobolevn: cc @flaeppe

flaeppe: This is fine IMO. The removal here goes away in my changes in #2252 too

flaeppe (approved):

sobolevn:
Thanks everyone!

PR #2277, opened 2024-07-25
Check model fields on filtering methods of queryset types

flaeppe:
I noticed that we should check some methods on both `Manager` and `QuerySet` since the `Manager` class is a bit special as it is a type "merged" from `BaseManager` and `QuerySet`.

@sobolevn I think this could be a nice addition before releasing `5.0.3`, or what do you think?

Thinking that the use cases from the test included here aren't too foreign:

```python
MyModel.objects.filter(id=1).filter(invalid=1)
MyModel.objects.filter(id=1).get(invalid=1)
MyModel.objects.filter(id=1).exclude(invalid=1)
MyModel.objects.filter(id=1).create(invalid=1)
```

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -33,7 +33,6 @@ from mypy.plugin import (
     DynamicClassDefContext,
     FunctionContext,
     MethodContext,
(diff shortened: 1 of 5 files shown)


sobolevn (approved):
Thanks, very good usecase.

Issue #1684, opened 2023-09-05
`type[Model]` no longer allows access to `.objects` in `django-stubs` 4.2.4


johnthagen:
# Bug report

## What's wrong

Prior to `django-stubs` 4.2.4 / `mypy` 1.5.1, the you could type hint a variable/class attribute as `type[Model]` and access `.objects` on it:

```py
from django.contrib.auth.models import User
from django.db.models import Model

user_type: type[Model] = User

users = user_type.objects.all()
```

Note that is the real usage, the variable is a class variable that is set by derived classes.

But as of the latest `django-stubs` and `mypy`, this generates a type check error:

```
error: "type[Model]" has no attribute "objects"  [attr-defined]
```

## How is that should be

No type check errors.

## System information

- OS: macOS 
- `python` version: 3.10.11
- `django` version: 4.2.5
- `mypy` version: 1.5.1
- `django-stubs` version: 4.2.4
- `django-stubs-ext` version: 4.2.2

Mypy configuration:

```toml
[tool.mypy]
ignore_missing_imports = true
strict = true

disallow_subclassing_any = false
disallow_untyped_decorators = false
warn_return_any = false

plugins = [
    "pydantic.mypy",
    "mypy_django_plugin.main",
]
```

flaeppe:
Yes, `models.Model` is no longer assumed to have the `objects` manager. It's not guaranteed to exist for a `models.Model`. You should use the `_default_manager` attribute instead: https://docs.djangoproject.com/en/4.2/topics/db/managers/#django.db.models.Model._default_manager

> If you’re writing some code that must handle an unknown model, for example, in a third-party app that implements a generic view, use this manager (or [_base_manager](https://docs.djangoproject.com/en/4.2/topics/db/managers/#django.db.models.Model._base_manager)) rather than assuming the model has an objects manager.

```python
from django.contrib.auth.models import User
from django.db.models import Model

user_type: type[Model] = User

users = user_type._default_manager.all()
```

johnthagen:
Thanks for the tip @flaeppe!

sobolevn:
I think that we might need to put this in the docs!

djbrown:
But when I switch to `model._default_manager` I get the following pylint error:
`W0212: Access to a protected member _default_manager of a client class (protected-access)`
So either I stay with `model.objects`  and add a mypy ignore line comment,
or go with `model._default_manager` and add a pylint ignore line comment.. 😞

[...]

PR #2280, opened 2024-07-26
Don't remove `objects` attribute from `Model` in plugin

### LINKED ISSUE #1744: Type[SomeConcreteModel] has no attribute "objects" -- but it does at runtime
# Bug report

unfortunately having a really hard time tracking this down -- I've seen #1684 already however this is a concrete model and not an abstract one

## What's wrong

I haven't been able to narrow this to a minimal case so I suspect it's something wrong with our setup -- here's at least a maximal reproduction:

```bash
set -euxo pipefail

rm -rf sentry

git clone -qq --depth=1 https://github.com/getsentry/sentry -b 23.9.1 >& /dev/null
cd sentry
python3.8 -m venv venv >& /dev/null
# cheaty editable install so you don't need node
echo "$PWD/src" > venv/lib/python3.8/site-packages/easy-install.pth
venv/bin/pip install -qq --no-deps -r requirements-dev-frozen.txt
venv/bin/pip install -qq --upgrade mypy django-stubs django-stubs-ext
venv/bin/mypy tests/sentry/db/models/fields/test_jsonfield.py
```

output:

```console
$ bash t.sh 
+ rm -rf sentry
+ git clone -qq --depth=1 https://github.com/getsentry/sentry -b 23.9.1
+ cd sentry
+ python3.8 -m venv venv
+ echo /tmp/y/sentry/src
+ venv/bin/pip install -qq --no-deps -r requirements-dev-frozen.txt
+ venv/bin/pip install -qq --upgrade mypy django-stubs django-stubs-ext
+ venv/bin/mypy tests/sentry/db/models/fields/test_jsonfield.py
tests/sentry/db/models/fields/test_jsonfield.py:58: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:64: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:68: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:69: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:119: error: "Type[JSONFieldWithDefaultTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:120: error: "Type[JSONFieldWithDefaultTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:124: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:125: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:126: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:127: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:128: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:129: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:131: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:133: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:135: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:139: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:142: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:146: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:147: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:148: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:150: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:151: error: "Type[JSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:154: error: "Type[BlankJSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:155: error: "Type[BlankJSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:159: error: "Type[BlankJSONFieldTestModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:164: error: "Type[CallableDefaultModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:165: error: "Type[CallableDefaultModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:169: error: "Type[CallableDefaultModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:170: error: "Type[CallableDefaultModel]" has no attribute "objects"  [attr-defined]
tests/sentry/db/models/fields/test_jsonfield.py:194: error: "Type[BlankJSONFieldTestModel]" has no attribute "objects"  [attr-defined]
Found 30 errors in 1 file (checked 1 source file)
```

will try and narrow it down to a minimal case but I've failed to do that so far

## How is that should be

I expect no errors in that file

## System information

- OS:
- `python` version: 3.8.16
- `django` version: 3.2.20
- `mypy` version: 1.5.1
- `django-stubs` version:  4.2.4

### LINKED ISSUE #2304: [5.0.3 regression] error: "type[RealmFilter]" has no attribute "objects"  [attr-defined]
When trying to upgrade to django-stubs 5.0.3 or 5.0.4 in [Zulip](https://github.com/zulip/zulip), I get 25 unexpected errors like `error: "type[RealmFilter]" has no attribute "objects"  [attr-defined]`. This is not #1684; `RealmFilter` is a normal concrete model class. The issue seems to depend on the structure of the project’s import graph.

Based on `git bisect`, it first occurs at

- 2f5a4946e0569e9bc69729fc59794a1ea4347b91 (#2276, cc @asottile).

I’ve reduced the issue to this minimized test case: https://github.com/andersk/no-attribute-objects

```console
$ mypy .
zerver/models/__init__.py:6: error: "type[RealmFilter]" has no attribute "objects"  [attr-defined]
Found 1 error in 1 file (checked 10 source files)
```

**`pyproject.toml`**

```toml
[tool.django-stubs]
django_settings_module = "zproject.settings"

[tool.mypy]
```

**`confirmation/__init__.py`** (empty)

**`confirmation/models.py`**

```python
from django.db import models

from zerver.models import Realm


class Confirmation(models.Model):
    realm = models.ForeignKey(Realm, on_delete=models.RESTRICT)
```

**`zerver/__init__.py`** (empty)

**`zerver/models/__init__.py`**

```python
from zerver.models.linkifiers import RealmFilter as RealmFilter
from zerver.models.realms import Realm as Realm
from zerver.models.streams import Stream as Stream
from zerver.models.users import UserProfile as UserProfile

RealmFilter.objects
```

**`zerver/models/linkifiers.py`**

```python
from django.db import models


class RealmFilter(models.Model):
    pass
```

**`zerver/models/realms.py`**

```python
from django.db import models


class Realm(models.Model):
    pass
```

**`zerver/models/streams.py`**

```python
from django.db import models

from zerver.models.realms import Realm
from zerver.models.users import UserProfile


class Stream(models.Model):
    realm = models.ForeignKey(Realm, on_delete=models.RESTRICT)
    creator = models.ForeignKey(UserProfile, on_delete=models.RESTRICT)
```

**`zerver/models/users.py`**

```python


class UserProfile(AbstractBaseUser, PermissionsMixin):
    pass
```

**`zproject/__init__.py`** (empty)

**`zproject/settings.py`**

```python
INSTALLED_APPS = [
    "django.contrib.auth",
    "confirmation",
    "zerver",
]
```

flaeppe:
Partially reverts #1672

#2276 started looking through the mro for manager attributes, but that also conflicts with `mypy_django_plugin.transformers.models.MetaclassAdjustments`, which removes `objects` from `django.db.models.Model`. Because if a lookup through the mro is done _before_ `MetaclassAdjustments` has run, the plugin fools itself that the `objects` attribute exists.

I figured that reverting the removal of `objects` had most value, since most people use the default behaviour and it seems that not having an `objects` is quite rare.

It's currently only our plugin users that gets to follow along Django implementing the default manager attribute as an abstract attribute. But instead of trying to provide our home built one I think we should wait for any built in support of abstract attributes and then see if we can get any last pieces to work via the plugin.

## Related issues

- Refs: #1672
- Refs: #2276
-  Closes: #1744 
- Refs: https://github.com/typeddjango/djangorestframework-stubs/pull/644#issuecomment-2252205083
- Refs: https://github.com/python/cpython/issues/106248
- Refs: #1684 
- Closes #2304

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -319,10 +319,17 @@ class AddManagers(ModelClassInitializer):
         assert manager_info is not None
         # Reparameterize dynamically created manager with model type
         manager_type = helpers.fill_manager(manager_info, Instance(self.model_classdef.info, []))
+        manager_node = self.model_classdef.info.get(manager_name)
+        if manager_node and isinstance(manager_node.node, Var):
+            manager_node.node.type = manager_type
         self.add_new_node_to_model_class(manager_name, manager_type, is_classvar=True)
 
     def run_with_model_cls(self, model_cls: Type[Model]) -> None:
         manager_info: Optional[TypeInfo]
+
+        def cast_var_to_classvar(symbol: Optional[SymbolTableNode]) -> None:
+            if symbol and isinstance(symbol.node, Var):
+                symbol.node.is_classvar = True
 
         incomplete_manager_defs = set()
         for manager_name, manager in model_cls._meta.managers_map.items():
@@ -345,7 +352,24 @@ class AddManagers(ModelClassInitializer):
 
             assert self.model_classdef.info.self_type is not None
             manager_type = helpers.fill_manager(manager_info, self.model_classdef.info.self_type)
-            self.add_new_node_to_model_class(manager_name, manager_type, is_classvar=True)
+            # It seems that the type checker fetches a Var from expressions, but looks
+            # through the symbol table for the type(at some later stage?). Currently we
+            # don't overwrite the reference mypy holds from an expression to a Var
+            # instance when adding a new node, we only overwrite the reference to the
+            # Var in the symbol table. That means there's a lingering Var instance
+            # attached to expressions and if we don't flip that to a ClassVar, the
+            # checker will emit an error for overriding a class variable with an
+            # instance variable. As mypy seems to check that via the expression and not
+            # the symbol table. Optimally we want to just set a type on the existing Var
+            # like:
+            # manager_node.node.type = manager_type
+            # but for some reason that doesn't work. It only works replacing the
+            # existing Var with a new one in the symbol table.
+            cast_var_to_classvar(manager_node)
+            if manager_fullname == manager_info.fullname and manager_node and isinstance(manager_node.node, Var):
+                manager_node.node.type = manager_type
+            else:
+                self.add_new_node_to_model_class(manager_name, manager_type, is_classvar=True)
 
         if incomplete_manager_defs:
             if not self.api.final_iteration:
@@ -360,6 +384,7 @@ class AddManagers(ModelClassInitializer):
                 # setting _some_ type
                 fallback_manager_info = self.get_or_create_manager_with_any_fallback()
                 if fallback_manager_info is not None:
+                    cast_var_to_classvar(self.model_classdef.info.get(manager_name))
                     assert self.model_classdef.info.self_type is not None
                     manager_type = helpers.fill_manager(fallback_manager_info, self.model_classdef.info.self_type)
                     self.add_new_node_to_model_class(manager_name, manager_type, is_classvar=True)
@@ -957,12 +982,6 @@ class MetaclassAdjustments(ModelClassInitializer):
             and not multiple_objects_returned.plugin_generated
         ):
             del ctx.cls.info.names["MultipleObjectsReturned"]
-
-        objects = ctx.cls.info.names.get("objects")
-        if objects is not None and isinstance(objects.node, Var) and not objects.plugin_generated:
-            del ctx.cls.info.names["objects"]
-
-        return
 
     def get_exception_bases(self, name: str) -> List[Instance]:
         bases = []
--- a/tests/typecheck/managers/test_managers.yml
+++ b/tests/typecheck/managers/test_managers.yml
@@ -761,3 +761,65 @@
                     generic_manager = models.Manager()
                     generic_manager_from_generic_queryset = GenericManagerFromGenericQuerySet()
                     generic_manager_from_populated_queryset = GenericManagerFromPopulatedQuerySet()
+
+# Regression test for #2304
+-   case: test_objects_managers_is_kept_with_specific_import_graph
+    main: |
+        from zerver.models import RealmFilter
+        reveal_type(RealmFilter.objects)  # N: Revealed type is "django.db.models.manager.Manager[zerver.models.linkifiers.RealmFilter]"
+    installed_apps:
+        - django.contrib.auth
+        - django.contrib.contenttypes
+        - confirmation
+        - zerver
+    files:
+        -   path: confirmation/__init__.py
+        -   path: confirmation/models.py
+            content: |
+                from django.db import models
+
+                from zerver.models import Realm
+
+                class Confirmation(models.Model):
+                    realm = models.ForeignKey(Realm, on_delete=models.CASCADE)
+        -   path: zerver/__init__.py
+        -   path: zerver/models/__init__.py
+            content: |
+                from zerver.models.linkifiers import RealmFilter as RealmFilter
+                from zerver.models.realms import Realm as Realm
+                from zerver.models.streams import Stream as Stream
+                from zerver.models.users import UserProfile as UserProfile
+
+                RealmFilter.objects
+        -   path: zerver/models/linkifiers.py
+            content: |
+                from django.db import models
+
+
+                class RealmFilter(models.Model):
+                    pass
+        -   path: zerver/models/realms.py
+            content: |
+                from django.db import models
+
+
+                class Realm(models.Model):
+                    pass
+        -   path: zerver/models/streams.py
+            content: |
+                from django.db import models
+
+                from zerver.models.realms import Realm
+                from zerver.models.users import UserProfile
+
+
+                class Stream(models.Model):
+                    realm = models.ForeignKey(Realm, on_delete=models.RESTRICT)
+                    creator = models.ForeignKey(UserProfile, on_delete=models.RESTRICT)
+        -   path: zerver/models/users.py
+            content: |
+                from django.contrib.auth.models import AbstractBaseUser, PermissionsMixin
+
+
+                class UserProfile(AbstractBaseUser, PermissionsMixin):
+                    pass
--- a/django-stubs/db/models/base.pyi
+++ b/django-stubs/db/models/base.pyi
@@ -36,8 +36,7 @@ class Model(metaclass=ModelBase):
     # and re-add them to correct concrete subclasses of 'Model'
     DoesNotExist: Final[type[ObjectDoesNotExist]]
     MultipleObjectsReturned: Final[type[BaseMultipleObjectsReturned]]
-    # This 'objects' attribute will be deleted, via the plugin, in favor of managing it
-    # to only exist on subclasses it exists on during runtime.
+
     objects: ClassVar[Manager[Self]]
 
     _meta: ClassVar[Options[Self]]
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -727,10 +727,9 @@
         -   path: myapp/models.py
             content: |
                 from django.db import models
-                from django.db.models.manager import BaseManager
                 class TransactionQuerySet(models.QuerySet):
                     pass
-                TransactionManager = BaseManager.from_queryset(TransactionQuerySet)
+                TransactionManager = models.Manager.from_queryset(TransactionQuerySet)
                 class Transaction(models.Model):
                     pk = 0
                     objects = TransactionManager()
@@ -742,7 +741,7 @@
                 Transaction().test()
 
 
--   case: foreign_key_relationship_for_models_with_custom_manager_unsolvable
+-   case: foreign_key_relationship_for_models_with_custom_manager_solvable_via_as_manager_type
     main: |
         from myapp.models import Transaction
     installed_apps:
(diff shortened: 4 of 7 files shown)


flaeppe:
Well, with these changes we end up with having to declare `ClassVar` for `objects` when declared explicitly.

No surprise, I suppose, but probably a bunch of changes required for many.

flaeppe:
It actually is surprising, we shouldn't get complaints on overriding class with instance variable in all the failing tests. I think that we're missing something in the plugin, gonna dig around and see

sobolevn:
@flaeppe can you please add a regression test from https://github.com/typeddjango/django-stubs/issues/2304 to this PR?

flaeppe:
Yeah sure

flaeppe:
@sobolevn I've tried out the test, locally, but it doesn't break on `master`. It's probably some difference between how our tests are run and how it's run in https://github.com/andersk/no-attribute-objects. Since that whole issue is depending on how `mypy` builds or decides to process some internal structure I would guess.

I can still add it in if you like but it's unfortunately not reproducing anything

flaeppe:
Test can be found in 336684105bc047bc37554799b5aeeb3322802433

mypy_django_plugin/transformers/models.py
```diff
@@ -345,7 +352,24 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+                # expression to a Var instance when adding a new node, we only overwrite
+                # the reference to the Var in the symbol table. That means there's a
+                # lingering Var instance attached to expressions and if we don't flip
+                # that to a ClassVar, the checker will emit an error for overriding a
+                # class variable with an instance variable. As mypy seems to check that
+                # via the expression and not the symbol table. Optimally we want to just
+                # set a type on the existing Var like
+                # manager_node.node.type = manager_type
+                # but for some reason that doesn't work. It only works replacing the
+                # existing Var with a new one in the symbol table.
```
flaeppe: This was a strange discovery I had to battle with for some time to find any sort of workaround.

I realised that the error "overriding a class variable with an instance variable" from mypy uses the body/expressions instances e.g. `AssignmentStmt.lvalues[0]`. But we don't overwrite the expressions when inserting a new `Var` in any symbol table. So the checker uses the lingering reference to a `Var` instance that is not the same as in the symbol table and that one doesn't set/have `is_classvar` the way we want.

tests/typecheck/managers/test_managers.yml
```diff
@@ -169,28 +169,29 @@
...
     files:
         -   path: myapp/__init__.py
         -   path: myapp/models.py
             content: |
+                from typing import ClassVar
                 from django.db import models
 
                 class AbstractPerson(models.Model):
-                    abstract_persons = models.Manager['AbstractPerson']()
+                    abstract_persons: ClassVar[models.Manager["AbstractPerson"]] = models.Manager['AbstractPerson']()
```
sobolevn: Why do we need this change?

tests/typecheck/managers/test_managers.yml
```diff
@@ -169,28 +169,29 @@
...
                 class PublishedBookManager(models.Manager['Book']):
                     pass
                 class AnnotatedBookManager(models.Manager['Book']):
                     pass
                 class Book(models.Model):
                     title = models.CharField(max_length=50)
-                    published_objects = PublishedBookManager()
-                    annotated_objects = AnnotatedBookManager()
+                    published_objects: ClassVar[PublishedBookManager] = PublishedBookManager()
```
sobolevn: This seems like annotations that we should not need.

sobolevn (commented):
I am a bit confused about new `ClassVar` annotations in tests.

tests/typecheck/managers/test_managers.yml
flaeppe: Looked around a bit and It comes from the plugin no longer forcing `is_classvar = True` on an all manager `Var`s

I don't think we should need to set `ClassVar` explicitly either, but mypy otherwise gives us an error on attribute access

```
main:2: error: Access to generic instance variables via class is ambiguous  [misc]
```

Not a big thing that the plugin forces `is_classvar = True`, but optimally I don't think it should be necessary. Not sure what to make of the error though, I don't see why it's a generic instance variable..

django-stubs/contrib/sites/managers.pyi
```diff
@@ -1,9 +1,8 @@
 from typing import TypeVar
 
-from django.contrib.sites.models import Site
 from django.db import models
 
-_T = TypeVar("_T", bound=Site)
+_T = TypeVar("_T", bound=models.Model)
```
sobolevn: Looks like my comment was not sent the first time :(
This feels unrelated?

flaeppe: Yeah it is. But a test on it errored on the `ClassVar` thingy and I realised that it was incorrectly typed. So I just corrected it. As otherwise I would've added some new incorrect type hints in the breaking test..

tests/typecheck/managers/test_managers.yml
flaeppe: Right, I think I've found the problem. Found that one exception that avoids the error accessing generic instance variable via class is "Self type wrapped in ClassVar"

Ref: https://github.com/python/mypy/blob/570b90a7a368f04c64f60af339d0ac1808c49c15/mypy/checkmember.py#L1065-L1070

But our plugin blindly adds `Self` as a generic argument to a manager, regardless of `is_classvar=True` or not.

https://github.com/typeddjango/django-stubs/blob/972a4e529b733820c2d9317310b279c6823b3f41/mypy_django_plugin/transformers/querysets.py#L42-L54

Which means that the error will appear for e.g. `Manager[Self]`

flaeppe: I suppose there's 2 ways out to keep working with `Self`:

1. Force `is_classvar = True`
2. Don't add a type argument `Self` when `is_classvar = False`

flaeppe: I went with alternative 1. and we're forcing ClassVar, as that's what we did previously. That should leave those parts unchanged I'm thinking.

django-stubs/contrib/sites/managers.pyi
flaeppe: I reverted this, might as well push it separately to keep things clean

flaeppe: Can be found in #2308

sobolevn (approved):
Let's also wait for @intgr's review. This part is very complex.
Thanks a lot for your work!

tests/typecheck/managers/test_managers.yml
flaeppe: Well, now I just found that I went down this exact rabbit hole before: https://github.com/typeddjango/django-stubs/pull/1649#discussion_r1313778655

Had completely forgotten about that..

delfick:
Without passing on an opinion about this PR, and not at all any kind of blocker, I just wanna take a moment and quietly point out the way I've been resolving this problem at my workplace (6 million line django app with over 2000 models, still on mypy 1.4, django 4.1) is I wrote this https://github.com/delfick/extended-mypy-django-plugin so I can say `Concrete[AbstractModel]` that gets translated into a union of all the concrete children (and `DefaultQuerySet[Model]` to find the default queryset for a model, or union of concrete default querysets). So essentially a type that represents everything on all the concrete children without taking away the things the abstract model does not have. I was able to use it to get all the errors down so we can do an upgrade. I intend on making a github issue here at some point pointing it out and asking for opinions. It ends up solving more problems than just "objects" getting attr-defined errors, but introduces a bunch of changes I imagine would take discussion if any of this gets put upstream here.

flaeppe:
It sounds interesting, feel free to open any discussions/issues as you feel relevant

sobolevn:
I like the idea of `Concrete` type. But, I think that from users' point of view, most `SomeAbstractModel` annotations are concrete. Not the other way around. So, we can add `AbstractModel[SomeAbstractModel]` to specify that we are indeed working with abstract model that might not have `.objects`, etc.

@delfick please, feel free to discuss / backport any features / ideas. I would be happy to help!

delfick:
@sobolevn it's more about saying "I want the concrete models that extend this abstract model". Especially useful when statically you know about more models than are available at runtime. I'll expand when I get to making a github issue. I wanted to get us onto newer versions of mypy/django/django-stubs before I did that because of how long this process is taking. Hopefully not too much longer now (also figuring out how to make the mypy plugin took an astounding amount of churn to make it so it works without causing problems)

though, rereading your comment, I get what you mean. I could be wrong but my instinct would be mypy wouldn't make that possible without enhancements to what hooks are available.

delfick:
I ended up creating that discussion https://github.com/typeddjango/django-stubs/discussions/2452 :)

PR #2281, opened 2024-07-26
fill QuerySet generics using the manager's model type

asottile:
new test was previously failing with:

```
__________________ test_from_queryset_with_concrete_subclass ___________________
/tmp/django-stubs/tests/typecheck/managers/querysets/test_from_queryset.yml:895: 
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Invalid output: 
E   Actual:
E     ...
E     main:4: note: Revealed type is "myapp.models.CustomQuerySet" (diff)
E     main:5: note: Revealed type is "D`2 = M`1"    (diff)
E   Expected:
E     ...
E     main:4: note: Revealed type is "myapp.models.CustomQuerySet[myapp.models.Concrete, myapp.models.Concrete]" (diff)
E     main:5: note: Revealed type is "myapp.models.Concrete" (diff)
E   Alignment of first line difference:
E     E: ... "myapp.models.CustomQuerySet[myapp.models.Concrete, myapp.models.Con...
E     A: ... "myapp.models.CustomQuerySet"...
E                                        ^
```

PR #2282, opened 2024-07-27
convert as_manager hooks to base class hook

asottile:
finished my idea from https://github.com/typeddjango/django-stubs/pull/2281#issuecomment-2253340475 !

I was getting hung up on overriding methods in inherited querysets -- now it pulls the return type from the base queryset and inherits from that directly

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -37,8 +37,7 @@ from mypy_django_plugin.transformers import (
 )
 from mypy_django_plugin.transformers.functional import resolve_str_promise_attribute
 from mypy_django_plugin.transformers.managers import (
-    construct_as_manager_instance,
-    create_new_manager_class_from_as_manager_method,
+    add_as_manager_to_queryset_class,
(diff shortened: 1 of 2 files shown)


flaeppe (commented):
I have one thought: doesn't this only apply if `as_manager` comes from `django.db.models.QuerySet` or if `as_manager` is generated by the plugin?

I just mean that the plugin behaviour is probably undefined if `as_manager` has been explicitly overridden by a user. And in that case we could just exit. Or what do you think?

It's probably a check between the attribute `symbol.plugin_generated` and perhaps `queryset_info.get_containing_type_info` if we want to do this.

I had a quick look here: https://grep.app/search?q=def%20as_manager%28&case=true&filter[lang][0]=Python and none with a runtime overrides `as_manager`. It's not a critical case in any way, was just a thought

flaeppe (approved):
Ah, nvm, I found the comment and exit now:

```python
# either a manual `as_manager` definition or this is a deferral pass
```

PR #2283, opened 2024-07-28
Check correct model on other side of many to many reverse filtering

### LINKED ISSUE #2284: ManyToMany relation wrongly validates values_list field name
# Bug report

The following code gives error when it should not

``` python
class Cube(Model):
    heat = IntegerField(default=0)


class Ice(Model):
    cubes = ManyToManyField(Cube, blank=True)

    def heat_sum(self) -> int:
        return sum(self.cubes.values_list("heat", flat=True)) # <-- Error
```

## What's wrong

mypy gives error

`error: Cannot resolve keyword 'heat' into field. Choices are: cube, cube_id, ice, ice_id, id  [misc]`

## How is that should be

No error

## System information

- OS: windows
- `python` version:  3.12.1
- `django` version: 5.0.3
- `mypy` version: 1.11.0
- `django-stubs` version: 5.0.3
- `django-stubs-ext` version: 5.0.3

flaeppe:
Depending on the manager type there could be different model arguments. For a `ManyRelatedManager` the argument is the `through` model, but we want to check the `to` model.

As a simple starter we stash the `to` model in the manager's metadata and set a precedence on fetching the `to` model from metadata.

## Related issues

- Refs https://github.com/typeddjango/django-stubs/issues/2272#issuecomment-2254502136
- Refs #2275 
- Closes #2284

### DIFF
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -1515,7 +1515,6 @@
                 class Author(models.Model):
                     ...
 
-
                 class Book(models.Model):
                     featured = models.BooleanField(default=False)
                     authors = models.ManyToManyField(Author)
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -1498,3 +1498,24 @@
 
                 class MyModel(models.Model):
                     others = models.ManyToManyField(Other)
+
+-   case: test_reverse_m2m_relation_checks_other_model
+    main: |
+        from myapp.models import Author
+        Author().book_set.filter(featured=True)
+        Author().book_set.filter(xyz=True)  # E: Cannot resolve keyword 'xyz' into field. Choices are: authors, featured, id  [misc]
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+
+                class Author(models.Model):
+                    ...
+
+
+                class Book(models.Model):
+                    featured = models.BooleanField(default=False)
+                    authors = models.ManyToManyField(Author)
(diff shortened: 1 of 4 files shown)


tests/typecheck/fields/test_related.yml
```diff
@@ -1480,9 +1480,9 @@
         MyModel.objects.get(xyz__isnull=False)  # E: Cannot resolve keyword 'xyz' into field. Choices are: id, others  [misc]
         MyModel.objects.exclude(xyz__isnull=False)  # E: Cannot resolve keyword 'xyz' into field. Choices are: id, others  [misc]
         other = Other()
-        other.mymodel_set.filter(xyz__isnull=True)  # E: Cannot resolve keyword 'xyz' into field. Choices are: id, mymodel, mymodel_id, other, other_id  [misc]
-        other.mymodel_set.get(xyz__isnull=True)  # E: Cannot resolve keyword 'xyz' into field. Choices are: id, mymodel, mymodel_id, other, other_id  [misc]
-        other.mymodel_set.exclude(xyz__isnull=True)  # E: Cannot resolve keyword 'xyz' into field. Choices are: id, mymodel, mymodel_id, other, other_id  [misc]
+        other.mymodel_set.filter(xyz__isnull=True)  # E: Cannot resolve keyword 'xyz' into field. Choices are: id, others  [misc]
+        other.mymodel_set.get(xyz__isnull=True)  # E: Cannot resolve keyword 'xyz' into field. Choices are: id, others  [misc]
+        other.mymodel_set.exclude(xyz__isnull=True)  # E: Cannot resolve keyword 'xyz' into field. Choices are: id, others  [misc]
```
flaeppe, line 1485: We actually had incorrect tests for the issue fixed here .. 🙂

flaeppe:
@sobolevn it might be a bunch of false negatives due to the issue here that will come from `5.0.3`. I suppose we could chose to do a release if we get a bunch of reported issues about incorrect errors reported for model fields via many to many field filtering.

tests/typecheck/fields/test_related.yml
```diff
@@ -1498,3 +1498,24 @@
...
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+
+                class Author(models.Model):
+                    ...
+
+
+                class Book(models.Model):
```
sobolevn: style nit

```suggestion
                class Author(models.Model):
                    ...

                class Book(models.Model):
```

sobolevn (approved):
Thanks for the quick fix!

PR #2292, opened 2024-07-29
match Field TypeVar variance in models.fields.related

asottile:
previously failing with:

```
_____________________________ foreign_key_subclass _____________________________
/home/asottile/workspace/django-stubs/tests/typecheck/fields/test_related.yml:90: 
E   Actual:
E     myapp/models:7: error: Variance of TypeVar "_ST" incompatible with variance in parent type  [type-var] (diff)
E     myapp/models:7: error: Variance of TypeVar "_GT" incompatible with variance in parent type  [type-var] (diff)
E   Expected:
E     (empty)
```

I'm not sure how this wasn't a problem before 5.0.4 for us but it certainly is now!

I'm also surprised mypy doesn't report this same error for `related.pyi` itself -- there must be some special-cased rules that `.pyi` files don't have to follow

### DIFF
--- a/django-stubs/db/models/fields/related.pyi
+++ b/django-stubs/db/models/fields/related.pyi
@@ -27,9 +27,9 @@ RECURSIVE_RELATIONSHIP_CONSTANT: Literal["self"]
 def resolve_relation(scope_model: type[Model], relation: str | type[Model]) -> str | type[Model]: ...
 
 # __set__ value type
-_ST = TypeVar("_ST")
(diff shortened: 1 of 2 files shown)


asottile:
failure looks like a github flake

sobolevn (approved):
Thank you! 

Source: https://github.com/typeddjango/django-stubs/blob/04eed10ca014c73c04115e645e6e8e019134c5e2/django-stubs/db/models/fields/__init__.pyi#L47-L52

PR #2294, opened 2024-07-30
try and use named arguments from caller for matching name

asottile:
test failing prior to change:

```
__________________ nullable_foreign_key_with_init_overridden ___________________
/home/asottile/workspace/django-stubs/tests/typecheck/fields/test_related.yml:806: 
E   Actual:
E     main:2: note: Revealed type is "myapp.models.B" (diff)
E   Expected:
E     main:2: note: Revealed type is "Union[myapp.models.B, None]" (diff)
E     E: ...te: Revealed type is "Union[myapp.models.B, None]"
E     A: ...te: Revealed type is "myapp.models.B"
```

the `FK` example matches a similar one in sentry

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -10,6 +10,7 @@ from mypy.mro import calculate_mro
 from mypy.nodes import (
     GDEF,
     MDEF,
+    ArgKind,
     AssignmentStmt,
     Block,
     ClassDef,
@@ -182,6 +183,12 @@ def get_call_argument_by_name(ctx: Union[FunctionContext, MethodContext], name: 
     Return the expression for the specific argument.
     This helper should only be used with non-star arguments.
     """
+    # try and pull the named argument from the caller first
+    for kinds, argnames, args in zip(ctx.arg_kinds, ctx.arg_names, ctx.args):
+        for kind, argname, arg in zip(kinds, argnames, args):
+            if kind == ArgKind.ARG_NAMED and argname == name:
+                return arg
+
     if name not in ctx.callee_arg_names:
         return None
     idx = ctx.callee_arg_names.index(name)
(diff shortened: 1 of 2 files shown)


mypy_django_plugin/lib/helpers.py
```diff
@@ -182,6 +183,12 @@ def get_call_argument_by_name(ctx: Union[FunctionContext, MethodContext], name:
     Return the expression for the specific argument.
     This helper should only be used with non-star arguments.
     """
+    # try and pull the named argument from the caller first
+    for kinds, argnames, args in zip(ctx.arg_kinds, ctx.arg_names, ctx.args):
+        for kind, argname, arg in zip(kinds, argnames, args):
+            if kind == ArgKind.ARG_NAMED and argname == name:
+                return arg
```
asottile, line 190: one could imagine a perverse class that does something like:

```python
class FK(ForeignKey[_ST, _GT]):
   def __init__(self, *args, **kwargs):
       kwargs['null'] = not kwargs.get('null', False)
```

but I'm thinking that's probably not going to happen and possibly ok to ignore?

asottile: I guess they could even do that today and trick the current code with:

```python
class FK(ForeignKey[_ST, _GT]):
    def __init__(self, *args, null: bool = False, **kwargs):
        null = not null
        super().__init__(*args, null=null, **kwargs)
```

so definitely not a problem imo :)

sobolevn: What do you think about reusing `_get_argument` from `mypy.plugins.common`? Would it solve this problem?

asottile: I think the current code is equivalent to that so it wouldn't help (_get_argument only looks at callee args)

sobolevn (approved):
Thanks!

PR #2298, opened 2024-07-31
Improve `get_context_object_name` on `SingleObjectMixin` and `MultipleObjectMixin`

flaeppe:
Get in an overload for method `get_context_object_name`

## Related issues

- Refs #873


django-stubs/views/generic/detail.pyi
```diff
@@ -7,7 +7,7 @@ from django.views.generic.base import ContextMixin, TemplateResponseMixin, View
 _M = TypeVar("_M", bound=models.Model)
 
 class SingleObjectMixin(Generic[_M], ContextMixin):
-    model: type[_M]
+    model: type[_M] | None
```
sobolevn: This can be a breaking change for a lot of code.

flaeppe: Yeah perhaps. Its default value is `None` though. So if you're declaring a view, subclassing `SingleObjectMixin` and also declare `model`. The type will be that model, `None` would go away.

Do you think we should add that as a test?

Any other, more generic, code expecting this kind of view should hopefully be a bit happy that `None` is covered.

django-stubs/views/generic/list.pyi
```diff
@@ -8,7 +8,7 @@ from django.views.generic.base import ContextMixin, TemplateResponseMixin, View
 
 _M = TypeVar("_M", bound=Model)
 
-class HasModel(Protocol):
+class _HasModel(Protocol):
```
sobolevn, line 12:
```suggestion
@type_check_only
class _HasModel(Protocol):
```

django-stubs/views/generic/detail.pyi
sobolevn: Yes, sounds good.

flaeppe: > [...] So if you're declaring a view, subclassing SingleObjectMixin and also declare model. The type will be that model, None would go away. [...]

Well, so I thought, `pyright` behaves differently in regards to this and keeps the original type 🙂

sobolevn: I think that we have to decide: practicality or purity. I think that the first one is more important in this case. `None` is just a default / sentinel value.

flaeppe: Sure, I don't have that strong opinions about the type here but I would consider it fairly easy to run in to an `AttributeError: 'NoneType' object has no attribute 'model'`

flaeppe: But if it was a larger issue, someone would've probably reported it by now. So it should probably work out just fine

sobolevn (approved):

Issue #2302, opened 2024-07-31
[5.0.3 regression] Can’t `filter` a `QuerySet` on an `alias`


andersk:
```py
from django.db import models

class Blog(models.Model):
    pass

class Entry(models.Model):
    blog = models.ForeignKey(Blog, on_delete=models.CASCADE)

def test() -> None:
    Blog.objects.alias(entries=models.Count("entry")).filter(entries__gt=5)
```

```
my_app/models.py:13: error: Cannot resolve keyword 'entries' into field. Choices are: entry, id  [misc]
Found 1 error in 1 file (checked 3 source files)
```

This was introduced by a28717df356fb767efb2d832b9146f759fd8cfd9 (#2277).

flaeppe:
I don't know if we should call this a regression? Any version prior to `5.0.3` just didn't try to check the `filter` call.

At least there's no plugin code that tries to do anything with `alias`

andersk:
From a user perspective, this was allowed in 5.0.2 and incorrectly disallowed in 5.0.3. Whatever you’d like to call it, we can’t upgrade to django-stubs ≥ 5.0.3 in [Zulip](https://github.com/zulip/zulip) until it’s addressed.

sobolevn:
Really interesting: there are no handlers for `alias` in https://github.com/typeddjango/django-stubs/blob/b76f9f4d9d4f8b4ce8f0f9f0938c81faaafc25c6/mypy_django_plugin/main.py#L168-L181

I will start digging into this!

sobolevn:
The problem happens because we use `entries__gt=5` in `filter`, which should be allowed. We need to add `alias()` support and infer field names from there.

PR #2306, opened 2024-08-01
Regression test for #2302

sobolevn:
Refs #2302

### DIFF
--- a/tests/typecheck/managers/querysets/test_filter.yml
+++ b/tests/typecheck/managers/querysets/test_filter.yml
@@ -258,6 +258,7 @@
 
 - case: alias_and_annotate_can_be_used_with_dynamic_new_fields
   main: |
+    from django.db import models
     from myapp.models import Blog
 
     Blog.objects.alias(
--- a/tests/typecheck/managers/querysets/test_filter.yml
+++ b/tests/typecheck/managers/querysets/test_filter.yml
@@ -267,6 +267,25 @@
         annotate_entries=models.Count("entry"),
     ).filter(alias_entries__gt=5, annotate_entries__gt=5)
 
+    Blog.objects.alias(
+        alias_entries=models.Count("entry"),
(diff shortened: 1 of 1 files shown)


sobolevn:
cc @flaeppe

sobolevn:
Right now this test fails with:

```
_________________ alias_and_annotate_can_be_used_with_dynamic_new_fields __________________
/Users/sobolev/Desktop/django-stubs/tests/typecheck/managers/querysets/test_filter.yml:282: 
E   Actual:
E     main:10: error: Cannot resolve keyword 'alias_entries' into field. Choices are: entry, id  [misc] (diff)
E     main:16: error: Cannot resolve keyword 'alias_entries' into field. Choices are: entry, id  [misc] (diff)
E     main:22: error: Cannot resolve keyword 'alias_e' into field. Choices are: entry, id  [misc] (diff)
E   Expected:
E     main:22: error: Cannot resolve keyword 'alias_e' into field. Choices are: entry, id  [misc] (diff)
E     E: main:22: error: Cannot resolve keyword 'alias_e' into field. Choices are...
E     A: main:10: error: Cannot resolve keyword 'alias_entries' into field. Choic...
E             ^
```

We need to:
- Check how can we easily support `.alias` (while keeping it distinct from `.annotate`)
- Make sure that the tests passx

Unrelated (I will create a new issue):

```python
    Blog.objects.alias(
        alias_entries=models.Count("entry"),
    ).annotate(
        annotate_entries=models.Count("entry"),
    ).filter(alias_wrong=5, annotate_wrong=5)
```

^ This code does not fail for some reason

flaeppe:
> This code does not fail for some reason

I would suspect that has to do with the plugin code for `.annotate`

sobolevn:
I created https://github.com/typeddjango/django-stubs/issues/2307 to track this separate issue.

PR #2308, opened 2024-08-02
Loosen bound on `CurrentSiteManager` manager type var

flaeppe:
The manager accepts any model. It's intended to work on the _relation_ __to__ `Site`, not the `Site` declaration itself

## Related issues

- Refs #2280


sobolevn (approved):
Yes, docs agree https://docs.djangoproject.com/en/5.0/ref/contrib/sites/#the-currentsitemanager

PR #2309, opened 2024-08-02
Remove `BaseManager.cache` attribute

flaeppe:
It doesn't exist at runtime and, for some reason, sneaked in via #1873 

## Related issues

- Refs #1873


sobolevn (approved):
Looks like it is added via django-cacheops.

PR #2311, opened 2024-08-02
Use `Any` for `ConnectionProxy` attributes

RealOrangeOne:
Whilst the type is generic, it is not a wrapper - `__getattr__` should not return the inner type.


sobolevn (approved):
Thanks!

Issue #754, opened 2021-11-18
`ManyToManyField` is not complete


sobolevn:
Django defines `ManyToManyField` with more precise types that we do.
Django 3.2: https://github.com/django/django/blob/3.2/django/db/models/fields/related.py#L1134-L1142
Our stubs: https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/fields/related.pyi#L180-L187

We need to fix this.

StasGC:
Hello!
I want to solve this.

flaeppe:
I think this has been closed via #909 and #1253

PR #2312, opened 2024-08-02
Declare `ManyToManyField.description` a `_StrOrPromise`

flaeppe:
The default set by Django is a translate:able string:

https://github.com/django/django/blob/617bcf611f3daa796e4054ba041089ece30a32fc/django/db/models/fields/related.py#L1342

## Related issues

- Refs #754

### DIFF
--- a/django-stubs/db/models/fields/related.pyi
+++ b/django-stubs/db/models/fields/related.pyi
@@ -242,7 +242,7 @@ _Through = TypeVar("_Through", bound=Model)
 _To = TypeVar("_To", bound=Model)
 
 class ManyToManyField(RelatedField[Any, Any], Generic[_To, _Through]):
-    description: str
+    description: _StrOrPromise
     has_null_arg: bool
     swappable: bool
 


sobolevn (approved):

django-stubs/db/models/fields/related.pyi
```diff
@@ -242,7 +242,7 @@ _Through = TypeVar("_Through", bound=Model)
 _To = TypeVar("_To", bound=Model)
 
 class ManyToManyField(RelatedField[Any, Any], Generic[_To, _Through]):
-    description: str
+    description: _StrOrPromise
```
sobolevn, line 245: Maybe it is a good time to use `StrOrPromise` instead?

flaeppe: Do you mean importing it from `django_stubs_ext`?

sobolevn: Yes, this is non blocking. We can do this later.

flaeppe: Perhaps we should do the whole file at once? To not introduce any ambiguity?

sobolevn: Yes, I agree :)
Maybe even for the whole project.

PR #2314, opened 2024-08-03
Don't populate a type arg for non generic managers

flaeppe:
We probably want a more sophisticated implementation of this to get things right when a type has multiple arguments, specifically bounded, etc. But as a first step, this can at least eliminate the obviously invalid ones.

## Related issues

- Refs https://github.com/typeddjango/django-stubs/pull/1649#issuecomment-1681846078
- Refs https://github.com/typeddjango/django-stubs/pull/1672#discussion_r1315182802

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -318,7 +318,9 @@ class AddManagers(ModelClassInitializer):
 
         assert manager_info is not None
         # Reparameterize dynamically created manager with model type
-        manager_type = Instance(manager_info, [Instance(self.model_classdef.info, [])])
+        manager_type = Instance(
+            manager_info, [Instance(self.model_classdef.info, [])] if manager_info.is_generic() else []
+        )
         self.add_new_node_to_model_class(manager_name, manager_type, is_classvar=True)
 
     def run_with_model_cls(self, model_cls: Type[Model]) -> None:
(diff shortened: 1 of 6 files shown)


mypy_django_plugin/transformers/models.py
```diff
@@ -361,11 +365,11 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
                 fallback_manager_info = self.get_or_create_manager_with_any_fallback()
                 if fallback_manager_info is not None:
                     assert self.model_classdef.info.self_type is not None
-                    self.add_new_node_to_model_class(
-                        manager_name,
-                        Instance(fallback_manager_info, [self.model_classdef.info.self_type]),
-                        is_classvar=True,
+                    manager_type = Instance(
+                        fallback_manager_info,
```
sobolevn: Let's please move this to a helper function / method.

mypy_django_plugin/transformers/models.py
```diff
@@ -445,7 +449,9 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
                     return None
             default_manager_info = generated_manager_info
 
-        default_manager = Instance(default_manager_info, [Instance(self.model_classdef.info, [])])
+        default_manager = Instance(
```
sobolevn: We have at least three places where this happens.

mypy_django_plugin/transformers/models.py
flaeppe: True, good point

sobolevn (approved):

PR #2316, opened 2024-08-04
Fixed bugs with RawQuerySet

### LINKED ISSUE #2236: `RawQuerySet`  has conflicting errors: "is not subscriptable" vs "Missing type parameter"
# Bug report

## What's wrong

Using `RawQuerySet` as type annotation is not possible. 
_Without_ the generic parameter, mypy fails with `is not subscriptable` error. 
_With_ the generic parameter, python throws `TypeError`.

```
from django.db.models.query import RawQuerySet
from models import MyModel

# TypeError: type 'RawQuerySet' is not subscriptable
def foo() -> RawQuerySet[MyModel]:                  
    return MyModel.objects.raw("SELECT id FROM my_model")

# mypy error: Missing type parameters for generic type "RawQuerySet"  [type-arg]
def bar() -> RawQuerySet:                           
    return MyModel.objects.raw("SELECT id FROM my_model")
```

## System information

- OS: Ubuntu 24.04
- `python` version: 3.12.3
- `django` version: 5.0.6
- `mypy` version: 1.10
- `django-stubs` version: 5.0.2
- `django-stubs-ext` version: 5.0.2

ashm-dev:
# I have made things!

## Related issues
Closes #2236

### DIFF
--- a/ext/django_stubs_ext/patch.py
+++ b/ext/django_stubs_ext/patch.py
@@ -22,6 +22,7 @@ from django.utils.connection import BaseConnectionHandler
 from django.views.generic.detail import SingleObjectMixin
 from django.views.generic.edit import DeletionMixin, FormMixin
 from django.views.generic.list import MultipleObjectMixin
+from django.db.models.query import RawQuerySet
 
 __all__ = ["monkeypatch"]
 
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thank you!

PR #2317, opened 2024-08-04
Add Typing for EmailField with Support for String Set and Get Types

### LINKED ISSUE #2212: `django.contrib.auth.models.User.email` allows assignment of an integer
# Bug report

`django.contrib.auth.models.User.email` allows assignment of an integer. It's not just a problem for the `User` model but for all instances of `EmailField`.

## What's wrong

```python
user = User()
user.email = 1
```

## How is that should be

There should be an "Incompatible types in assignment ..."

## System information

- OS:
- `python` version: `3.12.3`
- `django` version: `5.0.6`
- `mypy` version: `1.10.0`
- `django-stubs` version: `master`
- `django-stubs-ext` version: `master`

ashm-dev:
# I have made things!

## Related issues
Closes #2212

### DIFF
--- a/django-stubs/db/models/fields/__init__.pyi
+++ b/django-stubs/db/models/fields/__init__.pyi
@@ -356,7 +356,9 @@ class SlugField(CharField[_ST, _GT]):
         allow_unicode: bool = ...,
     ) -> None: ...
 
-class EmailField(CharField[_ST, _GT]): ...
+class EmailField(CharField[_ST, _GT]):
+    _pyi_private_set_type: str | Combinable
+    _pyi_private_get_type: str
 
 class URLField(CharField[_ST, _GT]):
     def __init__(
--- a/django-stubs/db/models/fields/__init__.pyi
+++ b/django-stubs/db/models/fields/__init__.pyi
@@ -358,7 +358,6 @@ class SlugField(CharField[_ST, _GT]):
 
 class EmailField(CharField[_ST, _GT]):
     _pyi_private_set_type: str | Combinable
-    _pyi_private_get_type: str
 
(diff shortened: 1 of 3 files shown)


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -356,7 +356,9 @@ class SlugField(CharField[_ST, _GT]):
         allow_unicode: bool = ...,
     ) -> None: ...
 
-class EmailField(CharField[_ST, _GT]): ...
+class EmailField(CharField[_ST, _GT]):
+    _pyi_private_set_type: str | Combinable
+    _pyi_private_get_type: str
```
```suggestion
    _pyi_private_set_type: str | Combinable
```

Since `CharField` already has `_pyi_private_get_type: str` set.

ashm-dev:
@sobolevn Removed the duplication, thanks for your help!)

django-stubs/db/models/sql/compiler.pyi
```diff
@@ -13,9 +13,9 @@ from django.db.models.sql.subqueries import AggregateQuery, DeleteQuery, InsertQ
 from django.utils.functional import cached_property
 from typing_extensions import TypeAlias
 
-_ParamT: TypeAlias = str | int
+_ParamT: TypeAlias = Any
```
sobolevn: This does not look related to EmailFiled change

sobolevn (changes requested):

django-stubs/db/models/sql/compiler.pyi
ashm-dev: Data from another PR accidentally got here, I'm sorry)

sobolevn (approved):
Thanks! 👍

PR #2318, opened 2024-08-04
Fixed sql types

### LINKED ISSUE #2213: db.models.sql.compiler.as_sql() return type
## What's wrong

The return type of as_sql(), set as `_AsSqlType`, results to `tuple[str, list[str | int]]`

## How is that should be

It should be `tuple[str, tuple[Any, ...]]`

## System information

- `django` version: 4.2.8 code = `return " ".join(result), tuple(params)`

ashm-dev:
# I have made things!

## Related issues
Closes #2213

### DIFF
--- a/django-stubs/db/models/sql/compiler.pyi
+++ b/django-stubs/db/models/sql/compiler.pyi
@@ -13,9 +13,9 @@ from django.db.models.sql.subqueries import AggregateQuery, DeleteQuery, InsertQ
 from django.utils.functional import cached_property
 from typing_extensions import TypeAlias
 
-_ParamT: TypeAlias = str | int
+_ParamT: TypeAlias = Any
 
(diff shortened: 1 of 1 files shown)


django-stubs/db/models/sql/compiler.pyi
```diff
@@ -13,9 +13,9 @@ from django.db.models.sql.subqueries import AggregateQuery, DeleteQuery, InsertQ
 from django.utils.functional import cached_property
 from typing_extensions import TypeAlias
 
-_ParamT: TypeAlias = str | int
+_ParamT: TypeAlias = Any
```
sobolevn, line 16: This whole module must be revisited. There're lots of incorrect annotations there :(

ashm-dev: Copy that, thank you! Then I'll close the PR for now)

PR #2319, opened 2024-08-04
Refactor support for `annotate` to utilise `mypy.types.Instance.extra_attrs`

### LINKED ISSUE #760: WithAnnotations Crash in 1.9.0
# Bug report

I have upgraded to 1.9.0 and I'm getting this crash when I run `mypy .` inside my src folder 

```
  File "/home/parallels/Development/quickcheck/vitualenv/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/home/parallels/Development/quickcheck/vitualenv/lib/python3.9/site-packages/mypy/__main__.py", line 11, in console_entry
  File "mypy/main.py", line 87, in main
  File "mypy/main.py", line 165, in run_build
  File "mypy/build.py", line 179, in build
  File "mypy/build.py", line 254, in _build
  File "mypy/build.py", line 2697, in dispatch
  File "mypy/build.py", line 3014, in process_graph
  File "mypy/build.py", line 3092, in process_fresh_modules
  File "mypy/build.py", line 1991, in fix_cross_refs
  File "mypy/fixup.py", line 26, in fixup_module
  File "mypy/fixup.py", line 90, in visit_symbol_table
  File "mypy/fixup.py", line 46, in visit_type_info
  File "mypy/fixup.py", line 92, in visit_symbol_table
  File "mypy/nodes.py", line 885, in accept
  File "mypy/fixup.py", line 137, in visit_var
  File "mypy/types.py", line 846, in accept
  File "mypy/fixup.py", line 161, in visit_instance
  File "mypy/types.py", line 846, in accept
  File "mypy/fixup.py", line 154, in visit_instance
  File "mypy/fixup.py", line 269, in lookup_qualified_typeinfo
  File "mypy/fixup.py", line 297, in lookup_qualified
  File "mypy/fixup.py", line 306, in lookup_qualified_stnode
  File "mypy/lookup.py", line 47, in lookup_fully_qualified
AssertionError: Cannot find component 'WithAnnotations[admin' for "django_stubs_ext.WithAnnotations[admin.models.MyModel, TypedDict({'count_loan_id': Any})]"
```

The crash is generated by this piece of code.

```
self.var1 = (
    MyModel.objects.filter(
        to_process_date__gte=timezone.localtime() - timedelta(days=5),
        last_attempted_datetime__isnull=False,
    )
    .values('loan_id')
    .annotate(count_loan_id=Count('loan_id'))
    .filter(count_loan_id__gte=3)
    .values_list('loan_id', flat=True)
)
```

When typing mypy doesn't crash

```
self.var1: 'QuerySet[Any]' = ...
```

## System information

- OS:
- `python` version: 3.9.2
- `django` version: 2.2.17
- `mypy` version: 0.910

flaeppe:
I figured out that the implementation in mypy for `functools.partial`, in https://github.com/python/mypy/pull/16939, added some pieces that we can utilise to get better support for `.annotate`.

Namely that https://github.com/python/mypy/pull/16939 added support for writing `Instance.extra_attrs` to cache.

There are 2 main pieces at play in this implementation:

1. Annotated attributes are written to `mypy.types.Instance.extra_attrs`. This allows `.annotate` calls to bring in temporary attributes on same model instances.
2. The approach in 1. does not help out when comparing instance `A´1` with `A´2` for annotated attributes. Essentially nothing in `.extra_attrs` is considered when comparing instances, there could be completely different attributes in there and mypy still considers the instances equal/compatible. To be able to resolve that, the plugin creates a subclass for each model that also has `django_stubs_ext.Annotations` as base with a generic parameter. This subclass will be replace any encounter of `WithAnnotations[<Model>, <TypedDict>]`, where `<TypedDict>` populates the generic parameter of the model subclass. This allows for falling back on mypy for type compatibility depending on annotated attributes.

The description in 2. is something like this in code

```python
class MyModel(models.Model):
    username = models.CharField(max_length=100)

class MyModel_AnnotatedWith(models.Model, django_stubs_ext.Annotations[django_stubs_ext.annotations._Annotations]):
    ...
```

Then with a use case example

```python
class AnnotatedAttrs(TypedDict):
    foo: int

def func(instance: django_stubs_ext.WithAnnotations[MyModel, AnnotatedAttrs]) -> int:
    return instance.foo

func(MyModel.objects.annotate(foo=Value("")).get())  # OK
func(MyModel.objects.annotate(bar=Value("")).get())  # Error

reveal_type(func)  # Revealed type is "def (instance: MyModel@AnnotatedWith[TypedDict('AnnotatedAttrs', {'foo': builtins.int})]) -> builtins.int"
```

The main downside here is the subclass and that it needs a name, which is set to `{Model name}@AnnotatedWith`

---

Most of this is backwards compatible with our old `WithAnnotations` except for the fact that;

- `WithAnnotations[MyModel]` is no longer supported, it didn't really make sense to ask for an annotation of nothing. Though this enabled support for accepting any attribute in the previous version, but IMO `# type: ignore` will make do for that.

It might be more compatibility issues but our test suite is kind of lacking cases for using `.annotate`.

Let me know what you think, any feedback is greatly appreciated here.

## Related issues

There are a whole bunch of issues reported in regards to `annotate`: https://github.com/typeddjango/django-stubs/issues?q=is%3Aissue+is%3Aopen+annotate

I'm not going to hunt down and include which repro cases are resolved with _this PR_, but was thinking to do that subsequently to close off resolved issues. Though I will include closing one (major) cache issue we've had forever, since that should no longer be any issue with this implementation.

- Closes #760

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,11 +211,11 @@ class MyManager(models.Manager["MyModel"]):
 
 ### How do I annotate cases where I called QuerySet.annotate?
 
-Django-stubs provides a special type, `django_stubs_ext.WithAnnotations[Model]`, which indicates that the `Model` has
-been annotated, meaning it allows getting/setting extra attributes on the model instance.
+Django-stubs provides a special type, `django_stubs_ext.WithAnnotations[Model, <Annotations>]`, which indicates that
+the `Model` has been annotated, meaning it requires extra attributes on the model instance.
 
-Optionally, you can provide a `TypedDict` of these attributes,
-e.g. `WithAnnotations[MyModel, MyTypedDict]`, to specify which annotated attributes are present.
+You should provide a `TypedDict` of these attributes, e.g. `WithAnnotations[MyModel, MyTypedDict]`, to specify which
+annotated attributes are present.
 
 Currently, the mypy plugin can recognize that specific names were passed to `QuerySet.annotate` and
 include them in the type, but does not record the types of these attributes.
@@ -235,21 +235,11 @@ class MyModel(models.Model):
     username = models.CharField(max_length=100)
 
 
-def func(m: WithAnnotations[MyModel]) -> str:
-    return m.asdf  # OK, since the model is annotated as allowing any attribute
-
-
-func(MyModel.objects.annotate(foo=Value("")).get(id=1))  # OK
-func(
-    MyModel.objects.get(id=1)
-)  # Error, since this model will not allow access to any attribute
-
-
 class MyTypedDict(TypedDict):
     foo: str
 
 
-def func2(m: WithAnnotations[MyModel, MyTypedDict]) -> str:
+def func(m: WithAnnotations[MyModel, MyTypedDict]) -> str:
     print(m.bar)  # Error, since field "bar" is not in MyModel or MyTypedDict.
     return m.foo  # OK, since we said field "foo" was allowed
 
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -220,7 +220,7 @@ class AddAnnotateUtilities(ModelClassInitializer):
         class MyModel(models.Model):
             ...
 
-        class MyModel_AnnotatedWith(MyModel, django_stubs_ext.Annotations[_Annotations]):
+        class MyModel@AnnotatedWith(MyModel, django_stubs_ext.Annotations[_Annotations]):
             ...
     """
 
@@ -228,7 +228,7 @@ class AddAnnotateUtilities(ModelClassInitializer):
         annotations = self.lookup_typeinfo_or_incomplete_defn_error("django_stubs_ext.Annotations")
         object_does_not_exist = self.lookup_typeinfo_or_incomplete_defn_error(fullnames.OBJECT_DOES_NOT_EXIST)
         multiple_objects_returned = self.lookup_typeinfo_or_incomplete_defn_error(fullnames.MULTIPLE_OBJECTS_RETURNED)
-        annotated_model_name = self.model_classdef.info.name + "_AnnotatedWith"
+        annotated_model_name = self.model_classdef.info.name + "@AnnotatedWith"
         annotated_model = self.lookup_typeinfo(self.model_classdef.info.module_name + "." + annotated_model_name)
         if annotated_model is None:
             model_type = fill_typevars_with_any(self.model_classdef.info)
@@ -1115,6 +1115,10 @@ def set_auth_user_model_boolean_fields(ctx: AttributeContext, django_context: Dj
 
 
 def handle_annotated_type(ctx: AnalyzeTypeContext, fullname: str) -> MypyType:
+    """
+    Replaces the 'WithAnnotations' type with a type that can represent an annotated
+    model.
+    """
     is_with_annotations = fullname == fullnames.WITH_ANNOTATIONS_FULLNAME
     args = ctx.type.args
     if not args:
@@ -1149,13 +1153,7 @@ def get_annotated_type(
     api: Union[SemanticAnalyzer, TypeChecker], model_type: Instance, fields_dict: TypedDictType
 ) -> ProperType:
     """
-
-    Get or create the type for a model for which you getting/setting any attr is allowed.
-
-    The generated type is an subclass of the model and django._AnyAttrAllowed.
-    The generated type is placed in the django_stubs_ext module, with the name WithAnnotations[ModelName].
-    If the user wanted to annotate their code using this type, then this is the annotation they would use.
-    This is a bit of a hack to make a pretty type for error messages and which would make sense for users.
+    Get a model type that can be used to represent an annotated model
     """
     if model_type.extra_attrs:
         extra_attrs = ExtraAttrs(
@@ -1180,7 +1178,7 @@ def get_annotated_type(
                 required_keys={*model_type.args[0].required_keys, *fields_dict.required_keys},
             )
     else:
-        annotated_model = helpers.lookup_fully_qualified_typeinfo(api, model_type.type.fullname + "_AnnotatedWith")
+        annotated_model = helpers.lookup_fully_qualified_typeinfo(api, model_type.type.fullname + "@AnnotatedWith")
 
     if annotated_model is None:
         return model_type
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -22,19 +22,29 @@ from mypy.nodes import (
     TypeInfo,
     Var,
 )
-from mypy.plugin import AnalyzeTypeContext, AttributeContext, CheckerPluginInterface, ClassDefContext
+from mypy.plugin import AnalyzeTypeContext, AttributeContext, ClassDefContext
 from mypy.plugins import common
 from mypy.semanal import SemanticAnalyzer
 from mypy.typeanal import TypeAnalyser
-from mypy.types import AnyType, Instance, ProperType, TypedDictType, TypeOfAny, TypeType, TypeVarType, get_proper_type
+from mypy.types import (
+    AnyType,
+    ExtraAttrs,
+    Instance,
+    ProperType,
+    TypedDictType,
+    TypeOfAny,
+    TypeType,
+    TypeVarType,
+    get_proper_type,
+)
 from mypy.types import Type as MypyType
-from mypy.typevars import fill_typevars
+from mypy.typevars import fill_typevars, fill_typevars_with_any
 
 from mypy_django_plugin.django.context import DjangoContext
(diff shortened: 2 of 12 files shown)


mypy_django_plugin/transformers/querysets.py
```diff
@@ -237,25 +241,48 @@ def extract_proper_type_queryset_annotate(ctx: MethodContext, django_context: Dj
...
+        #   reveal_type(x)  # MyModel_WithAnnotations[TypedDict({"foo": Any})]
+        #   # Then, on an attribute access of 'field' like
+        #   reveal_type(x.field)
+        #
+        # Now 'CharField.__get__', which is overloaded, is passed 'x' as the 'model'
+        # argument and mypy consider it ambiguous to decide which overload method to
+        # select due to the 'Any' in 'TypedDict({"foo": Any})'. But if we specify the
+        # 'Any' as 'TypeOfAny.special_form' mypy doesn't consider the model instance to
+        # contain 'Any' and the ambiguity goes away.
+        field_types = {name: AnyType(TypeOfAny.special_form) for name, _ in kwargs.items()}
```
flaeppe, line 276: I don't know if what I try to describe here is a mypy problem, I tried to reproduce this issue in mypy playground but wasn't able to.

This is as far as I got, but without getting any error: https://mypy-play.net/?mypy=latest&python=3.12&gist=e4745bfc8e2f42b02911adec2214e6e5

```python
from typing import Any, Generic, Mapping, TypeVar, overload, TypedDict

_Annotations = TypeVar("_Annotations", covariant=True, bound=Mapping[str, Any])

class Annotations(Generic[_Annotations]): ...

T = TypeVar("T")

class Field(Generic[T]):
    @overload
    def __get__(self, instance: None, owner: Any) -> None: ...
    @overload
    def __get__(self, instance: Model, owner: Any) -> T: ...
    def __get__(self, instance: Model | None, owner: Any) -> T | None: ...

class Model:
    field = Field[str]()

class Foo(TypedDict):
    foo: Any

class A(Model, Annotations[_Annotations]):
    ...
    
a = A[Foo]()
reveal_type(a)
reveal_type(a.field)
```

README.md
```diff
@@ -235,21 +235,11 @@ class MyModel(models.Model):
     username = models.CharField(max_length=100)
 
 
-def func(m: WithAnnotations[MyModel]) -> str:
```
sobolevn, line 238: I don't think that we should remove this feature. Because I see that a lot of users discuss this and use this. Maybe we can treat `WithAnnotations[MyModel]` as `WithAnnotations[MyModel, Any]` implicitly? Via type var defaults or something?

This way we can continue to support this code (maybe with a deprecation warning, why not). But breaking it - does not seem right to me :(

sobolevn: But, I agree that we should not have done this in the first place.

mypy_django_plugin/transformers/models.py
```diff
@@ -200,6 +210,50 @@ def run_with_model_cls(self, model_cls: Type[Model]) -> None:
...
+
+        class MyModel_AnnotatedWith(MyModel, django_stubs_ext.Annotations[_Annotations]):
+            ...
+    """
+
+    def run(self) -> None:
+        annotations = self.lookup_typeinfo_or_incomplete_defn_error("django_stubs_ext.Annotations")
+        object_does_not_exist = self.lookup_typeinfo_or_incomplete_defn_error(fullnames.OBJECT_DOES_NOT_EXIST)
+        multiple_objects_returned = self.lookup_typeinfo_or_incomplete_defn_error(fullnames.MULTIPLE_OBJECTS_RETURNED)
+        annotated_model_name = self.model_classdef.info.name + "_AnnotatedWith"
```
sobolevn: this can conflict with a real name. Let's use `ModelName@AnnotatedWith` as a name, it cannot exist in Python, so we are fine.

tests/typecheck/test_annotated.yml
```diff
@@ -11,3 +12,20 @@
         @dataclass(unsafe_hash=True)
         class RatingComposite:
             max_value: Annotated[int, IntegerType(min_value=1, max_value=10)] = 5
+
+        class Obj(TypedDict):
+            foo: int
+
+        class X:
+            x: Annotated[Blog, Obj]
+        reveal_type(X().x)  # N: Revealed type is "myapp.models.Blog"
```
sobolevn, line 21:
```suggestion
        reveal_type(X().x)  # N: Revealed type is "myapp.models.Blog"
        reveal_type(X().x.foo)  # N: Revealed type is "builtins.int"
```

sobolevn (commented):
Thank you! Very good idea!

mypy_django_plugin/transformers/models.py
flaeppe: Ah great idea

README.md
flaeppe: Yeah, I suppose, gonna have to see if I can get it working. The support for any/arbitrary attributes is a bit of work I think

tests/typecheck/test_annotated.yml
flaeppe: A bit confusing here, but `.foo` shouldn't be recognised as an attribute. I'm just testing that our hook that listens to `Annotated` doesn't act on any kind of `Annotated`.

But I'll add in what the expected output is

sobolevn: Oh sorry, you are right. `.foo` should not be an attribute.

README.md
sobolevn: We only need to support `Any`, no other arguments / types.

flaeppe: Yes, the `Any` type argument is fine. But that doesn't add support for accessing an arbitrary attribute, like the previous implementation supported.

Essentially you'll still get something like:

```python
def func(m: WithAnnotations[MyModel]) -> str:
    return m.asdf  # E: MyModel@AnnotatedWith[Any] has no attribute "asdf"
```

I don't have any solution that removes the "has no attribute" error.

flaeppe: Just to clarify: the current implementation replaces `WithAnnotations[MyModel]` with `MyModel`. So we support writing `WithAnnotations[MyModel]` but will now yield "has no attribute" errors on accessing attributes unknown to `MyModel`

sobolevn: We can add `__getattr__(self, name: str) -> Any` to `Model@AnnotatedWith[Any]` types. This way all known attributes will have correct types, all unknown ones can be used. I will also suggest to add an options `--strict-model-annotations` with `False` as default for backward compatibility. What do you think?

flaeppe: There's 1 problem that I see with that, unless I'm missing something, which is that `Model@AnnotatedWith[Any]` is an `Instance` and not a `TypeInfo`. But we need to add the `__getattr__` method to the `TypeInfo` though if we do, all instances of `Model@AnnotatedWith` gets the `__getattr__` method

sobolevn: Won't `.extra_attrs` do the job for single instance `__getattr__`? 🤔 
I am not sure, just asking.

flaeppe: `__getattr__` will work for the arbitrary attributes. But we still want/need to be able to match the model type on e.g. a function argument.

This whole thing is kind of twofold.

1. Subclassing the model and the generic TypedDict argument for proper matching for `WithAnnotations` as a function argument
2. `extra_attrs` for accessing declared annotated attributes e.g. inside a function body

`__getattr__` would help out with attributes accessed inside the body. But I can't find a way to add it and keep __1.__ working and not affect all other instances. I've also tried adding the `__getattr__` method to `extra_attrs` with no luck: `ExtraAttrs(attrs={"__getattr__": CallableType(...)}, ...)`

There's also the `TypeInfo.fallback_to_any` which is the same as subclassing `Any` or our `AnyAttrAllowed`. But same issue here, it affects the type and not instances of said type.

sobolevn: Thanks a lot for your research! Not really what I hoped for :(

In this case, let's do the cleanest solution and break `WithAnnotations[M]` usage in a way that we restrict known attributes.

flaeppe: I was also hoping to find a way that kept backwards compatibility. At least I'm hoping breaking this will make it a bit easier going forward.

Do you mean restrict known attributes like emitting `"has no attribute"` errors?

Then, calling wise, if we replace

```python
WithAnnotations[MyModel] -> MyModel@AnnotatedWith[Any]
```

We can require that `.annotate` has been called, but we don't care with what. Just like before. __I'm thinking this is the implementation we want__.

Else if we replace like (this is what we do now)

```python
WithAnnotations[MyModel] -> MyModel
```

This wouldn't distinguish between having called `.annotate` or not. Meaning that there's no requirement of having called `.annotate` but having done it is also accepted.

sobolevn (approved):
Let's go for it!

flaeppe:
Agreed, let's do it. 🚀

PR #2322, opened 2024-08-06
Do `models.Model` adjustments from `get_metaclass_hook`

### LINKED ISSUE #2304: [5.0.3 regression] error: "type[RealmFilter]" has no attribute "objects"  [attr-defined]

flaeppe:
The called hook expects/wants the class body analyzed, but the `get_customize_class_mro_hook` hook, that it was previously called from, invokes the hook before the class body has been analyzed..

Docs: https://mypy.readthedocs.io/en/stable/extending_mypy.html#current-list-of-plugin-hooks

> __get_customize_class_mro_hook()__ can be used to modify class MRO (for example insert some entries there) before the class body is analyzed.

## Related issues

- Closes #2304 
- Refs #2280

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -223,9 +223,6 @@ class NewSemanalDjangoPlugin(Plugin):
         return None
 
     def get_customize_class_mro_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
-        if fullname == fullnames.MODEL_CLASS_FULLNAME:
-            return MetaclassAdjustments.adjust_model_class
-
         sym = self.lookup_fully_qualified(fullname)
         if (
             sym is not None
@@ -235,6 +232,11 @@ class NewSemanalDjangoPlugin(Plugin):
             return reparametrize_any_manager_hook
         else:
             return None
+
+    def get_metaclass_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
+        if fullname == fullnames.MODEL_METACLASS_FULLNAME:
+            return MetaclassAdjustments.adjust_model_class
+        return None
 
     def get_base_class_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
         # Base class is a Model class definition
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks! I think that we didn't have this hook back in the days of the original code.

PR #2331, opened 2024-08-09
5.1: Deprecate CheckConstraint.check

93578237:
# I have made things!
> [The check keyword argument of CheckConstraint is deprecated in favor of condition.](https://docs.djangoproject.com/en/5.1/releases/5.1/#features-deprecated-in-5-1)

https://github.com/django/django/blob/ecf13f192de7c26a7606f5ba46a3773a9c84ef72/django/db/models/constraints.py#L137

Added overload to forbid check=None, condition=None at the same

### DIFF
--- a/django-stubs/db/models/constraints.pyi
+++ b/django-stubs/db/models/constraints.pyi
@@ -46,8 +46,8 @@ class CheckConstraint(BaseConstraint):
     check: Q | BaseExpression
     condition: Q | BaseExpression
 
-    @deprecated("The check keyword argument is deprecated in favor of condition and will be removed in Django 6.0")
     @overload
+    @deprecated("The check keyword argument is deprecated in favor of condition and will be removed in Django 6.0")
     def __init__(
         self,
         *,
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
Please, use `typing_extensions.deprecated` on the appropriate overload instead

django-stubs/db/models/constraints.pyi
```diff
@@ -44,11 +44,24 @@ class BaseConstraint:
 
 class CheckConstraint(BaseConstraint):
     check: Q | BaseExpression
+    condition: Q | BaseExpression
+
+    @deprecated("The check keyword argument is deprecated in favor of condition and will be removed in Django 6.0")
+    @overload
```
Viicos: PEP 702 [suggests](https://peps.python.org/pep-0702/#example) using `@overload` and then `@deprecated`

93578237: Updated

django-stubs/db/models/constraints.pyi
```diff
@@ -46,8 +46,8 @@ class CheckConstraint(BaseConstraint):
     check: Q | BaseExpression
     condition: Q | BaseExpression
 
-    @deprecated("The check keyword argument is deprecated in favor of condition and will be removed in Django 6.0")
     @overload
+    @deprecated("The check keyword argument is deprecated in favor of condition and will be removed in Django 6.0")
```
sobolevn, line 50: Please, add a pyright ignore to 
Снимок экрана 2024-08-22 в 12 46 29

sobolevn (approved):
Thanks!

PR #2332, opened 2024-08-09
Check annotated fields in `<QuerySet>.filter` call

### LINKED ISSUE #2307: `.filter` with wrong fields does not fail if `.annotate` is used
```yml
- case: repro
  main: |
    from django.db import models
    from myapp.models import Blog

    Blog.objects.annotate(
        annotate_entries=models.Count("entry"),
    ).filter(annotate_wrong__gt=5)

    Blog.objects.alias(
        alias_entries=models.Count("entry"),
    ).annotate(
        annotate_entries=models.Count("entry"),
    ).filter(alias_wrong__gt=5, annotate_wrong__gt=5)
  installed_apps:
    - myapp
  files:
    - path: myapp/__init__.py
    - path: myapp/models.py
      content: |
        from django.db import models

        class Blog(models.Model):
            pass

        class Entry(models.Model):
```

This does not look right.

flaeppe:
## Related issues

- Closes #2307

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -472,11 +472,23 @@ class DjangoContext:
             raise LookupsAreUnsupported()
         return self._resolve_field_from_parts(field_parts, model_cls)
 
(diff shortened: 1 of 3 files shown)


sobolevn (approved):
Thank you!

PR #2333, opened 2024-08-09
Enable `mypy.plugins.proper_plugin`

flaeppe:
Refs: https://mypy.readthedocs.io/en/stable/extending_mypy.html#useful-tools

This resulted in quite a lot of changes.. I think there are multiple places in our plugin functions where we should use `ProperType` as return type instead of `MypyType`(our alias for `mypy.types.Type`), which could perhaps remove a couple of `get_proper_type`.

Not at all sure how much impact this could have, if any.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -20,7 +20,8 @@ force_union_syntax = true
 
 
 plugins =
-    mypy_django_plugin.main
+    mypy_django_plugin.main,
+    mypy.plugins.proper_plugin
 
 # Ignore incomplete hints in yaml-stubs
 [mypy-yaml.*]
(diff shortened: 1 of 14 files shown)


mypy.ini
```diff
@@ -20,7 +20,8 @@ force_union_syntax = true
 
 
 plugins =
-    mypy_django_plugin.main
+    mypy_django_plugin.main,
+    mypy.plugins.proper_plugin
```
sobolevn, line 24: Thanks! I 100% forgot about it 😆 
But, using it in django-stubs was my motivation for adding it to the `mypy.` namespace!

sobolevn (approved):
Thanks!

mypy.ini
flaeppe: Haha, yeah I happened to come across the documentation and started checking it out

PR #2334, opened 2024-08-09
Use mypy's `make_optional_type` instead of shipping our own

flaeppe:
The builtin one is a bit more sophisticated


sobolevn (approved):

sobolevn:
Please, rebase :)

PR #2342, opened 2024-08-18
Add arbitrary named arguments to @receiver decorator

craigds:
`@receiver` accepts arbitrary keyword arguments - per [the docs](https://docs.djangoproject.com/en/5.1/topics/signals/#django.dispatch.receiver)

This comes in handy for some third party apps.


sobolevn (approved):
Thanks!

PR #2348, opened 2024-08-21
feat: Add missing stubs

BradleyKirton:
This PR adds stubs for `login_not_required` which is new in django 5.1 and the `_unicode_ci_compare` function in the `django.contrib.auth.forms` module.


django-stubs/contrib/auth/forms.pyi
```diff
@@ -15,6 +15,8 @@ from typing_extensions import TypeAlias
 UserModel: TypeAlias = type[_UserModel]
 _User = TypeVar("_User", bound=AbstractBaseUser)
 
+def _unicode_ci_compare(s1: str, s2: str) -> bool: ...
```
sobolevn: We generally tend to ignore protected names in stubs. Can you please explain the use-case for this one?

BradleyKirton: Hi :wave:

I made use of that function in a project and noticed the type information was missing in my editor.

I did consider just copying the function into my code given it is so small.

I can amend the commit without this stub if you like.

sobolevn: yes, please :)

sobolevn (approved):
Thanks!

PR #2349, opened 2024-08-22
Update `db.migrations.operations`

Viicos:
# I have made things!


django-stubs/db/migrations/operations/base.pyi
```diff
@@ -1,18 +1,28 @@
...
 from django.db.models import Model
 from typing_extensions import Self
 
+class OperationCategory(str, Enum):
+    ADDITION = "+"
+    REMOVAL = "-"
+    ALTERATION = "~"
+    PYTHON = "p"
+    SQL = "s"
+    MIXED = "?"
```
Viicos, line 17: Having the actual values seems to be [recommended](https://typing.readthedocs.io/en/latest/guides/writing_stubs.html#classes).

django-stubs/db/migrations/operations/base.pyi
```diff
@@ -1,18 +1,28 @@
...
-    reduces_to_sql: bool
-    atomic: bool
-    elidable: bool
-    serialization_expand_args: list[str]
+    reversible: ClassVar[bool]
+    reduces_to_sql: ClassVar[bool]
+    atomic: ClassVar[bool]
+    elidable: ClassVar[bool]
+    serialization_expand_args: ClassVar[list[str]]
+    category: ClassVar[OperationCategory | None]
```
Viicos, line 25: I believe they are all class vars?

Viicos:
The stubtest error is annoying.. Looks like mixing `str` and `Enum` doesn't play well

django-stubs/db/migrations/operations/base.pyi
```diff
@@ -1,18 +1,28 @@
...
     _constructor_args: tuple[Sequence[Any], dict[str, Any]]
```
sobolevn: Why do we have this protected member here? Is it still useful?

sobolevn (commented):
Thanks, just one question

django-stubs/db/migrations/operations/base.pyi
adamchainz: It was added in the big PR #909. I think we can remove it until a concrete case pops up.

```suggestion
```

adamchainz (approved):

adamchainz:
Thanks @Viicos and @sobolevn

PR #2351, opened 2024-08-24
Update `contrib.sessions.backends`

Viicos:
# I have made things!


django-stubs/contrib/sessions/backends/base.pyi
```diff
@@ -1,29 +1,61 @@
...
     def __init__(self, session_key: str | None = None) -> None: ...
+    async def aset(self, key: str, value: Any) -> None: ...
     @property
     def key_salt(self) -> str: ...
+    @overload
+    async def aget(self, key: str) -> Any | None: ...
+    @overload
+    async def aget(self, key: str, default: Any) -> Any: ...
+    @overload
+    async def aget(self, key: str, default: _T) -> Any | _T: ...  # type: ignore[misc]
```
sobolevn: I would suggest to remove this case, `Any | _T` is still `Any`

sobolevn (commented):
Thank you! A lot of useful fixes 👍

django-stubs/contrib/sessions/backends/base.pyi
Viicos: I copied the signatures over from typeshed, but indeed because `_VT` is `Any` we can simplify this.

Viicos: However, pyright may retain the type information:

image


For a better user experience, maybe this can be useful?

sobolevn: I would still prefer not to complicate the signature, cause it does not provide anything extra, but is still visible in the autosuggestions / docs / etc.

django-stubs/contrib/sessions/backends/base.pyi
```diff
@@ -1,29 +1,59 @@
...
+from typing import Any, TypeVar, overload
+
+from _typeshed import SupportsKeysAndGetItem
 
 VALID_KEY_CHARS: Any
 
 class CreateError(Exception): ...
 class UpdateError(Exception): ...
 
+_T = TypeVar("_T")
```
sobolevn: It is now unused.

django-stubs/contrib/sessions/backends/base.pyi
```diff
@@ -1,29 +1,59 @@
...
+    @overload
+    async def aget(self, key: str) -> Any | None: ...
+    @overload
+    async def aget(self, key: str, default: Any) -> Any: ...
+    @overload
+    async def apop(self, key: str) -> Any: ...
+    @overload
+    async def apop(self, key: str, default: Any) -> Any: ...
+    @overload
+    async def apop(self, key: str, default: _T) -> Any | _T: ...  # type: ignore[misc]
```
sobolevn:
```suggestion
```

sobolevn (approved):
Thanks, great work!

PR #2363, opened 2024-09-09
Allow None return form model Field.formfield()

adamchainz:
# I have made things!

`None` is allowed, for example, from `AutoFieldMixin`: https://github.com/django/django/blob/cdbd31960e0cf41063b3efac97292ee0ccc262bb/django/db/models/fields/__init__.py#L2828-L2829

`None`s mean “ignore this field” when constructing a model form: https://github.com/django/django/blob/cdbd31960e0cf41063b3efac97292ee0ccc262bb/django/forms/models.py#L236-L248

I opened [Django ticket #35748](https://code.djangoproject.com/ticket/35748) to document this possibility.

## Related issues

n/a


sobolevn (approved):
Thanks! TIL

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -205,7 +205,7 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
         form_class: type[forms.Field] | None = None,
         choices_form_class: type[forms.ChoiceField] | None = None,
         **kwargs: Any,
-    ) -> forms.Field: ...
+    ) -> forms.Field | None: ...
```
asottile-sentry, line 208: hmmm so this is triggering a bunch of false positives after upgrading django-stubs

it seems to me that `django.db.models.fields.Field.formfield` *always* returns a `Field` so this annotation is wrong (?) and subclasses should override it to be `| None` if applicable (?) (even though it would violate substitutability -- but such is django)

adamchainz: Did you read the associated ticket and docs change? Any `Field` subclass may return `None` from `formfield()` to remove it from forms.

In Django itself, `AutoFieldMixin` and `OneToOneField` use this feature. I have used this feature at least once in my MySQL package: https://github.com/adamchainz/django-mysql/blob/1ba0471c06bf7e3b47a1ee2c544fc3738020611e/src/django_mysql/models/fields/dynamic.py#L295-L299

Adding the `None` option the base class fixes substitutability, no?

asottile-sentry: yes but `fields.Field` does not return None, and my subclass of `fields.Field` inherits the same implementation which does not return `None` -- and now the type checker is mad about that

asottile-sentry: for example this: https://github.com/adamchainz/django-jsonfield/blob/939c5a7b8c5414d2b63aa141899c9059b8ce79ec/jsonfield/tests/test_fields.py#L55-L57

adamchainz: Sure.. I'm happy adding `assert formfield is not None` to such tests. That's just type-hinted life, in my experience.

PR #2374, opened 2024-09-17
Version 5.1.0 release

richardebeling:
Blindly following the instructions in [Contributing.md](https://github.com/typeddjango/django-stubs/blob/master/CONTRIBUTING.md#releasing-django-stubs). pre-commit passes, stubtest passes, I hope that means that this PR makes sense and there are no release blockers.

Version 5.0.4 broke with with Django 5.1 in one of the projects I'm working on due to the `check`/`condition` rename that was fixed in #2331 but not yet released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ We rely on different `django` and `mypy` versions:
 
 | django-stubs   | Mypy version | Django version | Django partial support | Python version |
 |----------------|--------------|----------------|------------------------|----------------|
+| 5.1.0          | 1.11.x       | 5.1            | 4.2                    | 3.8 - 3.12     |
 | 5.0.4          | 1.11.x       | 5.0            | 4.2                    | 3.8 - 3.12     |
 | 5.0.3          | 1.11.x       | 5.0            | 4.2                    | 3.8 - 3.12     |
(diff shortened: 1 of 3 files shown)


flaeppe (commented):
I'm suspecting that releasing at this point will result in quite a bit of reports regarding annotate/alias

I've been meaning to dig in to some of the more "normal" usage but haven't gotten around to it yet

Perhaps it's for the better to just release and see what happens

sobolevn:
> a bit of reports regarding annotate/alias

Sorry, I am out of context. What is this about?

flaeppe:
I might be misremembering, but I think I had some additional case that would be good to cover in our tests for annotate. Think I had encountered some simple usage that was breaking.

But I have to check if I wrote anything down or if it was from somewhere else

richardebeling:
@flaeppe any updates on release blockers?

flaeppe:
Sorry, I meant to comment here that I think it was along the lines of what #2376 resolved. So I don't have any blockers.

flaeppe (approved):

flaeppe:
New versions released just now:

- https://pypi.org/project/django-stubs/5.1.0/
- https://pypi.org/project/django-stubs-ext/5.1.0/

PR #2376, opened 2024-09-20
Allow lookups on `.annotate` fields

sterliakov:
# Allow lookups on fields coming from `.annotate()`

Django does not object when fields introduced via `QuerySet.annotate` are used in `.filter` with lookup chains:

```
from django.db import models

class Blog(models.Model): pass

Blog.objects.annotate(id2=models.F('id')).filter(id2__gt=10)
```

`mypy`, however, rejects this with a really cryptic error message:

```
error: Cannot resolve keyword 'id2' into field. Choices are: id, id2  [misc]
```

Since now all `.annotate` fields are resolved as plain `Any`, I think it's fair to allow chaining any lookups with them. This PR does exactly that: when a `id2__something` lookup is encountered and `id2` is not a model field, proceed with checking that `id2` is `.annotate` field and accept if so, ignoring the chained lookups.

### DIFF
--- a/tests/typecheck/managers/querysets/test_annotate.yml
+++ b/tests/typecheck/managers/querysets/test_annotate.yml
@@ -257,7 +257,7 @@
                     username = models.CharField(max_length=100)
 
 
--   case: annotate_currently_allows_lookups_of_non_existent_field
+-   case: annotate_rejects_lookups_of_non_existent_field
     main: |
         from myapp.models import User
         from django.db.models.expressions import F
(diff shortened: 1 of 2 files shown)


tests/typecheck/managers/querysets/test_annotate.yml
```diff
@@ -395,3 +395,22 @@
...
+- case: test_annotate_allows_any_lookups_in_filter
+  main: |
+    from django.db import models
+    from myapp.models import Blog
+
+    qs = Blog.objects.annotate(distance=0)
+    qs.filter(distance=10)
+    qs.filter(distance__lt=10)
+    qs.filter(distance__lt__lt=10)
+    qs.filter(distance__unknown_lookup=10)
```
flaeppe: Could we reveal `qs` and 1 `filter` at least so we know we're working with an expected type here (and not `Any` or something swallowing whatever)

flaeppe (commented):
Nice, just one fix with the tests

sterliakov:
Tangentially related to #2306 and #2302? This should get rid of lookup problem there, but still doesn't support `alias`.

I'll grab `.alias()` as well, it also causes problems in my codebase:)

flaeppe:
I think we need a third type argument on `QuerySet` to resolve alias?

Or at least that might be one alternative way

flaeppe (approved):

PR #2377, opened 2024-09-20
Break what apparently is a cycle involving custom User model and `QuerySet.as_manager()`

sterliakov:
# Break a cycle resulting in `[django-manager-missing]`

I'm not sure if my treatment is correct, perhaps I'm just treating symptoms of some deeper problem. However, this at least makes a test case extracted from my real system pass, and removes a dozen of `django-manager-missing` errors in my code (actually all of them!).

See the contributed test case for details - I don't really see how to summarize it any better.

There seems to be some cycle during semanal, where `base_as_manager.type` remains `None` during all iterations (yes, I tried deferring first). This seems to only affect the base `QuerySet` and require custom `AUTH_USER_MODEL`.

Prior to the fix, the test failed with the following:

```
main:2: note: Revealed type is "payments.models.UnknownManager[payments.models.Payout]" (diff)
payments/models:10: error: Couldn't resolve related manager 'payouts' for relation 'payments.models.Payout.triggered_by'.  [django-manager-missing] (diff)
payments/models:24: error: Could not resolve manager type for "payments.models.Payout.objects"  [django-manager-missing] (diff)
```

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -512,7 +512,13 @@ def add_as_manager_to_queryset_class(ctx: ClassDefContext) -> None:
     if base_as_manager is None:
         return
 
-    base_ret_type: Optional[TypeInfo] = None
+    manager_sym = semanal_api.lookup_fully_qualified_or_none(fullnames.MANAGER_CLASS_FULLNAME)
+    if manager_sym is None or not isinstance(manager_sym.node, TypeInfo):
+        return _defer()
+
+    manager_base = manager_sym.node
+    base_ret_type = manager_base
+
     if base_as_manager.fullname != f"{fullnames.QUERYSET_CLASS_FULLNAME}.as_manager":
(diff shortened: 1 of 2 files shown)


sterliakov:
@flaeppe honestly, I don't really like this one - feels like a hack. If you have some time, could you have a look into what might be going on here? I already spent ~2 days here, but can't figure out why `QuerySet.as_manager` remains unresolved after multiple semanal passes.

flaeppe:
I'm not sure you can expect something to get fully typed during semantic analysis? I think you need the semantic analysis of types, type analysis or whatever, and then use a hook for when mypy has decided to enter that phase.

I can't say I'm totally sure about this stuff, but perhaps it could be of some help or give some pointers or something

flaeppe (approved):
I think this looks fine. It adds special treatment to the `QuerySet.as_manager` method. Apparently because we can't expect it to have gotten types when the `get_base_class_hook` comes around.

We also add expectations on what its return value is, which is a bit of cheating, but it's not likely to change and a fair compromise.

PR #2380, opened 2024-09-24
Add django 5.1 to the test matrix and add its classifier

foarsitter:
Thank you for bringing the 5.1.0 release!

According to the table in the README.md, there is support for Django 5.1. To ensure we're aligned, I've added 5.1 to the test matrix and updated the classifier in setup.py. Additionally, I included a task to the contributiong guidelines in case there is support for new Python or Django versions.


sobolevn (approved):
Thanks!

PR #2384, opened 2024-09-27
create and re-use `TypeAlias`es and `TypeVar`s for "user" and "any user"

terencehonles:
# I have made things!

This change also adds a few user type vars and alises to cover the common use cases of ``User``, ``User | AnonymousUser``, and their ``TypeVar`` forms for using in generic contexts.


django-stubs/contrib/auth/models.pyi
```diff
@@ -38,34 +49,32 @@ class Group(models.Model):
...
     is_superuser = models.BooleanField()
     groups = models.ManyToManyField(Group)
     user_permissions = models.ManyToManyField(Permission)
 
-    def get_user_permissions(self, obj: _AnyUser | None = ...) -> set[str]: ...
-    def get_group_permissions(self, obj: _AnyUser | None = ...) -> set[str]: ...
-    def get_all_permissions(self, obj: _AnyUser | None = ...) -> set[str]: ...
-    def has_perm(self, perm: str, obj: _AnyUser | None = ...) -> bool: ...
-    def has_perms(self, perm_list: Iterable[str], obj: _AnyUser | None = ...) -> bool: ...
+    def get_user_permissions(self, obj: AbstractUser | None = ...) -> set[str]: ...
```
terencehonles: I'm not sure if there's enough tests against how this one is used, but this may want to stay wide as `_AnyUser` :thinking:

django-stubs/contrib/auth/models.pyi
```diff
@@ -12,9 +12,20 @@ from django.db.models.manager import EmptyManager
...
 from typing_extensions import Self, TypeAlias
 
-_AnyUser: TypeAlias = Model | AnonymousUser
+_AnyUser: TypeAlias = _User | AnonymousUser
 
-def update_last_login(sender: type[AbstractBaseUser], user: AbstractBaseUser, **kwargs: Any) -> None: ...
+# This is our "placeholder" type the mypy plugin refines to configured 'AUTH_USER_MODEL'
+# wherever it is used as a type. The most recognised example of this is (probably)
+# `HttpRequest.user`
+_User: TypeAlias = AbstractUser
```
sobolevn: Let's move this before `_AnyUser`, no need to create an extra inference cycle.

sobolevn (commented):
Generally - looks great! Just one note

flaeppe (commented):
I don't think we should change the base class to be `AbstractUser`. This is a section of what the "Specifying a custom user model" says in the docs:

> [...] The easiest way to construct a compliant custom user model is to inherit from [AbstractBaseUser](https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#django.contrib.auth.models.AbstractBaseUser). [AbstractBaseUser](https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#django.contrib.auth.models.AbstractBaseUser) provides the core implementation of a user model, including hashed passwords and tokenized password resets. [...]

https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#specifying-a-custom-user-model

sobolevn:
> I don't think we should change the base class to be AbstractUser.

As far as I understand, this is not the base class, this is the default type, unless `AUTH_USER_MODEL` is specified.

terencehonles:
That's interesting that they suggest that... The issue is that `AbstractBaseUser` does not include the `PermissionsMixin` and if we shouldn't be requiring `AbstractUser` we may want to create a type checking only protocol (not sure if that's easy to do as a subtype and I can play around with that).

It does look like there's probably some complications if we move the base class up (subclasses of `AbstractBaseUser` will not work as subclasses of `AbstractUser`), so I'll look into this more, but since the plugin is patching the type based on `AUTH_USER_MODEL` it may only be a problem if you aren't using mypy or haven't configured the plugin.

django-stubs/contrib/auth/models.pyi
terencehonles: done

terencehonles:
> I don't think we should change the base class to be `AbstractUser`. This is a section of what the "Specifying a custom user model" says in the docs:

I rolled back the change since it does affect a user who subclasses `AbstractBaseUser` based on the documentation. I can look into the permissions issue a little if I have time, but I had already fixed our custom plugin not pulling the right model so the replacement is working for us and we're not hit by `AbstractBaseUser` not implementing permissions.

This change still adds some helpful aliases so it can get merged in and the follow-up can be done independently.

terencehonles:
@sobolevn any chance I can get a re-review? I updated this to fix the conflicts, but it was already updated to remove the concern about shifting the base user. It now only provides a few useful type aliases that can be shared across the codebase.

High level changes:
- `_UserModel` renamed to `_User` since it's an _instance_ of the model not the model (type) itself
- `_UserType` shared `TypeVar` that represents a `_User` that should be used in generic context to capture the exact user type being passed into a function/class.
- `_AnyUser` alias of `_User | AnonymousUser` and its shared `TypeVar` `_AnyUserType` for the same reason as ^^^

django-stubs/contrib/auth/forms.pyi
```diff
@@ -15,8 +15,7 @@ from django.utils.functional import _StrOrPromise
 
 logger: Logger
 
-UserModel: TypeAlias = type[_UserModel]
-_User = TypeVar("_User", bound=AbstractBaseUser)
+UserModel: TypeAlias = type[_User]
```
sobolevn: This is a duplicate. We already gave `UserModel` in `django-stubs/contrib/auth/backends.pyi`

django-stubs/contrib/auth/models.pyi
```diff
@@ -12,9 +12,20 @@ from django.db.models.manager import EmptyManager
...
-def update_last_login(sender: type[AbstractBaseUser], user: AbstractBaseUser, **kwargs: Any) -> None: ...
+_AnyUser: TypeAlias = _User | AnonymousUser
+
+# These are only needed for generic classes in order to bind to a specific implementation
+_AnyUserType = TypeVar("_AnyUserType", bound=_AnyUser)  # noqa: PYI018
+
+# do not use the alias `_User` so the bound remains at `AbstractUser`
+_UserType = TypeVar("_UserType", bound=AbstractUser)
+
+def update_last_login(sender: type[_User], user: _User, **kwargs: Any) -> None: ...
```
```suggestion
def update_last_login(sender: UserModel, user: _User, **kwargs: Any) -> None: ...
```

django-stubs/contrib/auth/__init__.pyi
```diff
@@ -18,17 +17,15 @@ REDIRECT_FIELD_NAME: str
...
+def authenticate(request: HttpRequest | None = ..., **credentials: Any) -> _User | None: ...
+async def aauthenticate(request: HttpRequest | None = ..., **credentials: Any) -> _User | None: ...
+def login(request: HttpRequest, user: _User | None, backend: type[BaseBackend] | str | None = ...) -> None: ...
+async def alogin(request: HttpRequest, user: _User | None, backend: type[BaseBackend] | str | None = ...) -> None: ...
 def logout(request: HttpRequest) -> None: ...
 async def alogout(request: HttpRequest) -> None: ...
-def get_user_model() -> type[_UserModel]: ...
-def get_user(request: HttpRequest | Client) -> _UserModel | AnonymousUser: ...
-async def aget_user(request: HttpRequest | Client) -> _UserModel | AnonymousUser: ...
+def get_user_model() -> type[_User]: ...
```
sobolevn:
```suggestion
def get_user_model() -> UserModel: ...
```

django-stubs/contrib/auth/models.pyi
```diff
@@ -12,9 +12,20 @@ from django.db.models.manager import EmptyManager
...
 from typing_extensions import Self
 
-_AnyUser: TypeAlias = Model | AnonymousUser
+# This is our "placeholder" type the mypy plugin refines to configured 'AUTH_USER_MODEL'
+# wherever it is used as a type. The most recognised example of this is (probably)
+# `HttpRequest.user`
+_User: TypeAlias = AbstractBaseUser
 
-def update_last_login(sender: type[AbstractBaseUser], user: AbstractBaseUser, **kwargs: Any) -> None: ...
+_AnyUser: TypeAlias = _User | AnonymousUser
```
sobolevn, line 22: I would move `UserModel` here, since here we define all the aliases and then re-export it where needed.

django-stubs/contrib/auth/views.pyi
```diff
@@ -1,13 +1,13 @@
...
-from django.contrib.auth.base_user import _UserModel
 from django.contrib.auth.forms import AuthenticationForm
+from django.contrib.auth.models import _User
 from django.http.request import HttpRequest
 from django.http.response import HttpResponse, HttpResponseRedirect
 from django.views.generic.base import TemplateView
 from django.views.generic.edit import FormView
 
-UserModel: TypeAlias = type[_UserModel]
+UserModel: TypeAlias = type[_User]
```
sobolevn: This is again a duplicate, let's just import it.

sobolevn (commented):
Thanks for the ping! Mostly looks good.

django-stubs/contrib/auth/__init__.pyi
terencehonles: I'll look to see if there are any `type[_User]` that can be replaced with the `UserModel` alias

django-stubs/contrib/auth/forms.pyi
terencehonles: I assume these are part of Django's public API, so IIRC this would require `from django.contrib.auth.backends import UserModel as UserModel` to re-export it. You'd prefer this over re-declaring it? (it may not be as obvious it's defined if it's part of the imports instead of part of the code)

django-stubs/contrib/auth/models.pyi
terencehonles: Django doesn't declare a `UserModel` [here](https://github.com/django/django/blob/4a293eff6fb10a04de7c65ed705ca3c6a362a587/django/contrib/auth/models.py) so if you'd prefer this I would suggest we declare it as `_UserModel` and when re-exporting it (assuming you want that) we'd use `from django.contrib.auth.models import _UserModel as UserModel`

terencehonles (commented):
The suggestion seems reasonable, but a few comments and I just want to understand what you'd prefer before continuing.

sobolevn:
Yes, I prefer re-exporting over re-definition, because definition can change at some point :)

sobolevn:
(missclick, sorry)

terencehonles:
Ruff _really_ doesn't like re-exporting with a rename, but I think I finally got it to stop deleting the re-exports :unamused:

django-stubs/contrib/auth/__init__.pyi
terencehonles: I think there was just one other that you didn't mention

django-stubs/contrib/auth/backends.pyi
```diff
@@ -1,19 +1,16 @@
-from typing import Any, TypeAlias, TypeVar
+from typing import Any
 
-from django.contrib.auth.base_user import AbstractBaseUser, _UserModel
-from django.contrib.auth.models import AnonymousUser, Permission
+from django.contrib.auth.models import Permission, _AnyUser, _User, _UserType
+from django.contrib.auth.models import _UserModel as UserModel  # noqa: F401
```
sobolevn: Oh, this is not right. Let's use `from django.contrib.auth.models import _UserModel; UserModel = _UserModel`

Because `as` should have the same name for mypy. It is weird, I know :)

terencehonles: I didn't think mypy was that picky about the names, but it appears it is (I tested it out locally), and probably why ruff has the same logic

terencehonles:
updated

sobolevn (approved):
Thanks!

django-stubs/contrib/auth/models.pyi
```diff
@@ -12,9 +12,22 @@ from django.db.models.manager import EmptyManager
...
-def update_last_login(sender: type[AbstractBaseUser], user: AbstractBaseUser, **kwargs: Any) -> None: ...
+_UserModel: TypeAlias = type[_User]
+
+_AnyUser: TypeAlias = _User | AnonymousUser
+
+# These are only needed for generic classes in order to bind to a specific implementation
+_AnyUserType = TypeVar("_AnyUserType", bound=_AnyUser)  # noqa: PYI018
+
+# do not use the alias `_User` so the bound remains at `AbstractUser`
+_UserType = TypeVar("_UserType", bound=AbstractUser)
```
asottile-sentry, line 28: this bound seems incorrect -- or at least it breaks the usage in sentry

the docs for django seem to indicate AbstractBaseUser as the bound here ?

sobolevn: yes, this probably is more correct. thanks for noticing! 

btw, can we test compat with sentry somehow? like running `mypy-primer` but for `django`-related projects only? 🤔 

it should be possible if projects that you are talking about are opensource.

asottile-sentry: yep yep -- the repo in question is fully open source https://github.com/getsentry/sentry

there's a few problems that might be showstoppers though:
- we've not completely made it through our blocklist of files: https://github.com/getsentry/sentry/blob/f7c3a817a586f9e5807784e0deb2db1b634e919b/pyproject.toml#L114
- we're carrying a patch which undoes abstract `objects = ...`: https://github.com/getsentry/sentry-forked-django-stubs/commit/f96ab511b04410987d7685a2b8062c79e96aee2c
- we're carrying a set of patches because we (ab)use `cache` versions as strings: https://github.com/getsentry/sentry-forked-django-stubs/commit/28d1e15523fce9cea31a50537122be234e5c71be https://github.com/getsentry/sentry-forked-django-stubs/commit/3c735503ad99bb5bd7cd9a69be9fb14bf9d732b6

so I guess as long as it doesn't have to typecheck completely cleanly it should be fine?

asottile-sentry: created #2645 for the fix

django-stubs/contrib/auth/forms.pyi
```diff
@@ -114,28 +112,26 @@ class PasswordResetForm(forms.Form):
         extra_email_context: dict[str, str] | None = ...,
     ) -> None: ...
 
-class SetPasswordForm(forms.Form):
-    error_messages: _ErrorMessagesDict
+class SetPasswordForm(Generic[_UserType], SetPasswordMixin, forms.Form):
```
andersk, line 115: This is not generic at runtime, which prevents this from being subclassed.

```python
class MySetPasswordForm(SetPasswordForm):
    pass
```

mypy complains: `error: Missing type parameters for generic type "SetPasswordForm"  [type-arg]`

```python
class MySetPasswordForm(SetPasswordForm[User]):
    pass
```

Fails at runtime: `TypeError: 'DeclarativeFieldsMetaclass' object is not subscriptable`

sobolevn: @andersk can you please send a PR to `ext/`? So we can monkeypatch this object in runtime.

UnknownPlatypus: Maybe we should add a small test to ensure that ? Something parsing pyi file for ``generic` calls and ensuring they match the one declared in the ext ? I can try something like that

sobolevn: Sure, this would be a great addition. I have no idea on how to actually do that :)
Maybe a `stubtest` extra check? 🤔 

I can open an issue in mypy.

UnknownPlatypus: Would be nice if stubtest can catch that indeed

sobolevn: https://github.com/python/mypy/issues/19032

PR #2390, opened 2024-10-03
fix function overload of Library.simple_tag

asottile:
previously failing with:

```
/tmp/django-stubs/tests/typecheck/template/test_library.yml:68: 
E   Actual:
E     main:7: error: No overload variant of "simple_tag" of "Library" matches argument types "Callable[[str], str]", "str"  [call-overload] (diff)
E     main:7: note: Possible overload variants:     (diff)
E     main:7: note:     def [_C] simple_tag(self, func: _C) -> _C (diff)
E     main:7: note:     def simple_tag(self, takes_context: Optional[bool] = ..., name: Optional[str] = ...) -> Callable[[_C], _C] (diff)
E     main:7: note: Revealed type is "Any"          (diff)
E   Expected:
E     main:7: note: Revealed type is "def (s: builtins.str) -> builtins.str" (diff)
E     E: main:7: note: Revealed type is "def (s: builtins.str) -> builtins.str"...
E     A: main:7: error: No overload variant of "simple_tag" of "Library" matches ...
E        
```

real usage: https://github.com/getsentry/sentry/blob/a1acfad04851c894d29daa5e2ddd78d25a506be6/src/sentry/templatetags/sentry_assets.py#L15

implementation: https://github.com/django/django/blob/c334c1a8ff4579cdb1dd77cce8da747070ac9fc4/django/template/library.py#L121


sobolevn (approved):
Thanks!

PR #2392, opened 2024-10-04
Fix resolve_error_handler return type

ljodal:
`URLResolver.resolve_error_handler` should return a callable that takes in a request and returns a response, with no addition arguments.


sobolevn (approved):
Thanks!

PR #2397, opened 2024-10-08
Remove EOL Python 3.8 and add 3.13 to CI

sobolevn:


sobolevn:
Looks like 3.13 is not yet available in `setup-python`, let's wait a bit.

sobolevn:
We need psycopg2 support for 3.13 first.

PR #2400, opened 2024-10-09
Make `from_string()` and `get_template()` return backend-specific template

mthuurne:
Each template backend will only return instances of its own specific template class. For example, `django.template.backends.jinja2.Jinja2` only returns `django.template.backends.jinja2.Template` instances.

Each backend-specific template class has some unique features that a Django app or tool might be interested in.

I had to include a bunch of cleanups as well, because the original signatures of `_EngineTemplate.render()` and `django.template.backends.dummy.Template.render()` were incompatible.


mthuurne:
This code:
```py
class Template(string.Template, _EngineTemplate):
```
triggers this mypy error on Python 3.8:
```
django-stubs/template/backends/dummy.pyi:13: error: Metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases  [misc]
```

What would be the best solution?
- drop commit 509e36813f0ed418f8cbeba9b529292b678ecb49 that adds `_EngineTemplate` as a base class (since it's a protocol, we don't *need* to inherit from it)
- wait for `django-stubs` to drop Python 3.8 support (since it's EOL now)

For now, I'll revert the commit, to make sure that all the CI checks pass before further review.

django-stubs/template/backends/base.pyi
```diff
@@ -23,6 +21,6 @@ class BaseEngine:
 class _EngineTemplate(Protocol):
     def render(
         self,
-        context: Context | dict[str, Any] | None = ...,
```
sobolevn, line 26: Why do you remove this? `Context` can be passed to `render()`

mthuurne: See the commit comment of 2a070e0a3306875b2d8e436ca4f96083ae36224b: not all implementations of this `render()` method accept `Context` and the spec states that `dict` must be used.

The naming is a bit confusing, as there are multiple classes named `Template` and multiple `render()` methods with different interfaces. The implementer of `_EngineTemplate` is `django.template.backends.django.Template` and the `render()` method of that class doesn't accept `Context`, as it calls `make_context()` which explicitly rejects anything other than `None` and `dict`.

The `django.template.base.Template.render()` method does accept `Context` and in fact requires it, but that is not an implementation of the `_EngineTemplate` protocol, despite using the same class and method name.

sobolevn: Ok, I got it! Thanks!

sobolevn (approved):
Thanks!

PR #2113, opened 2024-05-02
5.0: Update `django.contrib.auth.base_user`

JaeHyuckSa:
# I have made things!
Update stubs for `django.contrib.auth.base_user` for Django 5.0.

- [x]  `django.contrib.auth.base_user`
  - [x]  `django.contrib.auth.base_user.AbstractBaseUser.get_session_auth_fallback_hash`was added

## Related issues

Refs
- https://github.com/typeddjango/django-stubs/issues/1493

Upstream PR
- https://github.com/django/django/pull/16631

PR #2407, opened 2024-10-15
Update type of the `get_session_auth_fallback_hash` method to the correct type

JaeHyuckSa:
# I have made things!

I've made this update based on the review for this PR. I also believe that this change is necessary.

[I believe this should also be Iterable[str] https://github.com/django/django/blob/c499184f198df8deb8b5f7282b679babef8384ff/django/contrib/auth/base_user.py#L138-L140](https://github.com/typeddjango/django-stubs/pull/2113#discussion_r1800477662)


sobolevn (approved):
Thank you!

PR #2411, opened 2024-10-20
Treat `QuerySet.alias` like `QuerySet.annotate`

### LINKED ISSUE #2302: [5.0.3 regression] Can’t `filter` a `QuerySet` on an `alias`

andersk:
Although this incorrectly allows some misuses of `alias` that only work at runtime with `annotate`, that is better than incorrectly forbidding all uses of `alias`.

- Fixes #2302.

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -169,6 +169,7 @@ class NewSemanalDjangoPlugin(Plugin):
             "values_list": partial(
                 querysets.extract_proper_type_queryset_values_list, django_context=self.django_context
             ),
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks! Further improvements of `alias()` are welcome!

PR #2419, opened 2024-10-24
Allow `None` on `view_on_site`

nijel:
# I have made things!

## Related issues

This makes it consistent with `get_view_on_site_url` signature where it is used.


django-stubs/contrib/admin/options.pyi
```diff
@@ -139,7 +139,7 @@ class BaseModelAdmin(Generic[_ModelT]):
     def has_view_or_change_permission(self, request: HttpRequest, obj: _ModelT | None = ...) -> bool: ...
     def has_module_permission(self, request: HttpRequest) -> bool: ...
     @property
-    def view_on_site(self) -> Callable[[_ModelT], str] | bool: ...
+    def view_on_site(self) -> Callable[[_ModelT], str | None] | bool: ...
```
sobolevn, line 142: Sorry, this does not seem correct. Pls, see https://github.com/django/django/blob/6dcab75d5d8c2ef18de15323930057e6fa9ec00f/django/contrib/admin/options.py#L358-L363

nijel: Sorry, but the linked code doesn't deal with the return value from the callable at all.

`view_on_site` can be either a boolean, or a callable returning a string or `None`. When callable returns `None`, `get_view_on_site_url` will return `None`, what is a valid return value here (as can be seen in the code you linked, which returns `None` as well when `view_on_site` is false).

sobolevn: My bad, I misread the type :)
Thanks for the clarification.

sobolevn (approved):
Thanks!

PR #2423, opened 2024-10-26
Update `contrib.admin.models.LogEntryManager`

JaeHyuckSa:
# I have made things!


sobolevn (commented):
Please, make `log_action` as deprecated: https://github.com/django/django/blob/373cb3037fe4e67adbac9ac43340391e859aa957/django/contrib/admin/models.py#L39

and type `queryset: Queryset[]` type param.

JaeHyuckSa:
> Please, make `log_action` as deprecated: https://github.com/django/django/blob/373cb3037fe4e67adbac9ac43340391e859aa957/django/contrib/admin/models.py#L39
> 
> and type `queryset: Queryset[]` type param.

I often come across methods marked with `RemovedInDjango60Warning`  in Django, and it seems these aren’t fully addressed in `django-stubs` yet — or is there something I might be missing? Would it be better to handle these using a `deprecated` decorator, or should we add a category to the allowlist?

If you could provide some guidance, I’ll go ahead and create a PR to address this more thoroughly.

sobolevn:
> Would it be better to handle these using a deprecated decorator

Yes, I think that we should apply this decorator.

JaeHyuckSa:
> > Would it be better to handle these using a deprecated decorator
> 
> Yes, I think that we should apply this decorator.

The suggested changes have been applied. -> https://github.com/typeddjango/django-stubs/pull/2458

sobolevn:
This now has merge conflicts :(

JaeHyuckSa:
> This now has merge conflicts :(

I resolved the conflict.

django-stubs/contrib/admin/models.pyi
```diff
@@ -21,6 +22,15 @@ class LogEntryManager(models.Manager[LogEntry]):
...
         change_message: Any = ...,
     ) -> LogEntry: ...
+    def log_actions(
+        self,
+        user_id: int,
+        queryset: QuerySet[Model],
+        action_flag: int,
+        change_message: str | list[Any] = "",
+        *,
+        single_object: bool = False,
```
sobolevn: Maybe this should be an `@overload`? When `single_object: Literal[True]` it will return `-> LogEntry`, otherwise `-> list[LogEntry]`.

JaeHyuckSa: Thank you for the review! This approach makes it clearer.

sobolevn (approved):
Thank you!

PR #2426, opened 2024-10-30
Update enums to reflect typing spec changes

hauntsaninja:
See https://typing.readthedocs.io/en/latest/spec/enums.html#defining-members

https://github.com/python/typing-council/issues/11

The next version of mypy will obey the spec change: https://github.com/python/mypy/pull/18068

pyright already requires this


sobolevn (approved):
Thank you! 👍

PR #2431, opened 2024-11-03
fix .values_list returns incorrect type for field with same name when selected from related model

### LINKED ISSUE #2336: `.values_list` returns incorrect type for field with same name when selected from related model
This wasn't a problem in `5.0.2` and thus I've bisected the issue to

- #2248 

---

If I run `values_list` and select a field on a related model, which happens to have the same name on the filtering model, types are not selected from the related model's field.

Repro case:

```yaml
-   case: handles_field_with_same_name_on_other_model
    main: |
        from myapp.models import A
# Below reveals "Tuple[Union[builtins.str, None]]" which is from field `A.name` but `B.name` is selected in `values_list`
        reveal_type(A.objects.values_list("b__name").get())  # N: Revealed type is "Tuple[builtins.str]"
    installed_apps:
        - myapp
    files:
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                from django.db import models

                class B(models.Model):
                    name = models.CharField()

                class A(models.Model):
                    b = models.ForeignKey(B, on_delete=models.CASCADE)
                    name = models.CharField(null=True)
```

## What's wrong

Results in an error:

```
main:2: note: Revealed type is "Tuple[Union[builtins.str, None]]"
```

Above are the types from `A.name` but the query selects from `B.name`

## How is that should be

main:2: note: Revealed type is "Tuple[builtins.str]"

Niicck:
# Fix .values_list bug that returns incorrect type for field with same name when selected from related model

`get_field_type_from_lookup` will now check the correct related_model to get a field's type. Previously, it was only ever checking the original model.

## Related issues
- Closes https://github.com/typeddjango/django-stubs/issues/2336

### DIFF
--- a/tests/typecheck/managers/querysets/test_values_list.yml
+++ b/tests/typecheck/managers/querysets/test_values_list.yml
@@ -325,7 +325,7 @@
 -   case: handles_field_with_same_name_on_other_model
     main: |
         from myapp.models import A
-        reveal_type(A.objects.values_list("b__name").get())  # N: Revealed type is "Tuple[builtins.str]"
+        reveal_type(A.objects.values_list("name", "b__name").get())  # N: Revealed type is "Tuple[builtins.int, builtins.str]"
     installed_apps:
         - myapp
     files:
@@ -339,4 +339,4 @@
 
                 class A(models.Model):
                     b = models.ForeignKey(B, on_delete=models.CASCADE)
-                    name = models.CharField(null=True)
+                    name = models.IntegerField()
--- a/tests/typecheck/managers/querysets/test_values_list.yml
+++ b/tests/typecheck/managers/querysets/test_values_list.yml
@@ -322,3 +322,21 @@
                 class Blog(models.Model):
                     num_posts = models.IntegerField()
                     text = models.CharField(max_length=100, blank=True)
+-   case: handles_field_with_same_name_on_other_model
+    main: |
+        from myapp.models import A
+        reveal_type(A.objects.values_list("b__name").get())  # N: Revealed type is "Tuple[builtins.str]"
+    installed_apps:
+        - myapp
+    files:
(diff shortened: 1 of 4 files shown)


tests/typecheck/managers/querysets/test_values_list.yml
```diff
@@ -322,3 +322,21 @@
                 class Blog(models.Model):
                     num_posts = models.IntegerField()
                     text = models.CharField(max_length=100, blank=True)
+-   case: handles_field_with_same_name_on_other_model
+    main: |
+        from myapp.models import A
+        reveal_type(A.objects.values_list("b__name").get())  # N: Revealed type is "Tuple[builtins.str]"
```
sobolevn: Let's use two names just to be sure they are different.

```suggestion
        reveal_type(A.objects.values_list("name", "b__name").get())  # N: Revealed type is "Tuple[builtins.int, builtins.str]"
```

tests/typecheck/managers/querysets/test_values_list.yml
```diff
@@ -322,3 +322,21 @@
...
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+
+                class B(models.Model):
+                    name = models.CharField()
+
+                class A(models.Model):
+                    b = models.ForeignKey(B, on_delete=models.CASCADE)
+                    name = models.CharField(null=True)
```
sobolevn: I suggest to change this field to `int` field, just to be sure that they are different.

sobolevn (commented):
Thanks!

tests/typecheck/managers/querysets/test_values_list.yml
Niicck: good call!

sobolevn (approved):
Thank you!

PR #2439, opened 2024-11-07
Fix typing of related manager methods

ngnpope:
Looking at the [related manager documentation][0] it isn't very clear that the `bulk` argument only applies to `RelatedManager` and the `through_defaults` argument only applies to `ManyRelatedManager`.

Based on the [related descriptors source][1], the following changes have been made in this commit:

- The `bulk` argument has been added to `.clear()` and `.aclear()` for `RelatedManager`
- The `bulk` argument has been removed from all methods of `ManyRelatedManager`
- Additional methods using `through_defaults` have been added to `ManyRelatedManager`.

[0]: https://docs.djangoproject.com/en/stable/ref/models/relations/
[1]: https://github.com/django/django/blob/042b381e2e37c0c37b8a8f6cc9947f1a2ebfa0dd/django/db/models/fields/related_descriptors.py


sobolevn (approved):
Thank you!

PR #2440, opened 2024-11-10
Fix type annotations for 3rd party models

### LINKED ISSUE #2341: Type Annotation Errors in Models Inheriting from 3rd Party Packages
# Bug report

## What's wrong

With the latest django-stubs v5.0.4, I've noticed strange behavior when using models that inherit from 3rd party packages. Specifically, mypy throws a "Need type annotation for <field_name>" error on model fields.

Here’s the scenario:

- I created a new project using `django-cookiecutter`.
- I added `mypy` v1.11.1, `django-stubs` v5.0.4, and tested with `Django` 4.2.11 and 5.0.8.
- I have the following model classes:

```
class FooBase(models.Model):
    name = models.CharField(max_length=255)

    class Meta:
        abstract = True

class Bar(FooBase):
    updated_at = models.DateTimeField(auto_now=True)
```

- When both classes are defined in `local_app/models.py`, there are no errors.
- However, if I move the `FooBase` class to a 3rd party package’s `models.py` (e.g., `site-packages/django_extensions/models.py` since that came with the cookiecutter install), I get a mypy error on the `updated_at` field of `Bar`.

This issue only started occurring with django-stubs v5.0.3 or v5.0.4 and mypy v1.11.0 or v1.11.1. It was working correctly with the previous versions (django-stubs v5.0.2 and mypy v1.10.1).

I am aware of issue #2011, which has been closed. However, the difference here is that the code was working fine with the prior django-stubs and mypy versions, but the error occurs after updating to the latest versions.

## How is that should be

Type checking should pass without errors when using models that inherit from 3rd party packages, just as it did with previous versions of django-stubs and mypy.

## System information

- OS: yes
- `python` version: 3.11.9
- `django` version: 5.0.8 & 4.2.11
- `mypy` version: 1.11.1
- `django-stubs` version: 5.0.4
- `django-stubs-ext` version: 5.0.4

### mypy Settings

```
[tool.mypy]
python_version = "3.11"
check_untyped_defs = true
ignore_missing_imports = true
warn_unused_ignores = true
warn_redundant_casts = true
warn_unused_configs = true
plugins = [
    "mypy_django_plugin.main",
]
```

Niicck:
# Fix type annotations for 3rd party models

The clean-up in [PR #2263](https://github.com/typeddjango/django-stubs/pull/2263) that got released in v5.0.3 ended up breaking type annotations for models that are subclasses of models imported from 3rd party libraries.

I restored the logic that checks a model's fullname in the list of registered django apps. We still need to do this because Django models imported from untyped 3rd party libraries will have no `metaclass_type`, so `helpers.is_model_type` will always fail for them.

## Related issues
- Closes: [#2341](https://github.com/typeddjango/django-stubs/issues/2341)
- Related to: [PR #2263](https://github.com/typeddjango/django-stubs/pull/2263)

### DIFF
--- a/tests/typecheck/models/test_3rd_party_models.yml
+++ b/tests/typecheck/models/test_3rd_party_models.yml
@@ -1,8 +1,15 @@
 -   case: handles_type_annotations_on_3rd_party_models
+    installed_apps:
+        - django_extensions
+        - myapp
     main: |
-        from django.db import models
-        from django_extensions.db.models import TimeStampedModel # type: ignore [import-untyped]
+        from myapp.models import A
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+                from django_extensions.db.models import TimeStampedModel # type: ignore[import-untyped]
 
-        class A(TimeStampedModel):
-            name = models.CharField()
-            count = models.IntegerField()
+                class A(TimeStampedModel):
+                    name = models.CharField()
--- a/tests/typecheck/models/test_3rd_party_models.yml
+++ b/tests/typecheck/models/test_3rd_party_models.yml
@@ -1,0 +1,8 @@
+-   case: handles_type_annotations_on_3rd_party_models
+    main: |
+        from django.db import models
+        from django_extensions.db.models import TimeStampedModel # type: ignore [import-untyped]
+
+        class A(TimeStampedModel):
+            name = models.CharField()
+            count = models.IntegerField()
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -526,6 +526,10 @@ def is_model_type(info: TypeInfo) -> bool:
     return info.metaclass_type is not None and info.metaclass_type.type.has_base(fullnames.MODEL_METACLASS_FULLNAME)
 
 
+def is_registered_model_type(info: TypeInfo, django_context: "DjangoContext") -> bool:
+    return info.fullname in {get_class_fullname(cls) for cls in django_context.all_registered_model_classes}
+
+
 def get_model_from_expression(
     expr: Expression,
     *,
--- a/mypy_django_plugin/transformers/fields.py
+++ b/mypy_django_plugin/transformers/fields.py
@@ -237,7 +237,10 @@ def transform_into_proper_return_type(ctx: FunctionContext, django_context: Djan
     assert isinstance(default_return_type, Instance)
 
     outer_model_info = helpers.get_typechecker_api(ctx).scope.active_class()
-    if outer_model_info is None or not helpers.is_model_type(outer_model_info):
+    if outer_model_info is None or (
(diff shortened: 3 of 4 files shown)


tests/typecheck/models/test_3rd_party_models.yml
```diff
@@ -0,0 +1,15 @@
+-   case: handles_type_annotations_on_3rd_party_models
+    installed_apps:
+        - django_extensions
+        - myapp
+    main: |
+        from myapp.models import A
```
sobolevn, line 6: please, add a `reveal_type` to `A().name`

tests/typecheck/models/test_3rd_party_models.yml
```diff
@@ -0,0 +1,15 @@
...
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+                from django_extensions.db.models import TimeStampedModel # type: ignore[import-untyped]
```
flaeppe, line 12: I actually disagree with the changes here. The main issue in #2341 is that `django-extensions` is a package that doesn't export any type hints.

The ignore comment I'm highlighting here is saying precisely that. What is experienced in #2341 is how mypy behaves when there are no type hints.

What is attempted to be repaired here is only the first line of issues, there could be an arbitrary amount of other problems to solve after this one, coming from it being an untyped package.

There are multiple other approaches to resolve this, that integrates better with the type checker. e.g.

- Add and export type hints in `django-extensions`
- Create a stubs package(`.pyi`) just like `django-stubs` for `django-extrnsions`
- Add stubs (`.pyi`-files) for `django-extensions` to your project, locally
- Perhaps `django-stubs` would consider including stubs in a similar fashion to `typeshed`, that includes/maintains a bunch of 3rd party stubs?

Regardless, I think merging the changes here is a step in the wrong direction as it's outside of `django-stubs` domain

rsiemens: ~I'm not sure I understand the "how mypy behaves when there are no type hints." bit.~

~Shouldn't `--follow-imports=normal` (the default) still allow for the top level class relationship to be constructed regardless of if the third part is typed or not?~

~I have models that inherit [`SealableModel`](https://github.com/charettes/django-seal/blob/1.4.4/seal/models.py#L33), but when I inspect my model during `transform_into_proper_return_type` it seems like all the context is lost.~

```
In [1]: outer_model_info.fullname
Out[1]: 'projects.models.Project'

In [2]: helpers.is_model_type(outer_model_info)
Out[2]: False

In [3]: outer_model_info.metaclass_type

In [4]: outer_model_info.is_metaclass()
Out[4]: True 
```

edit: I realize I have `ignore_missing_imports = true` in my mypy.ini file. The fact that things worked before was somewhat happenstance thanks to the `info.fullname in {get_class_fullname(cls) for cls in django_context.all_registered_model_classes}` (for better or worse as you pointed out)

PR #2444, opened 2024-11-15
Fix signature of `ModelAdmin.construct_change_message`

youtux:
The `form` argument was incorrectly typed.


sobolevn (commented):
Why do you think so? Is this documented? There's also one more place in `admin/utils.py` where the same type is used.

youtux:
The method belongs to generic ModelAdmin class; it makes no sense for it to take an AdminPasswordChange form. The implementation just require any instance of Form

sobolevn:
Got it, please address the second comment :)

youtux:
Done in ~~605cddf9~~ 80f199f5

sobolevn (approved):
Thanks!

PR #2457, opened 2024-12-06
Marking default_error_messages as a ClassVar

delfick:
This is only ever a dictionary defined on the class and not being defined as a ClassVar means in projects using django-stubs it's not possible to comply with https://docs.astral.sh/ruff/rules/mutable-class-default/

### DIFF
--- a/django-stubs/contrib/postgres/fields/array.pyi
+++ b/django-stubs/contrib/postgres/fields/array.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Iterable, Sequence
-from typing import Any, TypeVar
+from typing import Any, ClassVar, TypeVar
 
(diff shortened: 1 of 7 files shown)


sobolevn (approved):
You can add this new mypy error code https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-overrides-of-mutable-attributes-are-safe-mutable-override in the next PR, just to see how many errors we have. Maybe some of them are worth fixing. Like this one.

delfick:
@sobolevn `mypy --enable-error-code mutable-override` is a very funny command for the 6 million line django monolith I work on lol

```
Found 21195 errors in 5260 files (checked 37001 source files)
```

On django-stubs it appears only 200 errors :)

PR #2422, opened 2024-10-26
Update `contrib.admin.options.ModelAdmin`

JaeHyuckSa:
# I have made things!

PR #2458, opened 2024-12-06
Add deprecated decorators for multiple classes and methods

JaeHyuckSa:
# I have made things!
This PR adds deprecated decorators to address Django 6.0 deprecations, providing guidance for transitioning to updated methods as noted [comment](https://github.com/typeddjango/django-stubs/pull/2423#issuecomment-2439833406).

## Related PR 
- https://github.com/typeddjango/django-stubs/pull/2423
- https://github.com/typeddjango/django-stubs/pull/2422


sobolevn (approved):
Thank you!

django-stubs/contrib/admin/options.pyi
```diff
@@ -129,6 +129,7 @@ class BaseModelAdmin(Generic[_ModelT]):
     def get_prepopulated_fields(self, request: HttpRequest, obj: _ModelT | None = ...) -> dict[str, Sequence[str]]: ...
     def get_queryset(self, request: HttpRequest) -> QuerySet[_ModelT]: ...
     def get_sortable_by(self, request: HttpRequest) -> _DisplayT[_ModelT]: ...
+    @deprecated("The None value for the request parameter will be removed in Django 6.0.")
```
ngnpope, line 132: This one should be removed. `@deprecated` can only be used if the whole function is deprecated, not changes to types for individual parameters. This will cause the deprecation to be shown even if `request` is not `None`.

django-stubs/db/models/enums.pyi
```diff
@@ -21,6 +21,7 @@ else:
     class IntEnum(int, ReprEnum): ...
     class StrEnum(str, ReprEnum): ...
 
+@deprecated("ChoicesMeta is deprecated in favor of ChoicesType and will be removed in Django 6.0.")
```
ngnpope, line 24: I'm not sure that this works either as `ChoicesType` is aliased to `ChoicesMeta` so will show the same deprecation warning.

As an aside, we should probably also invert the alias so that we have `class ChoicesType(EnumType)` and `ChoicesMeta = ChoicesType` for the alias.

django-stubs/utils/html.pyi
```diff
@@ -19,6 +20,7 @@ def json_script(value: Any, element_id: str | None = None, encoder: type[JSONEnc
 
 # conditional_escape could use a protocol to be more precise, see https://github.com/typeddjango/django-stubs/issues/1474
 def conditional_escape(text: _StrOrPromise | SafeData) -> SafeString: ...
+@deprecated("Calling format_html() without passing args or kwargs is deprecated.")
```
ngnpope, line 23: Again, it's not possible to use `@deprecated` here without false positives.

django-stubs/utils/itercompat.pyi
```diff
@@ -1,3 +1,8 @@
 from typing import Any
 
+from typing_extensions import deprecated
+
+@deprecated(
+    "The django.utils.itercompat.is_iterable() is deprecated and will be removed in Django 6.0. Use isinstance(..., collections.abc.Iterable) instead."
```
ngnpope, line 6: There is a lot of `"The module.function() is deprecated..."` which doesn't read properly. In most cases we can probably drop the "The":

```suggestion
    "django.utils.itercompat.is_iterable() is deprecated and will be removed in Django 6.0. Use isinstance(..., collections.abc.Iterable) instead."
```

Please review all the other introductions of this.

sobolevn:
@ngnpope thanks for the detailed review! 
@JaeHyuckSa can you please address these comments?

JaeHyuckSa:
> @ngnpope thanks for the detailed review! @JaeHyuckSa can you please address these comments?

Thank you so much, @ngnpope, for your detailed and thoughtful review! I really appreciate the time and effort you put into it. I’m currently busy with work, but I’ll make sure to address your comments at the earliest possible opportunity :)

django-stubs/contrib/admin/options.pyi
Viicos: Alternatively, [deprecated overloads](https://typing.readthedocs.io/en/latest/spec/directives.html#example) can be used.

django-stubs/db/models/fields/reverse_related.pyi
```diff
@@ -77,6 +78,9 @@ class ForeignObjectRel(FieldCacheMixin):
         ordering: Sequence[str] = (),
     ) -> _ChoicesList: ...
     def is_hidden(self) -> bool: ...
+    @deprecated(
+        "The get_joining_columns() is deprecated and will be removed in Django 6.0. Use get_joining_fields() instead."
+    )
```
Viicos, line 83: To be consistent, the "The" prefix (and in other places) should probably be removed.

django-stubs/contrib/admin/options.pyi
ngnpope: Of course, yes!

PR #2459, opened 2024-12-06
base.pyi: use base class HttpResponseBase

dnet:
Previously, this was only a `Callable` but 125f31a81116f5f149c3e868d950608571e9a5e8 restricted the last parameter to `HttpResponse` which is unnecessarily strict, as it does not cover things like `StreamingHttpResponse` and its descendant `FileResponse` among others – while _view_ functions can return these as well. They have a common base class `HttpResponseBase` – I guess this would be a better choice here.

### DIFF
--- a/django-stubs/urls/base.pyi
+++ b/django-stubs/urls/base.pyi
@@ -1,7 +1,6 @@
 from collections.abc import Callable, Sequence
 from typing import Any, Literal
 
-from django.http.response import HttpResponse
 from django.urls.resolvers import ResolverMatch
 
 def resolve(path: str, urlconf: str | None = ...) -> ResolverMatch: ...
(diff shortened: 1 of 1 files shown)


django-stubs/urls/base.pyi
```diff
@@ -1,12 +1,11 @@
 from collections.abc import Callable, Sequence
 from typing import Any, Literal
 
-from django.http.response import HttpResponse
 from django.urls.resolvers import ResolverMatch
```
sobolevn, line 5:
```suggestion
from django.http.response import HttpResponseBase
from django.urls.resolvers import ResolverMatch
```

sobolevn (commented):
Thanks! But, CI is broken. I suggested a fix.

PR #2470, opened 2024-12-30
Add py.typed file for pylance support

ashwanthbalakrishnan5:
This is similar to [django-types](https://github.com/sbdchd/django-types/) but they are outdated.
By adding this file pylance can use these stubs instead - https://github.com/microsoft/pylance-release/issues/6029


sobolevn (approved):
Thanks!

PR #2472, opened 2025-01-03
Correct type hints for models.JSONField.from_db_value

angryfoxx:


sobolevn (commented):
Thanks for the PR! However, I don't think that we should include it.

We prefer not to include union return types. See https://github.com/python/mypy/issues/1693

PR #2368, opened 2024-09-11
Update _Choices type alias

joaoseckler:
# Updating the _Choices type alias

Django 5 allows model field choices to be callables, mappings or subclasses of `models.Choices`. This PR introduces these options in the stubs.

## Related issues

No related issues

PR #2476, opened 2025-01-09
Update models's fields' _Choices type

stianjensen:
Django 5 allows model field choices to be callables, mappings or subclasses of models.Choices. This commit introduces these options in the stubs.

Continuation of https://github.com/typeddjango/django-stubs/pull/2368, to fix a merge conflict, avoid issues with importing `Tuple` instead of using `tuple`, and avoid unnecessary formatting changes.

### DIFF
--- a/django-stubs/utils/choices.pyi
+++ b/django-stubs/utils/choices.pyi
@@ -1,4 +1,4 @@
-from collections.abc import Iterable, Iterator
+from collections.abc import Iterable, Iterator, Mapping
 from typing import Any, Protocol, TypeVar, type_check_only
 
 from typing_extensions import TypeAlias
@@ -6,6 +6,7 @@ from typing_extensions import TypeAlias
 _Choice: TypeAlias = tuple[Any, Any]
 _ChoiceNamedGroup: TypeAlias = tuple[str, Iterable[_Choice]]
 _Choices: TypeAlias = Iterable[_Choice | _ChoiceNamedGroup]
+_ChoicesMapping: TypeAlias = Mapping[Any, Any]  # noqa: PYI047
 
 @type_check_only
 class _ChoicesCallable(Protocol):
--- a/django-stubs/db/models/fields/__init__.pyi
+++ b/django-stubs/db/models/fields/__init__.pyi
@@ -9,12 +9,13 @@ from django import forms
 from django.core import validators  # due to weird mypy.stubtest error
 from django.core.checks import CheckMessage
 from django.db.backends.base.base import BaseDatabaseWrapper
-from django.db.models import Model
+from django.db.models import Choices, Model
 from django.db.models.expressions import Col, Combinable, Expression, Func
 from django.db.models.fields.reverse_related import ForeignObjectRel
 from django.db.models.query_utils import Q, RegisterLookupMixin
 from django.forms import Widget
-from django.utils.choices import BlankChoiceIterator, _Choice, _ChoiceNamedGroup, _Choices, _ChoicesCallable
+from django.utils.choices import BlankChoiceIterator, _Choice, _ChoiceNamedGroup, _ChoicesCallable, _ChoicesMapping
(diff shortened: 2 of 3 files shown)


django-stubs/utils/choices.pyi
```diff
@@ -1,11 +1,12 @@
-from collections.abc import Iterable, Iterator
+from collections.abc import Iterable, Iterator, Mapping
 from typing import Any, Protocol, TypeVar, type_check_only
 
 from typing_extensions import TypeAlias
 
 _Choice: TypeAlias = tuple[Any, Any]
 _ChoiceNamedGroup: TypeAlias = tuple[str, Iterable[_Choice]]
 _Choices: TypeAlias = Iterable[_Choice | _ChoiceNamedGroup]
+_ChoicesMapping: TypeAlias = Mapping[Any, Any] | Mapping[str, Mapping[Any, Any]]  # noqa: PYI047
```
sobolevn: Isn't it just `Mapping[Any, Any]`?

tests/typecheck/db/models/test_fields.yml
```diff
@@ -0,0 +1,159 @@
+-   case: db_models_fields_choices
```
sobolevn: The `tests/assert_type/db/models/fields/test_choices.py` test is enough, no need to duplicate it. Can you please test invalid types here? At least one.

stianjensen: I pushed an update now, was this what you had in mind?

sobolevn (commented):
Thanks! I would like to have one more pair of eyes :)
cc @intgr @flaeppe

stianjensen:
I see you made another release now – would be great if this made the cut for the next one!

tests/typecheck/db/models/test_fields.yml
```diff
@@ -0,0 +1,13 @@
+- case: db_models_fields_choices
+  main: |
+    from django.db import models
+
+    class MyModel(models.Model):
+        char1 = models.CharField[str, str](max_length=200, choices='test')
```
sobolevn: Why do we need type params here? Let's remove them.

sobolevn (commented):
The only question I have is about type params to fields.

stianjensen:
Yeah good point, that shouldn't be necessary. Updated!

tests/assert_type/db/models/fields/test_choices.py
```diff
@@ -0,0 +1,89 @@
...
+class TestModel(models.Model):
+    class TextChoices(models.TextChoices):
+        FIRST = "foo", "bar"
+        SECOND = "foo2", "bar"
+
+    class IntegerChoices(models.IntegerChoices):
+        FIRST = 1, "bar"
+        SECOND = 2, "bar"
+
+    char1 = models.CharField[str, str](max_length=5, choices=TextChoices, default="foo")
```
sobolevn: The same question about these ones.

sobolevn (commented):
Sorry, I've missed to comment on them during the prev review :(

tests/assert_type/db/models/fields/test_choices.py
stianjensen: Hah, no worries – I could have looked through for those myself, too!

stianjensen:
But this last change seems to break pyright tests?

sobolevn:
Ok, that's unfortunate. Let's move this test to mypy test only, because we don't officially support pyright and I don't know anything about how it works :(

stianjensen:
Cool! How do I make a test mypy only? 😅

sobolevn:
You move it to `tests/typecheck/db/models/test_fields.yml`

stianjensen:
Thanks for the help and patience!

Here's a new attempt, with proving a couple invalid cases first, then all the succesful asserts, all in a yaml test case for mypy!

sobolevn (approved):
Thanks!

PR #2093, opened 2024-04-29
Add missing type checking for `acreate`

jlost:
# I have made things!

## Related issues

#2090

PR #2477, opened 2025-01-10
Add type-checking to Manager.acreate

### LINKED ISSUE #2090: `Manager.acreate`: model isn't type checked
```py
from django.db import models

class Thing(models.Model):
    id = models.AutoField()

def create_thing_sync(name: str) -> None:
    Thing.objects.create(name="This line shows an error")

async def create_thing_async(name: str) -> None:
    Thing.objects.acreate(name="This one doesn't")
```

Error only appears for the first example, but it should show for both:

```sh
mypy thing.py
Unexpected attribute "name" for model "Thing"
```

I think the problem is in the mypy plugin. I'll take a look when I get the chance.

vitosamson:
This PR adds type-checking to kwargs for the `.acreate()` method. Currently `.create()` is type-checked but `.acreate()` is not.

## Related issues

Closes #2090 

This is my first time diving into this codebase so let me know if I missed anything. I used the test case from #2093.

### DIFF
--- a/tests/typecheck/models/test_create.yml
+++ b/tests/typecheck/models/test_create.yml
@@ -161,7 +161,7 @@
         import asyncio
         from myapp.models import User
         async def amain() -> None:
-            await User.objects.acreate(pk=1, name='Max', age=10)
+            reveal_type(await User.objects.acreate(pk=1, name='Max', age=10))  # N: Revealed type is "myapp.models.User"
             await User.objects.acreate(age=[])  # E: Incompatible type for "age" of "User" (got "List[Any]", expected "Union[float, int, str, Combinable]")  [misc]
     installed_apps:
         - myapp
--- a/tests/typecheck/models/test_create.yml
+++ b/tests/typecheck/models/test_create.yml
@@ -155,3 +155,21 @@
                     id = models.IntegerField(primary_key=True)
                 class MyModel3(models.Model):
                     default = models.IntegerField(default=return_int)
+
+-   case: default_manager_acreate_is_typechecked
+    main: |
+        import asyncio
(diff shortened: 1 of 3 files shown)


tests/typecheck/models/test_create.yml
```diff
@@ -155,3 +155,21 @@
                     id = models.IntegerField(primary_key=True)
                 class MyModel3(models.Model):
                     default = models.IntegerField(default=return_int)
+
+-   case: default_manager_acreate_is_typechecked
+    main: |
+        import asyncio
+        from myapp.models import User
+        async def amain() -> None:
+            await User.objects.acreate(pk=1, name='Max', age=10)
```
sobolevn:
```suggestion
            reveal_type(await User.objects.acreate(pk=1, name='Max', age=10))
```

sobolevn (commented):
Thanks! One test suggestion

sobolevn (approved):

vitosamson:
Thanks @sobolevn! Any idea when this might go out in a release?

sobolevn:
We do releases every couple of month. I think that somewhere before / around the start of Feb.

sobolevn:
@flaeppe oh, I double checked. Our latest release was in Oct. So, maybe we can do one?

vitosamson:
That would be great since this issue caused us to miss a bug. I saw in the contributing docs that anybody can open the release PR - I'm happy to get that ball rolling if it would help.

sobolevn:
@vitosamson please, go ahead :)

vitosamson:
@sobolevn done https://github.com/typeddjango/django-stubs/pull/2478

PR #2485, opened 2025-01-15
Fix `@deprecated` based on the review

### LINKED ISSUE #2484: Overzealous deprecation warnings
# Bug report

`django.utils.html.format_html` is marked deprecated as a whole by the stubs, but this is wrong.

## What's wrong

#2458 added deprecation markers for Django 6, but it is incomplete as per the comments there: https://github.com/typeddjango/django-stubs/pull/2458#pullrequestreview-2491840104
It was still merged, so now `format_html` and others are marked deprecated no matter what, which is not correct.
I'm posting this issue so it does not get lost.

sobolevn:
No tests, becase we have a blocker: https://github.com/python/mypy/issues/18474

CC @ngnpope, @Viicos, @JaeHyuckSa for reviews.

Closes https://github.com/typeddjango/django-stubs/issues/2484
Refs https://github.com/typeddjango/django-stubs/pull/2458


ngnpope (approved):
Thanks for fixing these up 👍🏻

PR #2487, opened 2025-01-16
Add "User.Meta" and "AbstractUser.Meta" classes

brianhelba:


brianhelba:
I'd welcome any guidance on whether there's a way to use `TypedModelMeta` (or its attributes) from within the main package.

brianhelba:
I didn't realize this was so difficult, and I hadn't seen the history of #2000 and #2112. I'll watch #2112 and wait for this to be resolved, as I do need it in my own project.

Closing.

PR #2492, opened 2025-01-23
Fix error with `null=True` and `Any` type param for `Field` types

### LINKED ISSUE #2403: Derived ForeignKey field type "is nullable but its generic get type parameter is not optional"
Hi.

I'm getting `is nullable but its generic get type parameter is not optional` for fields with derived `ForeignKey` field types.

Given my main model uses a "name" as primary key, I'd like to use `_name` instead of `_id` as attname and dbcolum suffix, and I think I can only overwrite dbcolumn in a fields constructor (not the attname).

Demo code:

```
class ForeignNameKey(models.ForeignKey):
    def get_attname(self) -> str:
        return '%s_name' % self.name


class Domain(models.Model):
    name = models.CharField(unique=True, max_length=255, blank=False, primary_key=True)


class Foo(models.Model):
    domain = ForeignNameKey(
        Domain,
        on_delete=models.RESTRICT,
        null=True,
    )
```

-> `error: ForeignNameKey is nullable but its generic get type parameter is not optional`

I tried understanding what the code from #2048 does, but couldn't figure out why it works with `models.ForeignKey` and not with my derived class. Also the error message is completely useless - I have no idea how to "fix" it (also unclear what is wrong; I don't think anything is actually wrong).

## System information

- OS: debian bookworm
- `python` version: 3.11.2
- `django` version: 5.1.2
- `mypy` version: 1.11.2
- `django-stubs` version: 5.1.0
- `django-stubs-ext` version: 5.1.0

sobolevn:
Closes https://github.com/typeddjango/django-stubs/issues/2403
Refs https://github.com/typeddjango/django-stubs/pull/2048

### DIFF
--- a/mypy_django_plugin/transformers/fields.py
+++ b/mypy_django_plugin/transformers/fields.py
@@ -166,11 +166,16 @@ def set_descriptor_types_for_field(
     if not (isinstance(mapped_set_type, UninhabitedType) or isinstance(mapped_get_type, UninhabitedType)):
         # always replace set_type and get_type with (non-Any) mapped types
         set_type = helpers.convert_any_to_type(mapped_set_type, set_type)
-        get_type = helpers.convert_any_to_type(mapped_get_type, get_type)
+        get_type = get_proper_type(helpers.convert_any_to_type(mapped_get_type, get_type))
 
         # the get_type must be optional if the field is nullable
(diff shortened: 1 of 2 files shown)


flaeppe (approved):
👍

PR #2497, opened 2025-02-03
Fix types for `TestCase.assertJSON{Not,}Equal`

nickstenning:
The implementation of this function just passes the `raw` value to `json.loads`, which [is typed as accepting][1] strings, bytes objects, or bytearrays.

Update the signature of `assertJSONEqual` and `assertJSONNotEqual` to match.

[1]: https://github.com/python/typeshed/blob/aac4394eb29d86797628b57d6f01f5e17f5ff83f/stdlib/json/__init__.pyi#L40

Please let me know if I missed important things from the contributor process?

### DIFF
--- a/django-stubs/test/testcases.pyi
+++ b/django-stubs/test/testcases.pyi
@@ -152,13 +152,13 @@ class SimpleTestCase(unittest.TestCase):
     def assertInHTML(self, needle: str, haystack: str, count: int | None = ..., msg_prefix: str = ...) -> None: ...
     def assertJSONEqual(
         self,
-        raw: str,
+        raw: str | bytes | bytearray,
         expected_data: dict[str, Any] | list[Any] | str | int | float | bool | None,
         msg: str | None = ...,
     ) -> None: ...
     def assertJSONNotEqual(
         self,
-        raw: str,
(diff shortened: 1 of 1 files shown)


tests/typecheck/test/test_testcase.yml
```diff
@@ -10,3 +10,30 @@
...
+              self.assertJSONNotEqual(b"{}", {"a": 1})
+
+-   case: testcase_assert_json_equal_bytearray
+    main: |
+      from django.test.testcases import TestCase
+
+      class ExampleTestCase(TestCase):
+          def test_method(self) -> None:
+              self.assertJSONEqual(bytearray(b"{}"), {})
+              self.assertJSONNotEqual(bytearray(b"{}"), {"a": 1})
```
sobolevn:
```suggestion
```

We don't need tests for simple annotations :)

sobolevn (commented):
Thanks!

tests/typecheck/test/test_testcase.yml
nickstenning: OK! I'll rebase those out.

sobolevn (approved):

PR #2498, opened 2025-02-03
DomainNameValidator types

HansAarneLiblik:
# I have made things!

I'm unable to use `django.core.validators.DomainNameValidator` in my project due to mypy complaining 
> error: Module "django.core.validators" has no attribute "DomainNameValidator"  [attr-defined]

This should hopefully fix that issue

## Related issues

- None. I can create one, if needed


sobolevn (approved):
Thanks!

PR #2509, opened 2025-02-11
fix: Add `context_data` attribute to _MonkeyPatchedW/ASGIResponse.

PacificGilly:
As detailed in Django's documentation [0], when a non-standard template backend is used, the `context_data` attribute is populated instead of `context` when the test API call is successful. I've seen this to be a populated attribute regardless of the backend template engine being used. Regardless, this near-hidden attribute is missing and should be included to help improve unit testing when validating that the response data contains the expected data.

[0] https://docs.djangoproject.com/en/stable/topics/testing/tools/#django.test.Response.context

## Related issues
None.

### DIFF
--- a/django-stubs/test/client.pyi
+++ b/django-stubs/test/client.pyi
@@ -189,7 +189,15 @@ class _MonkeyPatchedWSGIResponse(_WSGIResponse):
     request: dict[str, Any]
     client: Client
     templates: list[Template]
+
+    # `context` and `context_data` are populated based on whether you use the standard template
+    # backend or a custom one respectively.
+    # `context_data` only exists if the response was successful too as the return type changes.
+    # `HTTPResponse` when API failed and `TemplateResponse` when successful.
+    # https://docs.djangoproject.com/en/stable/topics/testing/tools/#django.test.Response.context
     context: ContextList | dict[str, Any]
+    context_data: ContextList | dict[str, Any]
+
     content: bytes
     resolver_match: ResolverMatch
     redirect_chain: list[tuple[str, int]]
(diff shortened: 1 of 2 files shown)


django-stubs/test/client.pyi
```diff
@@ -189,7 +189,15 @@ class _MonkeyPatchedWSGIResponse(_WSGIResponse):
...
     client: Client
     templates: list[Template]
-    context: ContextList | dict[str, Any]
+
+    # `context` and `context_data` are populated based on whether you use the standard template
+    # backend or a custom one respectively.
+    # `context_data` only exists if the response was successful too as the return type changes.
+    # `HTTPResponse` when API failed and `TemplateResponse` when successful.
+    # https://docs.djangoproject.com/en/stable/topics/testing/tools/#django.test.Response.context
+    context: ContextList | dict[str, Any] | None
```
sobolevn: I think that this can ruin valid use, basically you would have to do:

```python
assert response.context
assert response.context[key]
```

Right?

```
This attribute is only populated when using the [DjangoTemplates](https://docs.djangoproject.com/en/5.1/topics/templates/#django.template.backends.django.DjangoTemplates) backend. If you’re using another template engine, [context_data](https://docs.djangoproject.com/en/5.1/ref/template-response/#django.template.response.SimpleTemplateResponse.context_data) may be a suitable alternative on responses with that attribute.
```

PacificGilly: If you have a custom Template BE then context is null and you have to use context_data instead. From testing this out, I found context was always null on a 200 response regardless of template BE using Django 4.2

sobolevn: I am not against adding `context_data`, but I don't think we should add `None` type to these values.

PacificGilly: Ah okay, I see your point. I only added None type as that's what I saw during some TDD, but like you mentioned this might do more harm than good. Happy to remove the None types here 🙂

PacificGilly: Removed `None` now from both `context` and `context_data`.

sobolevn (approved):
Thank you!

PR #2515, opened 2025-02-19
Fix id and pk of AnonymousUser

### LINKED ISSUE #2513: Incorrect types for AnonymousUser
# Bug report

## What's wrong

Mypy is not reporting issues with the following:

```
def my_view(request: HttpRequest) -> HttpResponse:
    current_id_plus_one = request.user.id + 1
    return HttpResponse(current_id_plus_one)
```

## How is that should be

Mypy should throw an error because `request.user` can be an `AnonymousUser` which has [`id` and `pk` set to None](https://github.com/django/django/blob/611bf6c2e2a1b4ab93273980c45150c099ab146d/django/contrib/auth/models.py#L530). If I change the [stub](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/contrib/auth/models.pyi#L96) from `Any` to `None` things work as expected. I wanted to check if `Any` was used purposefully before opening a PR.

## System information

- OS:
- `python` version: `3.12.0`
- `django` version: `5.1.6`
- `mypy` version: `1.15.0`
- `django-stubs` version: `5.1.3`
- `django-stubs-ext` version: `5.1.3`.

stickperson:
# I have made things!

## Related issues

- Closes #2513

Would it also make sense to update `save()`, `delete()`, `set_password()`, and `check_password()` with the `NoReturn` type since they always raise?

### DIFF
--- a/django-stubs/contrib/auth/models.pyi
+++ b/django-stubs/contrib/auth/models.pyi
@@ -93,8 +93,8 @@ class AbstractUser(AbstractBaseUser, PermissionsMixin):
 class User(AbstractUser): ...
 
 class AnonymousUser:
-    id: Any
-    pk: Any
+    id: Literal[None]
+    pk: Literal[None]
     username: Literal[""]
     is_staff: Literal[False]
     is_active: Literal[False]
--- a/django-stubs/contrib/auth/models.pyi
+++ b/django-stubs/contrib/auth/models.pyi
@@ -93,8 +93,8 @@ class AbstractUser(AbstractBaseUser, PermissionsMixin):
 class User(AbstractUser): ...
 
 class AnonymousUser:
-    id: Literal[None]
-    pk: Literal[None]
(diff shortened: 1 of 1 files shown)


django-stubs/contrib/auth/models.pyi
```diff
@@ -93,8 +93,8 @@ class AbstractUser(AbstractBaseUser, PermissionsMixin):
 class User(AbstractUser): ...
 
 class AnonymousUser:
-    id: Any
-    pk: Any
+    id: Literal[None]
+    pk: Literal[None]
```
```suggestion
    id: None
    pk: None
```

sobolevn (commented):
Let's do one thing at a time

sobolevn (approved):
https://github.com/django/django/blob/65c46d6932c0956d2988d13ec3d9ab3ef9d96d61/django/contrib/auth/models.py#L529-L531

Thanks!

PR #2529, opened 2025-03-04
Fix `Func.function`

adamchainz:
# I have made things!

This attribute is allowed to be `None`, per [the source](https://github.com/django/django/blob/ae2736ca3bf4c6a27e23ee95530ad965b550d4cc/django/db/models/expressions.py#L1051). Allowing that ripples through to subclasses. I also fixed [`GeoFuncMixin`](https://github.com/django/django/blob/ae2736ca3bf4c6a27e23ee95530ad965b550d4cc/django/db/models/expressions.py#L1051), a mixin for `Func` subclasses.

## Related issues

N/A

### DIFF
--- a/django-stubs/db/models/expressions.pyi
+++ b/django-stubs/db/models/expressions.pyi
@@ -181,7 +181,7 @@ class OuterRef(F):
     def relabeled_clone(self, relabels: Any) -> Self: ...
 
 class Func(SQLiteNumericMixin, Expression):
-    function: str
+    function: str | None
     name: str
     template: str
     arg_joiner: str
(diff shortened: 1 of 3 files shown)


django-stubs/db/models/expressions.pyi
```diff
@@ -180,7 +180,7 @@ class OuterRef(F):
     def relabeled_clone(self, relabels: Any) -> Self: ...
 
 class Func(SQLiteNumericMixin, Expression):
-    function: str
+    function: str | None
```
flaeppe: Should we mark the default value during runtime in `Func` class as well?

```suggestion
    function: str | None = None
```

flaeppe (approved):

sobolevn (approved):

sobolevn:
It also have conflicts now :(

django-stubs/db/models/expressions.pyi
adamchainz: Thanks!

adamchainz:
> It also have conflicts now :(

fffffixed!

sobolevn:
Thank you! Great work!

PR #2532, opened 2025-03-07
Bump django to 5.1.7

sobolevn:
Closes https://github.com/typeddjango/django-stubs/pull/2531


django-stubs/contrib/admin/models.pyi
```diff
@@ -22,26 +22,13 @@ class LogEntryManager(models.Manager[LogEntry]):
...
-    @overload
-    def log_actions(
-        self,
-        user_id: int,
-        queryset: QuerySet[Model],
-        action_flag: int,
-        change_message: str | list[Any] = "",
-        *,
-        single_object: Literal[False] = False,
-    ) -> list[LogEntry]: ...
```
adamchainz, line 44: This change is getting reverted because it was backwards-incompatible:

https://code.djangoproject.com/ticket/36234

Can we revert here and pin at 5.1.8 instead?

sobolevn: Sure! As soon as 5.1.8 is released

sobolevn: I was surprised to see this change :)

PR #2535, opened 2025-03-12
Update for Model.[a]save() signature change

adamchainz:
# I have made things!

Update for this upstream change deprecating keyword arguments for the `save()` and `asave()` methods: https://github.com/django/django/commit/3915d4c70d0d7673abe675525b58117a5099afd3

## Related issues

...

### DIFF
--- a/django-stubs/db/models/base.pyi
+++ b/django-stubs/db/models/base.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Collection, Iterable, Sequence
-from typing import Any, ClassVar, Final, TypeVar, overload
+from typing import Any, ClassVar, Final, Never, TypeVar, overload
 
 from django.core.checks.messages import CheckMessage
 from django.core.exceptions import MultipleObjectsReturned as BaseMultipleObjectsReturned
@@ -71,6 +71,7 @@ class Model(metaclass=ModelBase):
     def get_constraints(self) -> list[tuple[type[Model], Sequence[BaseConstraint]]]: ...
     def save(
         self,
+        *args: Never,
         force_insert: bool | tuple[ModelBase, ...] = False,
         force_update: bool = False,
         using: str | None = None,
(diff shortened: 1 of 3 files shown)


django-stubs/db/models/base.pyi
```diff
@@ -71,13 +71,15 @@ class Model(metaclass=ModelBase):
     def get_constraints(self) -> list[tuple[type[Model], Sequence[BaseConstraint]]]: ...
     def save(
         self,
+        *args: Never,
```
adamchainz: Three things here...

First, I have only left the `*args` in to appease stubtest. We could perhaps ignore it instead, since the future is keyword-only, but then we might miss future keyword args.

Second, I picked `NoReturn` to stop typing users from passing the arguments. I think that's acceptable.

Third, [this ruff rule](https://docs.astral.sh/ruff/rules/no-return-argument-annotation-in-stub/) made me rewrite `NoReturn` as `Never`. But `Never` was only added in Python 3.11, while we still try to support down to 3.9. Is this okay? Does Mypy know the new name already on old Pythons, for stub files at least?

adamchainz: Okay, tests answered my third question. Python 3.9 failed, so I reverted to `NoReturn` and disabled that particular Ruff rule.

sobolevn (approved):

django-stubs/db/models/base.pyi
```diff
@@ -71,13 +71,15 @@ class Model(metaclass=ModelBase):
     def get_constraints(self) -> list[tuple[type[Model], Sequence[BaseConstraint]]]: ...
     def save(
         self,
+        *args: NoReturn,
```
sobolevn, line 74: Sorry, I for some reason accepted this without asking, but now I got really interested: why not just `*`?

adamchainz: That fails stubtest right now, since Django allows positional arguments still. Maybe that would be better for the types though, since users will probably want to adopt the new keyword-only format now?

sobolevn: Yeah, I think that we should add an allow rule for that. Can you please?

PR #2536, opened 2025-03-12
Update for template system checks change

adamchainz:
# I have made things!

Update for this upstream change which I proposed and reviewed: https://github.com/django/django/commit/d658a3162fbeb68d148d1b2fcf4da4fe1437eddb

The system checks for templates were pushed down to the engine classes, and mostly made private, so all that's left are the top-level check function and the engine's check method.

## Related issues

n/a


sobolevn (approved):
Great work!

PR #2538, opened 2025-03-12
Improve admin decorators

adamchainz:
# I have made things!

Following previous improvements (#1267, #1292), improve them some more, so we can drop the stubtest exception. The key is to still have the `function` argument but show that in the second case, it must be left with the default `None` to return a decorator.

## Related issues

n/a


sobolevn (approved):
Thanks!

PR #2544, opened 2025-03-12
Add full signatures to CheckConstraint.__init__ overloads

adamchainz:
# I have made things!

Following #2331 from @q0w which added the overloads to cover the upstream deprecation, add the missing args on each overload to stop stubtest errors that the signatures don't match.

## Related issues

n/a


sobolevn (approved):

sobolevn:
It has merge conflicts :(

adamchainz:
Fixed!

PR #2547, opened 2025-03-12
Add query_params argument to test tools from Django 5.1

adamchainz:
# I have made things!

From https://github.com/django/django/commit/a03593967f098cf8dab79065bcabbcebd461f05b .

## Related issues

n/a


django-stubs/test/client.pyi
```diff
@@ -6,6 +6,7 @@ from json import JSONEncoder
 from re import Pattern
 from types import TracebackType
 from typing import Any, Generic, Literal, NoReturn, TypedDict, TypeVar, type_check_only
+from urllib.parse import _QueryType
```
adamchainz: Is it okay to import this from typeshed?

https://github.com/python/typeshed/blob/8bc9a6536775465280a42745940575d2f5980078/stdlib/urllib/parse.pyi#L145C1-L147

Because ultimately `query_params` works by passing down to `urlencode()`.

But maybe that's too detailed and we should just go with `Mapping[Any, Any]` ?

sobolevn: Not really, it can be changed / renamed / removed in any typeshed version. It is better to copy-paste, in my opinon. Or use `Mapping[Any, Any]` as you said.

sobolevn (commented):
Everything, except the import, looks fine :)

django-stubs/test/client.pyi
adamchainz: I went with `Mapping[Any, Any]`.

PR #2550, opened 2025-03-14
Make Model.[a]save() keyword-only

adamchainz:
# I have made things!

Follow-up to https://github.com/typeddjango/django-stubs/pull/2535#discussion_r1992996956 .

## Related issues

n/a


sobolevn (approved):
Thanks for fixing!

PR #2553, opened 2025-03-15
fix #2552 by add ModelIterable to _need_generic

### LINKED ISSUE #2552: TypeError: type 'ModelIterable' is not subscriptable
# Bug report

## What's wrong
I am working on adding typing for django-polymorphic and encountered an issue with ModelIterable.

Here is the relevant code:

```
class PolymorphicModelIterable(Generic[T], ModelIterable[T]):
```

- If I don’t add TypeVar for ModelIterable, mypy complains about a `type-var` error.
- If I do add TypeVar, Python raises a TypeError:

```
TypeError: type 'ModelIterable' is not subscriptable
```

- However, if I add ModelIterable to _need_generic in patch.py, the issue is fixed.

## How is that should be

- ModelIterable should be recognized as generic when used with Generic[T] without causing a TypeError.
- mypy should not complain about a missing TypeVar if ModelIterable is used as a base class.

## System information

- OS: Mac
- `python` version: 3.11
- `django` version: 5.1.7
- `mypy` version: 1.13.0
- `django-stubs` version:  5.1.1

lucemia:
# I have made things!

- fix #2552

## Related issues
N/A

### DIFF
--- a/ext/django_stubs_ext/patch.py
+++ b/ext/django_stubs_ext/patch.py
@@ -16,7 +16,7 @@ from django.db.models.fields.related import ForeignKey
 from django.db.models.fields.related_descriptors import ReverseManyToOneDescriptor
 from django.db.models.lookups import Lookup
 from django.db.models.manager import BaseManager
-from django.db.models.query import QuerySet, RawQuerySet, ModelIterable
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks!

PR #2554, opened 2025-03-15
Add window frame exclusion from Django 5.1

adamchainz:
# I have made things!

From this commit: https://github.com/django/django/commit/e4d012ca05b23445f422b0400d6dd7aa85743885

## Related issues

n/a


django-stubs/db/models/expressions.pyi
```diff
@@ -298,13 +299,25 @@ class Window(SQLiteNumericMixin, Expression):
         output_field: Field | None = None,
     ) -> None: ...
 
+class WindowFrameExclusion(Enum):
+    CURRENT_ROW = "CURRENT ROW"
```
sobolevn, line 303: New style stub `Enum`s, really cool!

sobolevn (approved):
Thank you!

PR #2555, opened 2025-03-15
Update contrib.auth for Django 5.1

adamchainz:
# I have made things!

From https://github.com/django/django/commit/00e187961059a0e77403151d2bb38c217101d5af:
* `django.contrib.auth.base_user.BaseUserManager.make_random_password` removed

From https://github.com/django/django/commit/f64c528c17af91333fc6e96d91b5f13c0fb647d8:

* `django.contrib.auth.forms.AdminPasswordChangeForm.clean_password2` removed
* `django.contrib.auth.forms.BaseUserCreationForm.clean_password2` removed
* `django.contrib.auth.forms.SetPasswordForm.clean_new_password2` removed
* `django.contrib.auth.forms.SetPasswordMixin` added

From https://github.com/django/django/commit/e626716c28b6286f8cf0f8174077f3d2244f3eb3:

* `django.contrib.auth.forms.AdminPasswordChangeForm.usable_password_help_text` added

From https://github.com/django/django/commit/da22e6cb3cdb51f2f7ca45e1591d2c6a8be458e0:

* `django.contrib.auth.forms.SetUnusablePasswordMixin` added
* `django.contrib.auth.forms.AdminUserCreationForm` added

From https://github.com/django/django/commit/c7fc9f20b49b5889a9a8f47de45165ac443c1a21:

* `django.contrib.auth.middleware.LoginRequiredMiddleware` added

## Related issues

n/a

PR #2556, opened 2025-03-15
Improve process_view() methods

adamchainz:
# I have made things!

I noticed whilst working on the new `process_view()` added in #2555 that these two were inconsistent and incomplete, so expanding them now.

## Related issues

n/a


django-stubs/contrib/admindocs/middleware.pyi
```diff
@@ -10,6 +10,6 @@ class XViewMiddleware(MiddlewareMixin):
         self,
         request: HttpRequest,
         view_func: Callable[..., HttpResponseBase],
```
sobolevn, line 12: Just from the structure it looks like it can be a `ParamSpec` thing, can't it?

django-stubs/middleware/csrf.pyi
```diff
@@ -49,7 +49,11 @@ class CsrfViewMiddleware(MiddlewareMixin):
     def allowed_origin_subdomains(self) -> defaultdict[str, list[str]]: ...
     def process_request(self, request: HttpRequest) -> None: ...
     def process_view(
-        self, request: HttpRequest, callback: Callable | None, callback_args: tuple, callback_kwargs: dict[str, Any]
+        self,
+        request: HttpRequest,
+        callback: Callable[..., HttpResponseBase],
```
sobolevn, line 54: The same here.

django-stubs/contrib/admindocs/middleware.pyi
sobolevn: Never mind, it is not `*` and `**`

sobolevn (approved):

django-stubs/contrib/admindocs/middleware.pyi
adamchainz: I don't think so... `process_view()` methods don't call `view_func`. Their return values can come only from the middleware.

https://docs.djangoproject.com/en/5.1/topics/http/middleware/#process-view

PR #2560, opened 2025-03-15
Complete JSONObject

adamchainz:
# I have made things!

Cover the `as_*` methods, including `as_native` added in https://github.com/django/django/commit/81ccf92f154c6d9eac3e30bac0aa67574d0ace15 .

## Related issues

n/a

### DIFF
--- a/django-stubs/db/models/functions/comparison.pyi
+++ b/django-stubs/db/models/functions/comparison.pyi
@@ -1,9 +1,11 @@
 from typing import Any, ClassVar
 
+from django.db.backends.base.base import BaseDatabaseWrapper
 from django.db.models import Func
 from django.db.models.expressions import Combinable
 from django.db.models.fields import Field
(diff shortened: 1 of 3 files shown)


django-stubs/db/models/functions/comparison.pyi
```diff
@@ -16,8 +18,16 @@ class Collate(Func):
 class Greatest(Func): ...
 
 class JSONObject(Func):
-    def __init__(self, **fields: Any) -> None: ...
     output_field: ClassVar[JSONField]
+    def __init__(self, **fields: Any) -> None: ...
+    def as_sql(self, compiler: SQLCompiler, connection: BaseDatabaseWrapper, **extra_context: Any) -> _AsSqlType: ...  # type: ignore [override]
```
adamchainz, line 23: Incompatible override in Django itself, supporting other args only through kwargs

sobolevn (approved):

PR #2559, opened 2025-03-15
Update test.selenium for Django 5.1

adamchainz:
# I have made things!

From https://github.com/django/django/commit/f6629ee2c986d3bf59b4c1b3058f370a00bdc573:

* `django.test.selenium.ChangeWindowSize` added
* `django.test.selenium.SeleniumTestCase.desktop_size` added
* `django.test.selenium.SeleniumTestCase.mobile_size` added
* `django.test.selenium.SeleniumTestCase.small_screen_size` added

From https://github.com/django/django/commit/be56c982c0805f62a2948d9d7a3e21215c352174:

* `django.test.selenium.SeleniumTestCase.dark` added
* `django.test.selenium.SeleniumTestCase.rtl` added
* `django.test.selenium.SeleniumTestCase.take_screenshot` added
* `django.test.selenium.SeleniumTestCase.screenshots` added
* `django.test.selenium.screenshot_cases` added

From https://github.com/django/django/commit/39828fa7786a805f7a542abb929f352d3c5b722c:

* `django.test.selenium.SeleniumTestCase.high_contrast` added
* `django.test.selenium.SeleniumTestCase.set_emulated_media` added

From https://github.com/django/django/commit/ceaf1e2848583ba832cc74715da38c802b6b0671:

* `django.test.selenium.SeleniumTestCase.set_emulated_media` adjusted

## Related issues

n/a

PR #2561, opened 2025-03-15
Use AbstractContextManager instead of @contextmanager

adamchainz:
# I have made things!

Per [typing docs stubs guide](https://typing.python.org/en/latest/guides/writing_stubs.html#decorators).

## Related issues

Following [this comment](https://github.com/typeddjango/django-stubs/pull/2559#discussion_r1997396220).


sobolevn (approved):
Thank you!

PR #2565, opened 2025-03-16
Update _Choices type for forms

mrgalopes:
form.ChoiceField and other related fields accept the same input as the model's choices argument (https://docs.djangoproject.com/en/5.2/ref/forms/fields/#django.forms.ChoiceField.choices). This updates the typing for the form to be the same as the models'.

Let me know what you think. Thanks!

Refs #2476

### DIFF
--- a/django-stubs/db/models/fields/__init__.pyi
+++ b/django-stubs/db/models/fields/__init__.pyi
@@ -15,8 +15,7 @@ from django.db.models.fields.reverse_related import ForeignObjectRel
 from django.db.models.query_utils import Q, RegisterLookupMixin
 from django.db.models.sql.compiler import SQLCompiler, _AsSqlType, _ParamsT
 from django.forms import Widget
-from django.utils.choices import BlankChoiceIterator, _Choice, _ChoiceNamedGroup, _ChoicesCallable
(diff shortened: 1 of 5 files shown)


django-stubs/db/models/fields/__init__.pyi
```diff
@@ -9,14 +9,14 @@ from django import forms
...
+from django.db.models import Model
 from django.db.models.expressions import Col, Combinable, Expression, Func
 from django.db.models.fields.reverse_related import ForeignObjectRel
 from django.db.models.query_utils import Q, RegisterLookupMixin
 from django.db.models.sql.compiler import SQLCompiler, _AsSqlType, _ParamsT
 from django.forms import Widget
-from django.utils.choices import BlankChoiceIterator, _Choice, _ChoiceNamedGroup, _ChoicesCallable, _ChoicesMapping
-from django.utils.choices import _Choices as _ChoicesSequence
+from django.utils.choices import BlankChoiceIterator, _Choice, _ChoiceNamedGroup, _ChoicesCallable
+from django.utils.choices import _ChoicesInput as _Choices
```
sobolevn: Let's not do an alias here, it is a bit confusing.

sobolevn (approved):
Thanks!

PR #2568, opened 2025-03-17
Make GeneratedField parameter db_persist required

JasonLovesDoggo:
see https://github.com/django/django/blob/ef6a83789b310a441237a190a493c9586a4cb260/django/db/models/fields/generated.py#L27-L28


adamchainz:
Can you suggest a change upstream in Django too? It doesn't make sense to have a default of `None` when that doesn't work.

adamchainz (changes requested):
This fails stubtest because the signature no longer matches.

I think it's okay to add an exclusion in `allowlist.txt`, with an appropriate comment.

JasonLovesDoggo:
> This fails stubtest because the signature no longer matches.
> 
> I think it's okay to add an exclusion in `allowlist.txt`, with an appropriate comment.

Created https://github.com/django/django/pull/19281, will fix the failing tests when I get home in about 7 hours. 

Can you show me how you modify allowlist.txt?

JasonLovesDoggo:
Not sure how to fix https://github.com/typeddjango/django-stubs/actions/runs/13890824623/job/38879638893?pr=2568 as given the other testcases I cant see why that fails

JasonLovesDoggo:
My PR linked above has been merged, just needs the signature change here

adamchainz:
So your upstream PR was merged for Django 6.0. We currently target 5.1 here, and will update to target 5.2 after it's released. We'll need an allowlist entry here until 6.0.

adamchainz:
I have added the allowlist entry.

Your tests were failing because they declared an installed app `myapp` but didn't include source files for the app.

While reviewing the tests, I realized they're pretty redundant, and just checking basic type-checker behaviour, so I dropped them.

Thank you for the PRs here and upstream, will merge when green!

adamchainz (approved):

PR #2570, opened 2025-03-18
Remove unused code from the plugin

ngnpope:
# I have made things!

Removes a smattering of things that I found were no longer in use.

## Related issues

N/A


sobolevn (approved):
Thanks!

PR #2572, opened 2025-03-18
Improve stubs for `flatten` and `flatten_fieldsets`

ngnpope:
# I have made things!

These functions are implemented in a generic way that doesn't generally care about the content. The existing return type of `Callable | str` makes these awkward to use.

I've changed the use of `Sequence` in the `_FieldGroups` alias to use `_ListOrTuple`. This is another case where `ModelAdmin.fields` and the `"fields"` item in fieldsets are presumed by Django to be a `list` or `tuple`, e.g. there are some system checks that ensure this.

See https://github.com/django/django/blob/afbb8c709d40e77b3f71c152d363c5ad95ceec2d/django/contrib/admin/utils.py#L102-L120

## Related issues

N/A


ngnpope:
_(Failing check just needs a re-run.)_

django-stubs/contrib/admin/utils.pyi
```diff
@@ -31,8 +31,8 @@ def prepare_lookup_value(
 def build_q_object_from_lookup_parameters(parameters: dict[str, list[str]]) -> Q: ...
 def quote(s: int | str | UUID) -> str: ...
 def unquote(s: str) -> str: ...
-def flatten(fields: Any) -> list[Callable | str]: ...
-def flatten_fieldsets(fieldsets: Any) -> list[Callable | str]: ...
+def flatten(fields: _FieldGroups[_T]) -> list[_T]: ...
```
sobolevn: Can we please add a test case for this? It seems complex enough.

sobolevn (commented):
Thanks!

sobolevn (commented):
Looks like pyright is not happy with something :(

ngnpope:
> Looks like pyright is not happy with something :(

Yeah... Looking into that. It's hard to get the different type checkers to agree on pretty much anything at all it seems 😞

I've fixed a bunch of the pyright issues, ignored a few around the model fields that can't be handled until we support making fields generic as per #2214, and a handful still to check.

The good news is that, by adding some testing, I ended up simplifying the implementation.

ngnpope:
Success @sobolevn! Somehow it's turned out rather clean after my silly false starts! 😅

sobolevn (approved):
Awesome! Thanks!

PR #2576, opened 2025-03-19
Fix BaseModelAdmin.get_inlines() return type

### LINKED ISSUE #2574: `BaseModelAdmin.get_inlines()` return type doesn't match `ModelAdmin.inlines` type
# Bug report

## What's wrong

In `django-stubs/contrib/admin/options.pyi`, we have that `BaseModelAdmin.get_inlines()` should return a `list`:

```pyi
class BaseModelAdmin(Generic[_ModelT]):
    # ...
    def get_inlines(self, request: HttpRequest, obj: _ModelT | None) -> list[type[InlineModelAdmin]]: ...
```

But also that `ModelAdmin.inlines` should be a `list` _or a `tuple`_:

```pyi
    # ...
    inlines: ClassVar[_ListOrTuple[type[InlineModelAdmin]]]
```

And indeed the default implementation of `BaseModelAdmin.get_inlines()` simply returns the value of `inlines` (`django/contrib/admin/options.py`):
```py
    def get_inlines(self, request, obj):
        """Hook for specifying custom inlines."""
        return self.inlines
```

This means that we get a type error when overriding `get_inlines()` on a ModelAdmin, even if we use the exact default implementation. For example, this code yields an "Incompatible return value type" error:

```python
from typing import Type

from django.contrib.admin import ModelAdmin
from django.contrib.admin.options import InlineModelAdmin


class FooAdmin(ModelAdmin):
    def get_inlines(self, request, obj) -> list[Type[InlineModelAdmin]]:
        return self.inlines
```

## How it should be

1. The `BaseModelAdmin.get_inlines()` return type should at least allow the `ModelAdmin.inlines` types to be returned as-is, to match the default implementation of the hook, e.g.
```pyi
def get_inlines(self, request: HttpRequest, obj: _ModelT | None) -> _ListOrTuple[type[InlineModelAdmin]]: ...
```
2. The [Django docs](https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_inlines) specify that `get_inlines()` "is expected to return an iterable of inlines", so possibly one or both of the types in question should be `Iterable` rather than `_ListOrTuple`.

## System information

This is also an issue in django-stubs 4.2.x, where the hook still returns a `list`, but `inlines` is a `Sequence` rather than `_ListOrTuple`.

- OS: MacOS
- `python` version: 3.11.10
- `django` version: 5.1.7
- `mypy` version: 1.15.0

MaestroXXXVIII:
# I have made things!
I first changed the return type to Iterable for inlines as written in the django documentation.

I decided to use Iterable as described in the documentation, and it doesn't break anything and no error pops up. But the framework itself forbids us to work with types other than list, tuple. What do we want, to conform to the behavior of the framework or to what is written in its documentation?
I guess from the user's point of view it would be nice to tell the user in advance that this is a _ListOrTuple and not some iterated object?

So I think it would be logical to use a more specific type _ListOrTuple.

## Related issues
Fixes #2574


sobolevn (approved):
Yes, we should match the implementation, not docs. Thanks!

adamchainz (approved):
Thank you!

PR #2580, opened 2025-03-23
Run tests in parallel

UnknownPlatypus:
# I have made things!

This speed up the `django-stubs` test suite by using multiprocessing with [`pytest-xdist`](https://github.com/pytest-dev/pytest-xdist).

- Locally, I went from **~5min to ~25s** for the whole test suite.
- In CI, it went from **[~3min](https://github.com/typeddjango/django-stubs/actions/runs/14017953849/job/39246069086) to [~1min20s](https://github.com/typeddjango/django-stubs/actions/runs/14070003527/job/39401764007?pr=2580)** per shard

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,8 +69,6 @@ jobs:
       # Must match `shard` definition in the test matrix:
       - name: Run pytest tests
         run: PYTHONPATH='.' pytest --num-shards=4 --shard-id=${{ matrix.shard }} -n auto tests --durations=0
-        env:
-          PYTEST_XDIST_AUTO_NUM_WORKERS: 4
       - name: Run mypy on the test cases
         run: mypy --strict tests/assert_type
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,7 +70,7 @@ jobs:
       - name: Run pytest tests
         run: PYTHONPATH='.' pytest --num-shards=4 --shard-id=${{ matrix.shard }} -n auto tests
         env:
-          PYTEST_XDIST_AUTO_NUM_WORKERS: 2
+          PYTEST_XDIST_AUTO_NUM_WORKERS: 3
       - name: Run mypy on the test cases
         run: mypy --strict tests/assert_type
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,7 +68,7 @@ jobs:
 
       # Must match `shard` definition in the test matrix:
       - name: Run pytest tests
-        run: PYTHONPATH='.' pytest --num-shards=4 --shard-id=${{ matrix.shard }} -n auto tests --durations=0
+        run: PYTHONPATH='.' pytest --num-shards=4 --shard-id=${{ matrix.shard }} -n auto tests
       - name: Run mypy on the test cases
         run: mypy --strict tests/assert_type
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,6 +69,8 @@ jobs:
       # Must match `shard` definition in the test matrix:
       - name: Run pytest tests
         run: PYTHONPATH='.' pytest --num-shards=4 --shard-id=${{ matrix.shard }} -n auto tests
+        env:
+          PYTEST_XDIST_AUTO_NUM_WORKERS: 2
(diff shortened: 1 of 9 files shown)


adamchainz:
> 4 subprocess doesn't improve runtime, try 2

The free GHA runners only have 2 CPUs, so that would make sense.

adamchainz:
> The free GHA runners only have 2 CPUs, so that would make sense.

Wait, I'm wrong, [it's 4](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#choosing-github-hosted-runners) - hence `-n auto` chose 4.

UnknownPlatypus:
Yes I thought maybe there was some sort of CPU congestion but lowering to 2 makes it slower so there is some sort of overhead I'm missing.

scripts/tests_extension_hook.py
```diff
@@ -30,5 +30,8 @@ def django_plugin_hook(test_item: YamlTestItem) -> None:
         if "[mypy.plugins.django-stubs]" not in test_item.additional_mypy_config:
             test_item.additional_mypy_config += django_settings_section
 
+    # Disable caching in tests to allow parallel execution
+    test_item.disable_cache = True
```
sobolevn: I am not sure that we want to do that. Cache is very important in mypy. It is not a good idea to disable it :(

sobolevn:
If we want faster CI, we can increase the amount of shards. But, it will cosume more resources from GitHub. I think that 2m per job right now is very good result.

scripts/tests_extension_hook.py
UnknownPlatypus: Indeed, it was doing a lot more than I thought, i kinda forgot it also caches a lot of thing from stlib and dependencies. Thanks for pointing that out!

UnknownPlatypus:
> If we want faster CI, we can increase the amount of shards. But, it will cosume more resources from GitHub. I think that 2m per job right now is very good result.

To be fair my initial goal was to speed up local test runs, CI was a bonus since It's quite ok right now.

But anyway, I figured it out and fixed the last issue I had about a race condition when accessing the mypy cache from concurrent processes (I was facing this [known issue](https://github.com/python/mypy/issues/14521)). Setting a different `mypy_cache_dir` per processes did the trick.

This is really great!
- Locally, I went from **~5min to ~25s** for the whole test suite.
- In CI, it went from **[~3min](https://github.com/typeddjango/django-stubs/actions/runs/14017953849/job/39246069086) to [~1min20s](https://github.com/typeddjango/django-stubs/actions/runs/14070003527/job/39401764007?pr=2580)** per shard

.github/workflows/test.yml
```diff
@@ -68,7 +68,7 @@ jobs:
 
       # Must match `shard` definition in the test matrix:
       - name: Run pytest tests
-        run: PYTHONPATH='.' pytest --num-shards=4 --shard-id=${{ matrix.shard }} tests
+        run: PYTHONPATH='.' pytest --num-shards=4 --shard-id=${{ matrix.shard }} -n auto tests --durations=0
```
sobolevn:
```suggestion
        run: PYTHONPATH='.' pytest --num-shards=4 --shard-id=${{ matrix.shard }} -n auto tests
```

Let's remove this:

```
 ============================== slowest durations ===============================
```

sobolevn (approved):
Thank you! Now it looks really cool!

tests/typecheck/managers/querysets/test_from_queryset.yml
```diff
@@ -684,7 +683,6 @@
 # For details see: https://github.com/typeddjango/django-stubs/issues/1022
 -   case: from_queryset_custom_auth_user_model
     # Cache needs to be disabled to consistently reproduce the bug
```
ngnpope, line 685: This comment should be removed?

sobolevn: Or better `disable_cache: true` restored, good catch!

UnknownPlatypus: Yes my bad, I forgot to revert the disable_cache lines when cleaning up. Can send a pr later for that

adamchainz:
Thank you @UnknownPlatypus !

PR #2582, opened 2025-03-25
Improve typing of choices

ngnpope:
# I have made things!

So I finally got around to looking at improving the typing of `Choices`, etc.

I started out with some small bits of clean up:

- Prefer `ChoicesType` to `ChoicesMeta` naming
- Only add `__contains__` for `Choices` on Python < 3.12
- Standardize naming of modules under tests/assert_type

Added many more tests based on examples from Django's test suite:

- Make choices tests more comprehensive

Fixed some things that were incorrect (or didn't mirror what Django does):

- Simplify some imports and remove a lint violation 
- Fix `_enum_property` for Python < 3.11

Then made the main changes:

- Fix typing of labels for `Choices`
  - _Labels are translatable so should use `_StrOrPromise`_
- Fix typing of `do_not_call_in_templates` for `Choices`
  - _Using `@property` for this was actually kinda broken..._
  - _With Python 3.11 it was switched to `@enum.nonmember`..._
  - _It's better typed as `Literal[True]`_
- Fix typing of `Choices` with `__empty__`
  - _Introduces a hook in the plugin to detect `__empty__`..._
  - _...and make the value type in `.choices` and `.values` optional_
- Fix typing of Choices for custom mixed-in type
  - _Extends the new hook in the plugin to detect `Any`..._
  - _...and make the value type in `.choices`, `.values` and `.value` match the mixed-in base type_

It's my first time properly writing some mypy plugin stuff, so please check it carefully - it seems to work well though 😄 

## Related issues

N/A


mypy_django_plugin/transformers/choices.py
```diff
@@ -0,0 +1,87 @@
...
+
+    if not node.is_enum or not node.has_base(fullnames.CHOICES_CLASS_FULLNAME):
+        return default_attr_type
+
+    # Enums with more than one base will treat the first base as the mixed-in type.
+    base_type = node.bases[0] if len(node.bases) > 1 else None
+
+    # When `__empty__` is defined, the `.choices` and `.values` properties will include `None` for
+    # the blank choice which is labelled by the value of `__empty__`.
+    has_blank_choice = "__empty__" in node.names
```
sobolevn: Should this respect `__empty__` defined in base classes? It would be safer to use 

```suggestion
    has_blank_choice = node.get("__empty__")
```

tests/assert_type/db/models/test_enums.py
```diff
@@ -0,0 +1,132 @@
...
+    FEMALE = "F"
+    NOT_SPECIFIED = "X"
+
+    __empty__ = "(Undeclared)"
+
+
+class Medal(TextChoices):
+    GOLD = enum.auto()
+    SILVER = enum.auto()
+    BRONZE = enum.auto()
```
sobolevn, line 44: Do we care about a special case with `auto()` and `__empty__`?

tests/assert_type/db/models/test_enums.py
```diff
@@ -0,0 +1,132 @@
+import enum
+from typing import Literal, Optional
```
sobolevn: Let's convert this to mypy only test file (`.yml` based), because these tests will once be used by pyright as well.

And since this is a plugin-based feature, it won't be supported by pyright.

sobolevn (commented):
Thanks, a couple of questions :)

mypy_django_plugin/transformers/choices.py
ngnpope: I don't think so. When I implemented the choices stuff, it was only considered that `__empty__` would be defined on the class that introduced the members. I guess that it's _theoretically_ possible to do:

```python
class BaseEmptyChoices(models.TextChoices):
    __empty__ = "The Void"

class PlanetChoices(BaseEmptyChoices):
    MERCURY = enum.auto()
    ...
```

But I'm not sure it's a pattern to be encouraged.

tests/assert_type/db/models/test_enums.py
ngnpope: I don't think we need it - the two things don't interact.

tests/assert_type/db/models/test_enums.py
ngnpope: It's only a handful of cases that are affected by the plugin, for which there are `# pytest: ignore[...]` comments.

The majority is still useful for testing the stubs against pyright.

mypy_django_plugin/transformers/choices.py
sobolevn: > it's a pattern to be encouraged

Yeah, I undertand that. But, mypy is not about "encouriging", but about "it actually works". So, if this works in runtime, we should make it possible to check with mypy as well. 

Unless there are some big and complex problems.

tests/assert_type/db/models/test_enums.py
sobolevn: ok

tests/assert_type/db/models/test_enums.py
sobolevn: ok!

mypy_django_plugin/transformers/choices.py
ngnpope: Ok. I'll look into it.

ngnpope: Added this change 👍🏻

sobolevn (approved):
Thanks!

adamchainz:
Thanks @ngnpope , very solid work.

I have tried this on a project and I’m seeing a bit of churn due to labels changing to `StrOrPromise`, but that's ultimately the right type, so at least we're fixing it now.

ngnpope:
When testing this out on a larger repo, it's been seen that there are a couple of oversights here and perhaps some things that we can do to make the change a bit less imposing... I'm looking into that. Just a heads-up in case any reports come in.

sobolevn:
Thank you! 👍

ngnpope:
@sobolevn See #2646 for fixes.

PR #2584, opened 2025-03-26
Ignore __slotnames__

adamchainz:
# I have made things!

I noticed that we had a lot of `__slotnames__` entries in our allow lists.

I found that this is added by Python's pickling/copying machinery:

https://github.com/python/cpython/blob/5abff6960b4aecb0d5c81c7482cf3faa74e1983d/Lib/copyreg.py#L112-L161

https://github.com/python/cpython/blob/77d2fd441320ca365c34e21c475d3878f3d5d833/Objects/typeobject.c#L7036-L7083

It seems like it can be safely ignored.

I wondered if this attribute could be ignored by default in stubtest, or even added to `type` or `object` in typeshed (as `list[str]`), but I am not sure how sensible that is. It doesn't always exist, only after a pickle.

## Related issues

n/a


sobolevn (approved):
Thanks!

ngnpope:
> I wondered if this attribute could be ignored by default in stubtest...

It looks like we could just add this exclusion [here](https://github.com/python/mypy/blob/64b0a571e78c098316d69ebe280d8ea0d96a2737/mypy/stubtest.py#L1508-L1562)?

sobolevn:
yes, seems reasonable, ping me on PR :)

ngnpope:
> yes, seems reasonable, ping me on PR :)

https://github.com/python/mypy/pull/19077

PR #2589, opened 2025-03-27
Re-add explicit test cache disabling

UnknownPlatypus:
Addresses https://github.com/typeddjango/django-stubs/pull/2580#discussion_r2014010118 by bringing back the explicit disable_cache keys


adamchainz:
Thanks for coming back to it.

PR #2590, opened 2025-03-30
Feat/builtin models types

FallenDeity:
# I have made things!

This PR aims to add type hints to builtin model fields, i.e for example models in `contrib`, `admin`, `auth` etc.

Base generic fields are modified to use the `default=` from [PEP-696](https://peps.python.org/pep-0696/) to allow the following behaviour for models

```py
_ST_IntegerField = TypeVar("_ST_IntegerField", default=float | int | str | Combinable)
_GT_IntegerField = TypeVar("_GT_IntegerField", default=int)

class IntegerField(Field[_ST_IntegerField, _GT_IntegerField]):
    _pyi_private_set_type: float | int | str | Combinable
    _pyi_private_get_type: int
    _pyi_lookup_exact_type: str | int
```

```py
class Redirect(models.Model):
    id: models.AutoField
    pk: models.AutoField
    site: models.ForeignKey[Site | Combinable, Site]
    site_id: int
    old_path: models.CharField
    new_path: models.CharField
```
It eliminates the need to add generic arguments explicitly at each step whenever defining models, and ensures all the models being used internally have type hints.

This PR is not complete yet as I have a few questions to ask as I am not entirely clear about the whole process here are a few points I had doubts on, and wanted some feedback before proceeding

1. ~~Is there any particular reason models for `contrib/gis/db/backends` are typed Any and not included in `allowlist_todo.txt`, since they seem to have types [here](https://github.com/django/django/blob/c1a4fccf53437e480c42648f67277fb7e63ed4ab/django/contrib/gis/db/backends/postgis/models.py), should I type out the model fields as such~~ 
edit: models typed in line with source

```py
class OracleGeometryColumns(models.Model):
    table_name: models.CharField
    column_name: models.CharField
    srid: models.IntegerField
    objects: ClassVar[Manager[Self]]
```

2. ~~I noticed the presence of most of the inbuilt model fields and methods in `allowlist_todo.txt` so how would that be handled do I write tests under the `assert_type` and `typecheck/contrib` folder for these models and move them to `allowlist.txt` manually if they are type hinted and tested?~~ 
edit: Resolved created a category

3. ~~Final question, there is one case where the current model field typing system might need some explicit type hinting from the user in case of where fields become optional or with `null=True`, because with current system I don't think there is a way to infer types from the field parameters passed, Here is an example~~

edit: Resolved using `__new__` overloads

```py
text: models.CharField # get or return type is str, as expected
# But when params like blank=True, null=True are passed no way to autoinfer that and change _GT, _ST
# i.e without explicit typehints or generic params it would still show `str`, if we were to do `reveal_type`
# Redundant verbose example
text_nullable = models.CharField[Optional[Union[str, int, Combinable]], Optional[str]](max_length=100, null=True)
# A more generic user might have to do this
text_nullable = models.CharField[str | None, str | None](max_length=100, null=True)
```
This is fixable if we modify the `django-mypy-plugin` code but not sure if thats the best option, here is how one might go about it
```diff
def set_descriptor_types_for_field(
    ctx: FunctionContext, *, is_set_nullable: bool = False, is_get_nullable: bool = False
) -> Instance:
    default_return_type = cast(Instance, ctx.default_return_type)

    # Check for null_expr and primary key stuff
    ...

+  # We get expected nullable types here
    set_type, get_type = get_field_descriptor_types(
        default_return_type.type,
        is_set_nullable=is_set_nullable or is_nullable,
        is_get_nullable=is_get_nullable or is_nullable,
    )

    # reconcile set and get types with the base field class
    base_field_type = next(base for base in default_return_type.type.mro if base.fullname == fullnames.FIELD_FULLNAME)
    mapped_instance = map_instance_to_supertype(default_return_type, base_field_type)
+  # But mapped types give use the generic types we have in our fields without None
    mapped_set_type, mapped_get_type = tuple(get_proper_type(arg) for arg in mapped_instance.args)

    # bail if either mapped_set_type or mapped_get_type have type Never
    if not (isinstance(mapped_set_type, UninhabitedType) or isinstance(mapped_get_type, UninhabitedType)):
        # always replace set_type and get_type with (non-Any) mapped types
        set_type = helpers.convert_any_to_type(mapped_set_type, set_type)
        get_type = get_proper_type(helpers.convert_any_to_type(mapped_get_type, get_type))

-       # the get_type must be optional if the field is nullable
+      # Instead of nullable expression is set to True we make out mapped type optional
        if (is_get_nullable or is_nullable) and not (
            isinstance(get_type, NoneType) or helpers.is_optional(get_type) or isinstance(get_type, AnyType)
        ):
+            get_type = helpers.make_optional_type(get_type)
+           set_type = helpers.make_optional_type(set_type)
-            ctx.api.fail(
-               f"{default_return_type.type.name} is nullable but its generic get type parameter is not optional",
-                ctx.context,
-           )

    return helpers.reparametrize_instance(default_return_type, [set_type, get_type])
```
This was the only way I could think of going about it, and user needs to have the plugin installed.

## Related issues

- Refs: #2214 

## TODO

- [x] Add more tests
- [x] Work on the `contrib/gis/backends` models (most probably) 
- [x] Add overloads per field for `__new__` with `null`


django-stubs/contrib/admin/models.pyi
```diff
@@ -28,13 +31,19 @@ class LogEntryManager(models.Manager[LogEntry]):
         queryset: QuerySet[Model],
         action_flag: int,
         change_message: str | list[Any] = "",
+        *,
+        single_object: bool = False,
     ) -> list[LogEntry] | LogEntry: ...
 
 class LogEntry(models.Model):
+    id: models.AutoField
```
sobolevn, line 39: Right now we have both ways of annotating this: as `: models.AutoField` and as `= models.AutoField()`

Which one is better for both pyright and mypy?

django-stubs/contrib/auth/models.pyi
```diff
@@ -20,22 +22,68 @@ class PermissionManager(models.Manager[Permission]):
...
+    user_set: ManyToManyDescriptor[User, User_permissions]
     def natural_key(self) -> tuple[str, str, str]: ...
 
 class GroupManager(models.Manager[Group]):
     def get_by_natural_key(self, name: str) -> Group: ...
 
+# This is a model that only exists in Django's model registry and doesn't have any
+# class statement form. It's the through model between 'Group' and 'Permission'.
+@type_check_only
+class Group_permissions(models.Model):
```
```suggestion
class _Group_permissions(models.Model):
```

django-stubs/contrib/auth/models.pyi
```diff
@@ -20,22 +22,68 @@ class PermissionManager(models.Manager[Permission]):
...
+    pk: models.AutoField
+    group: models.ForeignKey[Group | Combinable, Group]
+    group_id: int
+    permission: models.ForeignKey[Permission | Combinable, Permission]
+    permission_id: int
+
+# This is a model that only exists in Django's model registry and doesn't have any
+# class statement form. It's the through model between 'User' and 'Group'.
+@type_check_only
+class User_groups(models.Model):
```
```suggestion
class _User_groups(models.Model):
```

django-stubs/contrib/auth/models.pyi
```diff
@@ -20,22 +22,68 @@ class PermissionManager(models.Manager[Permission]):
...
+    pk: models.AutoField
+    user: models.ForeignKey[User | Combinable, User]
+    user_id: int
+    group: models.ForeignKey[Group | Combinable, Group]
+    group_id: int
+
+# This is a model that only exists in Django's model registry and doesn't have any
+# class statement form. It's the through model between 'User' and 'Permission'.
+@type_check_only
+class User_permissions(models.Model):
```
```suggestion
class _User_permissions(models.Model):
```

django-stubs/contrib/flatpages/models.pyi
```diff
@@ -1,12 +1,31 @@
...
+
 from django.contrib.sites.models import Site
 from django.db import models
+from django.db.models.expressions import Combinable
+from typing_extensions import Self
+
+# This is a model that only exists in Django's model registry and doesn't have any
+# class statement form. It's the through model between 'FlatPage' and 'Site'.
+@type_check_only
+class FlatPage_sites(models.Model):
```
sobolevn:
```suggestion
class _FlatPage_sites(models.Model):
```

django-stubs/contrib/gis/db/models/fields.pyi
```diff
@@ -23,18 +23,66 @@ _ST = TypeVar("_ST")
...
+_Geom_MultiLineStringField = TypeVar("_Geom_MultiLineStringField", bound=MultiLineString, default=MultiLineString)
+
+_ST_MultiPolygonField = TypeVar("_ST_MultiPolygonField", default=MultiPolygon | Combinable)
+_GT_MultiPolygonField = TypeVar("_GT_MultiPolygonField", default=MultiPolygon)
+_Form_MultiPolygonField = TypeVar(
+    "_Form_MultiPolygonField", bound=forms.MultiPolygonField, default=forms.MultiPolygonField
+)
+_Geom_MultiPolygonField = TypeVar("_Geom_MultiPolygonField", bound=MultiPolygon, default=MultiPolygon)
+
+_ST_GeometryCollectionField = TypeVar("_ST_GeometryCollectionField", default=GeometryCollection | Combinable)
```
sobolevn: I propose moving these variables closer to the class that use them. It would be easier to review.

django-stubs/contrib/gis/db/models/fields.pyi
```diff
@@ -23,18 +23,66 @@ _ST = TypeVar("_ST")
...
+
 class SRIDCacheEntry(NamedTuple):
     units: Any
     units_name: str
     spheroid: str
     geodetic: bool
 
 def get_srid_info(srid: int, connection: Any) -> SRIDCacheEntry: ...
+@type_check_only
+class SpatialClassField(Generic[_Form_ClassT, _GEOM_ClassT]):
```
sobolevn:
```suggestion
class _SpatialClassField(Generic[_Form_ClassT, _GEOM_ClassT]):
```

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -54,6 +54,54 @@ _ST = TypeVar("_ST", contravariant=True)
...
+
+_ST_BooleanField = TypeVar("_ST_BooleanField", default=bool | Combinable)
+_GT_BooleanField = TypeVar("_GT_BooleanField", default=bool)
+
+_ST_IPAddressField = TypeVar("_ST_IPAddressField", default=str | Combinable)
+_GT_IPAddressField = TypeVar("_GT_IPAddressField", default=str)
+_ST_GenericIPAddressField = TypeVar("_ST_GenericIPAddressField", default=str | int | Callable[..., Any] | Combinable)
+_GT_GenericIPAddressField = TypeVar("_GT_GenericIPAddressField", default=str)
+
+_ST_DateField = TypeVar("_ST_DateField", default=str | date | Combinable)
```
sobolevn: The same: let's move these definitions closer to their usages.

sobolevn (commented):
Thanks, this is not a full review.

django-stubs/contrib/admin/models.pyi
FallenDeity: Probably just annotating as `: models.AutoField` any field for that matter as its just for annotation, any instance is unnecessary, both for mypy and pyright

FallenDeity:
Resolved most of the comments so far

Additionally wanted to ask do I go ahead and also add types for the `contrib/db/gis/backend` models, wrt to [here](https://github.com/django/django/blob/c1a4fccf53437e480c42648f67277fb7e63ed4ab/django/contrib/gis/db/backends/postgis/models.py)?

Edit: Added types and tests

django-stubs/contrib/auth/models.pyi
```diff
@@ -20,22 +22,68 @@ class PermissionManager(models.Manager[Permission]):
...
 
 class Permission(models.Model):
-    content_type_id: int
     objects: ClassVar[PermissionManager]
 
+    id: models.AutoField
+    pk: models.AutoField
     name = models.CharField(max_length=255)
-    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    content_type = models.ForeignKey[ContentType | Combinable, ContentType](ContentType, on_delete=models.CASCADE)
```
sobolevn: Can we please test that `Permission` works the way we want? It has both `=` and `:` model fields. They should be identical, right?

FallenDeity:
```py
from django.contrib.auth.models import Permission

perm = Permission()
reveal_type(perm.id)
reveal_type(perm.pk)
reveal_type(perm.name)
reveal_type(perm.content_type)
reveal_type(perm.codename)
reveal_type(perm.content_type_id)
reveal_type(perm.user_set)
reveal_type(perm.group_set)
```
```bash
$ mypy --config-file mypy.ini test.py 

test.py:5: note: Revealed type is "builtins.int"
test.py:6: note: Revealed type is "builtins.int"
test.py:7: note: Revealed type is "builtins.str"
test.py:8: note: Revealed type is "django.contrib.contenttypes.models.ContentType"
test.py:9: note: Revealed type is "builtins.str"
test.py:10: note: Revealed type is "builtins.int"
test.py:11: note: Revealed type is "django.db.models.fields.related_descriptors.ManyRelatedManager[django.contrib.auth.models.User, django.contrib.auth.models._User_permissions]"
test.py:12: note: Revealed type is "django.db.models.fields.related_descriptors.ManyRelatedManager[django.contrib.auth.models.Group, django.contrib.auth.models._Group_permissions]"
```
pyright:
image

Is this not what the expected result should be? Tests are present here https://github.com/FallenDeity/django-stubs/blob/c36ba71ea7346ef6b5fd2bc16923f277ff1d75a4/tests/assert_type/contrib/auth/test_auth_models.py#L44-L51 and https://github.com/FallenDeity/django-stubs/blob/c36ba71ea7346ef6b5fd2bc16923f277ff1d75a4/tests/typecheck/models/test_contrib_models.yml#L38-L40 which it passes.

Also whether we assign it as `: format` or `= format` doesn't really matter both usually yield to same results or types during type checking, I just added the generic type parameters to make sure the model is typed accurately as `ForeignKey` here expects them, otherwise `content_type` would be partially unknown. 

I did however notice a lot places where models are being initialized with fields `=` instead of `:` which should be the ideal approach, so we should probably make it consistent across the stubs, atleast the `models.pyi`, but some of them had some other field info like `max_length=`, `on_delete=` I wasn't sure if they were relevant or due to preferences, so I didn't mess with them too much

example:
```python
content_type: models.ForeignKey[ContentType | Combinable, ContentType]  # equivalent
```
What are your thoughts?

sobolevn: What about pyright?

FallenDeity: The text  is from the mypy run and the image is from pyright/pylance default linter for python it works for both

sobolevn: Awesome, thank you! This adds me extra confidence :)

scripts/stubtest/allowlist_todo.txt
```diff
@@ -174,15 +174,48 @@ django.contrib.gis.db.backends.mysql.introspection
...
+django.contrib.gis.db.backends.oracle.models.OracleSpatialRefSys.cs_name
+django.contrib.gis.db.backends.oracle.models.OracleSpatialRefSys.srid
+django.contrib.gis.db.backends.oracle.models.OracleSpatialRefSys.auth_srid
+django.contrib.gis.db.backends.oracle.models.OracleSpatialRefSys.auth_name
+django.contrib.gis.db.backends.oracle.models.OracleSpatialRefSys.wktext
+django.contrib.gis.db.backends.oracle.models.OracleSpatialRefSys.cs_bounds
 django.contrib.gis.db.backends.postgis.adapter.PostGISAdapter.prepare
 django.contrib.gis.db.backends.postgis.features.DatabaseFeatures.empty_intersection_returns_none
 django.contrib.gis.db.backends.postgis.features.DatabaseFeatures.supports_geography
+django.contrib.gis.db.backends.postgis.models.PostGISGeometryColumns.f_table_catalog
```
sobolevn: Why do we have these new items in stubs ignore?

FallenDeity: These are also models used internally by `django/contrib/gis` located in the backends previously they were typed `Any` so I added field types with reference to the main django code

```py
class OracleGeometryColumns(models.Model):
    table_name: models.CharField
    column_name: models.CharField
    srid: models.IntegerField
    objects: ClassVar[Manager[Self]]
```
https://github.com/django/django/blob/c1a4fccf53437e480c42648f67277fb7e63ed4ab/django/contrib/gis/db/backends/postgis/models.py

But this is also related to point number 2 I mentioned in the pr, about once the models are typed do I remove them from `allowlist_todo.txt` directly? Since if you check the list all models and their fields are mentioned there.

So I tried removing them and ran stubtest but I ran into the following
```bash
$ ./scripts/stubtest.sh 
error: django.contrib.admin.models.LogEntry.action_flag variable differs from runtime type django.db.models.query_utils.DeferredAttribute
Stub: in file /home/triyan/projects/django-stubs/django-stubs/contrib/admin/models.pyi:49
django.db.models.fields.PositiveSmallIntegerField[Union[builtins.float, builtins.int, builtins.str, django.db.models.expressions.Combinable], builtins.int]
Runtime:
<django.db.models.query_utils.DeferredAttribute object at 0x7fb7265e7d40>

error: django.contrib.gis.db.backends.postgis.models.PostGISGeometryColumns.f_table_catalog variable differs from runtime type django.db.models.query_utils.DeferredAttribute
Stub: in file /home/triyan/projects/django-stubs/django-stubs/contrib/gis/db/backends/postgis/models.pyi:8
django.db.models.fields.CharField[Union[builtins.str, builtins.int, django.db.models.expressions.Combinable], builtins.str]
Runtime:
<django.db.models.query_utils.DeferredAttribute object at 0x7fb7083eef30>

Found 2 errors (checked 882 modules)
```

Reason this happens is during runtime `Model.<some_field>` on the class itself would give a `DeferredAttribute` object but an instance of that model would provide a field with the actual expected typing

In this particular case I removed the following from the allowlist, both of which now have types

```diff
- django.contrib.admin.models.LogEntry.action_flag
- django.contrib.gis.db.backends.postgis.models.PostGISGeometryColumns.f_table_catalog
```

Hence why currently they are added as fields in `allowlist_todo.txt` but I think we should move this `allowlist.txt` under a category for model fields

With this pr we should pretty much be able to remove all `Model` instances from the `allowlist_todo.txt`

FallenDeity: Any thoughts on this? should we have a category for models in the `allowlist.txt` @sobolevn ?

sobolevn: Yes, allowlist with a category is good enough, thanks!

FallenDeity: Added a category in the allowlist

FallenDeity:
@sobolevn I think we should be all done, should we also update the `django_mypy_plugin` to handle `null=True`?
> there is one case where the current model field typing system might need some explicit type hinting from the user in case of where fields become optional or with null=True, because with current system I don't think there is a way to infer types from the field parameters passed, Here is an example
> ```py
> text: models.CharField # get or return type is str, as expected
> # But when params like null=True is passed no way to autoinfer that and change _GT, _ST
> # i.e without explicit typehints or generic params it would still show `str`, if we were to do `reveal_type`
> # Typehinting for null=True would require manual typehints like the following:
> # Redundant verbose example
> text_nullable = models.CharField[Optional[Union[str, int, Combinable]], Optional[str]](max_length=100, null=True)
> # A more generic user might have to do this
> text_nullable = models.CharField[str | None, str | None](max_length=100, null=True)
> ```

sobolevn:
`null=True` should already work 🤔

FallenDeity:
> `null=True` should already work 🤔

Thats because of the extension too, here is how the extension handles nulls, first it checks if the expression is nullable then based on that it fetches `set_type` and `get_type` making them nullable
```py
set_type, get_type = get_field_descriptor_types(
        default_return_type.type,
        is_set_nullable=is_set_nullable or is_nullable,
        is_get_nullable=is_get_nullable or is_nullable,
    )
```
after that it fetches mapped types which is the types if generic params are provided and replace the default types gotten above with the mapped types
```py
# bail if either mapped_set_type or mapped_get_type have type Never
    if not (isinstance(mapped_set_type, UninhabitedType) or isinstance(mapped_get_type, UninhabitedType)):
        # always replace set_type and get_type with (non-Any) mapped types
        set_type = helpers.convert_any_to_type(mapped_set_type, set_type)
        get_type = get_proper_type(helpers.convert_any_to_type(mapped_get_type, get_type))
```
the reason it works currently is because mapped types are `Never` and just the default get_type and set_type are returned
```py
from django.db import models
from datetime import date as Date

class MyUser(models.Model):
    date = models.DateField(null=True)
    my_date = models.DateField[Date, Date](null=True)

u = MyUser()
reveal_type(u.date) # unknown for pyright/pylance, date | None for mypy
reveal_type(u.my_date) # type date for both
```
```
mypy .\test.py
null_expr=<mypy.nodes.NameExpr object at 0x00000115725D7CE0>, is_nullable=True
set_type=Union[builtins.str, datetime.date, django.db.models.expressions.Combinable, None], get_type=Union[datetime.date, None]
mapped_set_type=Never, mapped_get_type=Never
null_expr=<mypy.nodes.NameExpr object at 0x00000115725D7F60>, is_nullable=True
set_type=Union[builtins.str, datetime.date, django.db.models.expressions.Combinable, None], get_type=Union[datetime.date, None]
mapped_set_type=datetime.date, mapped_get_type=datetime.date
set_type=datetime.date, get_type=datetime.date
test.py:7: error: DateField is nullable but its generic get type parameter is not optional  [misc]
test.py:11: note: Revealed type is "Union[datetime.date, None]"
test.py:12: note: Revealed type is "datetime.date"
```
As you can see if types are mapped the default expression types are overwritten which is whats happening since all our fields are now inherently mapped with generic type params

So to fix this I think we can just make the  mapped types optional too if the field is nullable

FallenDeity:
@sobolevn any thoughts on this? should we have the plugin enforce null for us if `null=True` is passed and make the mapped types nullable? I don't see any drawbacks to this approach

sobolevn:
> should we have the plugin enforce null for us if null=True is passed and make the mapped types nullable?

We 100% should, but I am pretty sure that it already works. If it does not, please open a new issue.

FallenDeity:
> > should we have the plugin enforce null for us if null=True is passed and make the mapped types nullable?
> 
> We 100% should, but I am pretty sure that it already works. If it does not, please open a new issue.

It works with the current fields types because the generic parameters are unknown as soon as you add those generic type parameters it overwrites the nullable expression with the generic parameters provided instead, which what's happening here since each field now have specified generic params such as `_ST_IntegerField`

I'll open a new issue regarding this and reference it here

sobolevn:
Sorry, I missed this message https://github.com/typeddjango/django-stubs/pull/2590#issuecomment-2797897225 

Reading!

sobolevn:
Oh, now I see: `my_date = models.DateField[Date, Date](null=True)` that's the cornercase that does not quite work. I think that we should not change `Date` here, because it is explicit. But, we can raise an error in our plugin that it is expected to have `| None` of `null=True` is specified.

FallenDeity:
> Oh, now I see: `my_date = models.DateField[Date, Date](null=True)` that's the cornercase that does not quite work. I think that we should not change `Date` here, because it is explicit. But, we can raise an error in our plugin that it is expected to have `| None` of `null=True` is specified.

But currently our internal fields are all typed with default types meaning our stubs already defines `IntegerField` as such for example `IntegerField[int, int]` and in such a case if someone does `IntegerField(null=True)` we need to make it nullable (which currently dosent happen due to reasons mentioned above), so now as a solution a user can either
- add explicit typehints wherever null is used `IntegerField[int | None, int | None](null=True)` in their code (user side)
- or the plugin handles and makes it nullable

The only cornercase in soln 2 (preferred to reduce redundancy) is that if user defines explicit typehints our plugin may override that and make it nullable if null=True is defined (imo thats expected behaviour) since currently I dont think there is a way to tell if the generic type params are coming from django-stubs types or user codebase, due to this raising an error might not work unless we can determine whether these explicit type hints are from stubs or user

maybe we can use inspect.getFile() or `.__module__` on the generic type params not sure if its reliable or a good way though

FallenDeity:
One thing I noticed is that if the generic types are user defined the following mypyc attrs are filled or populated in the mypyc plugin, information such a `line`, `column` etc. Otherwise if they are defined on our side in the stubs `__init__.pyi` these mypyc attrs fields are `-1` this can be a way to identify whether generic params were explicitly defined by the user and if so we leave them alone else if it's defined in the stubs we make them nullable if `null=True` 

Not sure how foolproof or reliable this method is any thoughts @sobolevn ?

https://pastebin.com/7MMsJTMs Here is an example where the types were user defined, you see that line no. and column is populated, whereas in stubs it would just show -1 for both
- user mapped types
```py
from django.db import models


class TestModel(models.Model):
    my_field = models.CharField[str, str](null=True)


t = TestModel()
reveal_type(t.my_field)
```
- stub mapped types
```py
from django.db import models


class TestModel(models.Model):
    my_field = models.CharField(null=True)


t = TestModel()
reveal_type(t.my_field)
```

Difference between the 2 different types of type definition can be seen here https://www.diffchecker.com/0ujD9s30/

FallenDeity:
@sobolevn any ideas? What are your thoughts on the different options we can approach with in regard to this, I feel like using some meta properties from mypyc might be a bit too hacky of a soln? We could alternatively just make them nullable regardless of where the generic params are coming from and users can force a typing.cast if they are sure it's not nullable via virtue of their code, since django internals wise the model field typing returned is accurate if `null=True` .

sobolevn:
Couple of thoughts: 

- This should be an error -> `models.CharField[str, str](null=True)`
Just like this case:

```python
class A[T]:
    def __init__(self, x: T) -> None:
        self.x = x
        
A[int]('a')
```

https://mypy-play.net/?mypy=latest&python=3.12&gist=ccb14fd0cc0bf9711ba4d513c16adf5a

- Checking for this in a plugin looks doable. We can start with this in a separate PR.

- We should not modify type params in the plugin, ever. It is hard, pretty unstable, error-prone.

FallenDeity:
@sobolevn I think there is a misunderstanding here check this mypy example for context:
https://mypy-play.net/?mypy=latest&python=3.12&gist=420f15b0129632baeaeae2225ce8299b

What I am trying to say is as of now the mypy plugin uses the fields `_pyi_private_get_type`, `_pyi_private_set_type` to infer types for fields, and currently they are made nullable **by the plugin** if `null=True` check this https://github.com/typeddjango/django-stubs/blob/c37850b62a7217129e75f9985da37c2870b6b069/mypy_django_plugin/lib/helpers.py#L275

This works well and good since currently the mapped types for generics are always Unknown or Never, since we didn't provide the generic params before. In which case currently the types are remapped or changed to the mapped get/set types, if the user ever provides them in **their code** since they take priority. https://github.com/typeddjango/django-stubs/blob/c37850b62a7217129e75f9985da37c2870b6b069/mypy_django_plugin/transformers/fields.py#L166-L169

Now I completely agree with you that above cases as you mentioned should raise errors, but right now the dilemma is that we are not using the default get/set types (i.e ones inferred from `_pyi_private_get_type` for example) https://github.com/typeddjango/django-stubs/blob/c37850b62a7217129e75f9985da37c2870b6b069/mypy_django_plugin/transformers/fields.py#L154 Instead we already have default mapped types instead from the TypeVars `default=` parameter, so the issue rises in the following points:
- What happens if `null=True` and user has not supplied any types like the following, since in this case the types are already provided internally from `_ST_CharField` and `_GT_CharField`.  These types are defined as such
https://github.com/typeddjango/django-stubs/blob/50dfc931ef47a938f152cd44e1ec03f9c5ca3109/django-stubs/db/models/fields/__init__.pyi#L307-L308
```py
class A(models.Model):
    text_nullable = models.CharField(null=True)
    
a = A()
# So now this is just str instead of str | None
reveal_type(a.text_nullable)
```
The above is basically the main issue that we are encountering, since we need to make our own inbuilt mapped types nullable somehow via the plugin. Which imo is a valid use case for the plugin, but as a side effect if we make this decision this can give rise to the following specific edge case
- Say now hypothetically we make the mapped types nullable as well based on `null=` which we are **already doing** for default types inferred ( `_pyi_private_get/set_type`) via the plugin. If in a particular case user supplies `null=True` and maps their own types, we are in risk or danger of making **their** mapped types nullable as well. Since I am unable to find a reliable of telling whether the types are coming from the user or internally (i.e `_ST/GT_CharField`)
```py
class A(models.Model):
    # No reliable way to tell if the mapped types are from user code here, or from the stubs .pyi file
    text_nullable = models.CharField[str, str](null=True)  # wrong typehint we should raise an error
   
a = A()
reveal_type(a.text_nullable)
```
I hope this clears it up a bit this is kind of a chicken and egg situation 😅

sobolevn:
It feels like we should not use `default=` from `TypeVar`, because it does not really work in this case with `null=True` modifing the default.

FallenDeity:
> It feels like we should not use `default=` from `TypeVar`, because it does not really work in this case with `null=True` modifing the default.

This isn't really a `default=` issue either the mapped types would be present no matter which method we use to pass generic arguments to the inbuilt model fields, even if we don't use default and manually type out the fields. This is more of a plugin issue since the plugin doesn't expect mapped generic types from the stub files and uses an internal logic to extract types from `_pyi_private_get/set_type` and considers them default. 
For pyright/pylance to pick up the types we would still need to map the fields to their generic argument at some point which the plugin is not meshing well with for this particular edge case.

@intgr @flaeppe you folks have had some prior experience with this in #2214 any ideas on how to approach this?

FallenDeity:
One solution to this can be to use `typing.overload` and write overloads for Field overloading the `__new__` method like the following, this also eliminates any plugin logic for handling nulls

- [Pyright Playground](https://pyright-play.net/?strict=true&code=GYJw9gtgBA%2BjwFcAuCQFM5QJYQA5hCSgEMA7UsJYpLMUgZwFgAoFnfQqJAT1y1IDmJelxZjmMAMoAVKAF4uAOmm80ANWIgAFACIp0nQEoWMAOKyFSZao3a95o%2BJYBjADbF6IgGJY0rgCZaVqZopGggWM4A2voANLDmALqGAFwsUBlQAAJWYABu4a5gxP7pmf5owLAwYQDucFpu9PEAVPGkCK6uKUoAMlhI4cSuUdIgCGjJUAC0AHxQPn7%2BMTJQAD5QAHJ0aPFmshvbYYlpzJnnUIpX4uc5ivmFxaVn5ZXVdQ1Nre2d3X0DQxGXmG9Em8gWILQhhm80WARW0j2SVOF0yV0UN1eVTgHxgjVczSgbSgHS6PQARmAwK5wcCCVCYQtfPC4glpIl1kylgjOUddmzeTsTmVUegUCBSFB6AhcOEtIZFDi0PU8U1jC8MiKoBVsTB%2BAMGqDXMBviTfhSqTSFHTQdC5lsdijUVKZXKFXB9UgGurxPoYABhAAWmjh-nBVhUstsuj9QZDzP8OniOuInSQcnoSBAnP4SHV%2BwDwZAofD1ijmhj5kL8aWSe1lVTrnTmZAPtYzDcHhEceLCa0oYR1d7SyR0iHoeSToy6K14gAHuCe6GtKTXHIbVCWOgCsMYDxZVo5%2BruIui8vV3IxhN1du0Lv92gtNx1QAvU81gLyrdoHeuPeqLQX3VIA)

```py

import typing as t


_ST = t.TypeVar("_ST")
_GT = t.TypeVar("_GT")


class Field(t.Generic[_ST, _GT]):
    @t.overload
    def __new__(cls, *, null: t.Literal[True]) -> Field[_ST | None, _GT | None]:
        ...

    @t.overload
    def __new__(cls, *, null: t.Literal[False] = False) -> Field[_ST, _GT]:
        ...

    def __new__(cls, *, null: bool = False) -> Field[_ST, _GT] | Field[_ST | None, _GT | None]:
        return super().__new__(cls)
    
    def __init__(self, *, null: bool = False) -> None:
        super().__init__()


_ST_CharField = t.TypeVar("_ST_CharField", default=str | int)
_GT_CharField = t.TypeVar("_GT_CharField", default=str)


class CharField(Field[_ST_CharField, _GT_CharField]):
    ...
    

x = CharField(null=False)
reveal_type(x)
y = CharField(null=True)
reveal_type(y)
z = CharField()
reveal_type(z)
```
image

edit: This seems to work with pyright but not mypy https://github.com/python/mypy/issues/15220 😭 , also this one https://github.com/python/mypy/issues/17251, If we change the order of methods and redefine the overloads locally for each field individually mypy seems to work (this is due to mypy using mro to pick between init and new for return types 😭 )

Overall after a bunch of edits this works on both mypy and pyright, https://mypy-play.net/?mypy=latest&python=3.12&gist=1f9176a8c2bef850b4e9a5d9af2e8211 only downside is `__new__` overloads needs to be defined for all fields individually, and any existing `__init__`'s need to be switched to `__new__`'s since mypy prefers `__init__` over `__new__` for a return type in case of a tie which doesn't make a lot of sense
https://github.com/python/mypy/blob/c724a6a806655f94d0c705a7121e3d671eced96d/mypy/typeops.py#L148-L149

FallenDeity:
@sobolevn I made the migration for fields to overload `__new__` based on `null`, I think all the objectives of the pr as mentioned initially are completed. As of now with this pr all models and fields would be typed using just pyright/pylance and mypy out of the box with `django-stubs` installed. The plugin is not necessary for field types anymore. Please review whenever you are free 😅 

Also thanks for your incredible patience, much appreciated 😄

tests/typecheck/fields/test_base.yml
```diff
@@ -81,15 +81,19 @@
...
     installed_apps:
         - myapp
     files:
         -   path: myapp/__init__.py
         -   path: myapp/models.py
             content: |
+                from typing import Optional, Union
                 from django.db import models
+                from django.db.models.expressions import Combinable
+
```
UnknownPlatypus: Why are these new imports required ?

tests/typecheck/models/test_contrib_models.yml
```diff
@@ -1,3 +1,16 @@
+-   case: contrib_admin_model_fields
```
UnknownPlatypus, line 1: These tests are redundant with the `assert_type` matching one ?
Do we need them still ?

tests/typecheck/models/test_extra_methods.yml
```diff
@@ -53,12 +53,17 @@
...
         MyUser().get_previous_by_date()  # E: "MyUser" has no attribute "get_previous_by_date"  [attr-defined]
+    monkeypatch: true
     installed_apps:
         - myapp
     files:
         -   path: myapp/__init__.py
         -   path: myapp/models.py
             content: |
+                from datetime import date
+                from typing import Optional, Union
```
UnknownPlatypus: Same, not sure why these are added

UnknownPlatypus (commented):
Gave a look because this is really exciting, noticed a few minor things so I left a few minor comments

tests/typecheck/fields/test_base.yml
FallenDeity: Oh yeah good catch! These are not required anymore since now we are overloading `__new__` previously I had manually typehinted the `null=True` test cases. Removed them now.

tests/typecheck/models/test_extra_methods.yml
FallenDeity: Same as above, have been removed

tests/typecheck/models/test_contrib_models.yml
FallenDeity: ok so from what I can see in workflow these tests are evaluated only using mypy via pytest (the ones under tests/typecheck/models/test_contrib_models.yml). The ones under `assert_type/` are evaluated using pyright https://github.com/typeddjango/django-stubs/blob/3e395fa5af4bd64e1aa0a3086c029fb57b3baed5/.github/workflows/test.yml#L123
```json
    "include": [
        "tests/assert_type/"
    ],
```
Ideally we would want pyright and mypy to produce same results for all model fields, so I think having them as separate tests help in case behavior of one of the type checkers change is the future or any possible changes in the stubs too which might affect this.

UnknownPlatypus: > The ones under assert_type/ are evaluated using pyright 

I believe they are also evaluated with mypy.  
```yml
      - name: Run mypy on the test cases
        run: mypy --strict tests/assert_type
```

But I do agree it's not 100% clear to me which type of test should be preferred, I feel like `assert_type` ones are better since they are mostly typechecker agnostic and easier to write (plain python file).

It would be nice to add some guideline around that in the contributing guide.

FallenDeity: fair enough, I didn't really take them out since one benefit of having them on pytest side is few people might already be familiar running them with just `pytest` but as opposed to running a few ci commands locally. But this might be more suited for a change in another pr along with a contributor guidlines update

huynguyengl99:
Hi @FallenDeity, I suggest defining the `__new__` overload for the Field which define/override the `__init__` only, like Field, DecimalField, CharField, etc. (not IntegerField, SmallIntegerField, etc. because they do not override `__init__` function). For example, here will be the definition of Field's `__new__` (we keep `__init__` as it is):

```python
@overload
def __new__(
    cls,
    null: Literal[False] = False,
    # ... other params
) -> Self: ...

@overload
def __new__(
    cls,
    null: Literal[True] = True,
    # ... other params
) -> Field[_ST | None, _GT | None]: ...
```

If we do that, the IntegerField and other ones which do not override `__init__` do not need to override the `__new__` again and again, so that the code will be more compact.

The problem I see is that if a field inherits and defines extra info, they might not work because we override the return type as Field type. I have tried to use generic Type[_F] for cls and return as `_F[_ST | None, _GT | None]` but unfortunately it does not work. So in that case, users need to define their `__new__` again to return their class.

Also I tried to use TypedDict to define params, like:

```python
class BaseFieldParams(TypedDict, total=False):
    verbose_name: _StrOrPromise | None
    name: str | None
    primary_key: bool
    max_length: int | None

class NonNullableParams(BaseFieldParams):
    null: Literal[False]

class NullableParams(BaseFieldParams):
    null: Literal[True]
```

and use like:

```python
def __init__(self, **kwargs: Unpack[BaseFieldParams]) -> None: ...

@overload
def __new__(cls, **kwargs: Unpack[NonNullableParams]) -> Self: ...

@overload
def __new__(cls: type[_F], **kwargs: Unpack[NullableParams]) -> Field[_ST | None, _GT | None]: ...
```

But it does not work, too. So I think the only working thing now is just to define the `__new__` function for all Field classes which declare `__init__` function. How do you feel about this suggestion?

FallenDeity:
@huynguyengl99 While what you suggested is what I would've preferred myself too, to avoid repeating the `__new__` signatures but unfortunately mypy requires explicit repetition, i.e overloaded signatures are not inherited.
- https://github.com/python/mypy/issues/5146
- https://github.com/python/typing/discussions/1254

huynguyengl99:
Hi @FallenDeity, I'm not sure if I understand you correctly, but unless we redefine the `__init__`, we need to redefine the `__new__`. Otherwise, for normal inheritance, like IntegerField inheriting from Field, we don't need to redefine the `__init__` function, and it will still work fine.

For example, this case will work following the mypy example from the link, due to not reimplementing fun:
```python
from typing import overload
class Foo:
    @overload
    def fun(self, s: str) -> str:
        ...
    @overload
    def fun(self, i: int) -> int:
        ...
    def fun(self, x):
        ...
class Bar(Foo):
    ...
Bar().fun(0.1)

#  error: No overload variant of "fun" of "Foo" matches argument type "float"
```

So, in our case, the Field without reimplementing `__init__`, like IntegerField:
```python
class IntegerField(Field[_ST, _GT]):
    _pyi_private_set_type: float | int | str | Combinable
    _pyi_private_get_type: int
    _pyi_lookup_exact_type: str | int
```
It should work normally because we don't reimplement the `__init__`, so it will inherit all the overloads of the base class, aka the Field.

For fields like DecimalField, which redefine `__init__`, we need to redefine all the `__new__` overloads.

Is that correct, or am I misunderstanding your point?

FallenDeity:
@huynguyengl99 there are a few issues with what you are suggesting if I am interpreting it correctly, you are suggesting something similar to this
- https://mypy-play.net/?mypy=latest&python=3.12&gist=12e36ca057975411b86d24fb02cd5588
but the above doesn't work due to mypy behaving weirdly with `__new__` overloads, in fact funnily enough the same method works if you replace new with another entirely different classmethod as such https://mypy-play.net/?mypy=latest&python=3.12&gist=41c3c8ff5dbae74584a98c84de18fa14

Second limitation is if we do define all the `__new__` in the base `Field` class all return types of subclassed fields would now be `Field[_ST_WhateverField, _GT_WhateverField]`, for this particular case `Self` was introduced but `Self` doesn't support type annotations and its recommended to go with explicit types https://peps.python.org/pep-0673/#use-in-generic-classes

image

So overall I think having explicit annotations is probably the best move, given similar scenarios and cases mentioned in typing specs and related issues. Also in a related scenario `django-types` a fork of `django-stubs` also seems to be doing something very similar to support inbuilt model typehints https://github.com/sbdchd/django-types/blob/main/django-stubs/db/models/fields/__init__.pyi

lemme know if I mentioned something inaccurate anywhere 😅

huynguyengl99:
Hi @FallenDeity, you are right. It's my fault because I only tested using VSCode/Pyright instead of mypy. It's quite interesting that Pyright can detect variable `y` as a nullable field.

image

However, in my opinion, re-annotating the `__new__` method for all fields, which would significantly increase the size of the type file (`__init__.pyi`), is not a good approach. So should we find a better solution for this?

FallenDeity:
I doubt a 1-2 kB increase is gonna affect the typechecker (pylance/pyright) by any noticeable amount. And even so the pyi files are indexed at the start when opening the workspace so it's one time very very negligible cost. But if you do stumble upon a better solution or alternative we could try out that too, unfortunately I don't think any other method is possible which keeps both mypy and pyright compatibility at the same time. There aren't too many advantages to an alternative other than reducing the repeating amount of boilerplate code (if that's possible)

huynguyengl99:
Hi @FallenDeity, I have another thought: what if we just implement the base case for fields like `IntegerField`/`CharField` without trying to automatically handle the `null` parameter through overloads?

For the normal case, we'd help users by providing proper types when they create a field without any parameters - those get/set types would be non-nullable by default. When users want nullable fields with `null=True`, they would need to explicitly specify the type parameters, like:

```python
my_field: IntegerField[int | None, int | None] = IntegerField(null=True)
```

The current mypy plugin approach works because of the custom plugin support, but this isn't standardized in any PEP yet. I think we shouldn't invest too much effort in complex workarounds until there's an official PEP for typing that enables dynamic type creation based on parameters passed to `__init__`.

This approach would:
1. Keep the type stubs simpler and more maintainable
2. Provide good defaults for the common case (non-nullable fields)
3. Still allow precise typing when needed through explicit type parameters
4. Avoid the complex `__new__` overload patterns that behave differently across type checkers

What do you think about this more conservative approach?

FallenDeity:
Both approaches have their own pros and cons, with a more conservative approach I would basically be repeating a lot of my typehints over multiple models (generally quite a few cols in table do tend to allow nullables in their db) so I would need to re type hint them. Which just seems counter intuitive to having the stubs in first place since technically the stub provides an incorrect typehint here

The more radical approach (currently implemented) provided accurate type hints rn but duplicates some parts of code over in the stubs code which in my opinion is an acceptable outcome and abstracts away from the user the need to tinker and add their own typehints. While I agree overloads are relatively complex they do however still exist and their usecase was intended as such displayed in this pr.

This is mostly a trade off between having user custom typehint their fields (a lot more larger scale impact) or handle it within the stubs itself with some amount of sophistication and complexity for the `Field`'s

At the end of the day this an opinionated decision and depends on the maintainers on which approach they prefer I think @sobolevn might be able to make a better call on this

PR #2592, opened 2025-04-01
Update Django 4.2 LTS version to 4.2.20

intgr:
Solves GitHub's Dependabot security alerts. The alerts are created because we still use an older Django version in requirements.txt when testing with older Python versions. There is no security impact, because Django is only used in CI and not exposed as a web service.

## Related issues

- Closes https://github.com/typeddjango/django-stubs/security/dependabot/26
- Previous one: #2369


adamchainz:
Thanks!

PR #2593, opened 2025-04-02
Update `django.contrib.gis.geoip2` stubs

ngnpope:
# I have made things!

From https://github.com/django/django/commit/f5f55b41afc43ef0cd661adcb48fc9ec1b2d4ac6:

- `django.contrib.gis.geoip2.GeoIP2.info` was removed.

From https://github.com/django/django/commit/60c7cb90f88c0819276da6407134ac311503fcad:

- `django.contrib.gis.geoip2.resources` was removed.

From https://github.com/django/django/commit/0ee2b8c326d47387bacb713a3ab369fa9a7a22ee:

- `django.contrib.gis.geoip2.base` was removed.
- `django.contrib.gis.geoip2` became a module instead of a package.

From https://github.com/django/django/commit/a93375e8ab26e7cb22c6ee7e4bf23bf2d6ed170d:

- `django.contrib.gis.geoip2.GEOIP_SETTINGS` was removed.

From https://github.com/django/django/commit/cc56c22a24ef717cc3111e92ca146136fa518d55:

- Support for querying using `IPv4Address` and `IPv6Address` was added.

From https://github.com/django/django/commit/f50184a84b885f4474d5030af968195219a360b9:

- `django.contrib.gis.geoip2.GeoIP2.city()` returns extra fields.
- `django.contrib.gis.geoip2.GeoIP2.country()` returns extra fields.

From https://github.com/django/django/commit/3fad712a91a8a8f6f6f904aff3d895e3b06b24c7:

- `django.contrib.gis.geoip2.SUPPORTED_DATABASE_TYPES` was added.
- `django.contrib.gis.geoip2.GeoIP2.is_city` was restored.
- `django.contrib.gis.geoip2.GeoIP2.is_country` was restored.

## Related issues

N/A


django-stubs/contrib/gis/geoip2.pyi
```diff
@@ -0,0 +1,75 @@
...
+    continent_code: str
+    continent_name: str
+    country_code: str
+    country_name: str
+    is_in_european_union: bool
+
+class GeoIP2Exception(Exception): ...
+
+class GeoIP2:
+    MODE_AUTO: Final[int]
```
sobolevn: Do we care about actual values here? Or just `int` is good enough?

django-stubs/contrib/gis/geoip2.pyi
```diff
@@ -0,0 +1,75 @@
...
+    def country_name(self, query: _QueryType) -> str: ...
+    def country(self, query: _QueryType) -> _CountryResponse: ...
+    @deprecated("coords() is deprecated and will be removed in Django 6.0. Use lon_lat() instead.")
+    def coords(self, query: _QueryType, ordering: Sequence[str] = ...) -> _CoordType: ...
+    def lon_lat(self, query: _QueryType) -> _CoordType: ...
+    def lat_lon(self, query: _QueryType) -> _CoordType: ...
+    def geos(self, query: _QueryType) -> Point: ...
+    @classmethod
+    @deprecated("open() is deprecated and will be removed in Django 6.0. Use GeoIP2() instead.")
+    def open(cls, full_path: Path | str | None, cache: int) -> Self: ...
```
sobolevn: Maybe `_typeshed.StrPath` instead of `Path | str`?

sobolevn (commented):
Thanks!

django-stubs/contrib/gis/geoip2.pyi
ngnpope: I tried `Final[Literal[0]]` but the linter wanted that changed to `Final` only 🤔

Unless we should do `MODE_AUTO: Final = 0`, etc?

django-stubs/contrib/gis/geoip2.pyi
ngnpope: TIL. Will check it out.

django-stubs/contrib/gis/geoip2.pyi
sobolevn: We can use `=`, sure :)

django-stubs/contrib/gis/geoip2.pyi
```diff
@@ -0,0 +1,75 @@
...
+
+class GeoIP2Exception(Exception): ...
+
+class GeoIP2:
+    MODE_AUTO: Final = 0
+    MODE_MMAP_EXT: Final = 1
+    MODE_MMAP: Final = 2
+    MODE_FILE: Final = 4
+    MODE_MEMORY: Final = 8
+    cache_options: Final[frozenset[Literal[0, 1, 2, 4, 8]]]
```
sobolevn:
```suggestion
    cache_options: Final[frozenset[int]]
```

django-stubs/contrib/gis/geoip2.pyi
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(
+        self,
+        path: StrPath | None = ...,
+        cache: Literal[0, 1, 2, 4, 8] = ...,
+        country: str | None = ...,
+        city: str | None = ...,
```
sobolevn:
```suggestion
        path: StrPath | None = None,
        cache: int = 0,
        country: str | None = None,
        city: str | None = None,
```

sobolevn (commented):
Thank you!

django-stubs/contrib/gis/geoip2.pyi
ngnpope: Oh, I did this because it'll only ever be a set of the above constants.

django-stubs/contrib/gis/geoip2.pyi
ngnpope: And is it not preferable to type `cache` here to only accept things that won't crash?

sobolevn: It is hard to infer `Literal[0, 1, 2, ...]`. For example, the simplest operations like `[0, 1][0]` or `cache_type = {"country": 0}; cache_type.get(name)` would infer `int` :(

ngnpope: Ah, yeah. Of course.

sobolevn (approved):
Thanks!

PR #2594, opened 2025-04-02
Allow tuples for some `TypedModelMeta` fields

ngnpope:
# I have made things!

Django supports both `list` and `tuple` in many places, including here.

## Related issues

N/A


sobolevn (approved):
Thanks!

PR #2595, opened 2025-04-05
Fix django.urls.base.reverse after 5.2

simwr872:
# I have made things!
Fixes types for `django.urls.base.reverse`.
See https://docs.djangoproject.com/en/5.2/releases/5.2/#urls (PR: https://github.com/django/django/pull/18848).

Probably good to get https://github.com/typeddjango/django-stubs/pull/2596 merged first to fix pipeline.


sobolevn (commented):
We are adding 5.2 support, please, fix the CI

django-stubs/urls/base.pyi
```diff
@@ -11,6 +12,9 @@ def reverse(
     args: Sequence[Any] | None = ...,
     kwargs: dict[str, Any] | None = ...,
     current_app: str | None = ...,
+    *,
+    query: QueryDict | dict[Any, Any] | Sequence[tuple[Any, Any]] | None = ...,
```
sobolevn: `query` supports: `QueryDict` or anything that `urlencode` supports. Which is defined as https://github.com/python/typeshed/blob/87f599dc8312ac67b941b5f2b47274534a1a2d3a/stdlib/urllib/parse.pyi#L136-L138

Let's copy this type alias with a link to the source and reuse it.

sobolevn (changes requested):
Thanks! One important thing.

sobolevn (approved):
Thanks!

PR #2596, opened 2025-04-06
Update django to 5.2

sobolevn:
Changes:
- dropped 4.2 support
- dropped 3.9 support, since 4.2 is now not supported and all django@5.x require `>=3.10`
- added new `todo` allowlist


django-stubs/core/management/commands/loaddata.pyi
```diff
@@ -1,15 +1,14 @@
...
-from typing import Callable, Literal
+from typing import Literal, TypeAlias
 
 from django.apps.config import AppConfig
 from django.core.management.base import BaseCommand
 from django.core.serializers.base import DeserializedObject
 from django.db.backends.base.base import BaseDatabaseWrapper
 from django.db.models.base import Model
 from django.utils.functional import cached_property
-from typing_extensions import TypeAlias
```
jnoring, line 12: Does `django/core/management/commands/shell.pyi` also need a definition for the new `get_auto_imports` method? I believe it just returns `list[str]`.

sobolevn: I would prefer not to add anything complex here. Because this PR is hard to review.
We can always add any new functions / methods / etc in the next PRs. After this one is merged.

flaeppe (approved):
👍

taraskuzyk:
When can we expect this to be released to PyPI?

sobolevn:
Yes, soon enough!

PR #2598, opened 2025-04-08
Update `django.core.validators` for Django 5.2

JaeHyuckSa:
# I have made things!
- added
    - `django.core.validators.DomainNameValidator.tld_no_fqdn_re`
    - `django.core.validators.EmailValidator.domain_re`
    - `django.core.validators.EmailValidator.hostname_re`
    - `django.core.validators.EmailValidator.tld_no_fqdn_re`

## Related issues
[django PR 18950](https://github.com/django/django/pull/18950)

### DIFF
--- a/django-stubs/core/validators.pyi
+++ b/django-stubs/core/validators.pyi
@@ -33,6 +33,7 @@ class DomainNameValidator(RegexValidator):
     ul: str
     hostname_re: str
     domain_re: str
+    tld_no_fqdn_re: str
     tld_re: str
     ascii_only_hostname_re: str
     ascii_only_domain_re: str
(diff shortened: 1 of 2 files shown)


django-stubs/core/validators.pyi
```diff
@@ -65,6 +66,9 @@ def validate_integer(value: float | str | None) -> None: ...
 class EmailValidator(_Deconstructible):
     message: _StrOrPromise
     code: str
+    domain_re: str
```
sobolevn, line 69: Ideally this should be converted to `ClassVar`, because there are are direct class-level usages https://github.com/django/django/blob/71a19a0e475165dbc14c1fe02f552013ee670e4c/django/core/validators.py#L212-L215

But, this can be done in a separate PR.

sobolevn (approved):
Thanks!

django-stubs/core/validators.pyi
JaeHyuckSa: Thank you, you had given me the same feedback before… but I applied it without thinking, just by looking at the codebase. I'll be more careful next time

sobolevn: Thanks a lot! I highly value your contributions.

PR #2604, opened 2025-04-11
Allow `MutableMapping` for `Form.Initial`

UnknownPlatypus:
# I have made things!

This allows modifying the `initial` dict of a `Form` / `ModelForm`. 
[This is a regular dict at runtime](https://github.com/django/django/blob/main/django/forms/forms.py#L94) so I think it make more sense to use `MutableMapping` here.

I was stuck with this error for a use case where I infer the initial value of a `ModelForm` field based on a field on the instance.
Something like that:

```python

    if self.instance.pk and self.instance.billing_address:
        self.initial["use_specific_billing_address"] = (
            self.instance.billing_address != self.instance.address
        )
    else:
         self.initial["use_specific_billing_address"] = False
```

## Related issues
None

### DIFF
--- a/django-stubs/forms/forms.pyi
+++ b/django-stubs/forms/forms.pyi
@@ -1,4 +1,4 @@
-from collections.abc import Iterable, Iterator, Mapping
+from collections.abc import Iterable, Iterator, MutableMapping
 from typing import Any, ClassVar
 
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks!

PR #2606, opened 2025-04-12
Update `django.db.models.sql` for Django 5.2

JaeHyuckSa:
# I have made things!

- added
    - django.db.models.sql.constants.ROW_COUNT
    - django.db.models.sql.query.Query.add_q
    - django.db.models.sql.query.Query.selected
    - django.db.models.sql.query.Query.try_transform

- removed
    - django.db.models.sql.where.SubqueryConstraint


sobolevn (approved):
Thanks!

PR #2608, opened 2025-04-13
Update `django.forms.widgets` for Django 5.2

federicobond:
# I have made things!

From https://docs.djangoproject.com/en/5.2/releases/5.2/#forms:

* Added ColorInput, SearchInput and TelInput widget classes
* Added Script class

### DIFF
--- a/django-stubs/forms/widgets.pyi
+++ b/django-stubs/forms/widgets.pyi
@@ -14,6 +14,21 @@ _OptAttrs: TypeAlias = dict[str, Any]
 
 class MediaOrderConflictWarning(RuntimeWarning): ...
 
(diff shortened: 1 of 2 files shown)


federicobond:
What should we do with the `error: django.forms.widgets.MediaAsset.__html__ is not present in stub`?

Should I add it to `allowlist_todo.txt` like the other `__html__` from `django.forms.Media.__html__`?

sobolevn (commented):
we can add `__html__`, maybe?

federicobond:
@sobolevn done!

sobolevn (approved):
Thanks!

PR #2627, opened 2025-04-25
Allow tuples for some `Options` fields

ngnpope:
# I have made things!

Django supports both `list` and `tuple` in many places, including here.

Follow up to 29e2ca847e04faa67645f1a5f7af6307a7e45aaf. I missed these 🤦🏻 

## Related issues

- #2594


sobolevn (approved):
Thanks!

PR #2635, opened 2025-04-27
Fix some `<type> is not present at runtime` errors

UnknownPlatypus:
# I have made things!
This fixes a few errors of incorrect symbols declared in the stubs that are not available at runtime (could cause `ImportError`s) :

Exemple stubtest error:
```
error: django.contrib.gis.forms.BoundWidget is not present at runtime
Stub: in file /home/thibaut/workspace/django-stubs/django-stubs/contrib/gis/forms/__init__.pyi:73
<TypeInfo django.forms.boundfield.BoundWidget>
Runtime:
MISSING

error: django.forms.BoundWidget is not present at runtime
Stub: in file /home/thibaut/workspace/django-stubs/django-stubs/forms/__init__.pyi:73
<TypeInfo django.forms.boundfield.BoundWidget>
Runtime:
MISSING

Found 2 errors (checked 884 modules)
```
and associated runtime error:
```python
>>> from django.forms import BoundWidget
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'BoundWidget' from 'django.forms' (/home/thibaut/workspace/django-stubs/.venv/lib/python3.11/site-packages/django/forms/__init__.py)

```

PR #2636, opened 2025-04-27
Fix missing at runtime errors (second batch)

UnknownPlatypus:
# I have made things!

Followup to #2635, review commit per commit is advised


sobolevn (approved):
Thanks!

adamchainz:
Thanks !

PR #2637, opened 2025-04-27
Move method defined on `BaseManager` to `Manager` to follow runtime behavior

UnknownPlatypus:
# I have made things!

This one is a bit more tricky, review per commit is advised. 

It does quite a lot of changes in test by switching from `BaseManager` to `Manager` which I think is good overall because [`Manager`](https://docs.djangoproject.com/en/5.2/topics/db/managers/) is documented, but `BaseManager` is not.

The issue was that `BaseManager` does not have every  `QuerySet` methods unless you instanciate it using `.from_queryset`.
So it was a bit inaccurate to declare them there. However, it does make sense on `Manager` because it's defined like that:

```python
class Manager(BaseManager.from_queryset(QuerySet)):
    pass
```


sobolevn (approved):
This looks like it solves a lot of todo entries.

cc @flaeppe to double check this pr.

flaeppe (approved):
Seems fine to me. I suppose anyone that's using, the undocumented, `BaseManager.from_queryset(...)` way would lose type hint support? Although it's not documented and I don't know of a use case where it needs to be used over `Manager.from_queryset(...)`

UnknownPlatypus:
> I suppose anyone that's using, the undocumented, BaseManager.from_queryset(...) way would lose type hint support?

I think yes, I'll add a test to confirm and maybe a small mention somewhere in the troubleshooting section from the readme ?

sobolevn:
No need for a readme notice :)
I don't think that this is popular enough.

PR #2640, opened 2025-04-28
Allow permission methods to take any model instance as obj

federicobond:
This was already the case for the auth backends.

See: https://github.com/typeddjango/django-stubs/blob/6917033b208de863e26b0b50e86ee75d507d67bb/django-stubs/contrib/auth/backends.pyi#L11-L22


sobolevn (commented):
Do you have a link to the docs / usecase for that? I can't really understand this change without a context, sorry :(

federicobond:
@sobolevn the original types were not correct. The `obj` parameter is designed for checking object-level permissions so it could always take any model instance.

This change just matches the types that flow to the authentication backend, where `user_obj` and `obj` have distinct types.

See docs here: https://docs.djangoproject.com/en/5.1/ref/contrib/auth/#django.contrib.auth.models.User.has_perm

sobolevn (approved):
Oh, now I see it, thank you!

johnthagen:
@sobolevn A new release that includes this PR would be greatly appreciated as 5.2.0 generated a lot of false positives due to the issue this PR fixes. Thanks.

Issue #2643, opened 2025-04-28
`Unable to resolve type of choices property` for `.value` on Choices


asottile-sentry:
# Bug report

here's a small example from sentry:

```python
from django.db import models

class ComparisonDeltaChoices(models.IntegerChoices):
    FIVE_MINUTES = (300, "5 minutes ago")
    FIFTEEN_MINUTES = (900, "15 minutes ago")
    ONE_HOUR = (3600, "1 hour ago")
    ONE_DAY = (86400, "1 day ago")
    ONE_WEEK = (604800, "1 week ago")
    ONE_MONTH = (2592000, "1 month ago")


print([choice.value for choice in ComparisonDeltaChoices])
```

## What's wrong

```console
$ mypy t3.py
t3.py:12: error: Unable to resolve type of choices property  [misc]
Found 1 error in 1 file (checked 1 source file)
```

## How is that should be

should resolve as `int`

## System information

- OS:
- `python` version: 3.13.1
- `django` version: 5.1.7
- `mypy` version: 1.15.0
- `django-stubs` version: 5.2.0 (well, technically patched 5.2.0)
- `django-stubs-ext` version: 5.2.0

asottile-sentry:
likely regressed in #2582 

cc @ngnpope

ngnpope:
Thanks for the report. I'm already working on a fix.

asottile-sentry:
#2644 fixes this particular case fwiw

ngnpope:
Thanks. I've got something that fixes a bunch of other related issues as well as this one. I'll have it up for review tomorrow.

PR #2646, opened 2025-04-29
Fixes to choices support in plugin

ngnpope:
# I have made things!

So I was far too naive in #2582 🤦🏻

I've run this on a very large repository and this now fixes all of the `Unable to resolve type of ... property` errors I encountered. I also made a change to narrow the label type from `_StrOrPromise` to `str` where all members labels and `__empty__` are not using lazy strings. Although the changes to use `_StrOrPromise` are correct according to the implementation, it can be annoying to fix up in large projects that are only using plain strings for labels. This change to narrow the type will vastly reduce the number of errors raised and make upgrading less painful.

## Related issues

- #2582
- #2643
- #2644


sobolevn (commented):
Wow! That's a lot of changes :)
I would like to see some second opinion from @adamchainz or @flaeppe 

Thanks!

delfick (commented):
Super thanks for making these changes! I love the way you've split it up, this is great :)

sobolevn (approved):

PR #2648, opened 2025-05-01
add null keyword argument to ManyToManyField init

### LINKED ISSUE #2505: ManyToManyField constructor type doesn't expect 'null' keyword argument, even though it can be passed
# Bug report

## What's wrong
When defining a model like this:

```python
from django.db import models

class Person(models.Model):
    first_name = models.TextField(max_length=30)
    last_name = models.TextField(max_length=30)
    friends = models.ManyToManyField("self", null=True, blank=True)
```
which is technically valid in Django, running `mypy` on this file with `django-stubs` installed produces the following error:
```console
$ mypy myapp/models.py --cache-dir /dev/null
myapp/models.py:7: error: Unexpected keyword argument "null" for "ManyToManyField"  [call-arg]
/home/user/workspace/django-example/venv/lib/python3.12/site-packages/django-stubs/db/models/fields/related.pyi:256: note: "ManyToManyField" defined here
```

It is passed as a kwarg from `ManyToManyField`'s [constructor](https://github.com/django/django/blob/main/django/db/models/fields/related.py#L1432-L1437) to `RelatedField`'s constructor, from where it is (still) [passed as a kwarg](https://github.com/django/django/blob/main/django/db/models/fields/related.py#L108) to `Field`'s constructor, where it is defined as an argument to the function ([link to relevant stub](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/fields/__init__.pyi#L164))

## How is that should be
The type stubs in `ManyToManyField` should probably reflect all arguments that can be passed as kwargs all the way down to `Field`'s constructor.

## System information

- OS: Linux x86_64
- `python` version: 3.12.4
- `django` version: 5.1.6
- `mypy` version: 1.15.0

lev-blit:
# I have made things!

Resolves #2505, added the `null` keyword argument to `ManyToManyField`'s `__init__`

### DIFF
--- a/tests/typecheck/fields/test_related.yml
+++ b/tests/typecheck/fields/test_related.yml
@@ -1597,3 +1597,12 @@
                     featured = models.BooleanField()
                     authors = models.ManyToManyField(Author)
                     objects = OtherManager()
+
+-   case: test_many_to_many_field_null
+    main: |
+        from django.db import models
+        class Person(models.Model):
+            friends = models.ManyToManyField("self", null=True, blank=True)
+
+        p: Person
+        reveal_type(p.friends)  # N: Revealed type is "django.db.models.fields.related_descriptors.ManyRelatedManager[main.Person, main.Person_friends]"
--- a/django-stubs/db/models/fields/related.pyi
+++ b/django-stubs/db/models/fields/related.pyi
@@ -272,6 +272,7 @@ class ManyToManyField(RelatedField[Any, Any], Generic[_To, _Through]):
         max_length: int | None = ...,
         unique: bool = ...,
         blank: bool = ...,
+        null: bool = ...,
         db_index: bool = ...,
         default: Any = ...,
         editable: bool = ...,


tests/typecheck/fields/test_related.yml
```diff
@@ -1597,3 +1597,9 @@
                     featured = models.BooleanField()
                     authors = models.ManyToManyField(Author)
                     objects = OtherManager()
+
+-   case: test_many_to_many_field_null
+    main: |
+        from django.db import models
+        class Person(models.Model):
+            friends = models.ManyToManyField("self", null=True, blank=True)
```
sobolevn, line 1605:
```suggestion
        class Person(models.Model):
            friends = models.ManyToManyField("self", null=True, blank=True)
            
        p: Person
        reveal_type(p.frields)    
```

sobolevn (commented):
> [null](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null) has no effect since there is no way to require a relationship at the database level.

Quoting docs from https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.ManyToManyField.swappable

Please, adjust a test case to be sure that we don't add `| None` to the type. You can take a look at other `null=True` tests for the reference.

sobolevn (approved):

PR #2650, opened 2025-05-02
Narrow `RegexValidator.regex` instance level type to `Pattern[str]`

UnknownPlatypus:
# I have made things!

This is cast to a lazy Pattern in the [`__init__`](https://github.com/django/django/blob/9d93e35c207a001de1aa9ca9165bdec824da9021/django/core/validators.py#L45) so I think it makes sens to narrow here. 

This fixes an issue I had at work when trying to access the regex pattern

```python
validate_not_gmail = RegexValidator(
    regex=r".+@(?!gmail).+\..+",
    message="The email address cannot contain gmail",
)
...
class MyForm(forms.Form):
    email = EmailField(
        label="Mail",
        required=True,
        widget=Input(
            input_type="email",
            attrs={
                "pattern": validate_not_gmail.regex.pattern, # error: Item "str" of "str | Pattern[str]" has no attribute "pattern"
                "title": validate_not_gmail.message,
            },
        ),
        validators=[validate_email_required, validate_not_gmail],
    )
```

This makes the type of `regex` on the class slightly off but I think it's better than having it slightly off for instances which people are usually working with.

### DIFF
--- a/django-stubs/core/validators.pyi
+++ b/django-stubs/core/validators.pyi
@@ -1,7 +1,7 @@
 from collections.abc import Callable, Collection, Sequence, Sized
 from decimal import Decimal
 from re import Pattern, RegexFlag
-from typing import Any, Generic, TypeAlias, TypeVar, overload, type_check_only
+from typing import Any, TypeAlias
 
 from django.core.files.base import File
 from django.utils.deconstruct import _Deconstructible
@@ -13,19 +13,8 @@ _Regex: TypeAlias = str | Pattern[str]
 
 _ValidatorCallable: TypeAlias = Callable[[Any], None]  # noqa: PYI047
 
-_ClassT = TypeVar("_ClassT")
-_InstanceT = TypeVar("_InstanceT")
-
-@type_check_only
-class _ClassOrInstanceAttribute(Generic[_ClassT, _InstanceT]):
-    @overload
-    def __get__(self, instance: None, owner: type[object]) -> _ClassT: ...
-    @overload
-    def __get__(self, instance: object, owner: type[object]) -> _InstanceT: ...
-    def __set__(self, instance: object, value: _InstanceT) -> None: ...
-
 class RegexValidator(_Deconstructible):
-    regex: _ClassOrInstanceAttribute[_Regex, Pattern[str]]
+    regex: Pattern[str]
     message: _StrOrPromise
     code: str
     inverse_match: bool
--- a/django-stubs/core/validators.pyi
+++ b/django-stubs/core/validators.pyi
@@ -14,7 +14,7 @@ _Regex: TypeAlias = str | Pattern[str]
 _ValidatorCallable: TypeAlias = Callable[[Any], None]  # noqa: PYI047
 
 class RegexValidator(_Deconstructible):
-    regex: _Regex  # Pattern[str] on instance, but may be str on class definition
+    regex: Pattern[str]
     message: _StrOrPromise
     code: str
     inverse_match: bool
--- a/django-stubs/core/validators.pyi
+++ b/django-stubs/core/validators.pyi
@@ -17,15 +17,15 @@ _ClassT = TypeVar("_ClassT")
 _InstanceT = TypeVar("_InstanceT")
 
 @type_check_only
-class ClassOrInstanceAttribute(Generic[_ClassT, _InstanceT]):
+class _ClassOrInstanceAttribute(Generic[_ClassT, _InstanceT]):
     @overload
-    def __get__(self, obj: None, owner: type[object]) -> _ClassT: ...
+    def __get__(self, instance: None, owner: type[object]) -> _ClassT: ...
     @overload
-    def __get__(self, obj: object, owner: type[object]) -> _InstanceT: ...
-    def __set__(self, obj: object, value: _InstanceT) -> None: ...
+    def __get__(self, instance: object, owner: type[object]) -> _InstanceT: ...
+    def __set__(self, instance: object, value: _InstanceT) -> None: ...
 
 class RegexValidator(_Deconstructible):
-    regex: ClassOrInstanceAttribute[_Regex, Pattern[str]]
+    regex: _ClassOrInstanceAttribute[_Regex, Pattern[str]]
     message: _StrOrPromise
     code: str
     inverse_match: bool
--- a/django-stubs/core/validators.pyi
+++ b/django-stubs/core/validators.pyi
@@ -1,7 +1,7 @@
 from collections.abc import Callable, Collection, Sequence, Sized
 from decimal import Decimal
 from re import Pattern, RegexFlag
-from typing import Any, TypeAlias
+from typing import Any, Generic, TypeAlias, TypeVar, overload, type_check_only
 
 from django.core.files.base import File
 from django.utils.deconstruct import _Deconstructible
@@ -13,8 +13,19 @@ _Regex: TypeAlias = str | Pattern[str]
 
 _ValidatorCallable: TypeAlias = Callable[[Any], None]  # noqa: PYI047
 
+_ClassT = TypeVar("_ClassT")
+_InstanceT = TypeVar("_InstanceT")
+
+@type_check_only
+class ClassOrInstanceAttribute(Generic[_ClassT, _InstanceT]):
+    @overload
+    def __get__(self, obj: None, owner: type[object]) -> _ClassT: ...
+    @overload
+    def __get__(self, obj: object, owner: type[object]) -> _InstanceT: ...
+    def __set__(self, obj: object, value: _InstanceT) -> None: ...
+
 class RegexValidator(_Deconstructible):
-    regex: Pattern[str]
+    regex: ClassOrInstanceAttribute[_Regex, Pattern[str]]
     message: _StrOrPromise
     code: str
     inverse_match: bool
--- a/scripts/stubtest/allowlist.txt
+++ b/scripts/stubtest/allowlist.txt
@@ -550,3 +550,9 @@ django.contrib.gis.forms.BaseModelFormSet.save_m2m
 
 # Dynamically generated in https://github.com/django/django/blob/0ee06c04e0256094270db3ffe8b5dafa6a8457a3/django/core/mail/backends/locmem.py#L24
 django.core.mail.outbox
+
+# We use a trick using a descriptor class to represent an attribute with different type at the class and instance level.
+# Here to narrow `str | Pattern[str]` at the class level to `Pattern[str]` at the instance level.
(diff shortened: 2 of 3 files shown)


django-stubs/core/validators.pyi
```diff
@@ -14,7 +14,7 @@ _Regex: TypeAlias = str | Pattern[str]
 _ValidatorCallable: TypeAlias = Callable[[Any], None]  # noqa: PYI047
 
 class RegexValidator(_Deconstructible):
-    regex: _Regex  # Pattern[str] on instance, but may be str on class definition
+    regex: Pattern[str]
```
sobolevn, line 17: You can use a descriptor with `__get__` here so you can make sure that instance and class-level annotations are different.

django-stubs/core/validators.pyi
```diff
@@ -13,8 +13,19 @@ _Regex: TypeAlias = str | Pattern[str]
 
 _ValidatorCallable: TypeAlias = Callable[[Any], None]  # noqa: PYI047
 
+_ClassT = TypeVar("_ClassT")
+_InstanceT = TypeVar("_InstanceT")
+
+@type_check_only
+class ClassOrInstanceAttribute(Generic[_ClassT, _InstanceT]):
```
sobolevn:
```suggestion
class _ClassOrInstanceAttribute(Generic[_ClassT, _InstanceT]):
```

django-stubs/core/validators.pyi
```diff
@@ -13,8 +13,19 @@ _Regex: TypeAlias = str | Pattern[str]
...
+    @overload
+    def __get__(self, obj: None, owner: type[object]) -> _ClassT: ...
+    @overload
+    def __get__(self, obj: object, owner: type[object]) -> _InstanceT: ...
+    def __set__(self, obj: object, value: _InstanceT) -> None: ...
+
 class RegexValidator(_Deconstructible):
-    regex: Pattern[str]
+    regex: ClassOrInstanceAttribute[_Regex, Pattern[str]]
```
sobolevn:
```suggestion
    regex: _ClassOrInstanceAttribute[_Regex, Pattern[str]]
```

django-stubs/core/validators.pyi
```diff
@@ -13,8 +13,19 @@ _Regex: TypeAlias = str | Pattern[str]
...
+_ClassT = TypeVar("_ClassT")
+_InstanceT = TypeVar("_InstanceT")
+
+@type_check_only
+class ClassOrInstanceAttribute(Generic[_ClassT, _InstanceT]):
+    @overload
+    def __get__(self, obj: None, owner: type[object]) -> _ClassT: ...
+    @overload
+    def __get__(self, obj: object, owner: type[object]) -> _InstanceT: ...
+    def __set__(self, obj: object, value: _InstanceT) -> None: ...
```
sobolevn: Do we need an overload for class / instance?

tests/assert_type/core/test_validators.py
```diff
@@ -0,0 +1,11 @@
+from re import Pattern
+from typing import assert_type
+
+from django.contrib.auth.validators import UnicodeUsernameValidator
+from django.core.validators import RegexValidator
+
+assert_type(RegexValidator.regex, str | Pattern[str])
+assert_type(RegexValidator().regex, Pattern[str])
```
sobolevn: Please, test `RegexValidator.regex = ...` and `RegexValidator().regex = ...`

UnknownPlatypus: I wanted to but was not sure how to do that for the assert_type tests, especially to test invalid assignments 

For exemple I want to ensure this:

`RegexValidator().regex = 3 # error`

sobolevn: `# type: ignore[assignment]` should work

django-stubs/core/validators.pyi
UnknownPlatypus: From what I've read in this thread https://discuss.python.org/t/support-different-type-for-class-variable-and-instance-variable/54198, descriptor only have a class level getter, setter and deleter only make sense at the instance level so this is probably the best we can get that satisfies [mypy](https://gist.github.com/mypy-play/5c08c14b4fc7fd44271d3966f5f53f3c) and [pyright](https://pyright-play.net/?pyrightVersion=1.1.353&pythonVersion=3.12&pythonPlatform=Linux&strict=true&locale=en&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDiApikSEgMYA0UAKokQIIA2SAhgM430JEBqbEDTAA3MszBsAJjQCqKJGBQAoZaEhQQRTNlz4ACmxgwyK5QH0ASkTREAHgC46DFuw5QAvFHmKUAbQ4YIUNjUwCggF0I5QBhZk4OWk9nXgEQAAoAIjiE2kyASmUASRRAthQKIiSvHn5BLJKyiqqC1Qp4jncczoB5EEaYcsrGY3IAIwBXE3TiUnIKP27EmgGhqoj8h2UoHagAAVFxSSlt3akiYChzc1sYa-SOImZgYTGAKycAOSUiYQB3OZOeC8Pxgd5ECgwDZQAC0AD4oEtaE4AHRo047A5iEASaQYqDnS7XW73R7PV4fKBgt4QmD-QGwBig8GQ6HwqCrZrIqBolH4wlXcyPO7mB5PF5U95Oam0mgiNjMCZEJycyq0fKwhHfUio9HKADEUAAIkgtJCoBMOGxbMp2gkoNZbHYBKwpEZcFtdpobPYnEs%2BqqmKMkJMTH4rD67DQQiYQP5AiAoqotGIFeZgUR0o77C6kG6YLgUVonYUU0Q0xms5Hc-ncOl8kXI4Vs86FXn3RkG8X7MkAIyqQ0lAAWZCQg2ato67h8FDA51kjzjbAgdVdHarTprHc2PL1ZYrDHSM7nRAXpmXq-bBZAjZLyn3zHTh%2BP88XKAvW%2Bv9dv9kKL9Pb4fm2tadj%2BdjJAATEAA) for the most likely usages.

I don't expect many user to try setting regex at the class level, looks like a convoluted way of defining a new class.

tests/assert_type/core/test_validators.py
```diff
@@ -0,0 +1,22 @@
...
+assert_type(UnicodeUsernameValidator().regex, Pattern[str])
+UnicodeUsernameValidator().regex = re.compile("")
+
+# expect "Pattern[str]"
+RegexValidator().regex = ""  # type: ignore[assignment] # pyright: ignore[reportAttributeAccessIssue]
+UnicodeUsernameValidator().regex = ""  # type: ignore[assignment] # pyright: ignore[reportAttributeAccessIssue]
+
+# expect "_ClassOrInstanceAttribute[Union[str, Pattern[str]], Pattern[str]]"
+RegexValidator.regex = "anything fails here"  # type: ignore[assignment] # pyright: ignore[reportAttributeAccessIssue]
+UnicodeUsernameValidator.regex = "anything fails here"  # type: ignore[assignment] # pyright: ignore[reportAttributeAccessIssue]
```
sobolevn: Hm, one more idea: will subtyping work correctly after this change?

We need to test two cases:

```python
class RegexSubtype(RegexValidator):
     regex = re.compile('abc')

class StrSubtype(RegexValidator):
      regex = 'abc'
```

UnknownPlatypus: **Nice catch, It causes the same errors, I suppose it's a blocker.** 

```python
tests/assert_type/core/test_validators.py:27: error: Incompatible types in assignment (expression has type "Pattern[str]", base class "RegexValidator" defined the type as "_ClassOrInstanceAttribute[Union[str, Pattern[str]], Pattern[str]]")  [assignment]
tests/assert_type/core/test_validators.py:31: error: Incompatible types in assignment (expression has type "str", base class "RegexValidator" defined the type as "_ClassOrInstanceAttribute[Union[str, Pattern[str]], Pattern[str]]")  [assignment]
```

This will probably be a blocker for #2615 too otherwise field subclass with custom widgets will raise errors (and this is quite a common pattern for custom fields).

----

I feel like maybe the initial version of this PR is the best we can get. 
We would have accurate type on the instance and a slightly too specific type on the class causing an issue for this case:

```python
class StrSubtype(RegexValidator):
      regex = 'abc' # error: Incompatible types in assignment (expression has type "str", base class "RegexValidator" defined the type as "Pattern[str]")  [assignment]
```
But it can be solved by using `re.compile("abc")`. I think it's mosty good because it's more explicit about the fact it's a regex.

----

The same applies for #2615 were we could do the same but it would require every field subclasses to pass instances and not classes (django already handle this correctly [when instanciating](https://github.com/django/django/blob/9d93e35c207a001de1aa9ca9165bdec824da9021/django/forms/fields.py#L145-L148))

```diff
class MyField(CharField):
-    widget = MyWidget
+    widget = MyWidget()

class MyEurosField(CharField):
    widget = MyWidget(unit="€") # Already good, it's an instance

```
This might cause a bit more churn but is not too bad to fix. 

What do you think @sobolevn ? Maybe you have other ideas on how to make the descriptor approach work ?

sobolevn: I propose to revert any descriptor-based changes, looks like it is too complex :(

scripts/stubtest/allowlist.txt
```diff
@@ -551,8 +551,9 @@ django.contrib.gis.forms.BaseModelFormSet.save_m2m
 # Dynamically generated in https://github.com/django/django/blob/0ee06c04e0256094270db3ffe8b5dafa6a8457a3/django/core/mail/backends/locmem.py#L24
 django.core.mail.outbox
 
-# We use a trick using a descriptor class to represent an attribute with different type at the class and instance level.
-# Here to narrow `str | Pattern[str]` at the class level to `Pattern[str]` at the instance level.
+# The type system does not allow to represent an attribute with different type at the class level (`str | Pattern[str]`)
+# and instance (`Pattern[str]`) level. But in order to have more precise type at the instance level, we restrict the types
```
sobolevn, line 555: > But in order to have more precise type at the instance level, we restrict the types

Sorry, we can't do that either. We can allow something that does not work to happen (false negative), but this will block valid useges (false positive). That's something we try to avoid, unless we are 100% sure that it is worth it.

UnknownPlatypus: Then I guess we cannot do more currently. At least I did not find a way. I tried another approach using `__new__` but could not get it to work properly for both case either.

I'll close the PR, we might be able to revisit once the type system allows to represent this.

Thanks for the review nonetheless!

UnknownPlatypus: Maybe #2615 will find a way

PR #2653, opened 2025-05-03
Fix missing at runtime 3

UnknownPlatypus:
More missing at runtime errors. Trims a few entries


mypy_django_plugin/main.py
```diff
@@ -123,10 +123,10 @@ def get_additional_deps(self, file: MypyFile) -> list[tuple[int, str, int]]:
                 # forward relations
                 self.django_context.get_model_related_fields(model_class),
                 # reverse relations - `related_objects` is private API (according to docstring)
-                model_class._meta.related_objects,  # type: ignore[attr-defined]
+                model_class._meta.related_objects,
             ):
                 try:
-                    related_model_cls = self.django_context.get_field_related_model_cls(field)
+                    related_model_cls = self.django_context.get_field_related_model_cls(field)  # type: ignore[arg-type]
```
sobolevn, line 129: question: what is the error here?

sobolevn (commented):
Thanks!

mypy_django_plugin/main.py
UnknownPlatypus: yeah this is because of the `itertools.chain` above, it does not narrow to `"RelatedField[Any, Any]" | ForeignObjectRel` but to a common base class of the 2.

Usually I solve that by passing the generic type to chain but that would mean I have to bring django imports, I was not sure it was ok, hence the type ignore comment. (But I totally forgot to post my comment asking for this, here it is I guess)

sobolevn: Ok, sounds fine!

sobolevn (approved):

PR #2654, opened 2025-05-04
Fix `django.http` errors

UnknownPlatypus:
Most notably this fixes the type for `QueryDict.pop` and `QueryDict.popitem`.
They both return a list of every item for the key, not a single item.

```python
>>> q = QueryDict("a=1&a=2&a=3", mutable=True)
>>> q.pop("a")
['1', '2', '3']
```

See https://docs.djangoproject.com/en/5.2/ref/request-response/#django.http.QueryDict.pop


sobolevn:
Please, resolve conflicts

sobolevn (approved):

PR #2655, opened 2025-05-04
Enable a few more ruff rules

UnknownPlatypus:
# I have made things! 

This does a few `ruff` related upgrades:
- ensure `assert_type` is imported from `typing_extension` (otherwise ci fails when testing with python 3.10).
- remove `PYI050` ignore -- ruff already knows it cannot run this rule since we declare `target-version 310`.
- document `PYI041`and `PYI021` ignores.
- remove `PYI024` ignore and fix existing violations.

I would also like to remove the `PYI043` ignore, I think it's valuable because `TypeAlias` and `TypeVar` are easily confused right now. See https://docs.astral.sh/ruff/rules/t-suffixed-type-alias/.
Let me know what you think, I can send a followup PR or document why we keep the ignore.

### DIFF
--- a/django-stubs/db/backends/base/introspection.pyi
+++ b/django-stubs/db/backends/base/introspection.pyi
@@ -1,17 +1,24 @@
-from collections import namedtuple
 from collections.abc import Iterable
-from typing import Any
+from typing import Any, NamedTuple
 
 from django.db.backends.base.base import BaseDatabaseWrapper
 from django.db.backends.utils import CursorWrapper
 from django.db.models.base import Model
 
-TableInfo = namedtuple("TableInfo", ["name", "type"])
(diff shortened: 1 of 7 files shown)


sobolevn (approved):
Thanks!

UnknownPlatypus:
> I would also like to remove the PYI043 ignore, I think it's valuable because TypeAlias and TypeVar are easily confused right now. See https://docs.astral.sh/ruff/rules/t-suffixed-type-alias/.
Let me know what you think, I can send a followup PR or document why we keep the ignore.


@sobolevn About this, should I go ahead and enable [PYI043](https://docs.astral.sh/ruff/rules/t-suffixed-type-alias/) ?

PR #2659, opened 2025-05-06
Add a test for missing generics in stubs

UnknownPlatypus:
# I have made things!

Followup to this discussion https://github.com/typeddjango/django-stubs/pull/2384#discussion_r2072873073

This introduce a test ensuring every generic we define in stubs are generic at runtime too.

This also adds the missing ones (`SetPasswordForm`, discussed above, being one of them)

<details>
<summary><a href="https://github.com/typeddjango/django-stubs/actions/runs/14863843217/job/41735483452?pr=2659">See initial failing run</a> </summary>

```
FAILED tests/test_generic_consistency.py::test_find_classes_inheriting_from_generic - AssertionError: classproperty is not patched in `ext/django_stubs_ext/patch.py`
  ConnectionProxy is not patched in `ext/django_stubs_ext/patch.py`
  Options is not patched in `ext/django_stubs_ext/patch.py`
  BaseIterable is not patched in `ext/django_stubs_ext/patch.py`
  ReverseOneToOneDescriptor is not patched in `ext/django_stubs_ext/patch.py`
  ForwardManyToOneDescriptor is not patched in `ext/django_stubs_ext/patch.py`
  SetUnusablePasswordMixin is not patched in `ext/django_stubs_ext/patch.py`
  SetPasswordForm(SetPasswordMixin, Form, BaseForm, RenderableFormMixin, RenderableMixin) is not patched in `ext/django_stubs_ext/patch.py`
  AdminPasswordChangeForm(SetUnusablePasswordMixin, SetPasswordMixin, Form, BaseForm, RenderableFormMixin, RenderableMixin) is not patched in `ext/django_stubs_ext/patch.py`
  SetPasswordMixin is not patched in `ext/django_stubs_ext/patch.py`
  ModelFormOptions is not patched in `ext/django_stubs_ext/patch.py`
```

</details>

PR #2661, opened 2025-05-07
Run mypy on every tests

UnknownPlatypus:
# I have made things!

Followup to https://github.com/typeddjango/django-stubs/pull/2659#discussion_r2075904793.

This is probably valuable to run mypy on every tests.


sobolevn (approved):
Why not :)

Issue #2665, opened 2025-05-08
`TextChoices` subclasses declared inside a class cause "Unable to resolve type of choices property" error


youtux:
# Bug report

## What's wrong

Having an enum that extends `TextChoices` declared inside another class (e.g. a model), results in the following error when running `mypy`:
```
file.py: error: Unable to resolve type of choices property  [misc]
```

Here's an example of such definition:
```python

class MyModel:
    class MyEnum(models.TextChoices):
        foo = "foo", "Foo"
        bar = "bar", "Bar"

a = MyEnum.foo
```

Running mypy against that:
```
file.py:9: error: Unable to resolve type of choices property  [misc]
```

## How is that should be

No error should happen.

## System information

- OS:
- `python` version: 3.12.9
- `django` version: 5.1.8
- `mypy` version: 1.13.0
- `django-stubs` version: 5.2.0
- `django-stubs-ext` version: 5.2.0

sobolevn:
cc @ngnpope

ngnpope:
Thanks for the heads-up. Will look into it.

ngnpope:
@youtux Could you try installing from the `master` branch? I think this should be fixed by #2646.

youtux:
the `master`  branch @ e566001 seems to be failing for me (both on mypy 1.13.0 and 1.15.0):
```
$ mypy file.py --show-traceback
.../lib/python3.12/site-packages/django-stubs/http/request.pyi:56: error: INTERNAL ERROR -- Please try using mypy master on GitHub:
https://mypy.readthedocs.io/en/stable/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 1.15.0
Traceback (most recent call last):
  File "mypy/semanal.py", line 7240, in accept
  File "mypy/nodes.py", line 1345, in accept
  File "mypy/semanal.py", line 3206, in visit_assignment_stmt
  File "mypy/semanal.py", line 3733, in process_type_annotation
  File "mypy/semanal.py", line 6894, in defer
AssertionError: Must not defer during final iteration
.../lib/python3.12/site-packages/django-stubs/http/request.pyi:56: : note: use --pdb to drop into pdb

```

youtux:
I managed to run `dmypy` instead with the `master` branch, but unfortunately I get a new error:
```
$ dmypy run
 dmypy run
Restarting: plugins changed
Daemon stopped
Daemon started
[...]

PR #2669, opened 2025-05-09
Restore support for mypy v1.13

ngnpope:
# I have made things!

Currently mypy v1.13 is the minimum supported version, but the fix made to the choices plugin code in #2646 made use of `TypeInfo.enum_members` which was introduced in v1.14. This commit adds a backport of the version introduced in v1.14 which can be removed when `django-stubs` bumps its minimum version.

## Related issues

- #2665
- #2646


sobolevn (approved):

UnknownPlatypus:
Should we test with the minimal supported mypy version in ci ? At least running the `mypy-self-check` jobs with mypy1.13 should help ? It would have caught this:

```
mypy_django_plugin/transformers/choices.py:36: error: "TypeInfo" has no attribute "enum_members"  [attr-defined]
```

sobolevn:
Yes, let's add this CI job :)

PR #2670, opened 2025-05-09
Fix a crash on mypy master branch

ngnpope:
# I have made things!

In https://github.com/python/mypy/pull/18818 the `msg` argument to `analyze_member_access()` was removed.
This will crash when mypy v1.16 gets released, so only pass `msg` if it is in the function signature.

## Related issues

N/A

### DIFF
--- a/mypy_django_plugin/transformers/functional.py
+++ b/mypy_django_plugin/transformers/functional.py
@@ -1,4 +1,3 @@
-import inspect
 from typing import Any
 
 from mypy.checkmember import analyze_member_access
@@ -7,8 +6,11 @@ from mypy.nodes import CallExpr, MemberExpr
 from mypy.plugin import AttributeContext
 from mypy.types import AnyType, Instance, TypeOfAny
 from mypy.types import Type as MypyType
+from mypy.version import __version__ as mypy_version
 
 from mypy_django_plugin.lib import helpers
+
+mypy_version_info = tuple(map(int, mypy_version.partition("+")[0].split(".")))
 
 
 def resolve_str_promise_attribute(ctx: AttributeContext) -> MypyType:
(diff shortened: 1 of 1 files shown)


ngnpope:
@sobolevn Perhaps we should also check mypy master in CI?

mypy_django_plugin/transformers/functional.py
```diff
@@ -20,6 +23,12 @@ def resolve_str_promise_attribute(ctx: AttributeContext) -> MypyType:
     str_info = helpers.lookup_fully_qualified_typeinfo(helpers.get_typechecker_api(ctx), "builtins.str")
     assert str_info is not None
     str_type = Instance(str_info, [])
+
+    # TODO: [mypy 1.16+] Remove this workaround for passing `msg` to `analyze_member_access()`.
+    extra: dict[str, Any] = {}
+    if "msg" in inspect.signature(analyze_member_access).parameters:  # mypy < 1.16
```
ngnpope: Bummer. Looks like `inspect.signature()` won't work because of mypyc...

```
ValueError: no signature found for builtin <built-in function analyze_member_access>
```

mypy_django_plugin/transformers/functional.py
```diff
@@ -1,12 +1,17 @@
...
 from mypy.errorcodes import ATTR_DEFINED
 from mypy.nodes import CallExpr, MemberExpr
 from mypy.plugin import AttributeContext
 from mypy.types import AnyType, Instance, TypeOfAny
 from mypy.types import Type as MypyType
+from mypy.version import __version__ as mypy_version
 
 from mypy_django_plugin.lib import helpers
 
+mypy_version_info = tuple(map(int, mypy_version.partition("+")[0].split(".")))
```
ngnpope, line 13: Switched to a version-based check. This is somewhat ugly, but I didn't want to add `packaging` as a dependency.

sobolevn (approved):
Yes, we need more mypy versions to be tested in CI, including `master` one.

stianjensen:
mypy 1.16 released now, will django-stubs cut a corresponding release with this fix soon?

sobolevn:
yes!

PR #2671, opened 2025-05-09
Run different version of mypy in ci

UnknownPlatypus:
# I have made things!

Run ci with supported mypy version ie  `1.13`, `1.14`, `1.15`, `1.16` and `master` branch on github.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.10', '3.11', '3.12', '3.13']
+        mypy-version: ['1.13', '1.14', '1.15', 'latest']
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
@@ -37,6 +38,14 @@ jobs:
 
       - name: Install dependencies
         run: uv sync --no-dev --group tests
+
+      - name: Override mypy version
+        run: |
+          if [ "${{ matrix.mypy-version }}" == "latest" ]; then
+            uv pip install git+https://github.com/python/mypy.git@master
+          else
+            uv pip install mypy==${{ matrix.mypy-version }}
+          fi
 
       - name: Run mypy on plugin code
         run: uv run mypy --strict mypy_django_plugin
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.10', '3.11', '3.12', '3.13']
-        mypy-version: ['1.13', '1.14', '1.15', 'latest']
+        mypy-version: ['1.13', '1.14', '1.15', '1.16', 'latest']
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -31,7 +32,12 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -U pip "setuptools<79.0.0" wheel
-          SETUPTOOLS_ENABLE_FEATURES=legacy-editable pip install -r ./requirements.txt
+          SETUPTOOLS_ENABLE_FEATURES=legacy-editable pip install -r ./requirements-base.txt
```
sobolevn: you can keep old `requirements.txt` and add `pip install -U "mypy==${{ matrix.mypy-version }}"`

UnknownPlatypus: I wanted to avoid installing 2 versions of mypy to gain some ci time.

Anyways, this should be much easier to handle once we revive the project.toml PR, might do that later

UnknownPlatypus: Sadly specifying multiple editable dependencies does not seem to be possible with a `pyproject.toml` file yet. From PEP 735:

> Special attention was devoted to attempting to standardize the expression of editable installations, as expressed by pip install -e and [PEP 660](https://peps.python.org/pep-0660/). However, although the creation of editable installs is standardized for build backends, the behavior of editables is not standardized for installers. Inclusion of editables in this PEP requires that any supporting tool allows for the installation of editables.
Therefore, although Poetry and PDM provide syntaxes for some of these features, they are considered insufficiently standardized at present for inclusion in Dependency Groups.

So I don't think we can completely remove the `requirements.txt` file without switching to something like [uv](https://docs.astral.sh/uv/concepts/projects/dependencies/#editable-dependencies).

I'll be happy to do that if you guys agree, it should also simplify the local setup to a simple `uv sync` command, making contributions easier. 
wdyt @sobolevn @adamchainz ?

sobolevn: sure, let's move to uv. current setup is broken for local development :(

UnknownPlatypus:
@sobolevn This is ready for review, was much easier now that we have uv.
I had to use `type: ignore[unused-ignore]` for type ignores that were required on old versions but not required in newer ones.
We should aim at removing them when we drop older mypy/python version support

sobolevn (commented):
I am not quite sure about that :(
It brings several new quirks for maintenance, not sure if it really helps though...

UnknownPlatypus:
> I am not quite sure about that :(
It brings several new quirks for maintenance, not sure if it really helps though...

Yeah I'm also not completely sure, but it would have caught https://github.com/typeddjango/django-stubs/issues/2665 earlier and would have probably also help with https://github.com/typeddjango/django-stubs/issues/2724#issuecomment-3062558245 were a mypy version bump introduced a regression in our plugin.

I feel like the kind of bugs related to mypy versions are the complicated ones, often related to the plugin.

Maybe others core maintainer of django-stubs have an opinion on the subject ?

ngnpope:
I fear this is about to become even more non-viable. In mypy 1.17 or 1.18 various options related to the formatting of types in errors are going to be obsolete. This means that we'll need to fix our tests - I have a branch ready with some of those changes which I'll open a PR for when the next release comes along... But, yeah, testing across versions will be nigh on impossible.

UnknownPlatypus:
That's sad but yeah, I suppose it would add too much of a burden

PR #2673, opened 2025-05-11
Update _AsSqlType to allow tuple[str, tuple]

jacobtylerwalls:
### Description
The Django ORM uses both tuples and lists for the params portion of the `as_sql()` return. This can cause problems ([ticket](https://code.djangoproject.com/ticket/35972)/[forum](https://forum.djangoproject.com/t/advanced-postgresql-functions-like-json-to-recordset-json-jsonb-to-recordset-jsonb/35542/5)/[PR](https://github.com/django/django/pull/19057)), but until the type is eventually narrowed to just `tuple[str, tuple]`  as described on the Django ticket, the stub should allow `tuple[str, tuple]` in addition to `tuple[str, list]` to reflect the current state of play.


sobolevn (approved):

sobolevn:
Thanks!

PR #2680, opened 2025-05-13
add mypy plugin support for Django's storage framework

terencehonles:
# I have made things!

This change correctly resolves ``django.core.files.storage.default_storage`` and ``django.contrib.staticfiles.storage.staticfiles_storage`` which are configured in the project's Django settings.

When ``django.core.files.storage.storages``, which has a dict like interface, is accessed with a literal value, then the type of the appropriate storage is returned.


django-stubs/core/files/storage/handler.pyi
```diff
@@ -1,15 +1,20 @@
...
 
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.functional import cached_property
 
 from .base import Storage
 
+@type_check_only
+class _StorageConfig(TypedDict):
+    BACKEND: str
+    OPTIONS: dict[str, Any] | None
```
sobolevn: `NotRequired`?

mypy_django_plugin/transformers/storage.py
```diff
@@ -0,0 +1,47 @@
...
+    if type_info is None:
+        if not ctx.api.api.final_iteration:
+            ctx.api.api.defer()
+            return PlaceholderType(fullname=fullname, args=[], line=ctx.context.line)
+        else:
+            ctx.api.fail(f'Cannot resolve storage backend "{fullname}"', ctx.context)
+            return ctx.type
+
+    assert type_info
+    return Instance(type_info, [])
```
sobolevn: What if this is a generic type? Would it be better to use `fill_typevars`?

mypy_django_plugin/main.py
```diff
@@ -98,8 +99,21 @@ def get_additional_deps(self, file: MypyFile) -> list[tuple[int, str, int]]:
         if file.fullname == "django.conf" and self.django_context.django_settings_module:
             return [self._new_dependency(self.django_context.django_settings_module, PRI_MED)]
 
+        # for settings.STORAGES["staticfiles"]
+        if file.fullname == "django.contrib.staticfiles.storage":
+            return [
+                self._new_dependency(self.django_context.settings.STORAGES["staticfiles"]["BACKEND"].rsplit(".", 1)[0])
```
sobolevn: Can these keys be missing?

tests/typecheck/core/test_storage.yml
```diff
@@ -0,0 +1,39 @@
...
+    reveal_type(storages["default"])  # N: Revealed type is "myapp.storage.MyDefaultStorage"
+    reveal_type(storages["custom"])  # N: Revealed type is "myapp.storage.MyStorage"
+    reveal_type(storages["staticfiles"])  # N: Revealed type is "django.contrib.staticfiles.storage.StaticFilesStorage"
+
+  custom_settings: |
+    from django.conf.global_settings import STORAGES as DEFAULT_STORAGES
+
+    STORAGES = {
+        **DEFAULT_STORAGES,
+        "default": {"BACKEND": "myapp.storage.MyDefaultStorage"},
```
sobolevn, line 23: What will happen if `MyDefaultStorage` is something invalid? Like instance of value `1` or `'abc'`

sobolevn (commented):
Thank you!

django-stubs/core/files/storage/handler.pyi
terencehonles: Thanks, wasn't aware of this (haven't used `TypedDict`s too much)

mypy_django_plugin/main.py
terencehonles: They are both set by default, and the `staticfiles` app has [a check that would raise `ImproperlyConfigured`](https://github.com/django/django/blob/6e36f7f78415e454dcde1d9ee33f2d2c7fdfdfad/django/contrib/staticfiles/checks.py#L5-L8), but it doesn't look like `default` has one. For Django 5.0 (in the depreciation period for  `DEFAULT_FILE_STORAGE` and `STATICFILES_STORAGE` settings), it puts those in the appropriate `STORAGES` entries, so those will be there and correct.

Having that said, I'll just make these defensive, and if type-checking an invalid project it will just no-op (since in this location it can't error).

mypy_django_plugin/transformers/storage.py
terencehonles: Good point

tests/typecheck/core/test_storage.yml
terencehonles: It will probably break, and I figured Django checked this, but I have made the plugin more defensive.

mypy_django_plugin/transformers/storage.py
```diff
@@ -0,0 +1,72 @@
...
+from mypy_django_plugin.django.context import DjangoContext
+from mypy_django_plugin.lib import helpers
+
+
+def get_storage_backend(alias: str, django_context: DjangoContext) -> str | None:
+    "Defensively look for a settings.STORAGES by its alias."
+
+    try:
+        fullname = django_context.settings.STORAGES[alias]["BACKEND"]
+        assert isinstance(fullname, str) and "." in fullname
```
sobolevn: Please, convert an `assert` to `if`, since we catch `AssertionError`

sobolevn (commented):
Please, add a test case for `pyright` in `assert_type` tests. We can't make a regression for non-mypy type-checkers :)

mypy_django_plugin/transformers/storage.py
terencehonles: I am also catching assertion error. I figured this was better than `if check: raise ValueError`, but do you not agree?

sobolevn:
```python
if not isinstance(fullname, str) or "." not in fullname:
    return None
```

?

terencehonles: I forgot that this code was no longer doing anything in the exception handler (since that was omitted in your quote). It was previously reporting a failure and that's why this code looked like this. I can adjust.

sobolevn: I mostly ask, because `assert` can be removed from code with `-OO`, it is fine for sanity-checks and type-narrowing, but I don't think that we can base logics on it :)

terencehonles:
> Please, add a test case for `pyright` in `assert_type` tests. We can't make a regression for non-mypy type-checkers :)

I'm not really sure how those are supposed to work or what I am supposed to put there, but hopefully that's sufficient :shrug:

tests/assert_type/contrib/staticfiles/test_storage.py
```diff
@@ -0,0 +1,8 @@
+from django.contrib.staticfiles.storage import ConfiguredStorage, StaticFilesStorage, staticfiles_storage
+from typing_extensions import assert_type
+
+# The plugin can figure out what these are (but pyright can't):
+assert_type(staticfiles_storage, StaticFilesStorage)  # pyright: ignore[reportAssertTypeFailure]
+
+# what pyright thinks these are:
+assert_type(staticfiles_storage, ConfiguredStorage)  # mypy: ignore[assert-type]
```
sobolevn, line 8: so, basically, right now `pyright` will infer `ConfiguredStorage` instead of `Any` and existing code can break, due to the fact that `ConfiguredStorage` is a `LazyObject` which does not really have right attributes. 

How can we fix that?

terencehonles: It wasn't `Any` before, but it can be if that's better... I just looked and mypy can't have its own type ignores, so I'm not sure how to ever make this work since mypy will always know more about the object than pyright does (even if it reports `Any` for pyright the assertion will fail, right?).

terencehonles: Any thoughts @sobolevn ?

PR #2684, opened 2025-05-14
fix: django.contrib.staticfiles for 5.2

### LINKED ISSUE #2675: Fix `django.contrib.staticfiles` for 5.2
https://github.com/typeddjango/django-stubs/blob/676509aaa09fc972105d794cc27fbc9783ec0e90/scripts/stubtest/allowlist_todo_django52.txt#L41-L45

We need to fix typing of these entries.

skv0zsneg:
# I have made things!

Hi. I'm work with stubs for the first time. Hope I made correct changes, at least locally `./scripts/stubtest.sh` and `pytests` is not complain on finders from `django.contrib.staticfiles` :) 

## Related issues

- Closes #2675


sobolevn (approved):
Thanks!

PR #2685, opened 2025-05-16
Add stub for `django.db.models.expressions.Expression.identity`

cjwatson:
I don't think we need to be very specific about the contents of the returned tuple; it's just used for comparison and hashing methods.


sobolevn (approved):
Thanks!

cjwatson:
Sorry I'm taking so many tries at this - I haven't managed to get the stub tests working locally so the iteration time is kind of long!  Hopefully this time should be closer.

cjwatson:
I eventually found https://github.com/python/mypy/pull/18959 that explains the errors I was running into.  I've added `allowlist.txt` entries for now; I assume you'll remove those in bulk later when upgrading mypy.

sobolevn:
We don't need this anymore, since mypy1.16 solved this problem :)

cjwatson:
Indeed!  Rebased to get rid of that.

sobolevn (approved):
Thanks!

PR #2688, opened 2025-05-18
Switch to pyproject.toml + PEP621

UnknownPlatypus:
Fixes #1570
Fixes #2641

This revives and update #1570 allowing us to use PEP621 and remove the `SETUPTOOLS_ENABLE_FEATURES=legacy-editable` we had to use and that broke ci recently.

The `pyproject.toml` was intially generated using [`hatch new --init`](https://hatch.pypa.io/latest/intro/#existing-project) and tweaked to keep existing ordering / comments.

I've checked what differs between the [pypi current](https://pypi.org/project/django-stubs/#files) dist and what is build here and found the following differences:
- we no longer ship the `tests` folders ( which is good ?)
- we no longer ship the `django_stubs.egg-info` folder in the sdist distribution which I believe is expected since it's a setuptools artifact

## Test plan

- [x] `python -m build ext/ && python -m build`
- [x] `pip install -r ./requirements.txt` && run the whole ci locally


ext/pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+name = "django-stubs-ext"
+# NB! For clarity, keep version major.minor.patch in sync with django-stubs.
+# It's fine to skip django-stubs-ext releases, but when doing a release, update this to newest django-stubs version.
+version = "5.2.0"
+description = "Monkey-patching and extensions for django-stubs"
+readme = "README.md"
+license = "MIT"
+license-files = ["LICENSE.md"]
+maintainers = [
+  { name = "Marti Raudsepp", email = "marti@juffo.org" },
```
sobolevn, line 15:
```suggestion
  { name = "Marti Raudsepp", email = "marti@juffo.org" },
  { name = "Nikita Sobolev", email = "mail@sobolevn.me" },
```

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+description = "Mypy stubs for Django"
+readme = "README.md"
+license = "MIT"
+license-files = ["LICENSE.md"]
+requires-python = ">=3.10"
+authors = [
+    { name = "Maksim Kurnikov", email = "maxim.kurnikov@gmail.com" },
+]
+maintainers = [
+    { name = "Marti Raudsepp", email = "marti@juffo.org" },
```
sobolevn:
```suggestion
    { name = "Marti Raudsepp", email = "marti@juffo.org" },
    { name = "Nikita Sobolev", email = "mail@sobolevn.me" },
```

sobolevn: We should add other active maintainers as well, but I don't know their emails :(

- https://github.com/flaeppe
- https://github.com/adamchainz

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+    "tomli; python_version < '3.11'",
+    # Types:
+    "typing-extensions>=4.11.0",
+    "types-PyYAML",
+]
+
+[project.optional-dependencies]
+compatible-mypy = ["mypy>=1.13,<1.16"]
+oracle = ["oracledb"]
+redis = ["redis"]
```
sobolevn: question: will `redis` + `types-redis` work better for our case?

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+    "Programming Language :: Python :: 3.13",
+    "Typing :: Typed",
+    "Framework :: Django",
+    "Framework :: Django :: 5.0",
+    "Framework :: Django :: 5.1",
+    "Framework :: Django :: 5.2",
+]
+dependencies = [
+    "django",
+    "asgiref",
```
sobolevn: I think that this can be removed, since 5.0 django has this requirement: https://github.com/django/django/blob/5.0/setup.cfg#L41

ext/pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+]
+authors = [
+  { name = "Simula Proxy" },
+]
+requires-python = ">=3.10"
+classifiers = [
+    "License :: OSI Approved :: MIT License",
```
sobolevn:
```suggestion
  "License :: OSI Approved :: MIT License",
```

sobolevn: and all others should have a proper spacing

UnknownPlatypus: Should I use https://github.com/tox-dev/pyproject-fmt ? I can add it to the pre-commit config so that this gets handled automatically

sobolevn: IFF it will not format headers, because it might turn 

```toml
[tool.ruff.lint.per-file-ignores]
```

into

```toml
[tool.ruff]
lint.per-file-ignores."tests/*.py" = ["INP001", "PGH003"]
```

which is very hard to read.

If it is configurable / disabled - sure!

ngnpope: I believe that the license trove classifiers are deprecated in favour of an SPDX license expression in `license`, so this could be removed:

```suggestion
```

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+license-files = ["LICENSE.md"]
+requires-python = ">=3.10"
+authors = [
+    { name = "Maksim Kurnikov", email = "maxim.kurnikov@gmail.com" },
+]
+maintainers = [
+    { name = "Marti Raudsepp", email = "marti@juffo.org" },
+]
+classifiers = [
+    "License :: OSI Approved :: MIT License",
```
ngnpope: I believe that the license trove classifiers are deprecated in favour of an SPDX license expression in `license`, so this could be removed:

```suggestion
```

ext/pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Typing :: Typed",
+    "Framework :: Django",
+    "Framework :: Django :: 5.0",
```
ngnpope: The table in the readme only expresses partial support for 5.1 for stubs v5.2:

```suggestion
```

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Typing :: Typed",
+    "Framework :: Django",
+    "Framework :: Django :: 5.0",
```
ngnpope: The table in the readme only expresses partial support for 5.1 for stubs v5.2:

```suggestion
```

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+    "Framework :: Django :: 5.2",
+]
+dependencies = [
+    "django",
+    "asgiref",
+    "django-stubs-ext>=5.2.0",
+    "tomli; python_version < '3.11'",
+    # Types:
+    "typing-extensions>=4.11.0",
+    "types-PyYAML",
```
ngnpope: Nit: I'd just normalise and sort these:

```suggestion
    "types-pyyaml",
    "typing-extensions>=4.11.0",
```

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+compatible-mypy = ["mypy>=1.13,<1.16"]
+oracle = ["oracledb"]
+redis = ["redis"]
+
+[project.urls]
+Homepage = "https://github.com/typeddjango/django-stubs"
+Funding = "https://github.com/sponsors/typeddjango"
+"Release notes" = "https://github.com/typeddjango/django-stubs/releases"
+
+[tool.hatch.build]
```
ngnpope, line 53: Nit: Sort tool block alphabetically, i.e. move after `[tool.codespell]`?

ngnpope:
Also wondering if we could replace `./scripts/stubtest.sh` with [hatch's scripts](https://hatch.pypa.io/latest/config/environment/overview/#scripts) so that you can just run `hatch run stubtest`? 🤔

UnknownPlatypus:
> Also wondering if we could replace ./scripts/stubtest.sh with hatch's scripts so that you can just run hatch run stubtest? 🤔

I think this should be in a separate pr and we might have to consider UV too for that.

This pr currently doesn't use hatch but only the hatchling build backend

ngnpope:
> This pr currently doesn't use hatch but only the hatchling build backend

Ah, yes, got muddled by the `[tool.hatch.build]` 🙃

> I think this should be in a separate pr and we might have to consider UV too for that.

If you're considering uv instead, then have you thought about the [uv build](https://docs.astral.sh/uv/configuration/build-backend/) backend instead of  hatchling?

sobolevn:
let's please not replace `sh` scripts with `hatch` 🙏 
I am using `sh` for 15+ years now and I have no idea what `hatch` is.

pyproject.toml
UnknownPlatypus: I've taken their emails from their public repos

ext/pyproject.toml
UnknownPlatypus: ye I think this is true, I've seen adam do that [in djade ](https://github.com/adamchainz/djade/pull/123)

I've removed it

ext/pyproject.toml
UnknownPlatypus: We are still testing against 5.0, 5.1 and 5.2 in ci so I believe I should amend the readme instead ? 

https://github.com/typeddjango/django-stubs/blob/master/.github/workflows/test.yml#L136

EDIT: I've done that

pyproject.toml
UnknownPlatypus: Why not!

pyproject.toml
UnknownPlatypus: I kinda want to keep it close to the rest of the package metadata since it's use to build. No strong feeling tho

ext/pyproject.toml
UnknownPlatypus: It doesn't seem configurable so I've just updated to spacing to 2 spaces everywhere

pyproject.toml
UnknownPlatypus: It probably makes sense since we have a few redis import in `django-stubs/core/cache/backends/redis.pyi`.

I did this change and this warning got raised by mypy `Warning: unused section(s) in mypy.ini: [mypy-cryptography.*]` so I've also removed this section that seem not necessary anymore

ext/pyproject.toml
ngnpope: Fair.

sobolevn (commented):
One last question: right now our `build-and-check-metadata` uses different flags from the final release https://github.com/typeddjango/django-stubs/blob/81321eb71707f97986b91b861724ec3ce7ee274d/.github/workflows/test.yml#L170-L171

Is it ok? Or should we adjust `python -m build` flags?

UnknownPlatypus:
> Is it ok? Or should we adjust python -m build flags?

I tested running `python3 -m build --sdist --wheel .` and `python3 -m build .` and got exactly the same sdist and wheel.

`python3 -m build --help` does say it's slightly different but recommend to not use any flags so I've removed them.

```
usage: python -m build [-h] [--version] [--verbose] [--sdist] [--wheel] [--outdir PATH] [--skip-dependency-check]
                       [--no-isolation | --installer {pip,uv}] [--config-setting KEY[=VALUE]]
                       [srcdir]

    A simple, correct Python build frontend.

    By default, a source distribution (sdist) is built from {srcdir}
    and a binary distribution (wheel) is built from the sdist.
    This is recommended as it will ensure the sdist can be used
    to build wheels.

    Pass -s/--sdist and/or -w/--wheel to build a specific distribution.
    If you do this, the default behavior will be disabled, and all
    artifacts will be built from {srcdir} (even if you combine
    -w/--wheel with -s/--sdist, the wheel will be built from {srcdir}).
 ```

sobolevn (approved):

ngnpope (approved):
👍🏻

sobolevn:
Thanks everyone!

PR #2691, opened 2025-05-25
Fix some django52

skv0zsneg:
# I have made things!

Fix some rows from django 5.2 allow list.  

## Related issues

Not sure is there any issue related to this PR :)


django-stubs/db/models/sql/compiler.pyi
```diff
@@ -120,12 +120,14 @@ class SQLCompiler:
         self, result_type: Literal["multi"] = "multi", chunked_fetch: bool = False, chunk_size: int = 100
     ) -> Iterable[list[Sequence[Any]]] | None: ...
     def explain_query(self) -> Iterator[str]: ...
+    def composite_fields_to_tuples(self, rows: Iterable[Any], expressions: Iterable[Any]) -> Iterator[Any]: ...
```
sobolevn: I think that expressions should be typed as `Iterable[Expression]` and this function return an `Iterator[list[tuple[Any, ...]]]`

django-stubs/db/models/sql/compiler.pyi
```diff
@@ -120,12 +120,14 @@ class SQLCompiler:
...
+    def has_composite_fields(self, expressions: Iterable[Any]) -> bool: ...
```
sobolevn:
```suggestion
    def has_composite_fields(self, expressions: Iterable[Expression]) -> bool: ...
```

django-stubs/db/models/options.pyi
```diff
@@ -143,3 +143,7 @@ class Options(Generic[_M]):
     def total_unique_constraints(self) -> list[UniqueConstraint]: ...
     @cached_property
     def db_returning_fields(self) -> list[Field[Any, Any]]: ...
+    @cached_property
+    def pk_fields(self) -> Any: ...
```
sobolevn:
```suggestion
    def pk_fields(self) -> list[Field[Any, Any]]: ...
```

django-stubs/db/models/fields/composite.pyi
```diff
@@ -60,3 +60,5 @@ class CompositePrimaryKey(Field):
     def __iter__(self) -> Iterator[Field | ForeignObjectRel | GenericForeignKey]: ...
     def __len__(self) -> int: ...
     def get_pk_value_on_save(self, instance: Model) -> tuple: ...  # actual type is tuple of field.value_from_object
+
+def unnest(fields: Iterable[Field]) -> list[Field]: ...
```
sobolevn:
```suggestion
def unnest(fields: Iterable[Field[Any, Any]]) -> list[Field[Any, Any]]: ...
```

django-stubs/db/models/sql/compiler.pyi
```diff
@@ -120,12 +120,14 @@ class SQLCompiler:
...
 
 class SQLInsertCompiler(SQLCompiler):
     query: InsertQuery
     returning_fields: Sequence[Any] | None
     returning_params: Sequence[Any]
-    def field_as_sql(self, field: Any, val: Any) -> _AsSqlType: ...
+    def field_as_sql(self, field: Any, get_placeholder: Callable, val: Any) -> _AsSqlType: ...
```
sobolevn: see https://github.com/django/django/blob/b373721af0e5c3de0986977ac07e3ad55061ecbe/django/db/models/sql/compiler.py#L1700 for `get_placeholder` type

django-stubs/forms/forms.pyi
```diff
@@ -32,6 +32,7 @@ class BaseForm(RenderableFormMixin):
     template_name_table: str
     template_name_ul: str
     template_name_label: str
+    bound_field_class: None
```
sobolevn:
```suggestion
    bound_field_class: type[BoundField] | None
```

django-stubs/forms/renderers.pyi
```diff
@@ -14,6 +14,7 @@ class BaseRenderer:
     form_template_name: str
     formset_template_name: str
     field_template_name: str
+    bound_field_class: None
```
sobolevn: same

django-stubs/db/models/options.pyi
skv0zsneg: There I got strange stub errors here. I've add  the returning type hint `list[Field[Any, Any]]` but check complain to it :( So I change it to `Any`.
```
def (self: django.db.models.options.Options[_M`1]) -> builtins.list[django.db.models.fields.Field[Any, Any]]
Runtime:
<django.utils.functional.cached_property object at 0x7f35733ddef0>
```

sobolevn: you should add all `cached_property` usages to `allowlist` for now. this is a bug in mypy

skv0zsneg: So, until mypy fix put it (and same) to allow list?

scripts/stubtest/allowlist_todo_django52.txt
```diff
@@ -97,19 +92,8 @@ django.db.models.expressions.ColPairs
 django.db.models.expressions.OrderBy.allows_composite_expressions
 django.db.models.expressions.OutputFieldIsNoneError
 django.db.models.fields.Field.has_db_default
-django.db.models.fields.composite.unnest
 django.db.models.fields.related.ForeignKey.cast_db_type
 django.db.models.fields.tuple_lookups
 django.db.models.functions.JSONObject.join
 django.db.models.functions.json.JSONObject.join
-django.db.models.options.Options.is_composite_pk
 django.db.models.options.Options.pk_fields
```
sobolevn: Please, move this to `allowlist.txt` and return removed property definition. This will be fixed in the next mypy release.

django-stubs/db/models/sql/compiler.pyi
```diff
@@ -120,12 +120,18 @@ class SQLCompiler:
...
+    ) -> Iterator[list[tuple[Any, ...]]]: ...
+    def has_composite_fields(self, expressions: Iterable[Expression]) -> bool: ...
 
 class SQLInsertCompiler(SQLCompiler):
     query: InsertQuery
     returning_fields: Sequence[Any] | None
     returning_params: Sequence[Any]
-    def field_as_sql(self, field: Any, val: Any) -> _AsSqlType: ...
+    def field_as_sql(
+        self, field: Any, get_placeholder: Callable[[Any, SQLInsertCompiler, BaseDatabaseWrapper], str], val: Any
```
sobolevn:
```suggestion
        self, field: Field[Any, Any] | None, get_placeholder: Callable[[Any, SQLInsertCompiler, BaseDatabaseWrapper], str], val: Any
```

sobolevn (approved):
Thanks!

PR #2698, opened 2025-05-31
Use `uv` to manage dev dependencies

UnknownPlatypus:
# I have made things!

Switch to [uv](https://docs.astral.sh/uv).
I've also used [PEP-735](https://peps.python.org/pep-0735/) dependency groups to split up the dev dependencies. This alows us to:
- install all dev dependencies locally with `uv sync`
- install a subset in ci jobs with `uv sync --no-dev --group pyright`

## Test plan
- [x] ci passes
- [x] do the contributing guide from scratch

## Notes

When running `uv sync`, I have the following warning we should probably address but I'm not sure which version we should use:
```sh
warning: Missing version constraint (e.g., a lower bound) for `django`
warning: Missing version constraint (e.g., a lower bound) for `tomli`
warning: Missing version constraint (e.g., a lower bound) for `types-pyyaml`
warning: Missing version constraint (e.g., a lower bound) for `oracledb`
warning: Missing version constraint (e.g., a lower bound) for `redis`
warning: Missing version constraint (e.g., a lower bound) for `types-redis`
```


pyproject.toml
```diff
@@ -45,15 +45,42 @@ compatible-mypy = ["mypy>=1.13,<1.16"]
...
+[dependency-groups]
+tests = [
+    # Dev tools:
+    "pytest==8.3.5",
+    "pytest-mypy-plugins==3.2.0",
+    "pytest-shard==0.1.2",
+    "pytest-xdist==3.7.0",
+    # Django deps:
+    "django==5.2.1",
+    "django-stubs[compatible-mypy,oracle,redis]",
```
sobolevn: let's use it as the last one

pyproject.toml
```diff
@@ -45,15 +45,42 @@ compatible-mypy = ["mypy>=1.13,<1.16"]
...
 Homepage = "https://github.com/typeddjango/django-stubs"
 Funding = "https://github.com/sponsors/typeddjango"
 "Release notes" = "https://github.com/typeddjango/django-stubs/releases"
 
+[tool.uv.sources]
+django-stubs = { workspace = true }
+django-stubs-ext = { path = "ext", editable = true }
+
 [tool.hatch.build]
 packages = ["django-stubs", "mypy_django_plugin"]
```
sobolevn, line 82: Do we need to keep `hatch` or not?

sobolevn (commented):
Thanks! Several questions.

pyproject.toml
UnknownPlatypus: This part is for the `hatchling` build backend, uv also provides created a build backend but it's not [production ready](https://docs.astral.sh/uv/configuration/build-backend/) yet. 

uv will use hatchling as the default build backend and it's also the recommended option by [packaging.python.org](https://packaging.python.org/en/latest/tutorials/packaging-projects/#choosing-a-build-backend)

sobolevn (approved):
Thanks!

PR #2701, opened 2025-06-04
Add "django.contrib.auth.management.commands.createsuperuser" annotations

brianhelba:


django-stubs/contrib/auth/management/commands/createsuperuser.pyi
```diff
@@ -1,9 +1,21 @@
...
 class NotRunningInTTYException(Exception): ...
 
+PASSWORD_FIELD: str
+
 class Command(BaseCommand):
+    UserModel: type[AbstractBaseUser]
+    username_field: Field
     stdin: Any
+    def get_input_data(self, field: Field, message: str, default: str | None = ...) -> str | None: ...
+    def _get_input_message(self, field: Field, default: str | None = ...) -> str: ...
```
sobolevn: We generally don't annotate protected functions, unless they are really required. Is this the case here?

django-stubs/contrib/auth/management/commands/createsuperuser.pyi
```diff
@@ -1,9 +1,21 @@
...
+    @cached_property
+    def username_is_unique(self) -> bool: ...
+    def _validate_username(self, username: str, verbose_field_name: str, database: str) -> str | None: ...
```
sobolevn: And here.

django-stubs/contrib/auth/management/commands/createsuperuser.pyi
brianhelba: They were only required for my personal use case.

I'm fine to remove them. I can always provide them downstream as local stubs if I really need them.

brianhelba:
@sobolevn PTAL again.

django-stubs/contrib/auth/management/commands/createsuperuser.pyi
brianhelba: > I can always provide them downstream as local stubs if I really need them.

Actually, maybe not, as it appears that stubs are resolved all-or-nothing for each module (including `__init__.py` files), so it's going to be too much to override the hierarchy (and all intermediate `__init__.py` files) from `django/contrib/auth/management/commands`.

Regardless, I'm not asking the private members to be put upstream.

sobolevn (approved):
Thanks!

PR #2704, opened 2025-06-06
Weaken response argument types in django.utils.cache

cjwatson:
These functions just look at response headers, so `HttpResponseBase` is perfectly fine.  This is useful when working with responses returned by methods such as `django.test.Client.get`.


sobolevn (approved):
Thanks!

PR #2705, opened 2025-06-07
change assertFormError form arg type to BaseForm

marxide:
# I have made things!

Changes the form arg type for `django.test.SimpleTestCase.assertFormError` from `django.forms.Form` to `django.forms.BaseForm`. Allows instances of other form classes that are not subclasses of `Form` to be used without raising type errors, e.g. `ModelForm`. This also better aligns with the type used for `assertFormSetError` which types the formset arg as `BaseFormSet`, not `FormSet`.


sobolevn (approved):
Thanks!

PR #2710, opened 2025-06-11
Fix signature of `LogEntryManager.log_actions`

youtux:
`LogEntryManager.log_actions` can take any `Iterable[T]` for the `queryset` param, not necessarily a `QuerySet[T]`. Indeed, Django itself invokes it with a `list[T]` in `ModelAdmin.log_addition`: https://github.com/django/django/blob/091f66e51aa900f7d7650529621bdc8e4b0dee68/django/contrib/admin/options.py#L945

### DIFF
--- a/django-stubs/contrib/admin/models.pyi
+++ b/django-stubs/contrib/admin/models.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Iterable
-from typing import Any, ClassVar
+from typing import Any, ClassVar, Literal, overload
 from uuid import UUID
 
 from django.db import models
@@ -22,6 +22,7 @@ class LogEntryManager(models.Manager[LogEntry]):
         action_flag: int,
         change_message: Any = ...,
     ) -> LogEntry: ...
+    @overload
     def log_actions(
         self,
         user_id: int | str | UUID,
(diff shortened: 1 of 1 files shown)


django-stubs/contrib/admin/models.pyi
```diff
@@ -25,12 +25,12 @@ class LogEntryManager(models.Manager[LogEntry]):
...
         self,
         user_id: int | str | UUID,
-        queryset: QuerySet[Model],
+        queryset: Iterable[Model],
         action_flag: int,
         change_message: str | list[Any] = "",
         *,
         single_object: bool = False,
-    ) -> list[LogEntry] | LogEntry: ...
+    ) -> list[LogEntry]: ...
```
sobolevn, line 44: It can, here: https://github.com/django/django/blob/091f66e51aa900f7d7650529621bdc8e4b0dee68/django/contrib/admin/models.py#L51

youtux: good one, reverted

ngnpope: You can use `@overload` to be more specific about the return type:

```python
    @overload
    def log_actions(
        self,
        user_id: int | str | UUID,
        queryset: Iterable[Model],
        action_flag: int,
        change_message: str | list[Any] = "",
        *,
        single_object: Literal[True],
    ) -> LogEntry: ...
    @overload
    def log_actions(
        self,
        user_id: int | str | UUID,
        queryset: Iterable[Model],
        action_flag: int,
        change_message: str | list[Any] = "",
        *,
        single_object: Literal[False] = False,
    ) -> list[LogEntry]: ...
    @overload
    def log_actions(
        self,
        user_id: int | str | UUID,
        queryset: Iterable[Model],
        action_flag: int,
        change_message: str | list[Any] = "",
        *,
        single_object: bool = False,
    ) -> list[LogEntry] | LogEntry: ...
```

youtux: done in 9c19640

sobolevn (approved):
Thanks!

PR #2714, opened 2025-06-17
fix stubs of HttpRequest read / readine

asottile:
regressed in #2654 cc @UnknownPlatypus


sobolevn (approved):
Optional: you can add a test to `tests/assert_type/http/test_request.py` if you want.

asottile:
> Optional: you can add a test to `tests/assert_type/http/test_request.py` if you want.

I thought about that but couldn't come up with anything that didn't just seem like a reimplement of the stub itself like

```python
def f(x: HttpRequest, y: int | None) -> None:
    x.read(y)
```

which just seems kinda useless

UnknownPlatypus:
Thanks @asottile , sorry for the churn

PR #2716, opened 2025-06-18
Fix `reveal_type` to return its argument

andersk:
`reveal_type` can be used in the middle of an expression, as in `func(reveal_type(obj))`, so it needs to return its argument at runtime.

By the way, [`typing.reveal_type`](https://docs.python.org/3/library/typing.html#typing.reveal_type) was added to the runtime library in Python 3.11, and also as [`typing_extensions.reveal_type`](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.reveal_type) for all Python versions in typing-extensions 4.1.0, so this monkey-patch is not really needed anymore—maybe we should turn it off by default.

### DIFF
--- a/ext/django_stubs_ext/patch.py
+++ b/ext/django_stubs_ext/patch.py
@@ -135,5 +135,5 @@ def monkeypatch(extra_classes: Iterable[type] | None = None, include_builtins: b
 
     # Add `reveal_type` and `reveal_locals` helpers if needed:
     if include_builtins:
-        builtins.reveal_type = lambda _: None
+        builtins.reveal_type = lambda obj, /: obj
         builtins.reveal_locals = lambda: None


sobolevn (approved):
Thanks!

sobolevn:
> maybe we should turn it off by default.

We might, but instead I would suggest using `builtins.reveal_type = typing_extensions.reveal_type`

It is not cool to crash mypy, when you left unimported `reveal_type` somewhere :(

ngnpope:
> By the way, [`typing.reveal_type`](https://docs.python.org/3/library/typing.html#typing.reveal_type) was added to the runtime library in Python 3.11, and also as [`typing_extensions.reveal_type`](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.reveal_type) for all Python versions in typing-extensions 4.1.0, so this monkey-patch is not really needed anymore—maybe we should turn it off by default.

I'd noticed this independently, so wanted to say that I'm +1 to this. It's not really something that's relevant to Django, so probably best to remove it given there are other ways to have it available at runtime. While there's no equivalent to `reveal_locals`, it's probably best to remove that too.

PR #2718, opened 2025-06-18
Add missing instance variable declarations for `RawQuerySet`

youtux:
Adding a few missing instance variable types

### DIFF
--- a/django-stubs/db/models/query.pyi
+++ b/django-stubs/db/models/query.pyi
@@ -203,7 +203,7 @@ class RawQuerySet(Iterable[_Model], Sized):
     def __init__(
         self,
         raw_query: RawQuery | str,
-        model: type[Model] | None = None,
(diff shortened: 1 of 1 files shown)


django-stubs/db/models/query.pyi
```diff
@@ -194,7 +194,12 @@ class QuerySet(Generic[_Model, _Row], Iterable[_Row], Sized):
     def __getitem__(self, s: slice) -> Self: ...
 
 class RawQuerySet(Iterable[_Model], Sized):
+    raw_query: RawQuery | str
+    model: type[_Model] | None
```
sobolevn, line 198: should `model` in `__init__` be also updated?

youtux: yep, done

sobolevn (approved):

PR #2719, opened 2025-06-18
restore iterability of ResolverMatch

### LINKED ISSUE #2713: regression in 5.2.1 -- ResolverMatch is no longer iterable
# Bug report

regressed in #2635 cc @UnknownPlatypus 

## What's wrong

https://github.com/getsentry/sentry/blob/fdd78facb166e55f73854d39ff9b89d3df9ff62a/src/sentry/api/client.py#L54-L55

now getting:

```
src/sentry/api/client.py:55: error: "ResolverMatch" object is not iterable  [misc]
src/sentry/api/client.py:107: error: Cannot determine type of "callback"  [has-type]
src/sentry/api/client.py:107: error: Cannot determine type of "callback_args"  [has-type]
src/sentry/api/client.py:107: error: Cannot determine type of "callback_kwargs"  [has-type]
```

## How is that should be

no errors

## System information

- OS:
- `python` version: 3.13.1
- `django` version: 5.2.1
- `mypy` version: 1.15.0
- `django-stubs` version: 5.2.1 (forked)
- `django-stubs-ext` version: 5.2.1 (forked)

asottile:
resolves #2713


sobolevn (approved):
Thanks!

PR #2733, opened 2025-07-08
add stubs for QuerySet._raw_delete

### LINKED ISSUE #2732: types for `QuerySet._raw_delete`?
it's private -- but potentially useful.  seems to be `(str) -> int`

https://code.djangoproject.com/ticket/9519#comment:30

asottile:
resolves #2732


django-stubs/db/models/query.pyi
```diff
@@ -122,6 +122,7 @@ class QuerySet(Generic[_Model, _Row], Iterable[_Row], Sized):
     async def ain_bulk(self, id_list: Iterable[Any] | None = None, *, field_name: str = "pk") -> dict[Any, _Model]: ...
     def delete(self) -> tuple[int, dict[str, int]]: ...
     async def adelete(self) -> tuple[int, dict[str, int]]: ...
+    def _raw_delete(self, using: str) -> int: ...
```
sobolevn:
```suggestion
    def _raw_delete(self, using: str | None) -> int: ...
```

`using` is passed to https://github.com/asottile/django-stubs/blob/d283970c1f816994f54e1e90c9a7c0752c79668e/django-stubs/db/models/sql/query.pyi#L93

Source: https://github.com/django/django/blob/6df19412aabb7d969f5eab4b2ff41269de89b233/django/db/models/query.py#L1252

sobolevn (approved):
Thanks!

PR #2734, opened 2025-07-11
Fix aliasing of choices in a class in choices plugin

### LINKED ISSUE #2665: `TextChoices` subclasses declared inside a class cause "Unable to resolve type of choices property" error

ngnpope:
Fixes another small case that was missed in the choices plugin implementation.

Another fix for a3f2a558fc6c5544b3b46a6e53e6f266272236f7 (#2582)
Follow up to c37850b62a7217129e75f9985da37c2870b6b069 (#2646)

Fixes #2665

### DIFF
--- a/mypy_django_plugin/transformers/choices.py
+++ b/mypy_django_plugin/transformers/choices.py
@@ -5,6 +5,7 @@ from mypy.types import (
     AnyType,
     Instance,
     LiteralType,
+    Overloaded,
     ProperType,
     TupleType,
     TypeOfAny,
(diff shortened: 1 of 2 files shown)


ngnpope:
@psagers Would you be able to test this fix in your project?

psagers:
Success. I think this is our last 5.2 blocker.

Thanks!

sobolevn (approved):
Thank you!

PR #2735, opened 2025-07-13
Change resolve_callables argument type to Mapping

rodmk:
The [implementation for `resolve_callables`](https://github.com/django/django/blame/94ebcf8366d62f6360851b40e9c4dfe3f71d202f/django/db/models/utils.py#L28) doesn't mutate the argument, and can be typed to use `Mapping` instead of `MutableMapping`.

## Related issues

N/A

## Test plan

- test action: https://github.com/rodmk/django-stubs/actions/runs/16245525594
- `pre-commit run --all-files` and `pytest -n auto` pass


sobolevn (approved):
Thanks, looks fair!

Issue #1845, opened 2023-11-16
custom queryset methods/types get lost following calls to `.values()`, `.values_list()`


bwo:
(and possibly other methods; in fact, I would bet any method that isn't annotated as returning `Self` in `query.pyi`.)

```python
#$ cat foo.py 
from typing import Self, TypeVar
from django.db import models

T = TypeVar("T", bound=models.Model)

class MyQuerySet(models.QuerySet[T]):
    def filter_field(self, x: list[str]) -> Self:
        return self.filter(field__in=x)

mgr = models.Manager.from_queryset(MyQuerySet)


class MyModelA(models.Model):
    # objects = models.Manager()
    objects = mgr()

    field = models.CharField()


reveal_type(MyModelA.objects.filter(id__gt=1).filter_field(["hello"]).first())
reveal_type(MyModelA.objects.filter(id__gt=1).values("field").filter_field(["hello"]).first())  # AttributeError. typed as Any
reveal_type(MyModelA.objects.filter(id__gt=1).filter_field(["hello"]).values("id").first())
```

Then running mypy:

```
foo.py:20: note: Revealed type is "Union[payment.models.MyModelA, None]"
foo.py:21 error: "_QuerySet[MyModelA, TypedDict({'field': str})]" has no attribute "filter_field"  [attr-defined]
foo.py:21: note: Revealed type is "Any"
foo.py:22: note: Revealed type is "Union[TypedDict({'id': builtins.int}), None]"
```

That is: the chaining works following `filter`, and the chaining works if `.values()` is called after a custom method. But it doesn't work if `.values()` is called first—note that it's lost the belief that the queryset is of type `MyQuerySet`.

I believe (as noted) that the root cause here is that `.values()` is annotated in `query.pyi` like this:

```python
    def values(self, *fields: str | Combinable, **expressions: Any) -> _QuerySet[_T, dict[str, Any]]: ... 
```

So of course it loses track of the actual type! I'm not sure if there's a good (or any) way to notate "the same type constructor as me, but differently parameterized" in mypy, though.

## System information

- OS: macos 13.5.2
- `python` version: 3.11
- `django` version: 4.2
- `mypy` version: 1.6.1 (also happens with 1.7 and 1.8-dev versions)
- `django-stubs` version: 4.2.6
- `django-stubs-ext` version: 4.2.5

hterik:
> note that it's lost the belief that the queryset is of type MyQuerySet.

[...]

PR #2739, opened 2025-07-14
Improve `.values()` types when expressions are used

UnknownPlatypus:
# I have made things!

This fixes the types when using `**expressions` inside `QuerySet.values(...)` calls.

Prior to this change, mypy would return the following wrong types in these cases:
```python
reveal_type(Blog.objects.values(foo=F("id"))) # Revealed type is "django.db.models.query.QuerySet[myapp.models.Blog, TypedDict({'id': builtins.int, 'num_posts': builtins.int, 'text': builtins.str})]"
reveal_type(Blog.objects.values("id", foo=F("id"))) # Revealed type is "django.db.models.query.QuerySet[myapp.models.Blog, TypedDict({'id': builtins.int})]"
reveal_type(Blog.objects.values("id", lower_text=Lower("text"))) # Revealed type is "django.db.models.query.QuerySet[myapp.models.Blog, TypedDict({'id': builtins.int})]"
```

See 22262d1ae232b6c354bf81f4fb1612d18e3d9bb7 for more context
## Related issues

I did not found an exact issue for this but #1845 is kinda related

Further improvement would be to infer the type for some builtin django combinable such as `F(...)` (`.values("id", id2=F("id"))` should both use the same inference code)

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -227,6 +227,41 @@ def gather_kwargs(ctx: MethodContext) -> dict[str, MypyType] | None:
     return kwargs
 
 
+def gather_expression_types(ctx: MethodContext) -> dict[str, MypyType] | None:
+    kwargs = gather_kwargs(ctx)
+    if not kwargs:
+        return None
+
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Looks great, thanks!

PR #2741, opened 2025-07-14
Improve `settings.TEMPLATE` types

### LINKED ISSUE #2731: Type of "TEMPLATES" is partially unknown
# Bug report

## What's wrong

Is unexpected that there are no availably typing's for settings.py and all its values, currently my settings.py file raises this error "CONSTANT is partially unknow" everywhere

Image

## How is that should be

It should automatically type settings.py based on the name of such file or offer typings that one can import (if the former way its impossible for any reason)

I am aware I could use `TEMPLATES: dict[str, Any]` but using `Any` defeats the purpose of typing all together.

UnknownPlatypus:
# I have made things!

Narrow the types slightly for `settings.TEMPLATE` with a `TypedDict`

We could maybe improve for the `OPTIONS` key too but the possible values differ depending on the backend. 
And people could have custom backend adding custom keys there so I'm not sure it's worth pursuing.

## Related issues

Fixes #2731

### DIFF
--- a/django-stubs/conf/global_settings.pyi
+++ b/django-stubs/conf/global_settings.pyi
@@ -7,7 +7,7 @@ from typing import Any, Literal, Protocol, TypeAlias, TypedDict, type_check_only
 
 from typing_extensions import NotRequired
 
-from django_stubs_ext.settings import _TemplatesSetting
+from django_stubs_ext.settings import TemplatesSetting
 
 _Admins: TypeAlias = list[tuple[str, str]]
 
@@ -116,7 +116,7 @@ EMAIL_TIMEOUT: int | None
 # List of strings representing installed apps.
 INSTALLED_APPS: list[str]
 
-TEMPLATES: list[_TemplatesSetting]
+TEMPLATES: list[TemplatesSetting]
 
 # Default form rendering class.
 FORM_RENDERER: str
--- a/django-stubs/conf/global_settings.pyi
+++ b/django-stubs/conf/global_settings.pyi
@@ -7,6 +7,8 @@ from typing import Any, Literal, Protocol, TypeAlias, TypedDict, type_check_only
 
 from typing_extensions import NotRequired
 
+from django_stubs_ext.settings import _TemplatesSetting
+
 _Admins: TypeAlias = list[tuple[str, str]]
 
 ####################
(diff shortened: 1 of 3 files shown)


sobolevn (commented):
we surely can expose it. please, fix the pyright failure :)

django-stubs/conf/global_settings.pyi
```diff
@@ -114,7 +115,15 @@ EMAIL_TIMEOUT: int | None
 # List of strings representing installed apps.
 INSTALLED_APPS: list[str]
 
-TEMPLATES: list[dict[str, Any]]
+@type_check_only
+class _TemplateConfig(TypedDict):
```
ngnpope: Nit: I'm sure there'll be more of these for other settings, so let's aim for `_{name.to_pascal_case()}Setting` as the format for these...

```suggestion
class _TemplatesSetting(TypedDict):
```

Then if we expose them via `django_stubs_ext` it'll be consistent.

ext/django_stubs_ext/settings.py
```diff
@@ -0,0 +1,16 @@
+from pathlib import Path
+from typing import Any, TypedDict, type_check_only
+
+from typing_extensions import NotRequired
+
+
+@type_check_only
+class _TemplatesSetting(TypedDict):
```
UnknownPlatypus: Should we keep the leading `_` for symbols we expose in `django_stubs_ext` ?

sobolevn: Nope, exposed types should be public.

UnknownPlatypus: This probably needs to be documented somewhere too, should I add details in the ext/README.md and link to it in the main readme ?

sobolevn (approved):
thanks!

ext/django_stubs_ext/settings.py
sobolevn: I think that this is a separate feature. We need to rework the whole `ext/README.md` file. It is very old and does not cover many use-cases.

ext/django_stubs_ext/settings.py
```diff
@@ -0,0 +1,16 @@
+from pathlib import Path
+from typing import Any, TypedDict, type_check_only
+
+from typing_extensions import NotRequired
+
+
+@type_check_only
+class TemplatesSetting(TypedDict):
```
sobolevn, line 8:
```suggestion
class TemplatesSetting(TypedDict):
    """Typing helper if you want to type `TEMPLATE` setting."""
```

PR #2742, opened 2025-07-15
BasePasswordHasher.safe_summary commonly uses StrOrPromise

jdelic:
All Django-provided subclasses of `BasePasswordHasher` return a dict in the form of

```py
{
     _('algorithm'): self.algorithm, ... 
}
```

so the key is commonly a `StrOrPromise`.

(I have not filed a separate issue for this, just the PR)


django-stubs/contrib/auth/hashers.pyi
```diff
@@ -32,7 +34,7 @@ class BasePasswordHasher:
     def verify(self, password: str, encoded: str) -> bool: ...
     def encode(self, password: str, salt: str) -> str: ...
     def decode(self, encoded: str) -> dict[str, Any]: ...
-    def safe_summary(self, encoded: str) -> dict[str, Any]: ...
+    def safe_summary(self, encoded: str) -> dict[str | StrOrPromise, Any]: ...
```
UnknownPlatypus:
```suggestion
    def safe_summary(self, encoded: str) -> dict[StrOrPromise, Any]: ...
```
StrOrPromise already contain str I believe ?

jdelic: That's a very good point. I just tested it, too and am now making the change.

django-stubs/contrib/auth/hashers.pyi
```diff
@@ -1,6 +1,8 @@
 from collections.abc import Callable
 from typing import Any
 
+from django_stubs import StrOrPromise
```
sobolevn:
```suggestion
from django.utils.functional import _StrOrPromise
```

django-stubs/contrib/auth/hashers.pyi
```diff
@@ -32,7 +34,7 @@ class BasePasswordHasher:
     def verify(self, password: str, encoded: str) -> bool: ...
     def encode(self, password: str, salt: str) -> str: ...
     def decode(self, encoded: str) -> dict[str, Any]: ...
-    def safe_summary(self, encoded: str) -> dict[str, Any]: ...
+    def safe_summary(self, encoded: str) -> dict[StrOrPromise, Any]: ...
```
sobolevn:
```suggestion
    def safe_summary(self, encoded: str) -> dict[_StrOrPromise, Any]: ...
```

jdelic:
@sobolevn I remember that importing `_StrOrPromise` from Django previously initialized settings. Is that not an issue anymore? ... but even more broadly: as this is a PR in `django-stubs` why wouldn't we use the type defined inside of django-stubs (which is public, unlike the private module member from Django itself)?

sobolevn:
Nope, it is fine in stubs :)

sobolevn (approved):

jdelic:
Thank you for merging. :bowing_man:

PR #2745, opened 2025-07-15
Minor Plugin improvements

UnknownPlatypus:
# I have made things!

I'm trying to grok what is actually going on in the mypy plugin (which me luck!) and found some minor areas for improvement. I figured it would still be valuable hence this PR.

I've left a few inline comments to explain the changes

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -289,11 +289,7 @@ class NewSemanalDjangoPlugin(Plugin):
         return None
 
     def get_type_analyze_hook(self, fullname: str) -> Callable[[AnalyzeTypeContext], MypyType] | None:
-        if fullname in {
-            "typing.Annotated",
-            "typing_extensions.Annotated",
-            "django_stubs_ext.annotations.WithAnnotations",
-        }:
+        if fullname in fullnames.ANNOTATED_TYPES_FULLNAMES:
             return partial(handle_annotated_type, fullname=fullname)
         elif fullname == "django.contrib.auth.models._User":
             return partial(get_user_model, django_context=self.django_context)
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -290,11 +290,11 @@ class NewSemanalDjangoPlugin(Plugin):
         return None
 
     def get_type_analyze_hook(self, fullname: str) -> Callable[[AnalyzeTypeContext], MypyType] | None:
-        if fullname in (
+        if fullname in {
             "typing.Annotated",
             "typing_extensions.Annotated",
             "django_stubs_ext.annotations.WithAnnotations",
-        ):
+        }:
             return partial(handle_annotated_type, fullname=fullname)
         elif fullname == "django.contrib.auth.models._User":
             return partial(get_user_model, django_context=self.django_context)
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -112,7 +112,7 @@ class NewSemanalDjangoPlugin(Plugin):
                 return []
             return [self._new_dependency(auth_user_module), self._new_dependency("django_stubs_ext")]
 
-        # ensure that all mentioned to='someapp.SomeModel' are loaded with corresponding related Fields
+        # ensure that all mentions to='someapp.SomeModel' are loaded with corresponding related Fields
         defined_model_classes = self.django_context.model_modules.get(file.fullname)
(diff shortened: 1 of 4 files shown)


mypy_django_plugin/main.py
```diff
@@ -147,7 +147,7 @@ def get_function_hook(self, fullname: str) -> Callable[[FunctionContext], MypyTy
                 return partial(fields.transform_into_proper_return_type, django_context=self.django_context)
 
             if helpers.is_model_type(info):
-                return partial(init_create.redefine_and_typecheck_model_init, django_context=self.django_context)
+                return partial(init_create.typecheck_model_init, django_context=self.django_context)
```
UnknownPlatypus, line 150: This function does not really redefine nor alter the return type, it only typechecks the args.

mypy_django_plugin/main.py
```diff
@@ -277,24 +277,23 @@ def get_attribute_hook(self, fullname: str) -> Callable[[AttributeContext], Mypy
...
-        if (
-            info
-            and info.has_base(fullnames.CHOICES_TYPE_METACLASS_FULLNAME)
-            and attr_name in ("choices", "labels", "values", "__empty__")
+        if info and (
+            (
+                info.has_base(fullnames.CHOICES_TYPE_METACLASS_FULLNAME)
+                and attr_name in {"choices", "labels", "values", "__empty__"}
+            )
+            or (info.has_base(fullnames.CHOICES_CLASS_FULLNAME) and attr_name in {"label", "value"})
```
UnknownPlatypus, line 285: This removes a duplicated `if info` check

mypy_django_plugin/main.py
```diff
@@ -277,24 +277,23 @@ def get_attribute_hook(self, fullname: str) -> Callable[[AttributeContext], Mypy
...
             return choices.transform_into_proper_attr_type
 
-        if info and info.has_base(fullnames.CHOICES_CLASS_FULLNAME) and attr_name in ("label", "value"):
-            return choices.transform_into_proper_attr_type
-
         return None
 
     def get_type_analyze_hook(self, fullname: str) -> Callable[[AnalyzeTypeContext], MypyType] | None:
-        if fullname in (
+        if fullname in {
```
UnknownPlatypus: membership checks are slightly faster with sets

mypy_django_plugin/transformers/auth.py
```diff
@@ -10,17 +10,18 @@
...
+    sym = api.lookup_fully_qualified(fullnames.ABSTRACT_BASE_USER_MODEL_FULLNAME)
+    assert isinstance(sym.node, TypeInfo)
+    return fill_typevars_with_any(sym.node)
+
+
 def get_user_model(ctx: AnalyzeTypeContext, django_context: DjangoContext) -> MypyType:
     assert isinstance(ctx.api, TypeAnalyser)
     assert isinstance(ctx.api.api, SemanticAnalyzer)
 
-    def get_abstract_base_user(api: SemanticAnalyzer) -> ProperType:
```
UnknownPlatypus, line 17: Creating function inside function is rather expensive and we pay the function creation each time the outer function is called, this was not required here since the inner func does not use closures

mypy_django_plugin/transformers/init_create.py
```diff
@@ -28,8 +28,12 @@ def get_actual_types(ctx: MethodContext | FunctionContext, expected_keys: list[s
 
 
 def typecheck_model_method(
-    ctx: FunctionContext | MethodContext, django_context: DjangoContext, model_cls: type[Model], method: str
-) -> MypyType:
+    ctx: FunctionContext | MethodContext,
+    django_context: DjangoContext,
+    model_cls: type[Model],
+    method: str,
+) -> None:
```
UnknownPlatypus, line 35: This function actually only typecheck and then return `ctx.default_return_type`

I've done this change to decouple typecheking and eventual return type alteration

mypy_django_plugin/main.py
sobolevn: Let's make it a constant then :)

sobolevn (commented):
Thank you! Mostly looks good, except one nit :)

mypy_django_plugin/main.py
UnknownPlatypus: Thanks, just added this to the `fullnames.py` file

sobolevn (approved):
Thanks! Planning to make a release tomorrow, we have a new mypy version support.

PR #2752, opened 2025-07-23
Change partition input to any iterable

bewing:
Expand partition signature to allow any iterable set of values, to allow
passing in QuerySets and similar.

# I have made things!


sobolevn (approved):
Thanks! Indeed!

PR #2754, opened 2025-07-24
content_type_extra contains bytes, not str.

alexmv:
Here's my test showing the runtime type:

```
#!/usr/bin/env python3
from django.core.files.uploadedfile import UploadedFile
from django.core.handlers.wsgi import WSGIRequest
from django.test.client import BOUNDARY, MULTIPART_CONTENT, FakePayload

settings.configure(DEBUG=True)

payload = FakePayload(
    "\r\n".join(
        [
            f"--{BOUNDARY}",
            "Content-Type: text/plain; key=value; charset=big5",
            'Content-Disposition: form-data; name="file"; filename="test.txt"',
            "",
            "body",
            f"--{BOUNDARY}--",
        ]
    )
)
request = WSGIRequest(
    {
        "REQUEST_METHOD": "POST",
        "CONTENT_LENGTH": len(payload),
        "CONTENT_TYPE": MULTIPART_CONTENT,
        "wsgi.input": payload,
    }
)

parsed_file = request.FILES["file"]
assert isinstance(parsed_file, UploadedFile)

# {'key': b'value', 'charset': b'big5'}
print(parsed_file.content_type_extra)
```

### DIFF
--- a/django-stubs/core/files/uploadedfile.pyi
+++ b/django-stubs/core/files/uploadedfile.pyi
@@ -6,7 +6,7 @@ from typing_extensions import Self
 class UploadedFile(File):
     content_type: str | None
     charset: str | None
-    content_type_extra: dict[str, str] | None
+    content_type_extra: dict[str, bytes] | None
     size: int | None  # type: ignore[assignment]
     name: str | None
     def __init__(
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks!

PR #2764, opened 2025-08-09
Use mypy `Instance.copy_modified`

UnknownPlatypus:
# I have made things!

Noticed mypy provides `Instance.copy_modified` which is a more accurate version of our `helpers.reparametrize_instance` so I've updated the usages.

The method has been around for at least 8 years in mypy (and initially looked exactly like our version) so I think it's should be safe to change given our mypy version support policy.


sobolevn (approved):
Thanks! True!

PR #2766, opened 2025-08-11
Annotate return type of reverse_lazy()

### LINKED ISSUE #2765: `reverse_lazy()` shouldn't be `Any`
# Bug report

## What's wrong

`django.urls.base.reverse_lazy()` is annotated as `Any`, which results in false negatives in cases like

```
def myview(request) -> HttpResponse:
     return reverse_lazy("some-url")
```

## How is that should be
`reverse_lazy()` should be annotated as `_StrPromise`.

## System information

- OS: Linux Mint 22.1
- `python` version: 3.12
- `django` version: 5.1
- `mypy` version: 1.17.1
- `django-stubs-ext` version:  5.2.2

pgcd:
# I have made things!
Update `urls/base.pyi` with an annotated `_StrPromise` return type for `reverse_lazy()`

## Related issues
Closes #2765

### DIFF
--- a/django-stubs/urls/base.pyi
+++ b/django-stubs/urls/base.pyi
@@ -4,6 +4,7 @@ from typing import Any, Literal, TypeAlias
 from django.http.request import QueryDict
 from django.http.response import HttpResponseBase
 from django.urls.resolvers import ResolverMatch
+from django.utils.functional import _StrPromise
 
 # https://github.com/python/typeshed/blob/87f599dc8312ac67b941b5f2b47274534a1a2d3a/stdlib/urllib/parse.pyi#L136-L138
 _QueryType: TypeAlias = (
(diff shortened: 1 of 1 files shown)


django-stubs/urls/base.pyi
```diff
@@ -21,9 +22,16 @@ def reverse(
...
-
+def reverse_lazy(
+    viewname: Callable[..., HttpResponseBase] | str | None,
+    urlconf: str | None = ...,
+    args: Sequence[Any] | None = ...,
+    kwargs: dict[str, Any] | None = ...,
+    current_app: str | None = ...,
+    *,
+    query: QueryDict | _QueryType | None = ...,
+    fragment: str | None = ...,
```
sobolevn:
```suggestion
    fragment: str | None = None,
```

We prefer to specify defaults when possible.

django-stubs/urls/base.pyi
```diff
@@ -21,9 +22,16 @@ def reverse(
     query: QueryDict | _QueryType | None = ...,
     fragment: str | None = ...,
 ) -> str: ...
-
-reverse_lazy: Any
-
+def reverse_lazy(
```
sobolevn, line 27: Please add comments to both `reverse` and `reverse_lazy` to be in sync.

sobolevn (commented):
Thanks! Good catch!

django-stubs/urls/base.pyi
```diff
@@ -13,17 +14,24 @@ _QueryType: TypeAlias = (
 def resolve(path: str, urlconf: str | None = ...) -> ResolverMatch: ...
 def reverse(
```
sobolevn, line 17:
```suggestion
# NOTE: make sure `reverse` and `reverse_lazy` objects are in sync:
def reverse(
```

pgcd: Sorry, I misunderstood :grimacing:

sobolevn (approved):
Thanks!

PR #2769, opened 2025-08-12
Change FormMixin.success_url to _StrOrPromise

### LINKED ISSUE #2768: `FormMixin.success_url` should allow for StrPromise
# Bug report

## What's wrong

`django.views.generic.edit.FormMixin.success_url` is annotated as `str | None`; after the change in https://github.com/typeddjango/django-stubs/pull/2766 this might result in unexpected failures.

## How is that should be

`django.views.generic.edit.FormMixin.success_url` should be annotated as `_StrOrPromise | None` instead

## System information

- OS: Linux Mint 22.1
- `python` version: 3.12 
- `django` version: 5.1
- `mypy` version: 1.17.1

pgcd:
# I have made things!

Update `django.views.generic.edit.FormMixin.success_url` to `_StrOrPromise | None`, to allow using `reverse_lazy()` as attribute value.

## Related issues

Closes https://github.com/typeddjango/django-stubs/issues/2768

### DIFF
--- a/django-stubs/views/generic/edit.pyi
+++ b/django-stubs/views/generic/edit.pyi
@@ -5,6 +5,7 @@ from django.forms.forms import BaseForm
 from django.forms.models import BaseModelForm
 from django.http import HttpRequest, HttpResponse
 from django.utils.datastructures import _ListOrTuple
+from django.utils.functional import _StrOrPromise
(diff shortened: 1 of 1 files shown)


ngnpope (approved):
Thanks!

I suspect that there are some other URL-related attributes that would benefit from the same. I'll follow up on that.

PR #2772, opened 2025-08-18
Fix ` generic get type parameter is not optional` false positive

### LINKED ISSUE #2724: error: DateTimeField is nullable but its generic get type parameter is not optional
# Bug report

## What's wrong

After upgrading to 5.2.1, I am getting the following issue for any field that has null=True argument.

```
error: DateTimeField is nullable but its generic get type parameter is not optional  [misc]
...
error: ForeignKey is nullable but its generic get type parameter is not optional  [misc]
...
error: CharField is nullable but its generic get type parameter is not optional  [misc]
```

As a temporary measure, I downgraded to 5.2.0 and mypy to 1.15.0.

## System information

- OS: MacOS Sequoia 15.5; Ubuntu 24.04.2 (GitHub Actions)
- `python` version: 3.13.4
- `django` version: 5.2.3
- `mypy` version: 1.16.1

UnknownPlatypus:
# I have made things!

(not ready for review)
Just a failing test for now. Apparently, we were bailing most of the time previously due to this check;

```
# bail if either mapped_set_type or mapped_get_type have type Never
if not (isinstance(mapped_set_type, UninhabitedType) or isinstance(mapped_get_type, UninhabitedType)):
```

which is not the case anymore after mypy 1.16

# Relate issue

Fixes #2724

### DIFF
--- a/tests/typecheck/fields/test_nullable.yml
+++ b/tests/typecheck/fields/test_nullable.yml
@@ -49,6 +49,27 @@
                     text_nullable = models.CharField(max_length=100, null=True)
                     text = models.CharField(max_length=100)
 
+-   case: nullable_field_with_strict_optional_false
+    main: |
+        from myapp.models import MyModel
+        reveal_type(MyModel().text_nullable)  # N: Revealed type is "builtins.str | None"
+    installed_apps:
+        - myapp
+    mypy_config: |
+      [mypy]
+      strict_optional = True
+
+      [mypy-myapp.*]
+      strict_optional = False
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
Just a quick note, that `strict_optional` is soft deprecated and not recommended. I am not sure that we want to add this support. We can add a doc note about it.

UnknownPlatypus:
I do agree this flag is a footgun, we are in the process of removing it at work but it's not that easy around Django models.

We have a bunch of fields that are initially empty but will always be filled at some point in the lifecycle of the object. We struggle to represent that without requiring a bunch of assertion to narrow types all around the place.

But yeah you are probably right that we should not invest time fixing this but just recommend to user to flip the flag

Issue #2681, opened 2025-05-13
How to correctly type manager functions that adds annotations?


DrakryggenLinus:
I am working with a few models where annotations are added in queryset functions. What would be the correct way to type the returned queryset to maintain the ability chain the functions while also applying typing for the annotations?

```
class FooQuerySet(models.QuerySet["FooModel"]):
    def with_bar(self) -> FooQuerySet:
        return self.annotate(bar="bar")
    def with_baz(self) -> FooQuerySet:
        return self.annotate(baz="baz")

FooManager = BaseManager.from_queryset(FooQuerySet)

class FooModel(models.Model):
    objects = FooManager()
```

Example usage:
`FooModel.objects.with_bar().with_baz()`
should return: "FooQuerySet[FooModel@AnnotatedWith[TypedDict({'bar': Any, 'baz': Any})], FooModel@AnnotatedWith[TypedDict({'bar': Any, 'baz': Any})]"

valmirRG:
Hi, were you able to find a solution for this?

UnknownPlatypus:
I think it probably need to be defined like that, the important bit beeing it needs to remain generic. We cannot bind the type var with `QuerySet["FooModel"]` since we intend to alter it in methods:

```python
class BarDict(TypedDict):
    bar: str

class BazDict(TypedDict):
    baz: str

_Model = TypeVar("_Model", bound=models.Model, covariant=True)

class FooQuerySet(models.QuerySet[_Model]):
    def with_bar(self) -> FooQuerySet[WithAnnotations[_Model, BarDict]]:
        return self.annotate(bar=Value("bar"))

    def with_baz(self) -> FooQuerySet[WithAnnotations[_Model, BarDict]]:
        return self.annotate(baz=Value("baz"))


FooManager = BaseManager.from_queryset(FooQuerySet)


class FooModel(models.Model):
    objects = FooManager()
```

Given that, I would expect the following to be correct:
```python
reveal_type(FooModel.objects)  # FooManager[FooModel]
reveal_type(FooModel.objects.all())  # FooQuerySet[FooModel]
reveal_type(FooModel.objects.with_bar())  # FooQuerySet[FooModel@AnnotatedWith[TypedDict({'bar': Any})]]
reveal_type(FooModel.objects.with_bar().with_baz())  # FooQuerySet[FooModel@AnnotatedWith[TypedDict({'bar': Any, 'baz': Any})]]
```

[...]

PR #2776, opened 2025-08-21
`Manager` with generic `QuerySet` using `TypeVar` defaults

### LINKED ISSUE #728: select_related changes the manager class
# Bug report

## What's wrong

`select_related` breaks custom managers.

Take the following example:

```python
class TransactionQuerySet(models.QuerySet):
    def with_status(self):
        return self.filter(status=...)


class TransactionManager(models.Manager):

    def get_queryset(self) -> TransactionQuerySet:
        return TransactionQuerySet(self.model, using=self._db)

    def with_status(self):
        return self.get_queryset().with_status()

    ...
    objects = TransactionManager()
```

The following code has no type errors:
```python
Transaction.objects.with_status()
```

But the following fails:
```python
Transaction.objects.select_related().with_status()

error: "_QuerySet[Transaction, Transaction]" has no attribute "with_status"
```

The revealed types:

```python
reveal_type(Transaction.objects)
reveal_type(Transaction.objects.select_related())

note: Revealed type is "core.core.models.Transaction_TransactionManager5[core.core.models.Transaction]"
note: Revealed type is "django.db.models.query._QuerySet[core.core.models.Transaction*, core.core.models.Transaction*]"
```

This problem started after I updated django-stubs from 1.8.0 to 1.9.0

## System information

- OS:
- `python` version: 3.9.7
- `django` version: 3.2.8
- `mypy` version: 0.910
- `django-stubs-ext` version: Not using it

### LINKED ISSUE #863: Making QuerySet classes part of the type of Manager classes

### LINKED ISSUE #1023: Could not resolve manager type for  "core.models.FooModel.objects" [django-manager-missing]

### LINKED ISSUE #1067: Mypy can't get custom manager type

### LINKED ISSUE #1918: Related Manager lose typing
# Bug report

Hi, not sure if this is a bug or my having wrong declarations.

## What's wrong

The typing of related_manager.all is off when I share a QuerySet between multiple models :

```python
-   case: related
    main: |
        reveal_type(MyModel.objects.all()) # N: Revealed type is "myapp.models.MyQuerySet[myapp.models.MyModel]"
        reveal_type(list(MyModel.objects.all())[0]) # N: Revealed type is "myapp.models.MyModel"
        for x in MyModel.objects.all():
            reveal_type(x) # N: Revealed type is "myapp.models.MyModel"
            reveal_type(x.relatedobject_set.all()) # N: Revealed type is "myapp.models.MyQuerySet[myapp.models.RelatedObject]"
            reveal_type(list(x.relatedobject_set.all())[0]) # N: Revealed type is "myapp.models.RelatedObject"
            for y in x.relatedobject_set.all():
                reveal_type(y) # N: Revealed type is "myapp.models.RelatedObject"

    installed_apps:
        - myapp
    files:
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                from django.db import models
                from typing import List, Dict
                from typing import TypeVar, Generic

                class MyQuerySet(models.QuerySet[T], Generic[T]): ...

                class MyModel(models.Model):

                class RelatedObject(models.Model):
                    parent = models.ForeignKey(MyModel, models.PROTECT)

```

```
E   Actual:
E     ...
E     main:6: note: Revealed type is "myapp.models.MyQuerySet" (diff)
E     main:7: note: Revealed type is "T`1"          (diff)
E     main:9: note: Revealed type is "T`1"          (diff)
E     myapp/models:15: error: Need type annotation for "objects"  [var-annotated] (diff)
E     myapp/models:20: error: Need type annotation for "objects"  [var-annotated] (diff)
E   Expected:
E     ...
E     main:6: note: Revealed type is "myapp.models.MyQuerySet[myapp.models.RelatedObject]" (diff)
E     main:7: note: Revealed type is "myapp.models.RelatedObject" (diff)
E     main:9: note: Revealed type is "myapp.models.RelatedObject" (diff)
```

```python

-   case: related_subclassing
    main: |
        reveal_type(MyModel.objects.all()) # N: Revealed type is "myapp.models.BaseQuerySet[myapp.models.MyModel]"
        reveal_type(list(MyModel.objects.all())[0]) # N: Revealed type is "myapp.models.MyModel"
        for x in MyModel.objects.all():
            reveal_type(x) # N: Revealed type is "myapp.models.MyModel"
            reveal_type(x.relatedobject_set.all()) # N: Revealed type is "myapp.models.BaseQuerySet[myapp.models.RelatedObject]"
            reveal_type(list(x.relatedobject_set.all())[0]) # N: Revealed type is "myapp.models.RelatedObject"
            for y in x.relatedobject_set.all():
                reveal_type(y) # N: Revealed type is "myapp.models.RelatedObject"

    installed_apps:
        - myapp
    files:
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                from django.db import models
                from typing import List, Dict
                from typing import TypeVar, Generic

                class BaseQuerySet(models.QuerySet[T], Generic[T]): ...

                    objects = BaseQuerySet.as_manager()
                    class Meta:
                        abstract = True

                class MyModel(BaseModel):
                    pass

                class RelatedObjectQuerySet(BaseQuerySet["RelatedObject"]):
                    pass

                class RelatedObject(BaseModel):
                    parent = models.ForeignKey(MyModel, models.PROTECT)

```

```
E   Actual:
E     ...
E     main:6: note: Revealed type is "myapp.models.BaseQuerySet" (diff)
E     main:7: note: Revealed type is "T`1"          (diff)
E     main:9: note: Revealed type is "T`1"          (diff)
E     myapp/models:13: error: Need type annotation for "objects"  [var-annotated] (diff)
E   Expected:
E     ...
E     main:6: note: Revealed type is "myapp.models.BaseQuerySet[myapp.models.RelatedObject]" (diff)
E     main:7: note: Revealed type is "myapp.models.RelatedObject" (diff)
E     main:9: note: Revealed type is "myapp.models.RelatedObject" (diff)
```

## How is that should be

The types should be `MyModel` or `RelatedObject` instead of `T`.

Note: If I remove the Generic I get `Any` instead of `T`.

## System information

- OS: osx
- `python` version: 3.11.3
- `django` version: Django==4.2.9
- `mypy` version: mypy==1.7.1
- `django-stubs` version: django-stubs==4.2.7
- `django-stubs-ext` version:  django-stubs-ext==4.2.7

### LINKED ISSUE #2602: Custom queryset typing is lost
# Bug report

## What's wrong

When defining managers with a custom queryset, the typing of said queryset is lost. Instead of having `MyQuerySet` we have `QuerySet[MyModel, MyModel]`.
This happens whether the manager is created manually, or using `Manager.from_queryset(...)`

```python
from django.db import models


    pass


class MyManager(models.Manager):
    def get_queryset(self) -> MyQuerySet:
        return MyQuerySet(self.model, using=self._db)


class MyModel(models.Model):
    objects_my_manager = MyManager()
    objects_from_queryset = models.Manager.from_queryset(MyQuerySet)()


reveal_type(MyModel.objects_my_manager.all())  # Revealed type is "QuerySet[MyModel, MyModel]"
reveal_type(MyModel.objects_from_queryset.all())  # Revealed type is "QuerySet[MyModel, MyModel]"
```

## How is that should be

I would like all manager methods that return a queryset (`.all()`, `.filter(...)`, `.only(...)` ....) to return `MyQuerySet`.
We are currently using a workaround very similar to the one described here https://github.com/typeddjango/django-stubs/issues/863

## System information

- OS: Ubuntu 22.04.5 LTS
- `python` version: 3.11.12
- `django` version: 4.2.11
- `mypy` version: 1.15.0

UnknownPlatypus:
# I have made things!

This is redo for #1270 using `TypeVar` defaults to implement the change in a non-breaking way.

This is not completely ready, I opening it to get some feedback and because there are still a few issues to address.

The PR currently passes tests but I've tried it on my work codebase and some related issues have resurfaced (not caused by this PR, but by the fact that a lot of places previously ignored are now checked)

Most notably:

1. :white_check_mark: `.annotate()` in custom queryset method looses type
     - #2681
     - #1049
     - a bit related to #2314
3. :white_check_mark: ~I had to do the same re-parametrization trick we do for `Manager` (cf #1169) which causes exactly the same kind of issue with overrides of `models.Manager.get_queryset` which are rather common because [documented](https://docs.djangoproject.com/en/5.2/topics/db/managers/#modifying-a-manager-s-initial-queryset)~ -> was done in #3217
4. :white_check_mark:  ~Also https://github.com/typeddjango/django-stubs/issues/795 was more present~ -- Fixed in #2779)

I think this is step in the right direction but I'm a bit afraid it might cause some churn if we don't address the related issues first. 

### Todos
- [ ] Maybe fix #286 ?
- [ ] Maybe fixes #287 ?

## Related issues
- Closes #863 (tracking Issue)
- Closes #1270 (supersede prior pr addressing #863)

Also fixes a bunch of related issues, I've added regression tests for them:
- Closes #2602 
- Closes #1067 
- Closes #1023 
- Closes #728 
- Closes #1918


django-stubs/db/models/fields/related_descriptors.pyi
```diff
@@ -93,16 +95,16 @@ class ReverseManyToOneDescriptor(Generic[_To]):
     field: ForeignKey[_To, _To]
     def __init__(self, rel: ManyToOneRel) -> None: ...
     @cached_property
-    def related_manager_cls(self) -> type[RelatedManager[_To]]: ...
+    def related_manager_cls(self) -> type[RelatedManager[_To, _To_QS]]: ...
     @overload
-    def __get__(self, instance: None, cls: Any | None = None) -> Self: ...
+    def __get__(self, instance: None, cls: Any | None = None) -> ReverseManyToOneDescriptor[_To, _To_QS]: ...
```
UnknownPlatypus, line 100: I wanted to return `Self[_To, _To_QS`] but it's [not supported](https://typing.python.org/en/latest/spec/generics.html#self:~:text=Note%20that%20we%20reject%20using%20Self%20with%20type%20arguments%2C%20such%20as%20Self%5Bint%5D%2E). It's probably fine because I don't expect people to subclass this ?

django-stubs/db/models/manager.pyi
```diff
@@ -8,13 +8,15 @@ from django.db.models.query import QuerySet, RawQuerySet
...
+_QS = TypeVar("_QS", bound=QuerySet[Any], covariant=True, default=QuerySet[_T])
 
-class BaseManager(Generic[_T]):
+class BaseManager(Generic[_T, _QS]):
     creation_counter: int
     auto_created: bool
     use_in_migrations: bool
     name: str
     model: type[_T]
+    _queryset_class: type[_QS]
```
UnknownPlatypus, line 19: I needed to expose this symbol because I use it in the plugin. 
If we don't want that, I can also add a type-ignore in the plugin code but I think it's fine to have it because it's a core component of a manager and very unlikely to disappear

mypy_django_plugin/transformers/models.py
```diff
@@ -887,21 +966,26 @@ def create_through_table_class(
         # Add the foreign key's '_id' field: <other_model>_id or to_<model>_id
         other_pk = self.get_pk_instance(m2m_args.to.model.type)
         helpers.add_new_sym_for_info(through_model, name=f"{to_name}_id", sym_type=other_pk.copy_modified())
+        return through_model
+
+    def add_through_table_managers(self, through_model: TypeInfo) -> None:
+        """The `self.manager_info` lookup might trigger a deferral pass so this has to be idempotent"""
```
UnknownPlatypus, line 972: This solves an issue that was already present but happening less often I suppose. With the typevar default, it started to cause some tests to fail.

Sometimes, `create_through_table_class` would fail creating the `objects` symbol, leaving an incomplete table definition and later mypy passes were never filling the missing part

mypy_django_plugin/transformers/managers.py
```diff
@@ -625,6 +624,39 @@ class MyManager(models.Manager[_T]): ...
...
+    Eg.
+
+        class MyQuerySet(models.QuerySet): ...
+
+    is interpreted as:
+
+        _T = TypeVar("_T", bound=Model, covariant=True)
+        class MyQuerySet(models.QuerySet[_T]): ...
+    """
+    reparametrize_generic_class(ctx, fullnames.QUERYSET_CLASS_FULLNAME)
```
UnknownPlatypus, line 662: I've added the automatic parametrisation because without it, a bunch of places where it was before `Queryset[MyModel]` became `MyQueryset[Any]` for untyped custom queryset.

There are still a lot of untyped django package around so this makes them work with little additional effort

django-stubs/db/models/manager.pyi
```diff
@@ -153,5 +163,9 @@ class ManagerDescriptor:
...
     def __get__(self, instance: Model, cls: type[Model] | None = None) -> NoReturn: ...
 
-class EmptyManager(Manager[_T]):
-    def __init__(self, model: type[_T]) -> None: ...
+# We have to define different typevars here otherwise it conflicts with the ones above
+_T2 = TypeVar("_T2", bound=Model, covariant=True)
+_QS2 = TypeVar("_QS2", bound=QuerySet[Any], covariant=True, default=QuerySet[_T2])
+
+class EmptyManager(Manager[_T2, _QS2]):
+    def __init__(self, model: type[_T2]) -> None: ...
```
UnknownPlatypus, line 171: I'm really not quite sure why but If I reuse the `T` and `_QS` typevar from above, some tests start to fail 
For some reason, it was affecting `Manager` inference and the `_T``1` queryset defaults where not all associated.

```diff
reveal_type(models.Manager[MyModel])
-note: Revealed type is "django.db.models.manager.Manager[django.db.models.base.Model, django.db.models.query.QuerySet[_T`1, _T`1]]"
+note: Revealed type is "django.db.models.manager.Manager[django.db.models.base.Model, django.db.models.query.QuerySet[django.db.models.base.Model, django.db.models.base.Model]]
```

rtpg:
@UnknownPlatypus do you want eyes on this? I had issues with annotations that ended up with me [writing this test](https://github.com/typeddjango/django-stubs/pull/3068), _maybe_ this branch could end up resolving those too

UnknownPlatypus:
@rtpg I plan on rebasing this pr latter this week now that most of the blocking issue mentioned in the description are resolved.

If you have some time, it would be great to resolve the [annotations in custom queryset issue](https://github.com/typeddjango/django-stubs/issues/2681). Without that, the current state of this PR will cause a lot of issues for such cases that were previously ignored

I'll integrate the test you proposed in this MR, I think it will work

PR #2779, opened 2025-08-22
Initial support for `to_attr` inference in `Prefetch` calls

### LINKED ISSUE #795: Prefetch's to_attr raises "Model" has no attribute "prefetched_field"
# Bug report

## What's wrong

```python
class Tag(models.Model):
    pass

class Article(models.Model):
    tags = models.ManyToManyField(to=Tag, related_name="articles", blank=True)

    def method(self):
        published_articles = Article.objects.exclude(pk=self.pk)
        for tag in self.tags.all().prefetch_related(
            Prefetch("articles", published_articles, to_attr="published_articles")
        ):
            print(tag.published_articles)
```

```
src/articles/models.py:147: error: "Tag" has no attribute "published_articles"
```

## How is that should be

`published_articles` should be recognized as a valid attribute for `tag` in the loop, as defined in the `to_attr` of the `Prefetch` instance.

## System information

- OS: macOS 12.1 (21C52)
- `python` version: Python 3.10.0
- `django` version: 4.0
- `mypy` version: mypy 0.930

UnknownPlatypus:
# I have made things!

A bit more involved than I expected. I was close to reuse some mypy plugin utils for the argument inference but they currently do no support class instantiation so I've adapted from it.
The logic is very similar to what is done in the `.annotate()` plugin code.

I left a few todos for improvements in later PR's, this one is already quite big.

<details>
<summary>todos

</summary>

### Correctly handle mixed `annotate`/`prefetch_related` on the same key -- :white_check_mark:  in #2791

Currently it applies the last annotation, in reality, only the annotate matter and the prefetch is ignored + we should probably raise type errors?

```yml
-   case: prefetch_related_conflict_with_annotate
    main: |
        from django.db.models import Prefetch, F
        from django.contrib.auth.models import User, Group

        # When mixing `.annotate(foo=...)` and `prefetch_related(Prefetch(...,to_attr=foo))`
        # The last annotate in the chain takes precedence (even if it is prior to the prefetch_related)
        # TODO: Whould be nice to raise errors here
        reveal_type(
            User.objects # N: Revealed type is "builtins.list[django.contrib.auth.models.Group]"  <-- Wrong, runtime type is F("username")
            .annotate(foo=F("username"))
            .prefetch_related(Prefetch("groups", Group.objects.all(), to_attr="foo"))
            .get().foo
        )
        reveal_type(
            User.objects # N: Revealed type is "Any"
            .prefetch_related(Prefetch("groups", Group.objects.all(), to_attr="foo"))
            .annotate(foo=F("username"))
            .get().foo
        )
    installed_apps:
        - django.contrib.auth
```

### Narrow type when no queryset is provided -- :white_check_mark: in #2786
Using the existing lookup_type infrastructure.
- create `get_model_info_from_qs_ctx`
- reuse / merge with `_extract_model_type_from_queryset`

```diff
# Prefetch with `to_attr` arg but without the `queryset` arg
# TODO: We should be able to resolve a more accurate type using existing lookup `resolve_lookup_expected_type` machinery
reveal_type(
    Article.objects.prefetch_related(
-     models.Prefetch("tags", to_attr="tags")).get().tags)  # N: Revealed type is "builtins.list[Any]"
+     models.Prefetch("tags", to_attr="tags")).get().tags)  # N: Revealed type is "builtins.list[myapp.models.Tag]"
    )
)
```

</details>

## Related issues

Fixes #795

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -6,7 +6,7 @@ from django.db.models.base import Model
 from django.db.models.fields.related import RelatedField
 from django.db.models.fields.reverse_related import ForeignObjectRel
 from mypy.checker import TypeChecker
-from mypy.nodes import ARG_NAMED, ARG_NAMED_OPT, AssignmentStmt, CallExpr, Expression, FuncDef, MypyFile, NameExpr
+from mypy.nodes import ARG_NAMED, ARG_NAMED_OPT, CallExpr, Expression
 from mypy.plugin import FunctionContext, MethodContext
 from mypy.types import AnyType, Instance, TupleType, TypedDictType, TypeOfAny, get_proper_type
 from mypy.types import Type as MypyType
@@ -34,42 +34,6 @@ def _extract_model_type_from_queryset(queryset_type: Instance, api: TypeChecker)
         model = get_proper_type(base_type.args[0])
         if isinstance(model, Instance) and helpers.is_model_type(model.type):
             return model
-    return None
-
-
-def _resolve_prefetch_call_expr(expr: Expression, api: TypeChecker) -> CallExpr | None:
-    """
-    Resolve a Prefetch call expression from either an inline call or a variable assignment.
-
-    Inline call:
-        MyModel.objects.prefetch_related(Prefetch("tags", Tag.objects.all(), to_attr="every_tags"))
-
-    Variable assignment:
-        tag_prefetch = Prefetch("tags", Tag.objects.all(), to_attr="every_tags")
-        MyModel.objects.prefetch_related(tag_prefetch)
-    """
-    if isinstance(expr, CallExpr):
-        return expr
-
-    if not isinstance(expr, NameExpr):
-        return None
-
-    for scope in reversed(api.scope.stack):
-        if isinstance(scope, FuncDef):
-            statements = scope.body.body
-        elif isinstance(scope, MypyFile):
-            statements = scope.defs
-        else:
-            continue
-
-        for stmt in statements:
-            if (
-                isinstance(stmt, AssignmentStmt)
-                and any(isinstance(lval, NameExpr) and lval.name == expr.name for lval in stmt.lvalues)
-                and isinstance(stmt.rvalue, CallExpr)
-            ):
-                return stmt.rvalue
-
     return None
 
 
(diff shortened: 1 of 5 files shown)


mypy_django_plugin/transformers/querysets.py
```diff
@@ -37,6 +37,42 @@ def _extract_model_type_from_queryset(queryset_type: Instance, api: TypeChecker)
...
+    for scope in reversed(api.scope.stack):
+        if isinstance(scope, FuncDef):
+            statements = scope.body.body
+        elif isinstance(scope, MypyFile):
+            statements = scope.defs
+        else:
+            continue
+
+        for stmt in statements:
+            if (
```
sobolevn: I see this as a potential footgun, for example, we can have something like:

```python
tag_prefetch = Prefetch("tags", Tag.objects.all(), to_attr="every_tags")

# ...

tag_prefetch = Prefetch("tags", Tag.objects.all(), to_attr="other_tags")
MyModel.objects.prefetch_related(tag_prefetch)
```

Maybe we should not support this pattern and require people to inline it?

mypy_django_plugin/transformers/querysets.py
```diff
@@ -373,3 +409,82 @@ def extract_proper_type_queryset_values(ctx: MethodContext, django_context: Djan
...
+        and isinstance((default_return_type := get_proper_type(ctx.default_return_type)), Instance)
+        and (api := helpers.get_typechecker_api(ctx))
+        and (model_type := _extract_model_type_from_queryset(ctx.type, api)) is not None
+        and ctx.args
+        and ctx.arg_types
+        and ctx.arg_types[0]
+    ):
+        return ctx.default_return_type
+
+    fields: OrderedDict[str, MypyType] = OrderedDict()
```
sobolevn: `dict` is ordered :)
We can drop all `OrderedDict` usages.

sobolevn (commented):
Thanks! Looks great, exept one thing :(

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: Ye I've noticed that too, this comes from the method created in 2019 above using that.

I was planning to address this in a followup pr if your fine with that ?

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: traversing statements in reverse could solve it ?

But I'm surprised mypy does not provide an utility somewhere to do this, I'm basically trying to do a simple variable lookup (that is currently quite bogus). 

What I would really like to support is Prefetch build from functions (I see this rather often) ie something like;
```python
def build_prefetches() -> Prefetch:
    return Prefetch(
        queryset=SocialNetwork.objects.filter(is_active=True),
       to_attr="active_socials"
    )

tag_prefetch = build_prefetches()
MyModel.objects.prefetch_related(tag_prefetch)
```

But I suppose unless we can have the required info in the `Prefetch` type itself, we will be chasing the actual implementation from one ref to another which will be a pain ?

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: But i guess I can still fix it in this pr and do another for the other occurences

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: Maybe Prefetch should be generic so that the plugin can extract information more easily? ie we could do 

```python
def build_prefetches() -> Prefetch[QuerySet[SocialNetwork], Literal["active_socials"]]:
    return Prefetch(
       queryset=SocialNetwork.objects.filter(is_active=True),
       to_attr="active_socials"
    )
```

Anyways, I'll remove the current implementation and defer this to another PR I think

sobolevn: Making it generic is fine :)

sobolevn (approved):
Looks great! Thank you!

PR #2783, opened 2025-08-25
Simplify parsing of bool call arguments

UnknownPlatypus:
# I have made things!

Add a small wrapper on top of `get_call_argument_by_name` to simplify parsing of boolean argument with default.
Still taking inspiration from some mypy plugin helpers, adapting to the `FunctionContext`/`MethodContext` use case


sobolevn (approved):
Thanks!

PR #2784, opened 2025-08-25
Improve `Model` inference from custom `QuerySet`/`Manager`

### LINKED ISSUE #1172: kwargs to the queryset "filter()" method not type checked in chained filter operations
# Bug report
## What's wrong
Kwargs passed into the `.filter()` method on a queryset only have type checking the _first_ time `filter` is called on the queryset. All subsequent calls to `filter` on the same queryset do not have their arguments checked. 

Here's an example file: 
```
from django.db import models

class Cheese(models.Model):
    a = models.BooleanField()
    
def f() -> None:
    Cheese.objects.filter(field=True).filter(field2=True)
    qs = Cheese.objects.all()
    qs.filter(field=True)

```
Running mypy outputs the following:
```
app/models.py:7: error: Cannot resolve keyword 'field' into field. Choices are: a, id  [misc]
Found 1 error in 1 file (checked 13 source files)
```
There is only 1 error caught by mypy here, but there should really be 3! The code above calls `filter` 3 times, each time with arguments that should fail type checking. 

I also tried the following other solutions, but only django-stubs 1.8.0 worked:
- upgrading django to 4.1.1
- upgrading and downgrading mypy
- django-stubs 1.9.0, 1.10.0, 1.11.0, 1.12.0
- pinning the latest master commit of django-stubs

## How it should be
Mypy should output 3 errors for the code sample above. Rerunning mypy with django-stubs 1.8.0 catches all 3 errors:
```
app/models.py:7: error: Cannot resolve keyword 'field' into field. Choices are: a, id  [misc]
app/models.py:7: error: Cannot resolve keyword 'field2' into field. Choices are: a, id  [misc]
app/models.py:9: error: Cannot resolve keyword 'field' into field. Choices are: a, id  [misc]
Found 3 errors in 1 file (checked 13 source files)
```

## System information

- OS: MacOS Big Sur 11.7
- `python` version: 3.8.9
- `django` version: 4.1.1
- `mypy` version: 0.981

Here is a minimum repo for reproducing this issue: https://github.com/Pastromhaug/django-stubs-filter-bug
At my company we have temporarily pinned django-stubs to 1.8.0 to work around this. 

Hope this is helpful! Happy to answer any questions.

UnknownPlatypus:
# I have made things!
_(review commit per commit should be easier)_

This MR unify the logic to go from a `QuerySet`/`Manager` mypy's `MethodContext` to the associated django Model infos.

This fixes the typechecking of `filter`/`exclude` on custom queryset which was previously broken because the hook was naively looking at the first type argument of the custom queryset (which does not exist). I've added a test that was failing prior to this PR and now passes. 

```python
class MyQuerySet(models.QuerySet[MyModel]):
    ...

# isinstance((model_type := get_proper_type(ctx.type.args[0])), Instance) fails in that case because ctx.type is `MyQuerySet` which does not have type arg, only the child `QuerySet` does.
```

I've also removed the `manager_to_model` metadata we stored on generate `ManyRelatedManager` TypeInfos. It's redundant with the first generic param of the underlying manager.

## Related issues

Fixes #1172 (I've added a regression test for this)


mypy_django_plugin/lib/helpers.py
```diff
@@ -201,6 +194,67 @@ def get_min_argument_count(ctx: MethodContext | FunctionContext) -> int:
     return sum(not kind.is_star() for kinds in ctx.arg_kinds for kind in kinds)
 
 
+class DjangoModel(NamedTuple):
```
sobolevn, line 197: Why not a `@dataclass`?

sobolevn (commented):
Great stuff, one question :)

mypy_django_plugin/lib/helpers.py
UnknownPlatypus: I've followed what we seem to be using for these little utils and it's also what mypy uses extensively internally.

I suppose because it's more lightweight, and immutable by default which makes it a good candidate for simple data grouping like that

asottile-sentry:
I'm seeing a weird error here that I think is related?  it seems to not be processing reverse model dependencies correctly?

```
src/sentry/releases/models.py:34: error: Cannot resolve keyword 'releasecommit' into field. Choices are: author, author_id, commitfilechange, date_added, date_created, id, key, message, organization_id, repository_id  [misc]
```

- https://github.com/getsentry/sentry/blob/e1829e8cbd0fe4179e137afbbee23b06d35d3eee/src/sentry/releases/models.py#L34
- https://github.com/getsentry/sentry/blob/e1829e8cbd0fe4179e137afbbee23b06d35d3eee/src/sentry/models/releasecommit.py#L20

___

also seems to not be processing `_id` automatic conversion:

```
src/sentry/users/models/user_option.py:62: error: Cannot resolve keyword 'project' into field. Choices are: id, key, organization_id, project_id, user, user_id, value  [misc]
```

- https://github.com/getsentry/sentry/blob/e1829e8cbd0fe4179e137afbbee23b06d35d3eee/src/sentry/users/models/user_option.py#L62
- https://github.com/getsentry/sentry/blob/e1829e8cbd0fe4179e137afbbee23b06d35d3eee/src/sentry/users/models/user_option.py#L204

asottile-sentry:
also maybe missing ManyToMany support?

```
src/sentry/uptime/subscriptions/subscriptions.py:520: error: Invalid field name "data_sources" in select_related lookup. Choices are: owner_team, project, workflow_condition_group  [misc]
```

- https://github.com/getsentry/sentry/blob/e1829e8cbd0fe4179e137afbbee23b06d35d3eee/src/sentry/uptime/subscriptions/subscriptions.py#L520
- https://github.com/getsentry/sentry/blob/e1829e8cbd0fe4179e137afbbee23b06d35d3eee/src/sentry/workflow_engine/models/detector.py#L55

sobolevn:
Yes, seems like a bug. Thank you! I opened https://github.com/typeddjango/django-stubs/issues/2817

PR #2786, opened 2025-08-26
Declare `Prefetch` as generic and do specialization in the plugin

UnknownPlatypus:
# I have made things!

Followup to #2779 to handle more `Prefetch` usages using a generic class over the prefetched `queryset` and `to_attr` value (with plugin specialization to LiteralString).

We now:
- handle calls defined outside of `prefetch_related` calls
- handle `Prefetch` classes created from external function or variable assignments
- correctly infer for case with the `to_attr` attribute filled but not the `queryset` attribute.
```diff
reveal_type(
    Article.objects.prefetch_related(
-     models.Prefetch("tags", to_attr="tags")).get().tags)  # N: Revealed type is "builtins.list[Any]"
+     models.Prefetch("tags", to_attr="tags")).get().tags)  # N: Revealed type is "builtins.list[myapp.models.Tag]"
    )
)
```

I've also expanded the tests with a test for annotated prefetched queryset

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -7,7 +7,7 @@ from django.db.models.fields.reverse_related import ForeignObjectRel
 from mypy.checker import TypeChecker
 from mypy.nodes import ARG_NAMED, ARG_NAMED_OPT, CallExpr, Expression
 from mypy.plugin import FunctionContext, MethodContext
-from mypy.types import AnyType, Instance, TupleType, TypedDictType, TypeOfAny, get_proper_type
+from mypy.types import AnyType, Instance, LiteralType, TupleType, TypedDictType, TypeOfAny, get_proper_type
 from mypy.types import Type as MypyType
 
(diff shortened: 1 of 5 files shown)


mypy_django_plugin/transformers/querysets.py
```diff
@@ -318,20 +318,38 @@ def extract_proper_type_queryset_values(ctx: MethodContext, django_context: Djan
...
-        # TODO: Infer the model type from the lookup in `Prefetch(lookup=..., to_attr=...)`
-        return None
     try:
         qs_type = get_proper_type(api.expr_checker.accept(queryset_expr))
     except Exception:
         return None
     if isinstance(qs_type, Instance):
-        return helpers.extract_model_type_from_queryset(qs_type, api)
+        return qs_type
+        # return helpers.extract_model_type_from_queryset(qs_type, api)
```
sobolevn: why a comment?

UnknownPlatypus: That's a woopsie, fixed

sobolevn (approved):
Thanks again!

PR #2788, opened 2025-09-01
Auto inherit from `TypedModelMeta`

UnknownPlatypus:
# I have made things!
(Review per commit advised, the first one is a minor cleanup)

Implement the suggestion in https://github.com/typeddjango/django-stubs/pull/1375#issuecomment-1440572037 providing automatic type-checking of Django models `Meta` class by artificially inserting `TypedModelMeta` in the mro.

The error message might be slightly confusing since the user did not add an explicit base class but I think is comprehensible enough and that the value outweigh the slightly weird message. For ex:

> Incompatible types in assignment (expression has type "int", base class "TypedModelMeta" defined the type as "bool")

I've also tested this change on a rather large Django codebase (1M python LoC) adding errors that were all caught properly.

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -262,8 +262,12 @@ class InjectAnyAsBaseForNestedMeta(ModelClassInitializer):
                 pass
     with
         class MyModel(models.Model):
-            class Meta(Any):
+            class Meta(TypedModelMeta):
                 pass
+
+    to provide proper typing of attributes in Meta inner classes.
+
+    If TypedModelMeta is not available, fallback to Any as a base
     to get around incompatible Meta inner classes for different models.
     """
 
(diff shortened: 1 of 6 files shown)


mypy_django_plugin/transformers/models.py
```diff
@@ -273,6 +277,12 @@ def run(self) -> None:
             return None
         meta_node.fallback_to_any = True
 
+        typed_model_meta_info = self.lookup_typeinfo(fullnames.TYPED_MODEL_META_FULLNAME)
+        if typed_model_meta_info and not meta_node.has_base(fullnames.TYPED_MODEL_META_FULLNAME):
+            # Insert TypedModelMeta just before `object` to leverage mypy's class-body semantic analysis.
+            meta_node.mro.insert(-1, typed_model_meta_info)
```
sobolevn, line 283: Do we need to add this base class to `ClassDef` as well?

sobolevn (commented):
Please, see why test is failing

mypy_django_plugin/transformers/models.py
UnknownPlatypus: I don't think because we could endup with multiple `TypedModelMeta` in the mro for multiple inheritance cases.

UnknownPlatypus:
> Please, see why test is failing

I think I found the issue, this was because `self._new_dependency("django_stubs_ext.db.models")` was not called for dangling abstract django models because of how `self.django_context.model_modules` collect modules (It uses django logic to get concrete models, and then walk the mro to find abstract ones, hence missing dangling abstract models)

So depending on how mypy was resolving the module deps, we could sometime check a file before `"django_stubs_ext.db.models"` was checked, hence the flaky test. And yeah, `abstract = 7` basically mean `abstract = True`.

I've updated the test to have concrete models at some point. Supporting dangling abstract model will add too much complexity, I don't think it's worth it

sobolevn (approved):
Great stuff! Thanks again!

PR #2789, opened 2025-09-01
Support star arg in `prefetch_related` calls

UnknownPlatypus:
# I have made things!

Expand #2786 to support cases where the `Prefetch` objects are passed using a star arg. 
I had something very similar to this at work:

```python
def _get_prefetches() -> tuple[
    Prefetch[QuerySet[Tag], Literal["every_tags"]],
    Prefetch[QuerySet[Tag], Literal["every_tags2"]],
]:
    return (
        Prefetch("tags", Tag.objects.all(), to_attr="every_tags"),
        Prefetch("tags", Tag.objects.all(), to_attr="every_tags"),
    )

##
my_model = (
    MyModel.objects
    .prefetch_related(*_get_prefetches(agent_main_group))
    .get()
)
```


sobolevn (approved):
Thank you!

PR #2791, opened 2025-09-02
Handle `.prefetch_related(..., to_attr="foo")` conflict with `.annotate(foo=...)`

UnknownPlatypus:
# I have made things!

Fixes one of the issue identified in the initial implementation (#2779).

The current fix does not correctly handle chained `prefetch_related` mapping to the same `to_attr` ie:

```python
reveal_type(
    Article.objects # N: Revealed type is "builtins.list[myapp.models.Group]"  <--- incorrect, should be `list[Article]`
     .prefetch_related(Prefetch("tags", Group.objects.all(), to_attr="foo"))
     .prefetch_related(Prefetch("related_articles", Article.objects.all(), to_attr="foo"))
     .get().foo
)
```

This should be possible to handle but require expanding the hacky usage of [`ExtraAttrs`](https://github.com/typeddjango/django-stubs/blob/cbe9bd63da1264ada0f3ae1ec1f1f52fc3c798b2/mypy_django_plugin/transformers/models.py#L1132-L1150) to record if the annotation come from a `.annotate()` or a `.prefetch_related()` call.

I'm not sure it's worth the complexity given the unlikeliness of this happening. Let me know what you think

### DIFF
--- a/tests/typecheck/managers/querysets/test_prefetch_related.yml
+++ b/tests/typecheck/managers/querysets/test_prefetch_related.yml
@@ -163,13 +163,15 @@
             User.objects
             .annotate(foo=F("username"))
             .prefetch_related(Prefetch("groups", Group.objects.all(), to_attr="foo"))
-            .get().foo
+            .get()
         )
-        reveal_type(annotate_into_prefetch)  # N: Revealed type is "Any"
+        reveal_type(annotate_into_prefetch)  # N: Revealed type is "django.contrib.auth.models.User@AnnotatedWith[TypedDict({'foo': Any})]"
         prefetch_into_annotate = (
             User.objects
             .prefetch_related(Prefetch("groups", Group.objects.all(), to_attr="foo"))
             .annotate(foo=F("username"))
-            .get().foo
+            .get()
         )
-        reveal_type(prefetch_into_annotate)  # N: Revealed type is "Any"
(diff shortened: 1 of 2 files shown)


tests/typecheck/managers/querysets/test_prefetch_related.yml
```diff
@@ -151,3 +151,25 @@
...
+
+        # When mixing `.annotate(foo=...)` and `prefetch_related(Prefetch(...,to_attr=foo))`
+        # The last annotate in the chain takes precedence (even if it is prior to the prefetch_related)
+        annotate_into_prefetch = (
+            User.objects
+            .annotate(foo=F("username"))
+            .prefetch_related(Prefetch("groups", Group.objects.all(), to_attr="foo"))
+            .get().foo
+        )
+        reveal_type(annotate_into_prefetch)  # N: Revealed type is "Any"
```
sobolevn: Is there any way to get something other than `Any`? Because `Any` can be returned not only from the correct case.

UnknownPlatypus: Very nice catch, I forgot the installed_apps so this was testing nothing and passing because we assert Any.
Thanks, I've change the code to asser on the `.get()` instead so we see the `WithAnnotation`

sobolevn (approved):
Thank you!

PR #2792, opened 2025-09-02
Improve `.select_related` stubs

UnknownPlatypus:
# I have made things!

Use overloads to represent the valid expected values passed to a `select_related` call.
It only actually support these 3 cases:

```python
Book.objects.select_related()                      # Select every FK's
Book.objects.select_related(None)                  # Reset any selected fields
Book.objects.select_related("author", "category")  # Select these FK's explicitly
```

ref: https://docs.djangoproject.com/en/5.2/ref/models/querysets/#select-related


sobolevn (approved):
Awesome! Thank you!

PR #2793, opened 2025-09-02
Improve `prefetch_related` types with an overload

UnknownPlatypus:
# I have made things!

Similar to #2792 but for `prefetch_related` on `QuerySet` and `RawQueryset` (with small improvements on `def raw()`) 

It accepts the same types of inputs, the only difference behing that a bare `prefetch_related()` does nothing compared to a bare `select_related` prefetching every possible relations. 

ref: https://docs.djangoproject.com/en/5.2/ref/models/querysets/#prefetch-related


django-stubs/db/models/manager.pyi
```diff
@@ -125,16 +125,17 @@ class Manager(BaseManager[_T]):
...
     @overload
     def select_related(self, *fields: str) -> QuerySet[_T]: ...
-    def prefetch_related(self, *lookups: Any) -> QuerySet[_T]: ...
+    @overload
+    def prefetch_related(self, clear: None, /) -> QuerySet[_T]: ...
+    @overload
+    def prefetch_related(self, *lookups: str | Prefetch[_PrefetchedQuerySetT, _ToAttrT]) -> QuerySet[_T]: ...
     def annotate(self, *args: Any, **kwargs: Any) -> QuerySet[_T]: ...
     def alias(self, *args: Any, **kwargs: Any) -> QuerySet[_T]: ...
     def order_by(self, *field_names: Any) -> QuerySet[_T]: ...
```
sobolevn:
```suggestion
    def order_by(self, *field_names: str) -> QuerySet[_T]: ...
```

or no?

django-stubs/db/models/query.pyi
```diff
@@ -168,16 +168,17 @@ class QuerySet(Generic[_Model, _Row], Iterable[_Row], Sized):
...
     @overload
     def select_related(self, *fields: str) -> Self: ...
-    def prefetch_related(self, *lookups: Any) -> Self: ...
+    @overload
+    def prefetch_related(self, clear: None, /) -> Self: ...
+    @overload
+    def prefetch_related(self, *lookups: str | Prefetch[_PrefetchedQuerySetT, _ToAttrT]) -> Self: ...
     def annotate(self, *args: Any, **kwargs: Any) -> Self: ...
     def alias(self, *args: Any, **kwargs: Any) -> Self: ...
     def order_by(self, *field_names: Any) -> Self: ...
```
sobolevn: and here?

tests/typecheck/managers/querysets/test_basic_methods.yml
```diff
@@ -18,7 +18,7 @@
         reveal_type(qs.none()) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.Blog, myapp.models.Blog]"
         reveal_type(qs.update_or_create()) # N: Revealed type is "tuple[myapp.models.Blog, builtins.bool]"
         reveal_type(qs.explain()) # N: Revealed type is "builtins.str"
-        reveal_type(qs.raw(qs.explain())) # N: Revealed type is "django.db.models.query.RawQuerySet[Any]"
+        reveal_type(qs.raw(qs.explain())) # N: Revealed type is "django.db.models.query.RawQuerySet[myapp.models.Blog]"
```
sobolevn, line 21: Is that correct? It returns a string AFAIK https://docs.djangoproject.com/en/5.2/ref/models/querysets/#explain

django-stubs/db/models/manager.pyi
UnknownPlatypus: I'm updating a few more in a followup pr (`order_by` / `defer` / `only` / `distinct`)

tests/typecheck/managers/querysets/test_basic_methods.yml
UnknownPlatypus: This will certainly not work but is correct typing-wise, `qs.explain()` is a `str` and is passed to `qs.raw(...)` which accept a `str`

django-stubs/db/models/manager.pyi
UnknownPlatypus: (this would be `str | OrderBy` here)

tests/typecheck/managers/querysets/test_basic_methods.yml
UnknownPlatypus: (you might have missed a parenthesis, the return of qs.expain is not what is revealed here)

sobolevn:
Please, resolve conflicts

UnknownPlatypus:
> Please, resolve conflicts
Done !

sobolevn:
And we have conflicts agains 😅

UnknownPlatypus:
> And we have conflicts agains 😅

Yeah woopsie, always not sure how to properly handle stacked pr (thats what i wanted to do to avoid conflict)

mypy_django_plugin/transformers/querysets.py
```diff
@@ -385,6 +385,9 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
         and ctx.args
         and ctx.arg_types
         and ctx.arg_types[0]
+        # Only process the correct overload, i.e.
+        #     def prefetch_related(self, *lookups: str | Prefetch[_PrefetchedQuerySetT, _ToAttrT]) -> Self: ...
+        and "lookups" in ctx.callee_arg_names
```
sobolevn: I think it would be better to use arg kind here, because name can be changed: since it is `*args`, it is not a requirement to keep this exact name.

UnknownPlatypus: The issue is that in both case `arg_kinds` is the same: `[[<ArgKind.ARG_POS: 0>]]`

Lookups overload:
image

None overload:
image

But maybe  `None not in ctx.callee_arg_names` is more reliable ?

sobolevn: yes, let's do this :)

UnknownPlatypus:
```suggestion
        and None not in ctx.callee_arg_names
```

sobolevn (approved):

PR #2794, opened 2025-09-03
Detect invalid `to_attr` target (alredy defined fields)

UnknownPlatypus:
# I have made things!

Trying to do that will either cause runtime errors or silently do nothing depending on the case.
The added branch is mutually exclusive with the annotate one because annotated fields are not stored on the type but on the `Instance` extra_attrs.

### DIFF
--- a/tests/typecheck/managers/querysets/test_prefetch_related.yml
+++ b/tests/typecheck/managers/querysets/test_prefetch_related.yml
@@ -172,7 +172,7 @@
     installed_apps:
         - django.contrib.auth
 
--   case: prefetch_related_conflict_with_annotate
+-   case: prefetch_related_to_attr_conflict_with_annotate
     main: |
         from django.db.models import Prefetch, F
         from django.contrib.auth.models import User, Group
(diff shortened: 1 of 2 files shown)


tests/typecheck/managers/querysets/test_prefetch_related.yml
```diff
@@ -195,3 +195,40 @@
...
     installed_apps:
         - django.contrib.auth
+
+
+-   case: prefetch_related_to_attr_conflict_with_model_attr
+    main: |
+        from django.db.models import Prefetch, F
+        from myapp.models import Article, Tag
+
+        Article.objects.prefetch_related(Prefetch("tags", Tag.objects.all(), to_attr="pk"))  # E: Attribute "pk" already defined on "Article"  [no-redef]
```
sobolevn, line 205: Let's test `Prefetch` conflicts maybe?

Something like 

```python
   Prefetch("tags", Tag.objects.all(), to_attr="new_attr"),
   Prefetch("tags", Tag.objects.all(), to_attr="new_attr"),
)
```

or 

```python
   Prefetch("tags", Tag.objects.all(), to_attr="new_attr"),
).prefetch_related(
   Prefetch("tags", Tag.objects.all(), to_attr="new_attr"),
)
```

UnknownPlatypus: This is not supported yet, @sobolevn do you mind if I defer this to another PR ? 

I would like to tackle other similar issues that are currently only caught at runtime:

```python
User.objects.prefetch_related("groups", Prefetch("groups")) # Ok
User.objects.prefetch_related("groups", Prefetch("groups", Group.objects.filter(name="test"))) # E: Not OK!
User.objects.prefetch_related(Prefetch("groups", Group.objects.filter(name="test")), "groups") # E: Not OK! (but OK at runtime :/)
```

cf django test for this kind of runtime error: 
https://github.com/django/django/blob/0a67611b81567687cd24d2be1c0e390931f06262/tests/prefetch_related/tests.py#L574-L619

sobolevn: Sure! Let's do it later

UnknownPlatypus: Also kinda related to https://github.com/typeddjango/django-stubs/pull/2791/files, do you think we would like hard errors on conflicting attr added via annotate + prefetch_related or `prefetch_related` chain ? ie:
```python
   Prefetch("tags", Tag.objects.all(), to_attr="new_attr"),
).prefetch_related(
   Prefetch("tags", Tag.objects.all(), to_attr="new_attr"),  # E: Attribute "new_attr" already defined on "Article"  [no-redef]
)
```
and 
```python
Article.objects.annotate(new_attr=F("tags")).prefetch_related(
   Prefetch("tags", Tag.objects.all(), to_attr="new_attr"),  # E: Attribute "new_attr" already defined on "Article"  [no-redef]
)
   Prefetch("tags", Tag.objects.all(), to_attr="new_attr"),
).annotate(new_attr=F("tags"))  # E: Attribute "new_attr" already defined on "Article"  [no-redef]
```

sobolevn: Yes, hard errors seem reasonable here

PR #2795, opened 2025-09-03
Improve type for `order_by` / `defer` / `only` / `distinct` / `order_by` / `explain` / `dates` / `datetimes`

UnknownPlatypus:
# I have made things!

Addresses https://github.com/typeddjango/django-stubs/pull/2793#discussion_r2316907181 and other method that could be improved


UnknownPlatypus:
@sobolevn Not sure why a job go cancelled, can you retry it ? I don't seem to have the permission for that

sobolevn:
@UnknownPlatypus btw, about that failing job :)
You made a lot of great prs!
Снимок экрана 2025-09-03 в 17 19 58

Would you like to join the TypedDjango team? :) 
It would mean that you can restart jobs / close issues / merge PRs / etc.

But, it does not require any extra involment from you, unless you want to :)
Thanks a lot for your hard work!

django-stubs/db/models/manager.pyi
```diff
@@ -108,19 +108,28 @@ class Manager(BaseManager[_T]):
...
     def none(self) -> QuerySet[_T]: ...
     def filter(self, *args: Any, **kwargs: Any) -> QuerySet[_T]: ...
     def exclude(self, *args: Any, **kwargs: Any) -> QuerySet[_T]: ...
     def complex_filter(self, filter_obj: Any) -> QuerySet[_T]: ...
     def count(self) -> int: ...
     async def acount(self) -> int: ...
-    def union(self, *other_qs: Any, all: bool = ...) -> QuerySet[_T]: ...
-    def intersection(self, *other_qs: Any) -> QuerySet[_T]: ...
-    def difference(self, *other_qs: Any) -> QuerySet[_T]: ...
+    def union(self, *other_qs: QuerySet[Model], all: bool = ...) -> QuerySet[_T]: ...
```
sobolevn, line 130:
```suggestion
    def union(self, *other_qs: QuerySet[_T], all: bool = ...) -> QuerySet[_T]: ...
```

I don't think that we can `.union` two different models, can't we?

UnknownPlatypus: We can! (But I would not recommend it because of how disturbing the result is) 

> union(), intersection(), and difference() return model instances of the type of the first QuerySet even if the arguments are QuerySets of other models. Passing different models works as long as the SELECT list is the same in all QuerySets (at least the types, the names don’t matter as long as the types are in the same order). In such cases, you must use the column names from the first QuerySet in QuerySet methods applied to the resulting QuerySet. 

ref: https://docs.djangoproject.com/en/5.2/ref/models/querysets/#union

UnknownPlatypus:
> Would you like to join the TypedDjango team? :)
It would mean that you can restart jobs / close issues / merge PRs / etc.

Would love too, I'm using django a lot at work and on personal projects and would love the story with types to be more smooth (hence the many recent pr's addressing some of the main blockers at work)  

> But, it does not require any extra involment from you, unless you want to :)
Thanks a lot for your hard work!

You're welcome, thanks for the many excellent reviews too!
Let's start small but if the need for extra involvement arise and it helps other maintainers, I'll be happy to help at some point

sobolevn:
Invite sent :)

django-stubs/db/models/manager.pyi
sobolevn: Ok! TIL :)

sobolevn (approved):

PR #2797, opened 2025-09-04
Detect more incompatible `to_attr` targets

UnknownPlatypus:
# I have made things!

Initial work to address https://github.com/typeddjango/django-stubs/pull/2794#discussion_r2318888629 and detect more incompatible `to_attr` targets.

The main changes in behavior are:
- We detect duplicate `to_attr` used in a single `prefetch_related`
- We detect duplicate `to_attr` from chained `prefetch_related`
- `.annotate` now also detect conflict with existing fields

The logic is now shared with the annotate hook, which means we now detect runtime annotate errors like [ValueError: The annotation 'username' conflicts with a field on the model.](https://github.com/django/django/blob/686a8a62ae7faba9c3b17080c3532b821e8cb1f3/django/db/models/query.py#L1734-L1737)

We still don't detect suttle errors like `User.objects.prefetch_related("groups", Prefetch("groups", Group.objects.filter(name="test")))` which is invalid because the queryset changed. 
I'm working on another PR to detect these and more but it's a bit more complicated because we need to track the `lookup` value to do so correctly

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -381,7 +381,7 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
         isinstance(ctx.type, Instance)
         and isinstance((default_return_type := get_proper_type(ctx.default_return_type)), Instance)
         and (api := helpers.get_typechecker_api(ctx))
-        and (model_type := helpers.extract_model_type_from_queryset(ctx.type, api)) is not None
+        and (qs_model := helpers.get_model_info_from_qs_ctx(ctx, django_context)) is not None
         and ctx.args
         and ctx.arg_types
         and ctx.arg_types[0]
@@ -437,11 +437,9 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
                     continue
 
                 if lookup_value := helpers.resolve_string_attribute_value(lookup_expr, django_context):
-                    if django_model := helpers.get_model_info_from_qs_ctx(ctx, django_context):
+                    if qs_model := helpers.get_model_info_from_qs_ctx(ctx, django_context):
                         try:
-                            observed_model_cls = django_context.resolve_lookup_into_field(
-                                django_model.cls, lookup_value
-                            )[1]
+                            observed_model_cls = django_context.resolve_lookup_into_field(qs_model.cls, lookup_value)[1]
                             if model_info := helpers.lookup_class_typeinfo(api, observed_model_cls):
                                 elem_model = Instance(model_info, [])
                         except (FieldError, LookupsAreUnsupported):
@@ -452,11 +450,11 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
             [elem_model if elem_model is not None else AnyType(TypeOfAny.special_form)],
         )
 
-        if model_type.type.get(to_attr_value):
+        if qs_model.typ.type.get(to_attr_value):
             ctx.api.fail(
-                f'Attribute "{to_attr_value}" already defined on "{model_type.type.name}"', ctx.context, code=NO_REDEF
+                f'Attribute "{to_attr_value}" already defined on "{qs_model.typ.type.name}"', ctx.context, code=NO_REDEF
             )
-        elif not (model_type.extra_attrs and to_attr_value in model_type.extra_attrs.attrs):
+        elif not (qs_model.typ.extra_attrs and to_attr_value in qs_model.typ.extra_attrs.attrs):
             # When mixing `.annotate(foo=...)` and `prefetch_related(Prefetch(...,to_attr=foo))`
             # The last annotate in the chain takes precedence (even if it is prior to the prefetch_related)
             # So only add the annotation here if it doesn't exist yet.
@@ -472,7 +470,7 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
         readonly_keys=set(),
     )
 
-    annotated_model = get_annotated_type(api, model_type, fields_dict=fields_dict)
+    annotated_model = get_annotated_type(api, qs_model.typ, fields_dict=fields_dict)
 
     # Keep row shape; if row is a model instance, update it to annotated
     # Todo: consolidate with `extract_proper_type_queryset_annotate` row handling above.
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -437,13 +437,12 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
                     continue
 
                 if lookup_value := helpers.resolve_string_attribute_value(lookup_expr, django_context):
-                    if qs_model := helpers.get_model_info_from_qs_ctx(ctx, django_context):
-                        try:
-                            observed_model_cls = django_context.resolve_lookup_into_field(qs_model.cls, lookup_value)[1]
-                            if model_info := helpers.lookup_class_typeinfo(api, observed_model_cls):
-                                elem_model = Instance(model_info, [])
-                        except (FieldError, LookupsAreUnsupported):
-                            pass
+                    try:
+                        observed_model_cls = django_context.resolve_lookup_into_field(qs_model.cls, lookup_value)[1]
+                        if model_info := helpers.lookup_class_typeinfo(api, observed_model_cls):
+                            elem_model = Instance(model_info, [])
+                    except (FieldError, LookupsAreUnsupported):
+                        pass
 
         value_type = api.named_generic_type(
             "builtins.list",
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -358,14 +358,15 @@ def gather_flat_args(ctx: MethodContext) -> list[tuple[Expression | None, Proper
     arguments when their type is a TupleType with statically known items.
     """
     lookups: list[tuple[Expression | None, ProperType]] = []
+    arg_start_idx = 0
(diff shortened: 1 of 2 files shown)


mypy_django_plugin/transformers/querysets.py
```diff
@@ -358,19 +356,58 @@ def gather_flat_args(ctx: MethodContext) -> list[tuple[Expression | None, Proper
...
     lookups: list[tuple[Expression | None, ProperType]] = []
+    arg_start_idx = 0
     for expr, typ, kind in zip(ctx.args[0], ctx.arg_types[0], ctx.arg_kinds[0], strict=False):
         ptyp = get_proper_type(typ)
         if kind == ARG_STAR:
             # Expand starred tuple items if statically known
             if isinstance(ptyp, TupleType):
-                for item_typ in ptyp.items:
-                    lookups.append((None, get_proper_type(item_typ)))
+                lookups.append((None, get_proper_type(ptyp.items[arg_start_idx])))
```
UnknownPlatypus, line 365: Before that we were returning every type in the tuple `len(ptyp.items)` times (which was a noop before but was now triggering `'Attribute "{attr_name}" already defined on "{model.typ.type.name}"'`

mypy_django_plugin/transformers/querysets.py
```diff
@@ -437,42 +474,32 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
                     continue
 
                 if lookup_value := helpers.resolve_string_attribute_value(lookup_expr, django_context):
-                    if django_model := helpers.get_model_info_from_qs_ctx(ctx, django_context):
```
UnknownPlatypus, line 440: We were already doing this fetch in the guard branch at the top of this function

tests/typecheck/managers/querysets/test_prefetch_related.yml
```diff
@@ -128,12 +130,13 @@
...
-        reveal_type(Article.objects.prefetch_related(first_prefetch, *prefetch_objs, third_prefetch).all()) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.Article@AnnotatedWith[TypedDict({'every_tags': builtins.list[myapp.models.Tag], 'every_tags2': builtins.list[myapp.models.Tag], 'every_tags3': builtins.list[myapp.models.Tag]})], myapp.models.Article@AnnotatedWith[TypedDict({'every_tags': builtins.list[myapp.models.Tag], 'every_tags2': builtins.list[myapp.models.Tag], 'every_tags3': builtins.list[myapp.models.Tag]})]]"
+        reveal_type(Article.objects.prefetch_related(fourth_prefetch, *prefetch_objs, third_prefetch).all()) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.Article@AnnotatedWith[TypedDict({'every_tags4': builtins.list[myapp.models.Tag], 'every_tags': builtins.list[myapp.models.Tag], 'every_tags2': builtins.list[myapp.models.Tag], 'every_tags3': builtins.list[myapp.models.Tag]})], myapp.models.Article@AnnotatedWith[TypedDict({'every_tags4': builtins.list[myapp.models.Tag], 'every_tags': builtins.list[myapp.models.Tag], 'every_tags2': builtins.list[myapp.models.Tag], 'every_tags3': builtins.list[myapp.models.Tag]})]]"
```
UnknownPlatypus, line 139: This test started to fail because `first_prefetch` was passed twice so it's a nice catch

tests/typecheck/managers/querysets/test_prefetch_related.yml
```diff
@@ -169,34 +173,34 @@
...
+        - django.contrib.auth
     main: |
         from django.db.models import Prefetch, F
         from django.contrib.auth.models import User, Group
 
         # When mixing `.annotate(foo=...)` and `prefetch_related(Prefetch(...,to_attr=foo))`
         # The last annotate in the chain takes precedence (even if it is prior to the prefetch_related)
         annotate_into_prefetch = (
-            User.objects
+            User.objects # E: Attribute "foo" already defined on "User@AnnotatedWith"  [no-redef]
```
UnknownPlatypus: I'm not super happy with the format `User@AnnotatedWith`, I would prefer: `User@AnnotatedWith[TypedDict({'foo': Any}]` but I did not found a way to trivially get that, str(...) returns the very verbose `django.contrib.auth.models.User@AnnotatedWith[TypedDict({'foo': Any})]`.

Do you think it's worth adding an utility to better format this ? Or did I missed a mypy util ?

tests/typecheck/managers/querysets/test_prefetch_related.yml
```diff
@@ -209,8 +213,8 @@
         Article.objects.prefetch_related(Prefetch("tags", Tag.objects.all(), to_attr="my_classmethod"))  # E: Attribute "my_classmethod" already defined on "Article"  [no-redef]
         Article.objects.prefetch_related(Prefetch("tags", Tag.objects.all(), to_attr="my_staticmethod"))  # E: Attribute "my_staticmethod" already defined on "Article"  [no-redef]
 
-    installed_apps:
-        - myapp
+        Tag.objects.prefetch_related(Prefetch("articles", Article.objects.all(), to_attr="articles"))  # E: Attribute "articles" already defined on "Tag"  [no-redef]
```
UnknownPlatypus: test reverse many-to-many too for good measure

tests/typecheck/managers/querysets/test_prefetch_related.yml
```diff
@@ -252,3 +256,23 @@
...
+           Prefetch("groups", Group.objects.all(), to_attr="new_attr"),
+        )
+        User.objects.prefetch_related( # E: Attribute "new_attr" already defined on "User@AnnotatedWith"  [no-redef]
+           Prefetch("groups", Group.objects.all(), to_attr="new_attr"),
+        ).prefetch_related(
+           Prefetch("groups", Group.objects.all(), to_attr="new_attr"),
+        )
+        User.objects.prefetch_related("groups", Prefetch("groups")) # Ok
+        User.objects.prefetch_related("groups", Prefetch("groups", Group.objects.filter(name="test"))) # TODO: Should be Not OK!
+        User.objects.prefetch_related(Prefetch("groups", Group.objects.filter(name="test")), "groups") # TODO: Should be Not OK! (but OK at runtime :/)
```
UnknownPlatypus, line 278: Will address in a followup pr

mypy_django_plugin/transformers/querysets.py
```diff
@@ -358,19 +356,58 @@ def gather_flat_args(ctx: MethodContext) -> list[tuple[Expression | None, Proper
...
+    """
+    Check if adding `attr_name` would conflict with existing symbols on `model`.
+
+    Args:
+        - model: The Django model being analyzed
+        - attr_name: The name of the attribute to be added
+        - fields: A mapping of field names to types currently being added to the model
+    """
+    is_conflicting_attr_value = bool(
+        # 1. Conflict with another symbol on the model.
```
sobolevn, line 385: Great docs!

mypy_django_plugin/transformers/querysets.py
```diff
@@ -358,19 +356,58 @@ def gather_flat_args(ctx: MethodContext) -> list[tuple[Expression | None, Proper
...
+        # Ex:
+        #     User.objects.prefetch_related(
+        #        Prefetch("groups", Group.objects.filter(name="test"), to_attr="new_attr"),
+        #        Prefetch("groups", Group.objects.all(), to_attr="new_attr"), # E: Not OK!
+        #     )
+        or (new_attrs is not None and attr_name in new_attrs)
+    )
+    if is_conflicting_attr_value:
+        ctx.api.fail(
+            f'Attribute "{attr_name}" already defined on "{model.typ.type.name}"',
```
```suggestion
            f'Attribute "{attr_name}" already defined on "{model.typ.type.fullname}"',
```

maybe? https://github.com/python/mypy/blob/39427835ea4ab9e5735e7e5956063026d1bcc4bb/mypy/nodes.py#L3537-L3539

tests/typecheck/managers/querysets/test_prefetch_related.yml
sobolevn: Added a suggestion

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: I believe we would end up with the fully qualified name ie:
 `django.contrib.auth.models.User@AnnotatedWith[TypedDict({'foo': Any})].` wichi quite verbose but maybe better ? (especially for annotated models)

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: Thanks ! I like to put some example for these because it's often not that easy to find the reasons from tests only

mypy_django_plugin/transformers/querysets.py
sobolevn: It is always better :)

sobolevn (approved):

asottile-sentry:
I believe this might not be correct for the `.annotate(...)` case

here's an example from sentry:

```
src/sentry/api/serializers/models/rule.py:207: error: Attribute "date_added" already defined on "sentry.models.rulefirehistory.RuleFireHistory"  [no-redef]
```

https://github.com/getsentry/sentry/blob/e1829e8cbd0fe4179e137afbbee23b06d35d3eee/src/sentry/api/serializers/models/rule.py#L209

(there's a few other examples in sentry as well where particular aggregations are used to produce a value with the same name as an existing field when making a grouped query)

sobolevn:
Hm, interesting, I haven't consider this use-case. So, basically, if we annotate an some existing `IntegerField` field with `int` value, there's no *type* error here. So, basically, we should instead detect type mismatches in annotated fields and report them.

sobolevn:
I opened https://github.com/typeddjango/django-stubs/issues/2818

PR #2799, opened 2025-09-05
Update mypy configuration

ngnpope:
I went through the mypy configuration and cleaned up a number of things:

- Removed a number of unnecessary options, e.g. already enabled by `strict`
- Removed `allow_redefinition` [^1]
- Enabled a number of additional error codes:
  - [x] `deprecated`
  - [x] `exhaustive-match`
  - [ ] `explicit-override` [^2]
  - [x] `ignore-without-code`
  - [ ] ~`mutable-override`~ [^3]
  - [x] `possibly-undefined`
  - [x] `redundant-expr`
  - [x] `redundant-self`
  - [x] `truthy-bool`
  - [x] `truthy-iterable`
  - [x] `unimported-reveal` [^4]
  - [x] `unused-awaitable`

[^1]: Perhaps we can enable `allow_redefinition_new` later or just wait for it to be the default...
[^2]: This was a bit more noisy in the stubs, so better done in a separate pull request.
[^3]: This just doesn't play nicely with Django, so not worth supporting.
[^4]: We should remove patching `reveal_type()` and `reveal_locals()` to `builtins` in `django_stubs_ext`, especially as `typing.reveal_type()` is available. With the addition of `unimported-reveal` in this pull request, we'll catch any stray uses in django-stubs so we can't make a bad release. Something for a follow up.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -22,6 +22,7 @@ enable_error_code =
     possibly-undefined,
     redundant-expr,
     redundant-self,
+    truthy-bool,
 
 show_traceback = true
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -19,6 +19,7 @@ enable_error_code =
     deprecated,
     exhaustive-match,
     ignore-without-code,
+    possibly-undefined,
 
 show_traceback = true
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -20,6 +20,7 @@ enable_error_code =
     exhaustive-match,
     ignore-without-code,
     possibly-undefined,
+    redundant-expr,
 
 show_traceback = true
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -18,6 +18,7 @@ disable_error_code = empty-body
 enable_error_code =
     deprecated,
     exhaustive-match,
+    ignore-without-code,
 
 show_traceback = true
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -6,14 +6,11 @@ incremental = true
 
 # Strictness:
 allow_redefinition = true
-check_untyped_defs = true
 # TODO: add type args to all generics
 disallow_any_generics = false
 # TODO: fix `Any` subclassing in `typeshed/builtins.pyi`
 disallow_subclassing_any = false
-ignore_missing_imports = false
 strict = true
-strict_bytes = true
 local_partial_types = true
 warn_unreachable = true
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -24,6 +24,7 @@ enable_error_code =
     redundant-self,
     truthy-bool,
     truthy-iterable,
+    unimported-reveal,
 
 show_traceback = true
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -15,7 +15,9 @@ local_partial_types = true
 warn_unreachable = true
 
 disable_error_code = empty-body
-enable_error_code = deprecated
+enable_error_code =
+    deprecated,
+    exhaustive-match,
 
 show_traceback = true
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -23,11 +23,6 @@ plugins =
     mypy_django_plugin.main,
     mypy.plugins.proper_plugin
 
-# Ignore incomplete hints in 3rd party stubs:
-[mypy-yaml.*]
-disallow_untyped_defs = false
-disallow_incomplete_defs = false
-
 # Our settings:
 [mypy.plugins.django-stubs]
 django_settings_module = scripts.django_tests_settings
--- a/mypy.ini
+++ b/mypy.ini
@@ -23,6 +23,7 @@ enable_error_code =
     redundant-expr,
     redundant-self,
     truthy-bool,
+    truthy-iterable,
 
 show_traceback = true
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -25,6 +25,7 @@ enable_error_code =
     truthy-bool,
     truthy-iterable,
     unimported-reveal,
+    unused-awaitable,
 
 show_traceback = true
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -21,6 +21,7 @@ enable_error_code =
     ignore-without-code,
     possibly-undefined,
     redundant-expr,
+    redundant-self,
 
 show_traceback = true
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -5,7 +5,6 @@
 incremental = true
 
 # Strictness:
-allow_redefinition = true
 # TODO: add type args to all generics
 disallow_any_generics = false
 # TODO: fix `Any` subclassing in `typeshed/builtins.pyi`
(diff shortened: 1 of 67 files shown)


ngnpope:
_(Converted to draft... Forgot to rebase!)_

ngnpope:
Rebased.

mypy_django_plugin/django/context.py
```diff
@@ -256,7 +256,7 @@ def get_expected_types(self, api: TypeChecker, model_cls: type[Model], *, method
 
                     expected_types[field_name] = model_set_type
 
-            elif contenttypes_in_apps and isinstance(field, GenericForeignKey):
+            elif contenttypes_in_apps and isinstance(field, GenericForeignKey):  # type: ignore[possibly-undefined]
```
sobolevn: We should not ignore this rule, unless there's a bug in mypy. It might actually be undefined :)

mypy_django_plugin/transformers/models.py
```diff
@@ -317,7 +317,7 @@ class AddPrimaryKeyAlias(AddDefaultPrimaryKey):
     def run_with_model_cls(self, model_cls: type[Model]) -> None:
         # We also need to override existing `pk` definition from `stubs`:
         auto_field = model_cls._meta.pk
-        if auto_field:
+        if auto_field:  # type: ignore[truthy-bool]
```
sobolevn: Should this be fixed instead? Or can we remove this `if`?

mypy_django_plugin/transformers/querysets.py
```diff
@@ -253,7 +253,7 @@ def extract_proper_type_queryset_annotate(ctx: MethodContext, django_context: Dj
             )
         elif isinstance(original_row_type, TupleType):
             fallback: Instance = original_row_type.partial_fallback
```
```suggestion
            fallback = original_row_type.partial_fallback
```

sobolevn (commented):
Thank you for this work! It is massive!

mypy_django_plugin/transformers/models.py
ngnpope: I think the issue here is that `Options.pk` can be `None` at certain points. Attempting to remove this branch causes mypy to crash. We have it typed as `Field` now which is why this violation occurs, but it used to be `Field | None` until e3f5b3d60.

The simple solution is to change this to check `if auto_field is not None:`.

mypy_django_plugin/django/context.py
ngnpope: Yeah, so this one is a bit of a mess...

```python
        contenttypes_in_apps = self.apps_registry.is_installed("django.contrib.contenttypes")
        if contenttypes_in_apps:
            from django.contrib.contenttypes.fields import GenericForeignKey
        ...
            elif contenttypes_in_apps and isinstance(field, GenericForeignKey):
                ...
        ...
```

It's protected by the check on `contenttypes_in_apps` at runtime, but mypy doesn't realise this and thinks that `GenericForeignKey` might still be undefined event though it's going to be imported when `contenttypes_in_apps` is true...

I'll have a play, but I suspect that we'll just be trading this for `[assignment,misc]`.

ngnpope: In this case, I'm not sure that the cure is better than the disease:

```diff
--- mypy_django_plugin/django/context.py
+++ mypy_django_plugin/django/context.py
@@ -199,9 +199,15 @@ class DjangoContext:
         raise ValueError("No primary key defined")
 
     def get_expected_types(self, api: TypeChecker, model_cls: type[Model], *, method: str) -> dict[str, MypyType]:
-        contenttypes_in_apps = self.apps_registry.is_installed("django.contrib.contenttypes")
-        if contenttypes_in_apps:
-            from django.contrib.contenttypes.fields import GenericForeignKey
+        class FakeGenericForeignKey:
+            pass
+
+        _GenericForeignKey: type[GenericForeignKey | FakeGenericForeignKey]
+
+        if self.apps_registry.is_installed("django.contrib.contenttypes"):
+            from django.contrib.contenttypes.fields import GenericForeignKey as _GenericForeignKey
+        else:
+            _GenericForeignKey = FakeGenericForeignKey
 
         expected_types = {}
         # add pk if not abstract=True
@@ -256,7 +262,7 @@ class DjangoContext:
 
                     expected_types[field_name] = model_set_type
 
-            elif contenttypes_in_apps and isinstance(field, GenericForeignKey):  # type: ignore[possibly-undefined]
+            elif isinstance(field, _GenericForeignKey):
                 # it's generic, so cannot set specific model
                 field_name = field.name
                 gfk_info = helpers.lookup_class_typeinfo(api, field.__class__)
```

I haven't pushed up this change, but have fixed the others. Feel free to apply this if you think it's worth it! 😅

tests/typecheck/test_settings.yml
```diff
@@ -12,7 +12,7 @@
         from base import *
         SECRET_KEY = 112233
         NUMBERS = ['one', 'two']
-        DICT = {}  # type: ignore
+        DICT = {}  # type: ignore[var-annotated]
```
UnknownPlatypus, line 16: Oh yeah we don't have ruff checks in these yml tests sadly :/ 
It should have been caught by https://docs.astral.sh/ruff/rules/blanket-type-ignore/

mypy_django_plugin/django/context.py
UnknownPlatypus: Could this be:

```suggestion
            elif contenttypes_in_apps:
                from django.contrib.contenttypes.fields import GenericForeignKey
                if isinstance(field, GenericForeignKey):
```

and we remove the if branch above ?

sobolevn: What about:

```python
elif contenttypes_in_apps:
    from django.contrib.contenttypes.fields import GenericForeignKey
    if isinstance(field, GenericForeignKey):
        ...
```

?

sobolevn: Ha! Perfect timing! 😆

tests/typecheck/test_settings.yml
UnknownPlatypus: (we also have a bunch of unused imports in these for sure)

sobolevn: It is also caught with `ignore-without-code` :)

mypy.ini
```diff
@@ -1,36 +1,33 @@
 # Regular configuration file (can be used as base in other projects, runs in CI)
 
 [mypy]
-# Modified in `tests.yml`
-incremental = true
```
sobolevn, line 5: Btw, I think that we need this to check both modes.

mypy_django_plugin/django/context.py
ngnpope: D'oh. That's much better.

mypy.ini
ngnpope: Can you explain? It's enabled by default, so having it in the config like this makes no difference AFAICT. We use `--no-incremental` to disable it when running mypy on the stubs and setting it to `false` in the necessary tests should still work fine.

sobolevn: The comment itself helps, so I would prefer to keep this one :)
This setting also caused a lot of trouble in the past.

ngnpope: Restored.

mypy_django_plugin/django/context.py
ngnpope: Changed.

ngnpope:
Updated.

sobolevn (approved):
Great stuff! Thanks a lot!

PR #2802, opened 2025-09-06
Updated `.pre-commit-config` and `CONTRIBUTING.MD` to Python 3.13

### LINKED ISSUE #2801: Pre-commits and CONTRIBUTING.md still using Python 3.12
I noticed two related Python version issues when setting up the repo:

### 1. pre-commit config
The current `.pre-commit-config.yaml` pins some hooks to `language_version: python3.12`.  
On a system with Python 3.13 (the latest stable release), running `pre-commit` fails with:

```bash
RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.12'
stderr: (none)
```
This forces contributors to install Python 3.12 even if they already have 3.13.

**Proposal:** Update those hooks to either:
- `language_version: python3.13` (explicit), or
- `language_version: python3` (portable, always uses system default).

---

### 2. CONTRIBUTING.md
The docs currently state:

> Important: right now we only run stubtest on Python 3.12 (because it is the latest released version at the moment)

Since Python 3.13 is now the latest release, this should be updated.  
Ideally we should also clarify if `stubtest` should always track “latest stable Python” going forward.

---

Happy to open a PR for both the `.pre-commit-config.yaml` and `CONTRIBUTING.md` and fix these

CodingWithSaksham:
# I have made things!

- Updated `.pre-commit-config.yaml` to drop  

```yaml
default_language_version:
  python: python3.12
```

as it’s redundant. `django-stubs` uses hooks that are already compatible with the latest stable Python versions.  
Keeping this pinned was only causing issues for new contributors, since `pre-commit install` failed with a `RuntimeError` if Python 3.12 wasn’t available.  

- Updated `CONTRIBUTING.md` to reflect Python 3.13 (the current stable release) instead of Python 3.12:  

> Important: right now we only run stubtest on Python 3.12 (because it is the latest released version at the moment), any other versions might generate different outputs. Any work to create per-version allowlists is welcome.

---

## Related issues
Closes #2801

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,10 +34,12 @@ In order to be able to continuously sync your fork with the origin repository's 
 To do so follow this [official github guide](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/syncing-a-fork).
 
 ### Dependency Setup
+
 We use [uv](https://github.com/astral-sh/uv) to manage our dev dependencies.
 To install it, see their [installation guide](https://docs.astral.sh/uv/getting-started/installation/)
 
 Once it's done, simply run the following command to automatically setup a virtual environment and install dev dependencies:
+
 ```bash
 uv sync
 source .venv/bin/activate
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -78,14 +80,15 @@ rm -r .mypy_cache
 Run `./scripts/stubtest.sh` to test that stubs and sources are in-line.
 
 We have some special files to allow errors:
+
 1. `scripts/stubtest/allowlist.txt` where we store things that we really don't care about: hacks, django internal utility modules, things that are handled by our plugin, things that are not representable by type system, etc
 2. `scripts/stubtest/allowlist_todo.txt` where we store all errors there are right now. Basically, this is a TODO list: we need to work through this list and fix things (or move entries to real `allowlist.txt`). In the end, ideally we can remove this file.
-3. `scripts/stubtest/allowlist_todo_django51.txt` where we store new errors from the Django 5.0 to 5.1 upgrade. This is an extra TODO list.
+3. `scripts/stubtest/allowlist_todo_django51.txt` where we store new errors from the Django 5.0 to 5.2 upgrade. This is an extra TODO list.
```
sobolevn:
```suggestion
3. `scripts/stubtest/allowlist_todo_django52.txt` where we store new errors from the Django 5.0 to 5.2 upgrade. This is an extra TODO list.
```

CONTRIBUTING.md
```diff
@@ -78,14 +80,15 @@ rm -r .mypy_cache
...
 
 You might also want to disable `incremental` mode while working on `stubtest` changes.
 This mode leads to several known problems (stubs do not show up or have strange errors).
 
-**Important**: right now we only run `stubtest` on Python 3.12 (because it is the latest released version at the moment), any other versions might generate different outputs. Any work to create per-version allowlists is welcome.
+**Important**: right now we only run `stubtest` on Python 3.13 (because it is the latest released version at the moment), any other versions might generate different outputs. Any work to create per-version allowlists is welcome.
```
sobolevn:
```suggestion
**Important**: right now we only run `stubtest` on the latest relased Python version, any other versions might generate different outputs. Any work to create per-version allowlists is welcome.
```

sobolevn (commented):
Thanks!

CONTRIBUTING.md
CodingWithSaksham: If we were to say this, cant we just remove this line completely? Seems quite redundant

sobolevn (approved):

PR #2803, opened 2025-09-06
Allow override of typing only class vars

UnknownPlatypus:
# I have made things!

Fixes a small oversight from #2794 I discoverd when testing master branch on my work codebase.

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -448,6 +448,7 @@ def add_new_class_for_module(
             var = Var(field_name, type=field_type)
             var.info = new_typeinfo
             var._fullname = new_typeinfo.fullname + "." + field_name
+            var.has_explicit_value = True
             new_typeinfo.names[field_name] = SymbolTableNode(
                 MDEF, var, plugin_generated=True, no_serialize=no_serialize
             )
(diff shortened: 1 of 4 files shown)


sobolevn (commented):
I am not sure that this is right. Annotations can be added for cases when some untyped base class is used. It provides some value, but users want to annotate the model for this attr.

Quite common pattern.

UnknownPlatypus:
But without this change, `.annotate(foo=...)` was causing errors if the model added a class level `foo: str`.
So I think my change accommodate for the use case you describe ? Or did I not understood correctly ?

sobolevn:
> But without this change, `.annotate(foo=...)` was causing errors if the model added a class level `foo: str`.

This looks like a correct behavior to me. 

Is `foo` a special placeholder for this `.annotate(foo=...)`?

UnknownPlatypus:
> Is foo a special placeholder for this .annotate(foo=...)?

Yes in my case this annotation is always present for this model (because the `prefetch_related(..., to_attr=...)` is in a custom queryset that is autoused) and adding this class level annotation was the only way to represent that prior to recent changes to support `to_attr` inference. 

But I do agree this is not accurate in the general case (when the custom queryset is not autoused). 
I'm fine closing this PR if it seem to specific, we can always revisit if issues are open with similar use cases 

This is again one of these cases we could solve if we tackle #2681

PR #2804, opened 2025-09-06
Add lookup validation in prefetch related

UnknownPlatypus:
# I have made things!

Further expand the validation of `prefetch_related` calls to check a bunch of error only checked at runtime.
I've used https://github.com/django/django/blob/0a67611b81567687cd24d2be1c0e390931f06262/tests/prefetch_related/tests.py#L574-L619 as a reference for these and tried to match the error messages.

To make this possible, I had to keep track of the `lookup` argument passed to `Prefetch` objects using the same specialization trick I used for the `to_attr` attribute. I'm not super happy with this approach, It get's a bit verbose but it works and cases where you pass around `Prefetch` objects are thin so the extra typing work should be manageable.

Doing so, I started to repeat a lot of stuff so I introduced a few helpers making the entrypoint `extract_prefetch_related_annotations` easier to parse.

PS: I tried reusing django's [prefetch_related_objects](https://github.com/django/django/blob/0231f71d31b3afa92b9b5b64ddad7efb9fab9a40/django/db/models/query.py#L2332) responsible of all these checks but It's highly coupled with the actual sql querying (and also make a bunch of errors lazy -- meaning absent at runtime if no items are prefetched!). So I've finally settled on small implementation in `check_valid_prefetch_related_lookup`


mypy_django_plugin/transformers/models.py
```diff
@@ -88,6 +88,7 @@ def create_new_var(self, name: str, typ: MypyType) -> Var:
         var._fullname = self.model_classdef.info.fullname + "." + name
         var.is_initialized_in_class = True
         var.is_inferred = True
+        var.has_explicit_value = True
```
UnknownPlatypus: See #2803, for this. I'm not sure how to do stacked pr from forks, I'll rebase to clear this once it lands

UnknownPlatypus:
This is ready for review @sobolevn, I've cleaned up the commits so review per commit should be easier.

django-stubs/db/models/query.pyi
```diff
@@ -15,9 +15,12 @@ _Model = TypeVar("_Model", bound=Model, covariant=True)
 _Row = TypeVar("_Row", covariant=True, default=_Model)  # ONLY use together with _Model
 _TupleT = TypeVar("_TupleT", bound=tuple[Any, ...], covariant=True)
 
-# The type of the prefetched model
+# The type of the lookup passed to Prefetch(...)
+# This will be specialized to a `LiteralString` in the plugin for further processing and validation
+_lookupT = TypeVar("_lookupT", bound=str, covariant=True)
```
sobolevn:
```suggestion
_LookupT = TypeVar("_LookupT", bound=str, covariant=True)
```

django-stubs/db/models/query.pyi
```diff
@@ -254,20 +257,20 @@ class RawQuerySet(Iterable[_Model], Sized):
...
-    def prefetch_related(self, *lookups: str | Prefetch[_PrefetchedQuerySetT, _ToAttrT]) -> Self: ...
+    def prefetch_related(self, *lookups: str | Prefetch[_lookupT, _PrefetchedQuerySetT, _ToAttrT]) -> Self: ...
     def resolve_model_init_order(self) -> tuple[list[str], list[int], list[tuple[str, int]]]: ...
     def using(self, alias: str | None) -> Self: ...
 
 # Deprecated alias of QuerySet, for compatibility only.
 _QuerySet: TypeAlias = QuerySet  # noqa: PYI047
 
-class Prefetch(Generic[_PrefetchedQuerySetT, _ToAttrT]):
+class Prefetch(Generic[_lookupT, _PrefetchedQuerySetT, _ToAttrT]):
```
sobolevn: Context: in latest published release `Prefetch` is not generic. So, it is fine to add this type var in the first position.

mypy_django_plugin/lib/helpers.py
```diff
@@ -362,6 +363,18 @@ def parse_bool(expr: Expression) -> bool | None:
     return None
 
 
+def get_literal_type(typ: Type) -> str | None:
```
```suggestion
def get_literal_str_type(typ: Type) -> str | None:
```

mypy_django_plugin/transformers/querysets.py
```diff
@@ -405,7 +462,71 @@ def check_conflicting_attr_value(
...
+    """Check if a lookup string resolve to something that can be prefetched"""
+    current_model_cls = django_model.cls
+    contenttypes_installed = django_context.apps_registry.is_installed("django.contrib.contenttypes")
+    for through_attr in lookup.split("__"):
+        rel_obj_descriptor = getattr(current_model_cls, through_attr, None)
+        if rel_obj_descriptor is None:
+            ctx.api.fail(
+                (
+                    f'Cannot find "{through_attr}" on "{current_model_cls.__name__}" object, '
+                    f'"{lookup}" is an invalid parameter to prefetch_related()'
```
```suggestion
                    f'"{lookup}" is an invalid parameter to "prefetch_related()"'
```

mypy_django_plugin/transformers/querysets.py
```diff
@@ -405,7 +462,71 @@ def check_conflicting_attr_value(
...
+            if contenttypes_installed:
+                from django.contrib.contenttypes.fields import GenericForeignKey
+
+                if isinstance(rel_obj_descriptor, GenericForeignKey):
+                    # Generic foreign keys can point to any model, so we use Model as the base type
+                    return True
+            ctx.api.fail(
+                (
+                    f'"{lookup}" does not resolve to an item that supports prefetching '
+                    f"- this is an invalid parameter to prefetch_related()"
```
```suggestion
                    f'- this is an invalid parameter to "prefetch_related()"'
```

sobolevn (approved):
Great work!

UnknownPlatypus:
Test on my work codebase revealed some deadcode and a few errors so pretty nice to see it useful too.

PR #2806, opened 2025-09-08
Validate `select_related` lookups

UnknownPlatypus:
# I have made things!

Similar to #2804, implement validation of the lookups passed to `.select_related(...)`.

The implementation is mostly extracted and adapted from `django.db.models.sql.compiler.SQLCompiler.get_related_selections`

PS: The `querysets.py` file is starting to get quite big, do you mind if I turn it into a module with a file per qs method ? (`select_related.py`, `prefetch_related.py` ...)

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -619,3 +619,79 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
         row_type = annotated_model
 
     return default_return_type.copy_modified(args=[annotated_model, row_type])
+
+
+def _get_select_related_field_choices(model_cls: type[Model]) -> set[str]:
+    """
+    Get valid field choices for select_related lookups.
+    Based on Django's SQLCompiler.get_related_selections._get_field_choices method.
+    """
+    opts = model_cls._meta
+
+    # Direct relation fields (forward relations)
+    direct_choices = (f.name for f in opts.fields if f.is_relation)
+
+    # Reverse relation fields (backward relations with unique=True)
+    reverse_choices = (f.field.related_query_name() for f in opts.related_objects if f.field.unique)
+    return {*direct_choices, *reverse_choices}
+
+
+def _validate_select_related_lookup(
+    ctx: MethodContext,
+    django_context: DjangoContext,
+    model_cls: type[Model],
+    lookup: str,
+) -> bool:
+    """Validate a single select_related lookup string."""
+    if not lookup.strip():
+        ctx.api.fail(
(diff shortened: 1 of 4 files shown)


mypy_django_plugin/transformers/querysets.py
```diff
@@ -619,3 +619,79 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
...
+def _validate_select_related_lookup(
+    ctx: MethodContext,
+    django_context: DjangoContext,
+    model_cls: type[Model],
+    lookup: str,
+) -> bool:
+    """Validate a single select_related lookup string."""
+    if not lookup.strip():
+        ctx.api.fail(
+            f'Invalid field name "{lookup}" in select_related lookup. Cannot be whitespace-only',
```
```suggestion
            f'Invalid field name "{lookup}" in select_related lookup',
```

I guess it is clear from the message.

mypy_django_plugin/transformers/querysets.py
```diff
@@ -619,3 +619,79 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
...
+
+    lookup_parts = lookup.split("__")
+    observed_model = model_cls
+    for i, part in enumerate(lookup_parts):
+        valid_choices = _get_select_related_field_choices(observed_model)
+
+        if part not in valid_choices:
+            ctx.api.fail(
+                f'Invalid field name "{part}" in select_related lookup. '
+                f"Choices are: {', '.join(sorted(valid_choices)) or '(none)'}",
```
sobolevn, line 661: There can be many choices. Let's limit them.

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: Fair enough

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: Are you sure ? Runtime display them all and it's usually quite handy to solve the issue.

Could be in a ctx.msg.note otherwise ? Similar to the verbose note printed when arguments are not matching an override

sobolevn: Ok, let's keep it unless someone complains :)

sobolevn (approved):
Are we ready for a release? :)

UnknownPlatypus:
> Are we ready for a release? :)

I have a few other ones coming for similar validation on other qs method, don't know If it's worth waiting for ?

sobolevn:
Sure, no rush :)
It is easier to release the whole package of features. Let' me know when to push the button :)

PR #2810, opened 2025-09-12
Bump mypy to 1.18.1

sobolevn:


sobolevn:
I want to make a release with new mypy version support.

PR #2819, opened 2025-09-12
Add `mypy_primer` workflow

### LINKED ISSUE #1362: Integrate mypy_primer into our CI
[mypy_primer](https://github.com/hauntsaninja/mypy_primer) is a project that runs mypy on lots of public open source repositories.

In CI, it can do a mypy run before and after some proposed changes in a pull request, and report the diff between those results. It's being used by typeshed and mypy itself among others. Example:

image

If we add this to django-stubs and djangorestframework-stubs, it would give us better visibility into possible breakage we could be causing and help troubleshoot issues earlier.

I have no immediate plans to implement this and no experience with mypy_primer. I opened this issue to facilitate some discussion and feedback.

Most projects from [mypy_primer's built-in project definitions](https://github.com/hauntsaninja/mypy_primer/blob/master/mypy_primer.py#L1066-L1084) wouldn't be very useuful for us, because they don't use Django. So we would probably want to create our own list of open source Django and django-rest-framework projects for mypy_primer. Suggestions for which ones to add would be helpful.

sobolevn:
Refs:
- https://github.com/numpy/numpy/blob/main/.github/workflows/mypy_primer.yml
- https://github.com/numpy/numpy/blob/main/.github/workflows/mypy_primer_comment.yml

Closes https://github.com/typeddjango/django-stubs/issues/1362

### DIFF
--- a/.github/workflows/mypy_primer.yml
+++ b/.github/workflows/mypy_primer.yml
@@ -52,7 +52,7 @@ jobs:
           (
             mypy_primer \
             --new v${MYPY_VERSION} --old v${MYPY_VERSION} \
-            --known-dependency-selector django \
+            --known-dependency-selector django-stubs \
             --old-prepend-path django-stubs_base --new-prepend-path django-stubs_to_test \
             --num-shards 1 --shard-index ${{ matrix.shard-index }} \
             --debug \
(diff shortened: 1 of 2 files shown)


UnknownPlatypus:
I don't know how this works but if we can check against sentry it would be nice, they seem to have a very diverse codebase that catches edge cases

sobolevn:
Yes, adding `sentry` to `mypy_primer` would be great! 👍

PR #2829, opened 2025-09-15
Fix `union` / `intersection` / `difference` to support QuerySet using `values_list` and `values`

### LINKED ISSUE #2825: 5.2.3 arg-type complaint for QuerySet.union
# Bug report

Hi,
Our project uses `union` to merge two querysets. This line of code now causes an `arg-type` complaint from MyPy after updating django-stubs to 5.2.3 or later (no issue when using 5.2.2). 

The pattern is lifted directly from the Django documentation for union:
https://docs.djangoproject.com/en/5.2/ref/models/querysets/#django.db.models.query.QuerySet.union

Here's a demo models.py that shows the issue:

```python
from django.db.models import Model, QuerySet, TextField


class Foo(Model):
    name = TextField(blank=False, null=False)


class Bar(Model):
    name = TextField(blank=False, null=False)


def foos_and_bars() -> QuerySet[Model, tuple[str]]:
    foos = Foo.objects.all().values_list("name")
    bars = Bar.objects.all().values_list("name")
    return foos.union(bars)
```

Here's what MyPy now has to say about this:

```shell
MyProject/models.py:15: error: Argument 1 to "union" of "QuerySet" has incompatible type "QuerySet[Bar, tuple[str]]"; expected "QuerySet[Model, Model]"  [arg-type]
```

I think the change that has introduced this behaviour is here:
https://github.com/typeddjango/django-stubs/commit/0087c115eda656f6fdc1faea0727a6a6fc2e47bb

The stub for `union` changed from:
`def union(self, *other_qs: Any, all: bool = False) -> Self: ...`
to:
`def union(self, *other_qs: QuerySet[Model], all: bool = False) -> Self: ...`

It looks like this new annotation for `*other_qs` is too narrow to allow for the output of the `qs.values_list()` method, even though this is the recommended pattern for using `qs.union()`.

A similar issue might now affect `qs.intersection()` and `qs.difference()`, but I haven't checked these myself.

## System information

- OS: MacOS Sequoia 15.6.1
- `python` version: 3.13.7 
- `django` version: 5.2.6
- `mypy` version: 1.18.1
- `django-stubs` version: 5.2.3
- `django-stubs-ext` version:

UnknownPlatypus:
Fixes #2825


tests/typecheck/managers/querysets/test_basic_methods.yml
```diff
@@ -70,12 +70,32 @@
...
+
+        reveal_type(foos.union(bars)) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.MyModel1, tuple[builtins.str]]"
+        reveal_type(foos.intersection(bars)) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.MyModel1, tuple[builtins.str]]"
+        reveal_type(foos.difference(bars)) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.MyModel1, tuple[builtins.str]]"
+
+        def union_values_list() -> models.QuerySet[models.Model, tuple[str]]:
+            union = foos.union(bars)
+            return union
+
+        foos_dict = MyModel1.objects.all().values("name")
```
sobolevn, line 104: let's test multiple values maybe?

tests/typecheck/managers/querysets/test_basic_methods.yml
```diff
@@ -70,12 +70,32 @@
...
+        bars_dict = MyModel2.objects.all().values("name")
+
+        reveal_type(foos_dict.union(bars_dict)) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.MyModel1, TypedDict({'name': builtins.str})]"
+        reveal_type(foos_dict.intersection(bars_dict)) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.MyModel1, TypedDict({'name': builtins.str})]"
+        reveal_type(foos_dict.difference(bars_dict)) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.MyModel1, TypedDict({'name': builtins.str})]"
```
sobolevn, line 109: Sorry, I can't take a look right now, so gonna ask: do we have tests for `.union` with full regular models?

UnknownPlatypus: Not a lot, this was not very well covered, I can expand the coverage taakingexemple from django docs + some edge cases I can think of

sobolevn: Thank you! 👍

UnknownPlatypus: I've expanded the tests a bit. We could be even more strict for weird cases mixing values and values_list with different number of field but I don't think it's worth supporting. It would require plugin code for very exotic use cases

tests/typecheck/managers/querysets/test_basic_methods.yml
```diff
@@ -70,12 +70,75 @@
...
+        from django.db import models
         from myapp.models import MyModel1, MyModel2
         kls: type[MyModel1 | MyModel2] = MyModel1
         reveal_type(kls.objects)  # N: Revealed type is "django.db.models.manager.Manager[myapp.models.MyModel1] | django.db.models.manager.Manager[myapp.models.MyModel2]"
         reveal_type(kls.objects.all())  # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.MyModel1, myapp.models.MyModel1] | django.db.models.query.QuerySet[myapp.models.MyModel2, myapp.models.MyModel2]"
         reveal_type(kls.objects.get())  # N: Revealed type is "myapp.models.MyModel1 | myapp.models.MyModel2"
+
+        # Regular QuerySet
+        foos = MyModel1.objects.all()
+        bars = MyModel2.objects.all()
```
sobolevn, line 85: Can we also please test structurally incompatible models? For now we don't do anything special with them.

sobolevn (commented):
Thanks a lot!

tests/typecheck/managers/querysets/test_basic_methods.yml
UnknownPlatypus: I can add a todo but yes we won't detect much currently (but runtime will most likely see errors like `The used SELECT statements have a different number of columns`)

sobolevn: Good enough for now! 🤝

palfrey:
I've just hit this one as well, and it seems to WFM

tests/typecheck/managers/querysets/test_basic_methods.yml
UnknownPlatypus: Done !

sobolevn (approved):

PR #2830, opened 2025-09-16
Add `strict_model_abstract_attrs` setting to allow `models.Model.objects` access

sobolevn:
We need this setting so we can test `sentry` in `mypy-primer`. To test it, we need to help Sentry's team to use our package, without their custom fork.

So, setting `strict_model_abstract_attrs = false` should do the trick for them.
It can be also helpful for other big projects to adopt `django-stubs`.

CC @flaeppe and @asottile-sentry

Refs https://github.com/hauntsaninja/mypy_primer/issues/202


sobolevn:
@UnknownPlatypus do you have any ideas on why this happens? :)

sobolevn:
For the record: I added `AUTH_USER_MODEL` for all cache entries. It solved the problem.

README.md
```diff
@@ -124,6 +124,14 @@ The supported settings are:
 
   Set to `false` if using dynamic settings, as [described below](https://github.com/typeddjango/django-stubs#how-to-use-a-custom-library-to-handle-django-settings).
 
+- `strict_model_abstract_attrs`, a boolean, default `true`.
+
+  Set to `false` if you want to keep `.objects`, `.DoesNotExist`,
+  and `.MultipleObjectsReturned` attributes on `models.Model` type.
+  [See here why](https://github.com/typeddjango/django-stubs?tab=readme-ov-file#how-to-use-typemodel-annotation-with-objects-attribute)
+  this is dangeours to do by default.
```
asottile-sentry:
```suggestion
  this is dangerous to do by default.
```

asottile-sentry (approved):
:pray: this will work great for us

flaeppe (approved):

PR #2836, opened 2025-09-21
Allow to annotate existing model field de-selected via prior `.values`/`.values_list` calls

### LINKED ISSUE #2818: 5.2.3 regression with `.annotate` that forbids overriding existing fields
Refs https://github.com/typeddjango/django-stubs/pull/2797

Quoting @asottile-sentry from https://github.com/typeddjango/django-stubs/pull/2797#issuecomment-3286657655

---

I believe this might not be correct for the `.annotate(...)` case

here's an example from sentry:

```
src/sentry/api/serializers/models/rule.py:207: error: Attribute "date_added" already defined on "sentry.models.rulefirehistory.RuleFireHistory"  [no-redef]
```

https://github.com/getsentry/sentry/blob/e1829e8cbd0fe4179e137afbbee23b06d35d3eee/src/sentry/api/serializers/models/rule.py#L209

(there's a few other examples in sentry as well where particular aggregations are used to produce a value with the same name as an existing field when making a grouped query)

---

Quoting myself from https://github.com/typeddjango/django-stubs/pull/2797#issuecomment-3286675283

---

Hm, interesting, I haven't consider this use-case. So, basically, if we annotate an some existing `IntegerField` field with `int` value, there's no *type* error here. So, basically, we should instead detect type mismatches in annotated fields and report them.

UnknownPlatypus:
# I have made things!

I've tried to avoid as much as possible false positive on loosely typed cases.

cc @asottile-sentry , would be awesome if can try to cherry-pick a1f80ead79dfcefb1d6e70bb11704b4330c1e3bb to see if it fixes your issue.

## Related issues

Fixes #2818

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -43,6 +43,7 @@ from mypy.typeanal import make_optional_type
 from mypy.types import (
     AnyType,
     CallableType,
+    ExtraAttrs,
     Instance,
     LiteralType,
     NoneTyp,
@@ -694,3 +695,32 @@ def get_model_from_expression(
 
 def fill_manager(manager: TypeInfo, typ: MypyType) -> Instance:
     return Instance(manager, [typ] if manager.is_generic() else [])
+
+
+def merge_extra_attrs(
+    base_extra_attrs: ExtraAttrs | None,
+    *,
+    new_attrs: dict[str, MypyType] | None = None,
+    new_immutable: set[str] | None = None,
+) -> ExtraAttrs:
+    """
+    Create a new `ExtraAttrs` by merging new attributes/immutable fields into a base.
+
+    If base_extra_attrs is None, creates a fresh ExtraAttrs with only the new values.
+    """
+    if base_extra_attrs:
+        return ExtraAttrs(
+            attrs={**base_extra_attrs.attrs, **new_attrs} if new_attrs is not None else base_extra_attrs.attrs.copy(),
+            immutable=(
+                base_extra_attrs.immutable | new_immutable
+                if new_immutable is not None
+                else base_extra_attrs.immutable.copy()
+            ),
+            mod_name=None,
+        )
+    else:
+        return ExtraAttrs(
+            attrs=new_attrs or {},
+            immutable=new_immutable,
+            mod_name=None,
+        )
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -718,9 +718,8 @@ def merge_extra_attrs(
             ),
             mod_name=None,
         )
-    else:
-        return ExtraAttrs(
-            attrs=new_attrs or {},
-            immutable=new_immutable,
-            mod_name=None,
-        )
+    return ExtraAttrs(
+        attrs=new_attrs or {},
+        immutable=new_immutable,
+        mod_name=None,
+    )
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -14,17 +14,7 @@ from mypy.checker import TypeChecker
 from mypy.errorcodes import NO_REDEF
 from mypy.nodes import ARG_NAMED, ARG_NAMED_OPT, ARG_STAR, CallExpr, Expression
 from mypy.plugin import FunctionContext, MethodContext
-from mypy.types import (
-    AnyType,
-    ExtraAttrs,
-    Instance,
-    LiteralType,
-    ProperType,
-    TupleType,
-    TypedDictType,
-    TypeOfAny,
-    get_proper_type,
-)
+from mypy.types import AnyType, Instance, LiteralType, ProperType, TupleType, TypedDictType, TypeOfAny, get_proper_type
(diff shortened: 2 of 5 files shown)


mypy_django_plugin/transformers/querysets.py
```diff
@@ -168,7 +178,13 @@ def extract_proper_type_queryset_values_list(ctx: MethodContext, django_context:
     row_type = get_values_list_row_type(
         ctx, django_context, django_model.cls, is_annotated=django_model.is_annotated, flat=flat, named=named
     )
-    return default_return_type.copy_modified(args=[django_model.typ, row_type])
+    ret = default_return_type.copy_modified(args=[django_model.typ, row_type])
+    if not named and (field_lookups := resolve_field_lookups(ctx.args[0], django_context)):
+        # For non-named values_list, the row type does not encode column names.
+        # Attach selected field names to the returned QuerySet instance so that
+        # subsequent annotate() can make an informed decision about name conflicts.
+        ret.extra_attrs = ExtraAttrs(attrs={}, immutable=set(field_lookups), mod_name=None)
```
sobolevn: Question: shouldn't we also set `attrs` when setting `immutable`?

sobolevn (commented):
LGTM, except one question :)

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: From what I understand of this `ExtraAttrs` thingy, it's not supposed to be used like that and we are kinda hacking around to add metadata per `Instance`. I can add the attrs dict but we will ignore it and I think mypy too :/

sobolevn: let's add this, because leaving this inconsistency will blow up at some point.

sobolevn: Btw, consider using `copy_with_extra_attr`? Or any other way of preserving existing `.extra_attrs`

UnknownPlatypus: > Btw, consider using copy_with_extra_attr? Or any other way of preserving existing .extra_attrs

We only set extra_attrs on this type here and the django behavior of chained `values_list` is that the last one override the others. I can add a comment to say that, this is not obvious reading the code

sobolevn: I would still prefer to preserve, some other plugin / feature might modify it.

UnknownPlatypus: @sobolevn I did some more digging, about how mypy internally deals with `extra_attrs`. Found relevant parts in the implementation of [add_any_attribute_to_type](https://github.com/python/mypy/blob/de2f375eb0da1c446e2aab1474c28b8d7707a848/mypy/checker.py#L7990-L8021) and in the [functools plugin](https://github.com/python/mypy/blob/de2f375eb0da1c446e2aab1474c28b8d7707a848/mypy/plugins/functools.py#L320-L330).

`attrs` and `immutable` seem to have different purposes. `attrs` to store key/value pairs of names/type, and `immutable` to store keys only (Used to store `Final` variables defined  at the module levels or as a way to pass around some boolean flags in the functools plugin)

So to answer your suggestions:

> Question: shouldn't we also set attrs when setting immutable?

Given the above, I think it's better to follow the functools plugin implementation and only use immutable. We really just want to passa round metadatas but are not adding actuals new attributes to the `Instance`

> Btw, consider using copy_with_extra_attr? Or any other way of preserving existing .extra_attrs

You're right, I've made the change! I introduced a new helper I could also use in `get_annotated_type`

mypy_django_plugin/lib/helpers.py
```diff
@@ -694,3 +695,32 @@ def get_model_from_expression(
...
+        return ExtraAttrs(
+            attrs={**base_extra_attrs.attrs, **new_attrs} if new_attrs is not None else base_extra_attrs.attrs.copy(),
+            immutable=(
+                base_extra_attrs.immutable | new_immutable
+                if new_immutable is not None
+                else base_extra_attrs.immutable.copy()
+            ),
+            mod_name=None,
+        )
+    else:
```
```suggestion
```

sobolevn (approved):
one nit :)

mypy_django_plugin/lib/helpers.py
UnknownPlatypus: I've made the change, should we add [RET505](https://docs.astral.sh/ruff/rules/superfluous-else-return/) from ruff to enforce this convention ? 

I can make a followup PR for this, it does affect  ~10 files. I also find these valueable:
```
  "RET504",  # Unnecessary assignment to {name} before return statement
  "RET505",  # Unnecessary {branch} after return statement
  "RET506",  # Unnecessary {branch} after raise statement
  "RET507",  # Unnecessary {branch} after continue statement
  "RET508",  # Unnecessary {branch} after break statement
```

sobolevn (approved):
Thanks a lot!

UnknownPlatypus:
btw, asotille recently left sentry so we wont probably hear from him on this issue.

sobolevn:
`master` branch is now broken :(
Please, take a look

UnknownPlatypus:
Seem to be a conflict with the branch removing automatically added contentypes, will send a fix

PR #2838, opened 2025-09-23
fix: allow to query decimal fields with int

### LINKED ISSUE #2831: DecimalField should accept ints
# Bug report

The type definitions for `DecimalField` are too strict and do not allow to query them with ints.

## What's wrong
We are enabling the plugin in one of our codebases and we are getting a bunch of errors of the form

```
error: Incompatible type for lookup 'total_amount': (got "int", expected "str | Decimal")  [misc]
```

This happens, because the type definition for `DecimalField. _pyi_lookup_exact_type` is too strict and does not include "float" like `_pyi_private_set_type`
https://github.com/typeddjango/django-stubs/blob/ed5046ec7395d4492147524f050e7bd9ae27396b/django-stubs/db/models/fields/__init__.pyi#L264-L266

## How is that should be
Code that works without typing should rarely be changed in order to accommodate types, unless of course there is an underlying typing issue. In this case I think no error should be raised as the conversion is safe, and at the very least we should add `int` to `DecimalField. _pyi_lookup_exact_type`.

## System information

- OS: macOS Sequoio 15.6.1
- `python` version: 3.13.3
- `django` version: 5.2.6
- `mypy` version: 1.17.1

sk-:
Django allow to query decimal fields with int, and given there should be no precision loss, we should accept ints to query such fields.

Note: even though we do accept floats when setting a value in a decimal field, I opted for only accepting ints in this PR, as to minimize the impact of the change.

## Related issues

Closes #2831

### DIFF
--- a/tests/typecheck/managers/querysets/test_filter.yml
+++ b/tests/typecheck/managers/querysets/test_filter.yml
@@ -45,6 +45,29 @@
                 class User(models.Model):
                     age = models.IntegerField()
 
+# Regression test for #2831
+-   case: filter_with_decimal
+    main: |
+        from myapp.models import MyModel
+        from decimal import Decimal
+
+        MyModel.objects.filter(a=1)
+        MyModel.objects.filter(a="1.0")
+        MyModel.objects.filter(a=Decimal(1.0))
+        MyModel.objects.filter(a=1.0)
+    out: |
+      main:7: error: Incompatible type for lookup 'a': (got "float", expected "str | int | Decimal")  [misc]
+
+    installed_apps:
+        - myapp
+    files:
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.db import models
+
+                class MyModel(models.Model):
+                    a = models.DecimalField()
 
 -   case: filter_with_multiple_fields
     main: |
--- a/tests/typecheck/managers/querysets/test_filter.yml
+++ b/tests/typecheck/managers/querysets/test_filter.yml
@@ -54,10 +54,7 @@
         MyModel.objects.filter(a=1)
         MyModel.objects.filter(a="1.0")
         MyModel.objects.filter(a=Decimal(1.0))
-        MyModel.objects.filter(a=1.0)
-    out: |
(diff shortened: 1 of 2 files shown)


UnknownPlatypus:
Hey, thanks for the PR! Could you add a small regression test next to this test showcasing correct filter values and ones that should raise error? 
https://github.com/typeddjango/django-stubs/blob/d8aa20d40cdb351cea76ea3e5c5a0bd8120fba62/tests/typecheck/managers/querysets/test_filter.yml#L33

sk-:
> Hey, thanks for the PR! Could you add a small regression test next to this test showcasing correct filter values and ones that should raise error?
> 
> https://github.com/typeddjango/django-stubs/blob/d8aa20d40cdb351cea76ea3e5c5a0bd8120fba62/tests/typecheck/managers/querysets/test_filter.yml#L33

Thanks @UnknownPlatypus, I had written the test but forgot to add it to the PR. Let me know if the test I added is enough.

tests/typecheck/managers/querysets/test_filter.yml
```diff
@@ -45,6 +45,29 @@
...
+        from myapp.models import MyModel
+        from decimal import Decimal
+
+        MyModel.objects.filter(a=1)
+        MyModel.objects.filter(a="1.0")
+        MyModel.objects.filter(a=Decimal(1.0))
+        MyModel.objects.filter(a=1.0)
+    out: |
+      main:7: error: Incompatible type for lookup 'a': (got "float", expected "str | int | Decimal")  [misc]
+
```
sobolevn:
```suggestion
        MyModel.objects.filter(a="1.0")
        MyModel.objects.filter(a=Decimal(1.0))
        MyModel.objects.filter(a=1.0)  # E: Incompatible type for lookup 'a': (got "float", expected "str | int | Decimal")  [misc]
```

UnknownPlatypus:
> Thanks @UnknownPlatypus, I had written the test but forgot to add it to the PR. Let me know if the test I added is enough.

Awesome, thanks @sk-

PR #2841, opened 2025-09-26
fix(utils/text): use proper type for `format_lazy` return type

### LINKED ISSUE #2840: The function `format_lazy` should return `_StrPromise`
# Bug report

The function `format_lazy` should return `_StrPromise` and not `_StrOrPromise`.

It is important when the output of this command is used with the function which expect `promise` only (for lazy evaluation of translation). This is done to notify the user if his translation will be evaluated incorrectly.

## What's wrong

We are receiving these type of errors:
```
error: Argument "my_argument" to "my_function" of "MyClass" has incompatible type "str | _StrPromise"; expected "_StrPromise"  [arg-type]
```

## How is that should be

No typing error should be returned as the `format_lazy` returning `promise` by itself

## System information

Not Applicable

sshishov:
Use `_StrPromise` instead of `_StrOrPromise` for `format_lazy` return type.

Closes #2840

### DIFF
--- a/django-stubs/utils/text.pyi
+++ b/django-stubs/utils/text.pyi
@@ -5,7 +5,7 @@ from re import Pattern
 from typing import ClassVar, TypeVar, overload
 
 from django.db.models.base import Model
-from django.utils.functional import SimpleLazyObject, _StrOrPromise, cached_property
+from django.utils.functional import SimpleLazyObject, _StrPromise, _StrOrPromise, cached_property
 
 _StrOrPromiseT = TypeVar("_StrOrPromiseT", bound=_StrOrPromise)
 _StrOrPromiseOrNoneT = TypeVar("_StrOrPromiseOrNoneT", bound=_StrOrPromise | None)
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
But, it can return `str`

```python
In [1]: from django.utils.text import format_lazy

In [2]: format_lazy('a{0}', 'b')
Out[2]: 'ab'
```

sshishov:
HI @sobolevn , it seems it just does not have the safety-net. But yeah if it can accept `str` and return `str` then we have to type it properly. We can close my MR with the statement that Working As Expected.

It is just useless to use `format_lazy` for non-lazy objects, you can use `f-string` or `str.format` instead.

To manage `mypy-error` in our code we will add the `internal` assert then, which should fail on dev runtime for us. It is also suitable 👍 

-- Sergei

sobolevn:
Maybe we can use overloads?

sshishov:
Let me try with overloads... will create MR today evening

UnknownPlatypus:
Superseded by #3133

PR #2845, opened 2025-09-29
Fix `RemoteUserMiddleware` missing methods and add test coverage

### LINKED ISSUE #2682: Wrong stub - django.contrib.auth.middleware.RemoteUserMiddleware
# Bug report

Wrongly specified inheritance and attributes.

## What's wrong

https://github.com/typeddjango/django-stubs/blob/2a89bb5e9ba7da7bbcb6f11dfa56b147d5aef5e2/django-stubs/contrib/auth/middleware.pyi#L30C1-L33C78

vs:

https://github.com/django/django/blob/f920937c8a63df6bea220e4386f59cdb45b2e355/django/contrib/auth/middleware.py#L93

## System information

- OS:
- `python` version: 3.13
- `django` version: 5.2.1
- `django-stubs` version: 5.2.0
- `django-stubs-ext` version: 5.2.0

emmanuel-ferdman:
## PR Summary
This PR fixes the `RemoteUserMiddleware` stub by adding the missing private methods `_remove_invalid_user()` and `_aremove_invalid_user()`, which exist in Django but were not in the stubs and caused mypy errors. It also adds tests for `AuthenticationMiddleware`, `RemoteUserMiddleware`, and `LoginRequiredMiddleware`, covering sync/async methods, attributes, and inheritance.

Fixes #2682.


django-stubs/contrib/auth/middleware.pyi
```diff
@@ -40,6 +40,8 @@ class RemoteUserMiddleware:
     def clean_username(self, username: str, request: HttpRequest) -> str: ...
     def process_request(self, request: HttpRequest) -> None: ...
     async def aprocess_request(self, request: HttpRequest) -> None: ...
+    def _remove_invalid_user(self, request: HttpRequest) -> None: ...
```
sobolevn, line 43: we generally don't add protected method to stubs, if they are required for some reason, please add a comment: why?

tests/typecheck/contrib/auth/test_middleware.yml
```diff
@@ -0,0 +1,166 @@
+-   case: test_authentication_middleware_basic
```
sobolevn, line 1: Sorry, but we don't test regular types, tests are expensive to run. Moreover, they are checked by mypy by default, so not required in most cases. Why do you need tests for this feature?

emmanuel-ferdman: You're right that these are regular types and don't need extensive testing. However, the tests were added because there was zero test coverage for any auth middleware classes. We can remove them.

django-stubs/contrib/auth/middleware.pyi
emmanuel-ferdman: I saw the open issue and reviewed the stub vs Django source code. The only missing items were these two protected methods (`_remove_invalid_user` and `_aremove_invalid_user`), as as a new user for this repo - I added them. If the issue is solved with this fix, I suggest closing it. Also, since you mentioned django-stubs generally doesn't include protected methods in stubs, please go ahead and close this PR as not needed. Otherwise please explain the issue and outline how it should be fixed.

emmanuel-ferdman:
Hi @sobolevn, thanks for the feedback! I agree with your comments, but could you please clarify what’s still needed for issue [#2682](https://github.com/typeddjango/django-stubs/issues/2682) to be closed? I’m considering closing this PR unless there’s something that still needs to be added to address that issue.

sobolevn:
In case you really need these two protected functions, please first specify why do you need them and prove that they will be stable at least for a while.

PR #2846, opened 2025-09-29
Update release documentation, workflow and build-and-check job to `uv`

intgr:
Aligns parts of `django-stubs` with the changes I have been making in `djangorestframework-stubs`:

* `build-and-check` changes: https://github.com/typeddjango/djangorestframework-stubs/pull/797
* `release` changes: https://github.com/typeddjango/djangorestframework-stubs/pull/838
* Docs: https://github.com/typeddjango/djangorestframework-stubs/pull/840

Changes:
* Using `uv build` instead of PyPI [`build`](https://pypi.org/project/build/) dependency
* Added `build-and-check` check to ensure `uv.lock` is up to date
* Updated from Python 3.12 to 3.13


intgr:
Related note: `djangorestframework-stubs` uses build system `uv_build` instead of `hatchling`. https://github.com/typeddjango/djangorestframework-stubs/blob/73ce41a8b87f83f08d33c5d4cda5f52b6835ee10/pyproject.toml#L58-L60

I'm not sure if there are any significant differences between the two. I can submit a PR to change it here too, if anybody thinks that's beneficial.

CONTRIBUTING.md
```diff
@@ -125,8 +125,9 @@ This rule helps us avoid tying in too closely to Django’s undocumented interna
...
       `patch` is sequentially increasing for each stubs release. Reset to `0` if `major.minor` was updated.
 
-    - Update the `version=` value within `setup(...)` call in **both** `setup.py` files. The versions must be in sync.
-    - Update `django-stubs-ext>=` dependency in root `setup.py` to the same version number.
+    - Update the `version =` value within `[project]` section in **both** `pyproject.toml` files. The versions must be in sync.
+    - Update `django-stubs-ext>=` dependency in root `pyproject.toml` to the same version number.
+    - Update lockfile, run: `uv lock`
     - Add a new row at the top of ['Version compatibility' table in README.md](README.md#version-compatibility).
     - Use pull request title "Version x.y.z release" by convention.
     - Add the correct classifiers to `setup.py` if support is added for a new Python or Django version
```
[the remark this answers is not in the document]
intgr:
```suggestion
    - Add the correct classifiers to `classifiers =` if support is added for a new Python or Django version
```

sobolevn (approved):

UnknownPlatypus:
> I'm not sure if there are any significant differences between the two. I can submit a PR to change it here too, if anybody thinks that's beneficial.

Don't have strong opinion on that, I think they both work well for this kind of simple package. 
I'll probably don't bother to swap since it's currently working and will probably for the years to come but if you'd prefer to unify and can make it work I'm fine with it too

.github/workflows/test.yml
```diff
@@ -172,16 +172,17 @@ jobs:
...
-          python3 -m build .
-          python3 -m build ext/
+          uv build .
+          uv build ext/
       - name: Check package metadata
         run: |
-          twine check --strict dist/*
-          twine check --strict ext/dist/*
+          uvx twine check --strict dist/*
+          uvx twine check --strict ext/dist/*
```
ngnpope, line 188: Can we get rid of twine in favour of `uv publish`?

ngnpope (commented):
> I'm not sure if there are any significant differences between the two. I can submit a PR to change it here too, if anybody thinks that's beneficial.

For my 2¢, I'm happy for this to be switched to the `uv-build` backend. One less explicit dependency (as `uv` already bundles its own build backend).

.github/workflows/test.yml
intgr: Publishing (`release.yml`) doesn't use twine (at least not directly), it uses `pypa/gh-action-pypi-publish@v1.13.0`.

This is just doing some checks on every PR build. [According to twine docs](https://twine.readthedocs.io/en/latest/#twine-check): "Checks whether your distribution’s long description will render correctly on PyPI"

ngnpope: 🤦🏻 Of course, scanned across that too quickly!

However, having now looked at [this comment](https://github.com/astral-sh/uv/issues/8641#issuecomment-3033058192) and [checked the source](https://github.com/pypa/twine/blob/974a8d8b1e63f65ab31936ae39a63adcdf3f47c6/twine/commands/check.py#L36) myself, I think we can just remove this step entirely - it only checks that rST can render properly and we're using markdown - so it's effectively a noop.

intgr: Will do in a follow-up

PR #2848, opened 2025-09-30
Switch to `uv_build` build system & cleanup

intgr:
Solving follow-ups from #2846 comments.

* Switched build system `hatchling` -> `uv_build`
* Removed useless `twine check` from `build-and-check` CI job

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -182,7 +182,3 @@ jobs:
         run: |
           uv build .
           uv build ext/
-      - name: Check package metadata
-        run: |
-          uvx twine check --strict dist/*
-          uvx twine check --strict ext/dist/*
--- a/ext/pyproject.toml
+++ b/ext/pyproject.toml
@@ -1,7 +1,3 @@
-[build-system]
-build-backend = "hatchling.build"
-requires = ["hatchling"]
-
 [project]
 name = "django-stubs-ext"
 # NB! For clarity, keep version major.minor.patch in sync with django-stubs.
@@ -43,5 +39,10 @@ Homepage = "https://github.com/typeddjango/django-stubs"
 Funding = "https://github.com/sponsors/typeddjango"
 "Release notes" = "https://github.com/typeddjango/django-stubs/releases"
 
-[tool.hatch.build]
-packages = ["django_stubs_ext"]
+[build-system]
+requires = ["uv_build>=0.8.22,<0.9.0"]
+build-backend = "uv_build"
+
+[tool.uv.build-backend]
+module-name = ["django_stubs_ext"]
+module-root = ""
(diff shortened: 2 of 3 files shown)


intgr:
Similar to `djangorestframework-stubs`, there are mostly frivolous differences in `.whl` file contents, comparing `python3 -m build` before with `uv build` after.

Such as `METADATA` file different order of fields; package name casefolding; whether single quotes or double quotes are used, etc.

Diffs: https://gist.github.com/intgr/dd8c503bc93148eef1f1e8632fff850e

pyproject.toml
```diff
@@ -1,7 +1,3 @@
-[build-system]
-requires = ["hatchling"]
-build-backend = "hatchling.build"
```
intgr, line 3: I never liked that this was at the top of the file; moved to below package metadata.

ext/pyproject.toml
```diff
@@ -43,5 +39,10 @@ Homepage = "https://github.com/typeddjango/django-stubs"
 Funding = "https://github.com/sponsors/typeddjango"
 "Release notes" = "https://github.com/typeddjango/django-stubs/releases"
 
-[tool.hatch.build]
-packages = ["django_stubs_ext"]
+[build-system]
+requires = ["uv_build>=0.8.22,<0.9.0"]
```
sobolevn, line 43: Do we need to be this specific? Will `"uv_build>=0.8"` work?

.github/workflows/test.yml
```diff
@@ -182,7 +182,3 @@ jobs:
         run: |
           uv build .
           uv build ext/
-      - name: Check package metadata
-        run: |
-          uvx twine check --strict dist/*
```
sobolevn, line 187: Does `uv` has an alternative with a dry-run?

ngnpope: See https://github.com/typeddjango/django-stubs/pull/2846#discussion_r2390681091 - `twine check` is only making sure that rST is valid in the long description, but we're using markdown.

ext/pyproject.toml
intgr: This is recommended by official docs https://docs.astral.sh/uv/concepts/build-backend/#using-the-uv-build-backend

I guess patchlevel versions may contain fixes, I see no reason to allow downgrading to more buggy versions given that we haven't tested with them.

UnknownPlatypus (approved):
Thanks, seem good to me

ngnpope (approved):

sobolevn (approved):

pyproject.toml
adamchainz: We can also lean on https://github.com/tox-dev/pyproject-fmt/ for a consistent sort

ext/pyproject.toml
intgr, line 43: Why isn't @dependabot updating this? 🤔 

It updated in djangorestframework-stubs:
* https://github.com/typeddjango/djangorestframework-stubs/pull/848

PR #2849, opened 2025-10-02
BaseForm.add_error: Use covariant types 

inducer:
Without this change, the user has to pass in exactly these types, since `list` and `dict` are invariant in the value type.

Please squash when merging.


sobolevn (approved):
Thanks!

PR #2851, opened 2025-10-04
Support `GenericPrefetch`

### LINKED ISSUE #2850: 5.2.3 regression: using GenericPrefetch breaks the plugin
# Bug report

## What's wrong

Using a `prefetch_related` with a [`GenericPrefetch`](https://docs.djangoproject.com/en/5.2/ref/contrib/contenttypes/#genericprefetch) (from `django.contrib.contenttypes`) seems to break the MyPy plugin.

Traceback:
```
version: 1.18.2
  File "mypy/checkexpr.py", line 6078, in accept
  File "mypy/checkexpr.py", line 6113, in accept_maybe_cache
  File "mypy/nodes.py", line 2248, in accept
  File "mypy/checkexpr.py", line 525, in visit_call_expr
  File "mypy/checkexpr.py", line 662, in visit_call_expr_inner
  File "mypy/checkexpr.py", line 1512, in check_call_expr_with_callee_type
  File "mypy/checkexpr.py", line 1605, in check_call
  File "mypy/checkexpr.py", line 1848, in check_callable_call
  File "mypy/checkexpr.py", line 1286, in apply_function_plugin
  File "/home/hartung/repos/olist-agents/.venv/lib/python3.13/site-packages/mypy_django_plugin/transformers/querysets.py", line 401, in specialize_prefetch_type
    lookup_type = _specialize_string_arg_to_literal(ctx, django_context, "lookup", default.args[0])
                                                                                   ~~~~~~~~~~~~^^^
IndexError: tuple index out of range
```

## System information

- OS:
- `python` version: 3.13.0
- `django` version: 5.2.6
- `mypy` version: 1.18.2
- `django-stubs` version: 5.2.5
- `django-stubs-ext` version: 5.2.5

UnknownPlatypus:
Fixes #2850 by adding proper typing to `GenericPrefetch` in stubs and slightly updating the plugin.

### DIFF
--- a/tests/typecheck/managers/querysets/test_prefetch_related.yml
+++ b/tests/typecheck/managers/querysets/test_prefetch_related.yml
@@ -384,3 +384,64 @@
                     subject_content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
                     subject_id = models.PositiveIntegerField()
                     subject = GenericForeignKey("subject_content_type", "subject_id")
+
+-   case: django_contrib_contenttypes_generic_prefetch
+    installed_apps:
+        - django.contrib.contenttypes
+        - myapp
+    main: |
+        from django.contrib.contenttypes.prefetch import GenericPrefetch
+        from myapp.models import Bookmark, Animal, TaggedItem
+        from typing_extensions import reveal_type
+
+        # Basic GenericPrefetch usage
+        prefetch = GenericPrefetch(
+            "content_object", [Bookmark.objects.all(), Animal.objects.only("name")]
+        )
+        reveal_type(prefetch)  # N: Revealed type is "django.contrib.contenttypes.prefetch.GenericPrefetch[Literal['content_object'], builtins.list[django.db.models.query.QuerySet[django.db.models.base.Model, django.db.models.base.Model]], builtins.str]"
+
+        # Using GenericPrefetch with prefetch_related
+        qs = TaggedItem.objects.prefetch_related(prefetch).all()
+        reveal_type(qs)  # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.TaggedItem, myapp.models.TaggedItem]"
+
+        # GenericPrefetch with to_attr
+        prefetch_with_attr = GenericPrefetch(
+            "content_object",
+            [Bookmark.objects.all(), Animal.objects.only("name")],
+            to_attr="prefetched_object"
+        )
+        qs_with_attr = TaggedItem.objects.prefetch_related(prefetch_with_attr).all()
+        reveal_type(qs_with_attr)  # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.TaggedItem@AnnotatedWith[TypedDict({'prefetched_object': builtins.list[Any]})], myapp.models.TaggedItem@AnnotatedWith[TypedDict({'prefetched_object': builtins.list[Any]})]]"
+        reveal_type(qs_with_attr.get().prefetched_object)  # N: Revealed type is "builtins.list[Any]"
+
+        # GenericPrefetch on invalid field (not a GenericForeignKey)
+        regular_fk_prefetch = GenericPrefetch(
+            "content_type",
+            [Bookmark.objects.all(), Animal.objects.only("name")],
+        )
+        TaggedItem.objects.prefetch_related(regular_fk_prefetch).all() # E: "content_type" on "TaggedItem" is not a GenericForeignKey, GenericPrefetch can only be used with GenericForeignKey fields  [misc]
+        regular_field_prefetch = GenericPrefetch(
+            "tag",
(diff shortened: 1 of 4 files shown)


tests/typecheck/managers/querysets/test_prefetch_related.yml
```diff
@@ -380,3 +380,64 @@
                     subject_content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
                     subject_id = models.PositiveIntegerField()
                     subject = GenericForeignKey("subject_content_type", "subject_id")
+
+-   case: django_contrib_contenttypes_generic_prefetch
+    installed_apps:
+        - django.contrib.contenttypes
```
sobolevn, line 390: please, add a test with `GenericPrefetch` but without this app

UnknownPlatypus: Any idea why [we automatically add](https://github.com/typeddjango/django-stubs/blob/8bd487b968fa2c5b5fae4139df72c79fa93d175d/scripts/tests_extension_hook.py#L15) `django.contrib.contenttypes` in every YamlTest ? x)

Makes it difficult to have a test without

sobolevn: Nope :)

sobolevn: It should be removed ideally. Too magical

UnknownPlatypus: Hum actually I think i understand why, if you remove this app, then any import in `django.contrib.admin` or `django.contrib.auth` also crashes like so:

```
  File "/home/thibaut/workspace/django-stubs/.venv/lib/python3.13/site-packages/django/contrib/admin/models.py", line 6, in <module>
    from django.contrib.contenttypes.models import ContentType
  File "/home/thibaut/.local/share/JetBrains/Toolbox/apps/pycharm-professional/plugins/python-ce/helpers/pydev/_pydev_bundle/pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
  File "/home/thibaut/workspace/django-stubs/.venv/lib/python3.13/site-packages/django/contrib/contenttypes/models.py", line 134, in <module>
    class ContentType(models.Model):
    ...<53 lines>...
            return (self.app_label, self.model)
  File "/home/thibaut/workspace/django-stubs/.venv/lib/python3.13/site-packages/django/db/models/base.py", line 136, in __new__
    raise RuntimeError(
    ...<3 lines>...
    )
RuntimeError: Model class django.contrib.contenttypes.models.ContentType doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

UnknownPlatypus: My test without `django.contrib.contenttypes` has the same issue because of the `from django.contrib.contenttypes.models import ContentType` import .
Even only having the `from django.contrib.contenttypes.fields import GenericForeignKey` causes this because it tries to import `django.contrib.contenttypes.models` and trigger the same issue.

The best i can do is this kind of test, not sure it's worth it ? wdyt? 

```yaml
-   case: uninstalled_django_contrib_contenttypes_generic_prefetch
    installed_apps:
        - myapp
    main: |
        from django.contrib.contenttypes.prefetch import GenericPrefetch
        from myapp.models import Bookmark, Animal, TaggedItem

        # Basic GenericPrefetch usage
        prefetch = GenericPrefetch(
            "content_object", [Bookmark.objects.all(), Animal.objects.only("name")]
        )
        reveal_type(prefetch)  # N: Revealed type is "django.contrib.contenttypes.prefetch.GenericPrefetch[Literal['content_object'], builtins.list[django.db.models.query.QuerySet[django.db.models.base.Model, django.db.models.base.Model]], builtins.str]"

        # Using GenericPrefetch with prefetch_related
        qs = TaggedItem.objects.prefetch_related(prefetch).all() # E: Cannot find "content_object" on "TaggedItem" object, "content_object" is an invalid parameter to "prefetch_related()"  [misc]
        reveal_type(qs)  # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.TaggedItem, myapp.models.TaggedItem]"


    files:
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                from django.db import models

                class Bookmark(models.Model):
                    url = models.URLField()

                class Animal(models.Model):

                class TaggedItem(models.Model):
                    tag = models.CharField(max_length=100)
```

sobolevn:
Merged https://github.com/typeddjango/django-stubs/pull/2853
Now this test can be added :)

UnknownPlatypus:
> Now this test can be added :)

Done!

sobolevn (approved):

sobolevn:
Thanks a lot!

PR #2853, opened 2025-10-04
Remove magically added `django.contrib.contenttypes` installed app in tests

UnknownPlatypus:
Instead, explicitly declare it when required.

See discussion in https://github.com/typeddjango/django-stubs/pull/2851#discussion_r2404147314


sobolevn (approved):
Thanks a lot!

PR #2854, opened 2025-10-05
Fix merge conflict + re-run pre-commit on all files

UnknownPlatypus:
Fixes master ci (due to a conflict between two recent merge)

cf https://github.com/typeddjango/django-stubs/pull/2836#issuecomment-3368832222


UnknownPlatypus:
cc @sobolevn

PR #2855, opened 2025-10-05
Enable more ruff rules

UnknownPlatypus:
# I have made things!

(_Review commit per commit should be easier_)

Followup to #2852, adding a few ruff rules I often use.


tests/assert_type/contrib/admin/test_utils.py
```diff
@@ -30,7 +30,7 @@ class PersonTupleAdmin(admin.ModelAdmin[Person]):
 
 
 class PersonFieldsetListAdmin(admin.ModelAdmin[Person]):
-    fieldsets = [
+    fieldsets = [  # noqa: RUF012
```
sobolevn: I propose to ignore this rule

UnknownPlatypus: I don't know, I feel it can catch future errors but sadly because of the `ClassVar` / `TypeVar` limitation, these few admin fields are not declared as `Classvar`. and trigger this rule.

What about ignoring for files matching the glob `"*/admin*.py"` ?

sobolevn: These are tests, they are not real code. I don't see a valid reason to force this rule on tests files at all.

UnknownPlatypus: Ye that make sense, but we probably still want it in the python plugin code for ex.

mypy_django_plugin/lib/helpers.py
```diff
@@ -75,11 +76,11 @@ class DjangoTypeMetadata(TypedDict, total=False):
 
 
 def get_django_metadata(model_info: TypeInfo) -> DjangoTypeMetadata:
-    return cast(DjangoTypeMetadata, model_info.metadata.setdefault("django", {}))
+    return cast("DjangoTypeMetadata", model_info.metadata.setdefault("django", {}))
```
intgr, line 79: Nit: Not opposed to this, but my first instinct is that these changes reduce readability.

From Ruff docs `runtime-cast-value (TC006)` https://docs.astral.sh/ruff/rules/runtime-cast-value/#why-is-this-bad

> In some cases where `cast()` is used in a hot loop, this rule may also help avoid overhead from repeatedly evaluating complex type expressions at runtime.

I guess that's a point sometimes. But doesn't really apply when the type expression is a simple type reference.

tests/assert_type/contrib/admin/test_utils.py
intgr: So, ignore for tests using `[tool.ruff.lint.per-file-ignores]`?

mypy_django_plugin/lib/helpers.py
UnknownPlatypus: > Nit: Not opposed to this, but my first instinct is that these reduce readability.

Agree, but I'm even wondering if we should really be using cast in these cases, a `type: ignore` makes more sense here for example and completely removes the function call

The perf argument does seem compelling since we might be in hot loops when running mypy

intgr (approved):

sobolevn (approved):

PR #2808, opened 2025-09-11
Validation of the `fields` argument to `bulk_update` 

UnknownPlatypus:
# I have made things!

Mostly mirrors validation logic from `django.db.models.query.QuerySet.bulk_update`.
Tests are inspired from `django/tests/queries/test_bulk_update.py`

I could have called a dummy `bulk_update` with no objects to reuse django internal logic but it felt brittle with risk of someday actually do database query so I preferred to re-implement the few checks

PR #2856, opened 2025-10-05
Also typecheck `aget` calls

UnknownPlatypus:
# I have made things!

We missed this async variant so it's argument where not typechecked, now they are :partying_face: 

cf https://github.com/typeddjango/django-stubs/pull/2808#discussion_r2339293315


sobolevn (approved):
Thanks! Good catch!

PR #2857, opened 2025-10-05
Use same `uv` workspace for `django-stubs-ext` also

intgr:
Previously, both the top-level project and the `ext` subdirectory had their own `.venv`, their own `uv.lock` (although `ext/uv.lock` was never committed to git).

With a shared workspace, these now only exist at the top level.

Also, for the release workflow, the top-level `dist/` output directory is now shared -- no more `ext/dist/`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,21 +16,17 @@ jobs:
     permissions:
       id-token: write
     steps:
-      - name: Setup python to build package
-        uses: actions/setup-python@v6
+      - name: Install uv & Python
+        uses: astral-sh/setup-uv@v6
         with:
-          python-version: '3.12'
-      - name: Install build
-        run: python -m pip install build
+          python-version: "3.13"
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
       - name: Build ext package
-        run: python -m build ext/
+        run: uv build --package=django-stubs-ext
       - name: Publish ext to PyPI
         uses: pypa/gh-action-pypi-publish@v1.13.0
-        with:
-          packages-dir: ext/dist/
 
   build-and-publish:
     runs-on: ubuntu-latest
@@ -49,6 +45,6 @@ jobs:
         with:
           fetch-depth: 0
       - name: Build package
-        run: uv build
+        run: uv build --package=django-stubs
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@v1.13.0
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -180,5 +180,4 @@ jobs:
         run: env --unset=UV_FROZEN uv lock --check
       - name: Build
         run: |
-          uv build .
-          uv build ext/
+          uv build --all
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -179,5 +179,4 @@ jobs:
       - name: Ensure uv.lock is up to date
         run: env --unset=UV_FROZEN uv lock --check
       - name: Build
-        run: |
-          uv build --all
+        run: uv build --all-packages
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,7 +74,10 @@ Funding = "https://github.com/sponsors/typeddjango"
 
 [tool.uv.sources]
 django-stubs = { workspace = true }
-django-stubs-ext = { path = "ext", editable = true }
+django-stubs-ext = { workspace = true }
+
+[tool.uv.workspace]
+members = ["ext"]
 
 [build-system]
 requires = ["uv_build>=0.8.22,<0.9.0"]
--- a/uv.lock
+++ b/uv.lock
@@ -2,6 +2,12 @@ version = 1
 revision = 3
 requires-python = ">=3.10"
 
+[manifest]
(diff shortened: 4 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -16,21 +16,17 @@ jobs:
...
-          python-version: '3.12'
-      - name: Install build
-        run: python -m pip install build
+          python-version: "3.13"
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
       - name: Build ext package
-        run: python -m build ext/
+        run: uv build --package=django-stubs-ext
```
intgr, line 27: * Oops -- this was omitted in the previous PR https://github.com/typeddjango/django-stubs/pull/2848

.github/workflows/release.yml
```diff
@@ -16,21 +16,17 @@ jobs:
...
       - name: Publish ext to PyPI
         uses: pypa/gh-action-pypi-publish@v1.13.0
-        with:
-          packages-dir: ext/dist/
```
intgr, line 33: Top-level `dist/` directory is now used for both packages.

.github/workflows/release.yml
UnknownPlatypus: no worries, they behave the same anyway

.github/workflows/release.yml
```diff
@@ -49,6 +45,6 @@ jobs:
         with:
           fetch-depth: 0
       - name: Build package
-        run: uv build
+        run: uv build --package=django-stubs
```
intgr, line 48: In theory, we could use `uv build --all` here and publish in one GitHub job.

But I couldn't figure out whether `pypa/gh-action-pypi-publish` supports atomic uploads for multiple different packages.

So the original concern for splitting GitHub jobs may be still valid. --> https://github.com/typeddjango/django-stubs/pull/2183#issuecomment-2133327132

> A problem that then shows up is that if e.g. django-stubs-ext succeeds but django-stubs fails, it's not possible to rerun the release action.

.github/workflows/release.yml
UnknownPlatypus: how does it look inside ? 

Is there a risk we upload part of django-stub-ext in django-stubs or the other way around ?

And does it mean that a bare `uv publish` will publish both ?

intgr: > Is there a risk we upload part of django-stub-ext in django-stubs or the other way around ?

Not as long as `uv build` respects the `--package=` switch. It only creates package files for one or the other package, not both.

> And does it mean that a bare uv publish will publish both ?

Seems like it just looks at the `dist/` directory and uploads whatever is present.

```
% uv publish --dry-run
error: No files found to publish

% uv build --package=django-stubs
...
Successfully built dist/django_stubs-5.2.5.tar.gz
Successfully built dist/django_stubs-5.2.5-py3-none-any.whl
% uv publish --dry-run
Checking 2 files against https://upload.pypi.org/legacy/
Enter username ('__token__' if using a token): ^C

% uv build --package=django-stubs-ext
...
Successfully built dist/django_stubs_ext-5.2.5.tar.gz
Successfully built dist/django_stubs_ext-5.2.5-py3-none-any.whl
% uv publish --dry-run
Checking 4 files against https://upload.pypi.org/legacy/
Enter username ('__token__' if using a token): ^C
```

intgr: > how does it look inside ?

After running both build commands it has 4 files:

```
django_stubs-5.2.5-py3-none-any.whl
django_stubs-5.2.5.tar.gz
django_stubs_ext-5.2.5-py3-none-any.whl
django_stubs_ext-5.2.5.tar.gz
```

UnknownPlatypus: Then I suppose a simple uv publish might be enough here, I'm using [this action](https://github.com/UnknownPlatypus/djangofmt/blob/main/.github/workflows/release-publish-pypi.yml#L15) for one of my packages ? 

We can probably test with test pypi `uv publish --publish-url https://test.pypi.org/legacy/`

.github/workflows/release.yml
sobolevn: I think one job is even better: because now the first one can finish and the second one can fail.

sobolevn (commented):
I like the idea! Let's switch the release action as well :)

.github/workflows/release.yml
intgr: > I suppose a simple uv publish might be enough here

We don't use manually configured PyPI credentials like the one linked above. There's some authentication magic that happens between GitHub Actions and PyPI.

No idea how it works, I'd rather not touch it myself. :)

sobolevn:
or not :)
because there are different envs and pypi config is not that simple, let's leave it as-is with two jobs for now. sorry!

intgr:
Once we merge this, I will make an "alpha" release to verify that I haven't broken the release workflow.

If it succeeds, I'll create a stable release as well.

.github/workflows/release.yml
UnknownPlatypus: > There's some authentication magic that happens between GitHub Actions and PyPI.

oh right, that's probably good to keep indeed

sobolevn (approved):

PR #2860, opened 2025-10-06
Fix PK field lookup for Django < 5.2

ngnpope:
The `pk_fields` attribute was added in Django 5.2 to support composite primary keys. Fall back to using the `pk` attribute in earlier versions of Django.

Regression in 736e30f591d59d044248c15097bfd999ae659985.

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -784,9 +784,9 @@ def _validate_bulk_update_field(
         ctx.api.fail(f'"{method}()" can only be used with concrete fields. Got "{field_name}"', ctx.context)
         return False
 
-    all_pk_fields = set(opts.pk_fields)
+    all_pk_fields = set(getattr(opts, "pk_fields", [opts.pk]))
     for parent in opts.all_parents:
-        all_pk_fields.update(parent._meta.pk_fields)
+        all_pk_fields.update(getattr(parent._meta, "pk_fields", [parent._meta.pk]))
 
     if field in all_pk_fields:
         ctx.api.fail(f'"{method}()" cannot be used with primary key fields. Got "{field_name}"', ctx.context)


sobolevn (approved):
Thanks!

sobolevn:
Question: what would be the easiest way to catch this in CI?

ngnpope:
> Question: what would be the easiest way to catch this in CI?

Not sure there is one without running the test suite against multiple versions of Django. That could also prove challenging as we may have to somehow flag certain tests as only compatible with a subset of versions. I guess that using django-stubs 5.2.x with a version of Django earlier than 5.2 is only best-effort anyway?

sobolevn:
Yes, we can try 5.0 as the lower bound, but we should not care about typing errors, only crashes. Because different django versions might give different typing errors, which we can't fully support and maintain.

mypy_django_plugin/transformers/querysets.py
```diff
@@ -784,7 +784,7 @@ def _validate_bulk_update_field(
         ctx.api.fail(f'"{method}()" can only be used with concrete fields. Got "{field_name}"', ctx.context)
         return False
 
-    all_pk_fields = set(opts.pk_fields)
+    all_pk_fields = set(getattr(opts, "pk_fields", [opts.pk]))
     for parent in opts.all_parents:
         all_pk_fields.update(parent._meta.pk_fields)
```
UnknownPlatypus: We probably need to do the same here ?

UnknownPlatypus:
> Yes, we can try 5.0 as the lower bound, but we should not care about typing errors, only crashes. Because different django versions might give different typing errors, which we can't fully support and maintain.

Seem like a decent compromise, is it easy to distinguish between a crash and regular mypy errors when using `pytest-mypy-plugins` ?

btw, we don't have any other typing errors (apart from the above crash, and `test_find_classes_inheriting_from_generic` that would need special casing) when running the whole test suite with django==5.0.0

```
____________________________________________________________________________________________________ bulk_update_invalid_fields ____________________________________________________________________________________________________
[gw8] linux -- Python 3.13.1 /home/thibaut/workspace/django-stubs/.venv/bin/python3
/home/thibaut/workspace/django-stubs/tests/typecheck/managers/querysets/test_bulk_update.yml:100: 
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Critical error occurred
_____________________________________________________________________________________________________ bulk_update_valid_fields _____________________________________________________________________________________________________
[gw5] linux -- Python 3.13.1 /home/thibaut/workspace/django-stubs/.venv/bin/python3
/home/thibaut/workspace/django-stubs/tests/typecheck/managers/querysets/test_bulk_update.yml:4: 
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Critical error occurred
________________________________________________________________________________________________ queryset_basic_methods_return_type ________________________________________________________________________________________________
[gw2] linux -- Python 3.13.1 /home/thibaut/workspace/django-stubs/.venv/bin/python3
/home/thibaut/workspace/django-stubs/tests/typecheck/managers/querysets/test_basic_methods.yml:2: 
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Critical error occurred
===================================================================================================== short test summary info ======================================================================================================
FAILED tests/test_generic_consistency.py::test_find_classes_inheriting_from_generic - AttributeError: module 'django.contrib.auth.forms' has no attribute 'SetUnusablePasswordMixin'
FAILED tests/typecheck/managers/querysets/test_bulk_update.yml::bulk_update_invalid_fields - 
FAILED tests/typecheck/managers/querysets/test_bulk_update.yml::bulk_update_valid_fields - 
FAILED tests/typecheck/managers/querysets/test_basic_methods.yml::queryset_basic_methods_return_type - 
```

mypy_django_plugin/transformers/querysets.py
ngnpope: 🤦🏻 Yes. Sorry - rushed this as I don't have much time right now. Pushed an update.

sobolevn:
> is it easy to distinguish between a crash and regular mypy errors when using `pytest-mypy-plugins`?

We can add any logic there, this is my project :)

Right now - I don't think that it is that easy. But doable with some `bash` magic.
https://github.com/typeddjango/pytest-mypy-plugins/blob/06a9e766441185301d602b27c794d6f80827edcb/pytest_mypy_plugins/item.py#L237-L239

UnknownPlatypus:
Ye maybe greping for "pytest_mypy_plugins.utils.TypecheckAssertionError: Critical error occurred" and return the correct exit code based on that is enough

UnknownPlatypus (approved):

sobolevn:
Would anyone like to work on such a CI job? :)

UnknownPlatypus:
> Would anyone like to work on such a CI job? :)

Will do, should not be that hard

intgr:
Was this a regression in django-stubs 5.2.6? Should we do a fix release ASAP?

ngnpope:
> Was this a regression in django-stubs 5.2.6? Should we do a fix release ASAP?

Yes, it was. Probably worthwhile as it causes a crash and I don't think we've got any other fixes in the pipeline right now to wait for.

intgr:
It's released https://pypi.org/project/django-stubs/5.2.7/

UnknownPlatypus:
thanks @intgr , should we [yank](https://docs.pypi.org/project-management/yanking/) the 5.2.6 release on pypi ?

intgr:
I don't think it warrants yanking. In any case I don't have privileges to do it.

PR #2862, opened 2025-10-07
Run tests against the lowest supported django version

UnknownPlatypus:
# I have made things!

Followup to https://github.com/typeddjango/django-stubs/pull/2860#issuecomment-3370448900

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -77,6 +77,36 @@ jobs:
       - name: Run pytest tests
         run: uv run pytest --num-shards=4 --shard-id=${{ matrix.shard }} -n auto tests
 
+  test-older-django:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        shard: [0, 1, 2, 3]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v5
+      - name: Setup system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install binutils libproj-dev gdal-bin
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+      - name: Set up Python 3.13
+        uses: actions/setup-python@v6
+        with:
+          python-version: 3.13
+
+      - name: Install dependencies
+        run: uv sync --no-dev --group tests
+
+      - name: Downgrade Django
+        run: uv pip install django==5.0.0
+
+      # Exit 1 only if ont of the pytest error is a mypy crash
+      - name: Run pytest tests
+        run: "! uv run pytest --num-shards=4 --shard-id=${{ matrix.shard }} -n auto tests | grep 'pytest_mypy_plugins.utils.TypecheckAssertionError: Critical error occurred'"
+
 
   stubtest:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -77,6 +77,36 @@ jobs:
...
+
+      - name: Install dependencies
+        run: uv sync --no-dev --group tests
+
+      - name: Downgrade Django
+        run: uv pip install django==5.0.0
+
+      # Exit 1 only if ont of the pytest error is a mypy crash
+      - name: Run pytest tests
+        run: "! uv run pytest --num-shards=4 --shard-id=${{ matrix.shard }} -n auto tests | grep 'pytest_mypy_plugins.utils.TypecheckAssertionError: Critical error occurred'"
```
sobolevn, line 108: what does `!` mean here?

mypy_django_plugin/transformers/querysets.py
```diff
@@ -784,7 +784,7 @@ def _validate_bulk_update_field(
         ctx.api.fail(f'"{method}()" can only be used with concrete fields. Got "{field_name}"', ctx.context)
         return False
 
-    all_pk_fields = set(getattr(opts, "pk_fields", [opts.pk]))
```
sobolevn, line 787: Please, don't forget to remove after the CI will fail :)

.github/workflows/test.yml
UnknownPlatypus: > If the reserved word ‘!’ precedes the pipeline, the exit status is the logical negation of the exit status as described above

This inverts the error code of the pipeline. See https://www.gnu.org/software/bash/manual/bash.html#Pipelines-1

This is useful here because grep exit `0` on match and exit `1` otherwise but since we want to error on matches, we want to invert it and `!` does that

mypy_django_plugin/transformers/querysets.py
mlee528, line 1: Oh, this is great. My team is currently on Django 4.2, and this caused an issue while upgrading our dependencies, since `all_parents` isn’t available in 4.2 😢

UnknownPlatypus:
This is ready for review.

- here is a [failing pipeline](https://github.com/typeddjango/django-stubs/actions/runs/18306222905/job/52123973797?pr=2862) with #2860 reverted
- I then reverted the change and pipeline was [still failing](https://github.com/typeddjango/django-stubs/actions/runs/18354674177/job/52283474369?pr=2862) on another api only added in [django 5.1](https://github.com/django/django/pull/17893)
- This is now fixed

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: ye, just discovered this. Great we test for this now.

We'll try to include this fix in a release shortly

mlee528: Sweet, thank you:)

sobolevn (approved):
Great! Let's try this! 

Thanks a lot for your help.

allanlewis:
Are we able to get a release with this change? I think it will fix things with my project. (The previous version works fine in the meantime.)

PR #2869, opened 2025-10-11
fix: allow `django.shortcuts.resolve_url` to receive a str-promise

### LINKED ISSUE #2867: `LOGIN_URL` declared as `_StrOrPromise`, should be just `str`
`LOGIN_URL` is declared as `_StrOrPromise` in `django-stubs/conf/global_settings.pyi`. But according to the docs at https://docs.djangoproject.com/en/5.2/ref/settings/#login-url it should be a `str`:

> The URL or [named URL pattern](https://docs.djangoproject.com/en/5.2/topics/http/urls/#naming-url-patterns) where requests are redirected for login

Even internally it's passed to `django.shortcuts.resolve_url` (see source code for `user_passes_test`). When I do it (`resolve_url(settings.LOGIN_URL)`), I get:

```
Argument 1 to "resolve_url" has incompatible type "str | _StrPromise"; expected "Callable[..., Any] | Model | str"
```

- `python` version: 3.11.12
- `django` version: 5.2.7
- `mypy` version: 1.13.0

alexei:
Fixes https://github.com/typeddjango/django-stubs/issues/2867


UnknownPlatypus (approved):
Thanks !

PR #2870, opened 2025-10-11
fix: add type declarations for some missing settings

### LINKED ISSUE #2868: 'Settings' object has no attribute 'ROOT_URLCONF'
`settings.ROOT_URLCONF` has existed since forever -- or at least from:
1.8 https://docs.djangoproject.com/en/1.8/ref/settings/#root-urlconf to:
5.2 https://docs.djangoproject.com/en/5.2/ref/settings/#root-urlconf

- `python` version: 3.11.12
- `django` version: 5.2.7
- `mypy` version: 1.13.0

alexei:
Fixes https://github.com/typeddjango/django-stubs/issues/2868

### DIFF
--- a/django-stubs/conf/global_settings.pyi
+++ b/django-stubs/conf/global_settings.pyi
@@ -77,6 +77,10 @@ MANAGERS: _Admins
 # Content-Type header.
 DEFAULT_CHARSET: str
 
+# A dictionary of modules containing serializer definitions (provided as strings), keyed by a string identifier for
+# that serialization type. For example, to define a YAML serializer, use: `{"yaml": "path.to.yaml_serializer"}`
+SERIALIZATION_MODULES: dict[str, str]
+
 # Email address that error messages come from.
 SERVER_EMAIL: str
 
@@ -170,6 +174,9 @@ ABSOLUTE_URL_OVERRIDES: dict[str, Any]
 #    ]
 IGNORABLE_404_URLS: list[Pattern[str]]
 
+# A string representing the full Python import path to your root URLconf, for example "mydjangoapps.urls".
+ROOT_URLCONF: str
+
 # A secret key for this particular Django installation. Used in secret-key
 # hashing algorithms. Set this in your settings, or Django will complain
 # loudly.
@@ -188,6 +195,9 @@ MEDIA_ROOT: str
 # URL that handles the media served from MEDIA_ROOT.
 # Examples: "http://example.com/media/", "http://media.example.com/"
 MEDIA_URL: str
+
+# The ID, as an integer, of the current site in the `django_site` database table.
+SITE_ID: int
 
 # Absolute path to the directory static files should be collected to.
 # Example: "/var/www/example.com/static/"
--- a/django-stubs/conf/global_settings.pyi
+++ b/django-stubs/conf/global_settings.pyi
@@ -77,9 +77,9 @@ MANAGERS: _Admins
 # Content-Type header.
 DEFAULT_CHARSET: str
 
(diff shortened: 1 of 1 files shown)


alexei:
Looks like these new additions result in "not present at runtime" errors. Is it because these properties are not defined by default? If so, what's the course of action?

django-stubs/conf/global_settings.pyi
```diff
@@ -189,6 +199,9 @@ MEDIA_ROOT: str
 # Examples: "http://example.com/media/", "http://media.example.com/"
 MEDIA_URL: str
 
+# The ID, as an integer, of the current site in the `django_site` database table.
+SITE_ID: int
```
```python
In [1]: from django.conf import settings

In [2]: settings.SITE_ID
Cell In[2], line 1
----> 1 settings.SITE_ID

File /.venv/lib/python3.13/site-packages/django/conf/__init__.py:83, in LazySettings.__getattr__(self, name)
     81     self._setup(name)
     82     _wrapped = self._wrapped
---> 83 val = getattr(_wrapped, name)
     85 # Special case some settings which require further modification.
     86 # This is done here for performance reasons so the modified value is cached.
     87 if name in {"MEDIA_URL", "STATIC_URL"} and val is not None:

AttributeError: 'Settings' object has no attribute 'SITE_ID'
```

It is not always present.

django-stubs/conf/global_settings.pyi
```diff
@@ -96,6 +100,9 @@ DATABASE_ROUTERS: list[str | Router]
 # to a module that defines an EmailBackend class.
 EMAIL_BACKEND: str
 
+# The directory used by the file email backend to store output files.
+EMAIL_FILE_PATH: str
```
sobolevn: It is not present for me by default:

```python
In [4]: settings.EMAIL_FILE_PATH
Cell In[4], line 1
----> 1 settings.EMAIL_FILE_PATH

File .venv/lib/python3.13/site-packages/django/conf/__init__.py:83, in LazySettings.__getattr__(self, name)
     81     self._setup(name)
     82     _wrapped = self._wrapped
---> 83 val = getattr(_wrapped, name)
     85 # Special case some settings which require further modification.
     86 # This is done here for performance reasons so the modified value is cached.
     87 if name in {"MEDIA_URL", "STATIC_URL"} and val is not None:

AttributeError: 'Settings' object has no attribute 'EMAIL_FILE_PATH'
```

django-stubs/conf/global_settings.pyi
```diff
@@ -77,6 +77,10 @@ MANAGERS: _Admins
 # Content-Type header.
 DEFAULT_CHARSET: str
 
+# A dictionary of modules containing serializer definitions (provided as strings), keyed by a string identifier for
+# that serialization type. For example, to define a YAML serializer, use: `{"yaml": "path.to.yaml_serializer"}`
+SERIALIZATION_MODULES: dict[str, str]
```
sobolevn: Also missing for me :(

alexei:
@sobolevn I get it now: because these attributes are by default undefined, one needs to access them via e.g. `getattr(settings, "ROOT_URLCONF", None)` as opposed to `settings.ROOT_URLCONF`.

What do you think about mentioning them anyway? I could add some comments explaining why `ROOT_URLCONF` etc. are not declared. So if someone's looking, they can learn why they don't exist and how to deal with it. Similar in spirit to how `MESSAGE_LEVEL` and `MESSAGE_TAGS` are mentioned:

https://github.com/typeddjango/django-stubs/blob/532be331b44708b3c032a2e985024c48dec31b24/django-stubs/conf/global_settings.pyi#L449-L450

sobolevn:
Comments are always welcome!

sobolevn (approved):

PR #2871, opened 2025-10-12
fix: import all forms widgets

alexei:
I found `ColorInput`, `MediaDefiningClass`, `Script`, `SearchInput` and `TelInput`, although declared, were missing from `django.forms`. I guess this can be avoided by importing everything from `widgets`, just like Django does:

https://github.com/django/django/blob/315dbe675df338ae66c8fa43274a76ecbed7ef67/django/forms/__init__.py#L11


sobolevn (approved):

sobolevn:
Thanks!

UnknownPlatypus:
Shouldn't `stubtest` have caught that ?

sobolevn:
Good question 🤔

intgr:
stubtest doesn't know which imports in original code are intended for re-export, and which imports are implementation details and not meant for user code. Python stubs strictly distinguish the two kinds of usages.

PR #2874, opened 2025-10-13
Make `View`, `TemplateView`, and `TemplateResponseMixin` generic on response type

### LINKED ISSUE #2873: `as_view` always returns a callable that returns a base response, which is not specific enough
# Bug report

## What's wrong

The return type of `as_view` is always a `def (*Any, **Any) -> django.http.response.HttpResponseBase`, so the response returned from the view function is always `django.http.response.HttpResponseBase`. This means that several attributes are missing from, for example, a template response:

```python

from django.views.generic import TemplateView
from django.test import RequestFactory


class MyView(TemplateView):
    template_name = "template.html"


rf = RequestFactory()
request = rf.get("/")

view = MyView.as_view()
reveal_type(view)

response = view(request)
reveal_type(response)

print(response.rendered_content)
```

```console
$ mypy t.py
t.py:13: note: Revealed type is "def (*Any, **Any) -> django.http.response.HttpResponseBase"
t.py:16: note: Revealed type is "django.http.response.HttpResponseBase"
t.py:18: error: "HttpResponseBase" has no attribute "rendered_content"  [attr-defined]
```

## How is that should be

Calling `as_view` on a `TemplateView` returns a function that returns a `django.template.response.TemplateResponse`. The type checker should know this.

## System information

- OS: macOS 15.7
- `python` version: 3.13.7
- `django` version: 5.2.7
- `mypy` version: 1.18.2
- `django-stubs-ext` version: 5.2.7

samueljsb:
This allows us to specify that the default response type for template
views is a `TemplateResponse`.

N.B. The `LogoutView` may return a response that is not a template
response (i.e. a redirect), so we annotate it as returning a less
specific type. We have not been more specific about exactly which types
it returns because doing so might require it to also become generic to
support subclasses that return different specific response types.

We could take this further and make more generic view subclasses generic,
but wanted to verify this approach with y'all before pushing too far ahead. 

## Related issues

Fixes #2873

### DIFF
--- a/django-stubs/contrib/auth/views.pyi
+++ b/django-stubs/contrib/auth/views.pyi
@@ -25,7 +25,7 @@ class LoginView(RedirectURLMixin, FormView[AuthenticationForm]):
     extra_context: Any
     def get_redirect_url(self) -> str: ...
 
-class LogoutView(RedirectURLMixin, TemplateView):
+class LogoutView(RedirectURLMixin, TemplateView[HttpResponse]):
     next_page: str | None
     redirect_field_name: str
     extra_context: Any
--- a/django-stubs/contrib/auth/views.pyi
+++ b/django-stubs/contrib/auth/views.pyi
@@ -29,7 +29,7 @@ class LoginView(RedirectURLMixin, FormView[_AuthForm]):
     redirect_authenticated_user: bool
     extra_context: Mapping[str, Any] | None
 
-class LogoutView(RedirectURLMixin, TemplateView):
+class LogoutView(RedirectURLMixin, TemplateView[HttpResponse]):
     next_page: str | None
     redirect_field_name: str
     extra_context: Mapping[str, Any] | None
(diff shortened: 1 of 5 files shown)


UnknownPlatypus:
The mypy primmer error is a genuine error on their end, they even do `assert isinstance(result, TemplateResponse)` so we know result is a `TemplateResponse` here

```python
    @override
    def get(
        self, request: HttpRequest, *args: object, article: str = "", **kwargs: object
    ) -> HttpResponse:
        # Hack: It's hard to reinitialize urls.py from tests, and so
        # we want to defer the use of settings.POLICIES_DIRECTORY to
        # runtime.
        if self.policies_view:
            self.path_template = f"{settings.POLICIES_DIRECTORY}/%s.md"

        documentation_article = self.get_path(article)
        http_status = documentation_article.article_http_status
        result = super().get(request, article=article)
        assert isinstance(result, TemplateResponse)
        for callback in self._post_render_callbacks:
            result.add_post_render_callback(callback)
        if http_status != 200:
            result.status_code = http_status
        return result
```

PR #2881, opened 2025-10-21
CI: Add pre-commit config to update `uv.lock`

intgr:


intgr:
Looks like it doesn't work because there's no network access from `pre-commit.ci`? Not sure whether it's a bug or on purpose.

UnknownPlatypus:
> Looks like it doesn't work because there's no network access from pre-commit.ci? Not sure whether it's a bug or on purpose.

It's on purpose and part of [the reasons why pre-commit.ci is very fast](https://pre-commit.ci/lite#differences-with-full-version). We could use pre-commit.ci lite to make this work but I'm not sure it's worth ? Only maintainer will change packages and we all use pre-commit locally.

It's probably fine to skip it in pre-commit.ci since it's already checked in the `build-and-check` step ?
```yaml
# .pre-commit-config.yaml
ci:
  skip:
    - uv-lock
```

intgr:
> Only maintainer will change packages and we all use pre-commit locally.

I found this pre-commit integration and thought it would be one less manual step when making a release. I don't use pre-commit locally. But it's not a big deal at all.

UnknownPlatypus:
> I found this pre-commit integration and thought it would be one less manual step when making a release.

I'm very happy to have this, I personally always have this in my pre-commit setup if I use uv.

> I don't use pre-commit locally

You should try :) It's rather handy (if people keep the hooks fast!)

UnknownPlatypus (approved):

UnknownPlatypus:
btw we should probably freeze the pre-commit deps for security / reproducibility.
A simple `pre-commit autoupdate --freeze` should do the trick (and pre-commit understand that and will do the same when submitting updates)

PR #2888, opened 2025-10-31
Remove `str` from enum label types when we know for sure the labels are lazy

### LINKED ISSUE #2887: Why do enum labels have type _StrOrPromise rather than _StrPromise?
Hi there! First of all, thank you for creating django-stubs - I'm a maintainer of a type checker called [pyrefly](https://github.com/facebook/pyrefly), and we've been building support for django on top of django-stubs.

I had a question about some of the asserted types in [tests/assert_type/db/models/test_enums.py](https://github.com/typeddjango/django-stubs/blob/master/tests/assert_type/db/models/test_enums.py). The `gettext_lazy` function is annotated as returning `_StrPromise`, but calls to `gettext_lazy` have asserted types of `_StrOrPromise`, e.g.:
```
class Suit(IntegerChoices):
    DIAMOND = 1, _("Diamond")
    SPADE = 2, _("Spade")
    HEART = 3, _("Heart")
    CLUB = 4, _("Club")
assert_type(Suit.CLUB.label, _StrOrPromise)  # why is this not _StrPromise?

class Vehicle(IntegerChoices):
    CAR = 1, "Carriage"
    TRUCK = 2
    JET_SKI = 3

    __empty__ = _("(Unknown)")
assert_type(Vehicle.__empty__, _StrOrPromise)  # why is this not _StrPromise?
```
Is there some Django magic I'm not aware of that is widening these types? Or is mypy inferring types that are less precise than they could be?

I'm pretty unfamiliar with Django, so apologies if this is a silly question. Thanks in advance!

rchen152:
# I have made things!

Improves the mypy plugin to drop `str` from the types of labels created via `gettext_lazy`.

Instead of calculating and passing around a `bool` indicating whether an enum has any lazy label, the code now calculates a `_LabelLaziness` enum with three members - `LAZY`, `NON_LAZY`, and `MIXED` - and drops `str` from the label type for `_LabelLaziness.LAZY`.

## Related issues

Closes https://github.com/typeddjango/django-stubs/issues/2887

### DIFF
--- a/mypy_django_plugin/transformers/choices.py
+++ b/mypy_django_plugin/transformers/choices.py
@@ -1,4 +1,4 @@
-from enum import auto, Enum
+from enum import Enum, auto
 
 from mypy.nodes import MemberExpr, NameExpr, SuperExpr, TypeAlias, TypeInfo, Var
 from mypy.plugin import AttributeContext
--- a/mypy_django_plugin/transformers/choices.py
+++ b/mypy_django_plugin/transformers/choices.py
@@ -1,3 +1,5 @@
+from enum import auto, Enum
+
 from mypy.nodes import MemberExpr, NameExpr, SuperExpr, TypeAlias, TypeInfo, Var
 from mypy.plugin import AttributeContext
 from mypy.typeanal import make_optional_type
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Great stuff! Thanks a lot!

ngnpope:
Thanks for the improvement! I completely overlooked that we can be more specific here. 😅

PR #2893, opened 2025-11-10
Switch Dependabot to Renovate

intgr:
I already migrated `djangorestframework-stubs` to Renovate and it's working fine. Though it did take me some time to get the configuration right for "automerge" of lock file maintenance.

* Related PR: https://github.com/typeddjango/djangorestframework-stubs/pull/865

@sobolevn mentioned in https://github.com/typeddjango/djangorestframework-stubs/issues/863#issuecomment-3510540661 that `django-stubs` also has "problems with uv in django-stubs. dependabot does not even create PRs there". So this might resolve those.

But for this to work, a maintainer with admin access (sobolevn) needs to enroll this repository at https://developer.mend.io/


.github/renovate.json
```diff
@@ -0,0 +1,13 @@
+{
+  "dependencyDashboard": true,
+  "mode": "full",
+  "lockFileMaintenance": {
+    "enabled": true,
+    "automerge": true,
+    "automergeType": "branch",
+    "schedule": ["* * * * 1"],
+    "automergeSchedule": ["* * * * 2"]
```
intgr, line 9: This would update all unpinned dependencies in `uv.lock` once per week:

* Branch is created every Monday.
* If the branch build succeeded, Renovate will automatically merge it without a PR on Tuesday.
* If build failed, a PR will be opened with the changes.

.github/renovate.json
```diff
@@ -0,0 +1,13 @@
+{
+  "dependencyDashboard": true,
```
intgr, line 2: If this is enabled, Renovate will maintain a GitHub issue about the status of various dependency updates, it can also be used to control its behavior via checkboxes. Example:

* https://github.com/typeddjango/djangorestframework-stubs/issues/868

.github/workflows/test.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - "renovate/**"
```
intgr, line 7: This is needed for Renovate "automerge" to be able to verify branch changes without a PR.

.github/renovate.json
intgr: In the future, `packageRules` can be used to configure similar automerge behavior for a wider range of packages.

UnknownPlatypus:
I'm +1

UnknownPlatypus (approved):
I'm +1 for renovate, lgtm

.github/dependabot.yml
```diff
@@ -1,18 +0,0 @@
...
-  directory: "/"
-  schedule:
-    interval: weekly
-
-- package-ecosystem: uv
-  directory: "/ext"
-  schedule:
-    interval: weekly
-
-- package-ecosystem: github-actions
```
sobolevn, line 13: Will `renovate` also update `github-actions`?

UnknownPlatypus: Yes, It's detected automatically, see https://github.com/typeddjango/djangorestframework-stubs/issues/868#:~:text=Detected%20dependencies,-github for example and the `Detected dependencies` section

sobolevn (approved):

intgr:
@sobolevn don't forget you need to grant Renovate access to the repo also. https://developer.mend.io/github/typeddjango

.github/workflows/test.yml
sobolevn: This causes to double all our CI jobs, see https://github.com/typeddjango/django-stubs/pull/2898

It starts 83 checks :(

intgr: Ah, good point. There are two options:

1. Configure GitHub Actions concurrency to terminate the branch job when PR job starts.
2. Only whitelist `renovate/lock-file-maintenance` for the branch CI.

Option (2) looks simpler -- but the downside is that if we were to expand "automerge" to other package updates, then the list of branches needs updating manually.

ISTM that we actually could use automerge for most version bumps -- in those PRs we usually just check that CI is passing. So automation could save time.

intgr: Opened a PR for option (2) https://github.com/typeddjango/django-stubs/pull/2899

PR #2908, opened 2025-11-18
make code in contrib/admin/filters.pyi properly use generics

Vulwsztyn:
# I have made things!

I improved usage of generics in `contrib/admin/filters.pyi`

## Related issues

I haven't been able to find any.


django-stubs/contrib/admin/filters.pyi
```diff
@@ -21,59 +21,68 @@ class _ListFilterChoices(TypedDict):
     query_string: str
     display: _StrOrPromise
 
-class ListFilter:
+_ModelT = TypeVar("_ModelT", bound=Model)
+_ST = TypeVar("_ST", contravariant=True)
+_GT = TypeVar("_GT", covariant=True)
```
UnknownPlatypus, line 26: Can we / should we share these typevar definition with the other places with matching ones ?

django-stubs/contrib/admin/filters.pyi
```diff
@@ -21,59 +21,68 @@ class _ListFilterChoices(TypedDict):
...
+
+class ListFilter(Generic[_ModelT]):
     title: _StrOrPromise | None
     template: str
     request: HttpRequest
     used_parameters: dict[str, bool | datetime | str]
     def __init__(
-        self, request: HttpRequest, params: dict[str, str], model: type[Model], model_admin: ModelAdmin
+        self, request: HttpRequest, params: dict[str, str], model: type[Model], model_admin: ModelAdmin[_ModelT]
```
UnknownPlatypus: Should model also be modelT?

UnknownPlatypus (commented):
A few open questions

django-stubs/contrib/admin/filters.pyi
Vulwsztyn: yeah, will fix

django-stubs/contrib/admin/filters.pyi
Vulwsztyn: No preference, I have not seen anyone doing it tho

PR #2914, opened 2025-11-25
Run pyrefly over test cases

rchen152:
# I have made things!

This PR proposes running the [pyrefly](https://github.com/facebook/pyrefly) type checker (which I am a maintainer of) over the assert_type tests. Pyrefly has [built-in django support](https://pyrefly.org/en/docs/django/) that uses django-stubs, and running pyrefly over these tests to make sure it stays compatible with the stubs would be helpful for us.

Notes:
* I've configured pyrefly to ignore `# type: ignore` comments and added `# pyrefly: ignore` comments where needed, following what was done for pyright.
* In `test_enums.py`, I've changed some of the `assert_type`s to accept both mypy and pyrefly's output, in cases in which both look reasonable to me. Happy to revert this change if the preference is to keep mypy's output as the source of truth.

## Related issues

Refs https://github.com/typeddjango/django-stubs/issues/579

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -171,6 +171,31 @@ jobs:
           version: PATH
           project: ./pyrightconfig.testcases.json
 
+  run-pyrefly:
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.12']
+      fail-fast: false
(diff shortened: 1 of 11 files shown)


sobolevn (approved):
Awesome! Thanks a lot!

Having more typecheckers is totally a great thing for the project.

PR #2921, opened 2025-11-29
Renovate: mypy updates to widen accepted version range

intgr:
* Solves comment at https://github.com/typeddjango/django-stubs/pull/2918/files
* Already tested on `djangorestframework-stubs`:
  * https://github.com/typeddjango/djangorestframework-stubs/pull/879
  * https://github.com/typeddjango/djangorestframework-stubs/pull/877

Change how Renovate bumps `mypy` dependency.

When updating to mypy 1.19, previously it would change `>=1.13,<1.19` -> `>=1.19,<1.20`.

With this change, it will change to `>=1.13,<1.20`.

Doc: https://docs.renovatebot.com/configuration-options/#rangestrategy


pyproject.toml
```diff
@@ -55,7 +55,6 @@ tests = [
   "jinja2==3.1.6",
   "pyyaml==6.0.3",
   # typing:
-  "mypy==1.19.0",
```
intgr, line 58: I suspect this won't work well with `"rangeStrategy": "widen"`

Removed this pinning in `tests` dependency group, as now we have `uv.lock` to pin dev versions. And the lockfile is updated every week by Renovate already.

sobolevn (approved):
Awesome, thanks for the fix!

.github/workflows/mypy_primer.yml
```diff
@@ -24,7 +24,8 @@ jobs:
         with:
           path: django-stubs_to_test
           fetch-depth: 0
-      - uses: actions/setup-python@v6
+      - name: Install uv & Python
```
UnknownPlatypus: I believe this is better to have both setup-python and setup-uv because setup-python caches python better

See https://docs.astral.sh/uv/guides/integration/github/#setting-up-python

UnknownPlatypus (approved):
Just a small nip around python setup but good to me for the rest

PR #2922, opened 2025-11-30
fix: allow `EmailMultiAlternatives` in `outbox`

alexei:
According to the [Django docs on sending email](https://docs.djangoproject.com/en/5.2/topics/email/#in-memory-backend) the `outbox` attribute is:

> a list with an `EmailMessage` instance for each message that would be sent.

When you send single-message emails, they are direct instances of `EmailMessage`. When the emails are multipart, they are instances of `EmailMultiAlternatives`. While `EmailMultiAlternatives` inherits from `EmailMessage`, declaring `outbox` as `list[EmailMessage` is only partially correct, because the [`EmailMultiAlternatives`](https://docs.djangoproject.com/en/5.2/topics/email/#django.core.mail.EmailMultiAlternatives) introduces some properties and methods of its own, such as `alternatives`. If you access any of them, the type checker will complain. I think, thus, the correct type is:

```python
outbox: list[EmailMessage | EmailMultiAlternatives]
```

### DIFF
--- a/django-stubs/core/mail/__init__.pyi
+++ b/django-stubs/core/mail/__init__.pyi
@@ -50,7 +50,7 @@ def mail_managers(
     html_message: str | None = None,
 ) -> None: ...
 
-outbox: list[EmailMessage]
+outbox: list[EmailMessage | EmailMultiAlternatives]
 
 __all__ = [
     "DEFAULT_ATTACHMENT_MIME_TYPE",
--- a/django-stubs/core/mail/__init__.pyi
+++ b/django-stubs/core/mail/__init__.pyi
@@ -50,7 +50,7 @@ def mail_managers(
     html_message: str | None = None,
 ) -> None: ...
 
-outbox: list[EmailMessage | EmailMultiAlternatives]
+outbox: Sequence[EmailMessage | EmailMultiAlternatives]
 
 __all__ = [
     "DEFAULT_ATTACHMENT_MIME_TYPE",
--- a/django-stubs/core/mail/__init__.pyi
+++ b/django-stubs/core/mail/__init__.pyi
@@ -1,4 +1,4 @@
-from collections.abc import Iterable, MutableSequence, Sequence
+from collections.abc import Iterable, Sequence
 from typing import Any
 
 from django.utils.functional import _StrOrPromise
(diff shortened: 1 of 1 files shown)


alexei:
On another note, I'm not sure `EmailMultiAlternatives.alternatives` is correctly typed as `list[tuple[bytes | EmailMessage | Message | SafeMIMEText | str, str]]`, when Django seems to create a list of `EmailAlternative`s:

https://github.com/django/django/blob/85586052e8d1d9f160b9f1b351e61a787a8e4bed/django/core/mail/message.py#L479-L481

django-stubs/core/mail/__init__.pyi
```diff
@@ -50,7 +50,7 @@ def mail_managers(
     html_message: str | None = None,
 ) -> None: ...
 
-outbox: list[EmailMessage]
+outbox: list[EmailMessage | EmailMultiAlternatives]
```
sobolevn, line 53: Maybe `Sequence`?

alexei: Done

sobolevn:
Sorry, my suggestion was not correct. Let's revert this

alexei:
@sobolevn see https://github.com/typeddjango/django-stubs/pull/2922/commits/844362ca5ef6dd2d1c96da867888f635934d2084 -- I switched to `MutableSequence` because we need to support e.g. `append`:

https://github.com/django/django/blob/dba622ebc1f6a6700b75303a65f8a334bd46bd8e/django/core/mail/backends/locmem.py#L31

django-stubs/core/mail/__init__.pyi
```diff
@@ -1,4 +1,4 @@
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable, MutableSequence, Sequence
```
sobolevn:
```suggestion
from collections.abc import Iterable, Sequence
```

django-stubs/core/mail/__init__.pyi
```diff
@@ -50,7 +50,7 @@ def mail_managers(
     html_message: str | None = None,
 ) -> None: ...
 
-outbox: list[EmailMessage]
+outbox: MutableSequence[EmailMessage | EmailMultiAlternatives]
```
sobolevn:
```suggestion
outbox: list[EmailMessage | EmailMultiAlternatives]
```

sobolevn (approved):
Thanks!

PR #2931, opened 2025-12-04
Django 6.0: Add stubs for built-in CSP support

federicobond:
Documentation: https://docs.djangoproject.com/en/6.0/ref/csp/

I reused a trick from `django-stubs/db/models/enums.pyi` to make it compatible with Python < 3.11.


federicobond:
Apparently the stubtest job needs the Django 6.0 dependency update to be merged first.

intgr:
> Django 6.0 dependency update to be merged first

* ✅ merged https://github.com/typeddjango/django-stubs/pull/2939

intgr:
Please look at stubtest output https://github.com/typeddjango/django-stubs/actions/runs/20098667950/job/57663468906?pr=2931

And update stubs or stubtest allowlist files `scripts/stubtest/allowlist_*.txt` as needed.

django-stubs/middleware/csp.pyi
```diff
@@ -0,0 +1,11 @@
+from typing import TYPE_CHECKING
+
+from django.utils.csp import CSP as CSP
+
+if TYPE_CHECKING:
```
intgr: No need for `if TYPE_CHECKING` in `.pyi` files -- they are parsed by type checkers, never by Python runtime.

federicobond:
Should I include undocumented functions like `get_nonce` and `build_policy` in the type stubs or allow them in the allowlist?

intgr:
Yes, add them to stubs. If you know they are undocumented, it's worth adding a code comment about it.

django-stubs/conf/global_settings.pyi
```diff
@@ -541,3 +541,9 @@ SECURE_REDIRECT_EXEMPT: list[str]
 SECURE_REFERRER_POLICY: str
 SECURE_SSL_HOST: str | None
 SECURE_SSL_REDIRECT: bool
+
+##################
+# CSP MIDDLEWARE #
+##################
+SECURE_CSP: dict[str, Any] = {}
+SECURE_CSP_REPORT_ONLY: dict[str, Any] = {}
```
intgr: Optional: Looks like you could define a `TypedDict` for these. But it's also fine if the 1st PR doesn't fine-tune everything.

But do use `from _typeshed import Incomplete` instead of `Any` in that case.

```suggestion
SECURE_CSP: dict[str, Incomplete] = {}
SECURE_CSP_REPORT_ONLY: dict[str, Incomplete] = {}
```

django-stubs/utils/csp.pyi
```diff
@@ -0,0 +1,24 @@
...
+
+if sys.version_info >= (3, 11):
+    from enum import StrEnum
+else:
+    from enum import Enum
+
+    class ReprEnum(Enum): ...  # type: ignore[misc]
+    class StrEnum(str, ReprEnum): ...  # type: ignore[misc]
+
+class CSP(StrEnum):
```
intgr: Use underscore so that these aren't visible to users.

```suggestion
    from enum import StrEnum as _StrEnum
else:
    from enum import Enum

    class _ReprEnum(Enum): ...  # type: ignore[misc]
    class _StrEnum(str, _ReprEnum): ...  # type: ignore[misc]

class CSP(_StrEnum):
```

django-stubs/conf/global_settings.pyi
```suggestion
SECURE_CSP: dict[str, Any]
SECURE_CSP_REPORT_ONLY: dict[str, Any]
```

sobolevn: We might also type it as `Sequence[str]` instead of `Any` maybe?

django-stubs/views/decorators/csp.pyi
```diff
@@ -0,0 +1,7 @@
+from collections.abc import Callable
+from typing import Any, TypeVar
+
+_F = TypeVar("_F", bound=Callable[..., Any])
+
+def csp_override(config: dict[str, Any]) -> Callable[[_F], _F]: ...
+def csp_report_only_override(config: dict[str, Any]) -> Callable[[_F], _F]: ...
```
intgr: Also use `Incomplete`.

Decorators are a little tricky. Would be good to have an `assert_type` test for both, such as

```python
@csp_override({...some actual example value...})
def my_view1(request: HttpRequest) -> HttpResponse: ...

assert_type(my_view1(HttpRequest()), HttpResponse)  # Check that decorated return type was preserved
```

intgr (commented):
Thanks!

django-stubs/conf/global_settings.pyi
intgr: Apparently `SECURE_CSP_REPORT_ONLY = {"report-uri": "/path/to/reports-endpoint/"}` is a valid value, so it's not always `Sequence`.

federicobond: Would `Sequence[str] | str` cover all use cases?

tests/assert_type/views/test_decorators.py
```diff
@@ -0,0 +1,25 @@
...
+        "script-src": ["'self'", "'unsafe-inline'"],
+    }
+)
+def my_view(request: HttpRequest) -> HttpResponse: ...
+
+
+@csp_report_only_override(
+    {
+        "default-src": ["'self'"],
+        "script-src": ["'self'", "'unsafe-inline'"],
```
intgr, line 19: Let's test the `"report-uri": "/path/to/reports-endpoint/"` too

django-stubs/conf/global_settings.pyi
intgr: I'm not sure. Check the docs and if that is inconclusive, maybe check source too.

federicobond: From the docs:

> The setting must be a mapping (typically a dictionary) of directive names to their values. Each key should be a valid CSP directive such as default-src or script-src. The corresponding value can be a list, tuple, or set of source expressions or URLs to allow for that directive. If a set is used, it will be automatically sorted to ensure consistent output in the generated headers.

Based on this, I updated the type for the policies to be `Mapping[str, Collection[str] | str]`. Let me know if that works.

federicobond:
I've updated the PR based on the review comments. All build jobs pass now.

sobolevn (approved):

intgr (approved):
Awesome, thanks for your help.

PR #2935, opened 2025-12-07
Fix infinite semantic analysis loop when using from_queryset

### LINKED ISSUE #2373: Maximum semantic analysis iteration count reached error with django-stubs 5.0.3 & 5.0.4
# Bug report

## What's wrong

When running django-stubs 5.0.2, mypy goes through without error. But when updating django-stubs to 5.0.3, I’m getting the following error:
`error: INTERNAL ERROR: maximum semantic analysis iteration count reached`.

The traceback is huge, and not really helping, so my next step is to have a minimal reproduction, but I wanted to raise the bug in case someone else is running into this as well. I’ll also try to review what’s happening between the 2 versions to understand where the issue can be.

## System information

- OS: Linux 6.10.5
- `python` version:
- `django` version: 5.0.9 (but same error with 5.1.1)
- `mypy` version: 1.11.2 (but same error with 1.10.1)
- `django-stubs` version: 5.0.3
- `django-stubs-ext` version:  5.0.3

It works fine with:
- `django` version 5.1.1
- `mypy` version 1.11.2
- `django-stubs` version 5.0.2
- `django-stubs-ext` version 5.0.2

lee3jjang:
# I have made things!

## Related issues

- Closes #2373

## Summary

This PR fixes the "maximum semantic analysis iteration count reached" error that was introduced in django-stubs 5.0.3.

### Problem

When `create_manager_info_from_from_queryset_call()` returns `None`, the code unconditionally calls `add_symbol_table_node()` with a `PlaceholderNode` on every iteration. This sets mypy's `progress = True`, preventing the semantic analysis loop from terminating.

The issue was introduced in PR #2228 as a hack for python/mypy#17402, but the implementation causes infinite iterations when the `from_queryset` call cannot be resolved immediately.

### Solution

Check if a `PlaceholderNode` already exists before adding a new one. This prevents redundant symbol table modifications that trigger infinite iterations while still maintaining the original hack for python/mypy#17402.

### Changes

- Added a check for existing `PlaceholderNode` before calling `add_symbol_table_node()` in `create_new_manager_class_from_from_queryset_method()`

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -334,9 +334,12 @@ def create_new_manager_class_from_from_queryset_method(ctx: DynamicClassDefConte
     new_manager_info = create_manager_info_from_from_queryset_call(semanal_api, ctx.call, ctx.name)
     if new_manager_info is None:
         if not ctx.api.final_iteration:
-            # XXX: hack for python/mypy#17402
-            ph = PlaceholderNode(ctx.api.qualified_name(ctx.name), ctx.call, ctx.call.line, becomes_typeinfo=True)
-            ctx.api.add_symbol_table_node(ctx.name, SymbolTableNode(GDEF, ph))
+            # Only add PlaceholderNode if it doesn't already exist to prevent
+            # infinite semantic analysis iterations (fixes #2373)
+            if not (manager_sym and isinstance(manager_sym.node, PlaceholderNode)):
+                # XXX: hack for python/mypy#17402
+                ph = PlaceholderNode(ctx.api.qualified_name(ctx.name), ctx.call, ctx.call.line, becomes_typeinfo=True)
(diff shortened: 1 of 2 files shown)


UnknownPlatypus:
This would be nice to have a test showing the max iteration issue so that we can be sure this resolves it. See #2228 for an exemple

lee3jjang:
I tried extensively to create a test that triggers the exact "maximum semantic analysis iteration count reached" error, but it was difficult because `create_manager_info_from_from_queryset_call()` returning `None` is transient - mypy resolves forward references after a few iterations.

The test `test_from_queryset_repeated_deferral_with_forward_metaclass` does exercise the `None` path (verified with debug output), but the forward metaclass resolves quickly before hitting max iterations.

**Real-world validation**: This fix resolved the error in my production codebase that was failing with django-stubs 5.0.3+.

sobolevn (approved):
Looks correct to me, thank you!

UnknownPlatypus (approved):

PR #2939, opened 2025-12-09
Tests: Update to Django 6.0; fall back to 5.2 on Python <3.12

intgr:
* Closes  #2929
* Depending on Python version, Django 6.0 or 5.2 will be used in `tests` dependency group.
* Workflows: Updated test matrix to use Django 6.0 x Python 3.12, 3.13
* Stubtest:
  * Only running with Django 6.0 now.
  * Created new `allowlist_todo_django60.txt` file.
  * Some old entries could also be removed.
* Tests: Updated some tests for Django 6.0 and excluded from running with older Django


intgr:
## It worked! :exploding_head: :partying_face:

tests/typecheck/models/test_create.yml
```diff
@@ -81,12 +81,13 @@
...
 
 -   case: optional_id_fields_for_create_is_error_if_not_autofield
+    skip: __import__("django").VERSION < (6, 0)
     main: |
         from myapp.models import Publisher, Book
 
         Book.objects.create(id=None)  # E: Incompatible type for "id" of "Book" (got "None", expected "float | int | str | Combinable")  [misc]
         Book.objects.create(publisher=None)  # E: Incompatible type for "publisher" of "Book" (got "None", expected "Publisher | Combinable")  [misc]
-        Book.objects.create(publisher_id=None)  # E: Incompatible type for "publisher_id" of "Book" (got "None", expected "Combinable | int | str")  [misc]
+        Book.objects.create(publisher_id=None)  # E: Incompatible type for "publisher_id" of "Book" (got "None", expected "float | int | str | Combinable")  [misc]
```
intgr, line 90: This is a bit surprising -- where does this `float` come from? But I don't care enough to spend more time on it, seems mostly harmless.

sobolevn (commented):
Great stuff! 

Now we should decide if we are ready to jump into 6.0 support cicle. If we are - merge this right ahead!

intgr:
> Now we should decide if we are ready to jump into 6.0 support cicle.

We don't necessarily want to advertise official Django 6.0 support yet. Not until we have more of the new APIs covered.

But this should be merged regardless -- it's crucial to at least begin work on adding 6.0 APIs.

intgr:
@UnknownPlatypus Do you want to take a look as well?

UnknownPlatypus (approved):

PR #2950, opened 2025-12-13
Fix dispatch_uid type: str -> _AnyHashable

craigds:
# I have made things!

Django accepts any hashable value for `dispatch_uid` (used as dict key).

### DIFF
--- a/django-stubs/db/models/signals.pyi
+++ b/django-stubs/db/models/signals.pyi
@@ -3,6 +3,7 @@ from collections.abc import Callable
 from django.apps.registry import Apps
 from django.db.models.base import Model
 from django.dispatch import Signal
+from django.dispatch.dispatcher import _AnyHashable
 
 class_prepared: Signal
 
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
I would rather use `Any` except `Hashable`, because this type is problematic. `(1, 2)` is hashable, but `([], )` is not :(

craigds:
hm, django's docstring explicitly says ["anything hashable"](https://github.com/django/django/blob/0174a85770356fd12e4c8daa42a4f1c752ae00e6/django/dispatch/dispatcher.py#L110-L113)

I guess while Hashable can't catch all runtime failures (like tuples containing lists), obviously neither can Any - and Hashable matches django's documented contract. Using Any loses the intent that this value must be usable as a dict key.

sobolevn:
`Hashable` type is mostly not usable also :(
Due to the fact that it is not sound in any way.

So, that why we can do it like this: `_AnyHashable: TypeAlias = Any` and use `_AnyHashable` there.

craigds:
I'm not really following... Is the objection to Hashable based on the fact that it can't detect `([],)` isn't hashable at type-check time?

I don't see how that could be considered worse than Any - which will clearly never detect that anything isn't hashable; it'll just accept everything

sobolevn:
`Hashable` can also fail with unhashable subtypes of hashable types and vice a versa.
This is also our unofficial policy in `typeshed` as well: https://github.com/python/typeshed/blob/9340d1a9124a76df512f7d64f095a8399910838e/stdlib/typing.pyi#L520-L522

Basically, this type is not very far from `Any`, but we are not sure if it is worth using it.

craigds:
that is indeed a strange thing that `hash(object())` works but `hash(list())` doesn't, although using Any instead of Hashable  seems a bit like abstaining from voting in a close election because both candidates are bad even though one is clearly worse...

But I'll change this to the _AnyHashable alias so you can merge this, it doesn't really matter to me

sobolevn (approved):
Thanks!

craigds:
you too, thanks for engaging with me calmly on that 😄 👍

Issue #2944, opened 2025-12-11
CALL TO ACTION: Adding Django 6.0 stubs


intgr:
### 🚀 Adding Django 6.0 stubs

We have just updated to Django 6.0 in `django-stubs` test suite, but most new APIs are still missing stubs coverage, and changes to existing APIs are not covered.

If you have interest in contributing, **please help by opening PRs** (check that a PR does not exist already for your topic).

### 👉 Starting points

* [Django 6.0 release notes](https://docs.djangoproject.com/en/6.0/releases/6.0/)
* ~~[scripts/stubtest/allowlist_todo_django60.txt](https://github.com/typeddjango/django-stubs/blob/master/scripts/stubtest/allowlist_todo_django60.txt) file~~ all items have been fixed

And by the way, we are still missing some things from previous Django versions too! Check earlier release note, other `allowlist_todo*.txt` files and https://github.com/typeddjango/django-stubs/issues/1493#issuecomment-2105730206

### 🤝 You help us, we help you

If want to help but you are not sure how to implement something in stubs:
* [Check tutorials listed here](https://github.com/typeddjango/django-stubs/blob/master/CONTRIBUTING.md#tutorials)
* [Or just open a discussion topic](https://github.com/typeddjango/django-stubs/discussions), feel free to ping @intgr in the message.

intgr:
@sobolevn Let's also include this call to action message in the next Release Notes, to get more visibility.

Tosinibikunle:
Hi, I would like to work on this!

sobolevn:
Please, go ahead! We need a lot of help :)

intgr:
Just pick one part at a time - don't try to do it all in one go 🙂

intgr:
Version 5.2.9 released, release notes: https://github.com/typeddjango/django-stubs/releases/tag/5.2.9

TOMAL5085:
Hi! I'm a GSoC 2026 aspirant and I'd like to help with the Django 6.0 stubs. I noticed that many modules are being updated so I'll focus on adding missing definitions for HttpRequest and HttpResponse based on the 6.0 release notes. I'll start with a small PR for these specifically.

TOMAL5085:
Update: I have submitted a PR to add the fallback argument to HttpRequest.get_signed_cookie: #3060

ahmedasar00:
Hi
I'm interested in contributing to Django as part of GSoC 2026. 
I've been exploring the codebase and trying to understand how 
[...]

PR #2954, opened 2025-12-15
Update version compatibility metadata

intgr:


intgr:
Ah, Renovate didn't change Python versions used in test matrix, will update that as well

README.md
```diff
@@ -48,6 +48,7 @@ We rely on different `django` and `mypy` versions:
 
 | django-stubs   | Mypy version | Django version | Django partial support | Python version |
 |----------------|--------------|----------------|------------------------|----------------|
+| (unreleased)   | 1.13 - 1.19  | 5.2            | 6.0, 5.1, 5.0          | 3.10 - 3.14    |
```
sobolevn, line 51:
```suggestion
| (unreleased)   | 1.13 - 1.19  | 6.0            | 5.2, 5.1, 5.0          | 3.10 - 3.14    |
```

sobolevn (commented):
I would put it like this. Because now we already have some parts of 6.0 and we can't claim to support 5.2, because it will allow code that does not work on 5.2 :(

But, some more work is required to support 6.0 to claim it

README.md
intgr: I think we're still closer to 5.2 than we are to fully supporting 6.0, for now.

I'd like to make a release at one point, to put out the "announcement" #2944 to perhaps get more help from the community.

But it's also an option to make a release from a branch, excluding all 6.0 changes.

sobolevn: It would be ideal to make a release with 5.2 and 3.14 as the last 5.2 release and then move on to 6.0 changes. And make the announcement. But, this will require a new branch with manual cherry-picking of commits.

intgr:
These changes were done in #3008

PR #2956, opened 2025-12-16
Update `get_auto_imports()` to also allow returning `None`

jamesbeith:
Returning `None` is the documented way to permanently disable automatic imports.

https://docs.djangoproject.com/en/5.2/howto/custom-shell/#customizing-shell-auto-imports

There is also specific handling of `None` within Django's `shell` management command.

https://github.com/django/django/blob/5.2.9/django/core/management/commands/shell.py#L170-L172


intgr (approved):
Thanks!

PR #2959, opened 2025-12-16
Make LoginView generic over its form class

### LINKED ISSUE #2946: LoginView should be generic and only bounded to AuthenticationForm
# Bug report

Consider this simplified example for customisation of login functionality:
```py
from django.contrib.auth.views import LoginView
from django.http import HttpResponse


class MyLoginForm(AuthenticationForm):
     ...


class MyLoginView(LoginView):
     form_class = MyLoginForm

     def form_valid(self, form: MyLoginForm) -> HttpResponse:
         ...
```

## What's wrong

Currently mypy (v1.19.0 ATM) fails to type check this together with django-stubs (v5.2.8) and yells:
```
error: Argument 1 of "form_valid" is incompatible with supertype "django.views.generic.edit.FormMixin"; supertype defines the argument type as "AuthenticationForm"  [override]
note: This violates the Liskov substitution principle
note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

This is probably because the LoginView has the hardcoded type variable when inheriting from FormMixin:
https://github.com/typeddjango/django-stubs/blob/51beeee9763f2f81dbb48327efb673341feae509/django-stubs/contrib/auth/views.pyi#L21

## How is that should be

At runtime this works (although haven't checked this exact MRE, but we have a similar, but more verbose setup in one of our projects).

I think the fix is to have a type var on that view and bound it to `AuthenticationForm`.
Like so (new syntax):
```py
class LoginView[T: AuthenticationForm](RedirectURLMixin, FormView[T]):
```

## System information

- OS: OSX
- `python` version: 3.12.12
- `django` version: 5.2
- `mypy` version: 1.19.0
- `django-stubs` version: 5.2.8

slafs:
Closes #2946

### DIFF
--- a/django-stubs/contrib/auth/views.pyi
+++ b/django-stubs/contrib/auth/views.pyi
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, TypeVar
 
 from django.contrib.auth.forms import AuthenticationForm
 from django.contrib.auth.models import _User, _UserModel
@@ -8,6 +8,7 @@ from django.views.generic.base import TemplateView
 from django.views.generic.edit import FormView
 
 UserModel = _UserModel
+_AuthForm = TypeVar("_AuthForm", bound=AuthenticationForm, default=AuthenticationForm)
 
 class RedirectURLMixin:
     next_page: str | None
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Looks good to me! Thanks!

PR #2961, opened 2025-12-17
Add stubs for django.contrib.postgres.search.Lexeme and CombinedLexeme

federicobond:
`LexemeCombinable` is a bit more difficult to implement because it's imcompatible with `Combinable`, as it would return a `CombinedLexeme` instead of a `Combinable` for the relevant dunder methods. Not sure how to handle that.

### DIFF
--- a/django-stubs/contrib/postgres/search.pyi
+++ b/django-stubs/contrib/postgres/search.pyi
@@ -2,7 +2,7 @@ from typing import Any, ClassVar, TypeAlias
 
 from django.db.backends.base.base import BaseDatabaseWrapper
 from django.db.models import Expression, Field, FloatField, TextField
-from django.db.models.expressions import Combinable, CombinedExpression, Func
+from django.db.models.expressions import Combinable, CombinedExpression, Func, Value
 from django.db.models.lookups import Lookup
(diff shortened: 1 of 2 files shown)


django-stubs/contrib/postgres/search.pyi
```diff
@@ -142,3 +142,18 @@ class TrigramWordDistance(TrigramWordBase): ...
...
+    def __init__(
+        self,
+        value: Any,
+        output_field: Any | None = None,
+        *,
+        invert: bool = False,
+        prefix: bool = False,
+        weight: Any | None = None,
+    ) -> None: ...
+    def as_sql(self, compiler: SQLCompiler, connection: BaseDatabaseWrapper) -> _AsSqlType: ...
```
sobolevn: Do we need `as_sql` here? It is defined in `BaseException` https://github.com/typeddjango/django-stubs/blob/dc4b59d6a0faaec1bc762add53e4905c3a1588a4/django-stubs/db/models/expressions.pyi#L71

federicobond: Good catch. I've updated the PR to remove `as_sql`. All build jobs pass.

sobolevn (approved):

PR #2966, opened 2025-12-18
Django 6.0: require kwargs for optional parameters in mail functions

federicobond:

> All optional parameters (fail_silently and later) must be passed as keyword arguments to [get_connection()](https://docs.djangoproject.com/en/dev/topics/email/#django.core.mail.get_connection), [mail_admins()](https://docs.djangoproject.com/en/dev/topics/email/#django.core.mail.mail_admins), [mail_managers()](https://docs.djangoproject.com/en/dev/topics/email/#django.core.mail.mail_managers), [send_mail()](https://docs.djangoproject.com/en/dev/topics/email/#django.core.mail.send_mail), and [send_mass_mail()](https://docs.djangoproject.com/en/dev/topics/email/#django.core.mail.send_mass_mail).

https://docs.djangoproject.com/en/dev/releases/6.0/#positional-arguments-in-django-core-mail-apis

### DIFF
--- a/django-stubs/core/mail/__init__.pyi
+++ b/django-stubs/core/mail/__init__.pyi
@@ -16,12 +16,13 @@ from .message import make_msgid as make_msgid
 from .utils import DNS_NAME as DNS_NAME
 from .utils import CachedDnsName as CachedDnsName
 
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
This will cause some false-positives on older versions, but I am fine with this cost.

PR #2967, opened 2025-12-19
Django 6.0: Add Django Tasks framework

federicobond:
I wrote stubs for the Django Task framework introduced in Django 6.0. They still need some polish but feedback is very welcome.

If you are reviewing this, pay special attention to the Task class generic parameters which have an influence in how many other methods are typed.

*Update:* this now passes all tests, I had to put some items in the allowlist and conditionally monkeypatch the tasks classes based on the Django version. There are some rough edges there, but things are looking good.


django-stubs/tasks/base.pyi
```diff
@@ -0,0 +1,137 @@
...
+    enqueued_at: datetime | None
+    started_at: datetime | None
+    finished_at: datetime | None
+    last_attempted_at: datetime | None
+    args: list[Any]
+    kwargs: dict[str, Any]
+    backend: str
+    errors: list[TaskError]
+    worker_ids: list[str]
+    def __init__(
```
sobolevn: Do we need to annotate `__init__` in `dataclass`es? Maybe just use `@dataclass(kw_only=True)`?

scripts/stubtest/allowlist_todo_django60.txt
```diff
@@ -202,15 +202,23 @@ django.forms.ClearableFileInput.use_fieldset
...
-django.tasks.backends.base
-django.tasks.backends.dummy
-django.tasks.backends.immediate
-django.tasks.base
 django.tasks.checks
-django.tasks.exceptions
-django.tasks.signals
+django.tasks.default_task_backend
+django.tasks.task
+django.tasks.Task.__match_args__
```
sobolevn: Question: why are `__match_args__` and `__post_init__` ignored?

tests/typecheck/tasks/test_tasks.yaml
```diff
@@ -0,0 +1,23 @@
...
+        from typing_extensions import reveal_type
+        from django.tasks import Task, task
+
+        @task
+        def test_task(x: int) -> int:
+            return x + 1
+
+        reveal_type(test_task) # N: Revealed type is "django.tasks.base.Task[[x: builtins.int], builtins.int]"
+        reveal_type(test_task.call(1)) # N: Revealed type is "builtins.int"
+        reveal_type(test_task.enqueue(1)) # N: Revealed type is "django.tasks.base.TaskResult[[x: builtins.int], builtins.int]"
```
sobolevn:
```suggestion
        reveal_type(test_task.enqueue) # N: ...
```

Please, test the signature of the task instead

tests/typecheck/tasks/test_tasks.yaml
```diff
@@ -0,0 +1,23 @@
...
+    main: |
+        from typing_extensions import reveal_type
+        from django.tasks import Task, task
+
+        @task
+        def test_task(x: int) -> int:
+            return x + 1
+
+        reveal_type(test_task) # N: Revealed type is "django.tasks.base.Task[[x: builtins.int], builtins.int]"
+        reveal_type(test_task.call(1)) # N: Revealed type is "builtins.int"
```
sobolevn:
```suggestion
        reveal_type(test_task.call) # N: Revealed type is ...
```

tests/typecheck/tasks/test_tasks.yaml
```diff
@@ -0,0 +1,23 @@
...
+        reveal_type(test_task.enqueue(1)) # N: Revealed type is "django.tasks.base.TaskResult[[x: builtins.int], builtins.int]"
+
+-   case: task_with_priority
+    main: |
+        from typing_extensions import reveal_type
+        from django.tasks import Task, task
```
sobolevn:
```suggestion
        from django.tasks import task
```

django-stubs/tasks/base.pyi
federicobond: You are right, this was a leftover from the initial stubs.

scripts/stubtest/allowlist_todo_django60.txt
federicobond: Leftover from when I was not using `@dataclass` in the stubs

scripts/stubtest/allowlist_todo_django60.txt
```diff
@@ -202,15 +202,12 @@ django.forms.ClearableFileInput.use_fieldset
...
-django.tasks
-django.tasks.backends
-django.tasks.backends.base
-django.tasks.backends.dummy
-django.tasks.backends.immediate
-django.tasks.base
 django.tasks.checks
-django.tasks.exceptions
-django.tasks.signals
+django.tasks.default_task_backend
```
sobolevn, line 181: Question: why are these items in `todo.txt`?
- If these are expected errors, move them to regular `allowlist.txt` (without `todo`) and comment why they are here
- If they are just not fixed yet, what is blocking us from fixing them?

federicobond: Honestly because I wasn't aware of how to use each list, so I just threw those items back to where they came from.

   * `django.tasks.default_task_backend` could be moved to allowlist_todo.txt with the other ConnectionProxy instances, like django.core.cache.cache and django.db.connection.
   * `django.tasks.task` decorator emits an error because one of the overloads expects a non-None function value when the decorator accepts None at runtime (for the other overload). Could move it to allowlist.txt
   * `django.tasks.backends.base.BaseTaskBackend.enqueue` should be typed as an abstract method but not sure if that's going to be problematic since default_task_backend is typed as BaseTaskBackend. Can move it to allowlist.txt but not sure.

tests/typecheck/tasks/test_tasks.yaml
```diff
@@ -0,0 +1,23 @@
...
+        from typing_extensions import reveal_type
+        from django.tasks import task
+
+        @task
+        def test_task(x: int) -> int:
+            return x + 1
+
+        reveal_type(test_task) # N: Revealed type is "django.tasks.base.Task[[x: builtins.int], builtins.int]"
+        reveal_type(test_task.call) # N: Revealed type is "def (x: builtins.int) -> builtins.int"
+        reveal_type(test_task.enqueue) # N: Revealed type is "def (x: builtins.int) -> django.tasks.base.TaskResult[[x: builtins.int], builtins.int]"
```
intgr, line 12: Optional: it looks like these tests could be replaced with `assert_type` tests? The advantage is that you can use IDE features to edit them, they work with multiple typecheckers and they are much faster to run.

federicobond: I tried that initially but I was not able to construct a valid expression for the type of Task or TaskResult with concrete ParamSpec arguments.

How do I write `django.tasks.base.Task[[x: builtins.int], builtins.int]` as a valid Python expression? `django.tasks.base.Task[[int], int]` did not work.

scripts/stubtest/allowlist_todo_django60.txt
federicobond: I moved the task decorator to `allowlist.txt` with proper reasoning.

I left the `default_task_backend` in todo mirroring the other ConnectionProxy instances. I also left enqueue method as todo for now.

tests/typecheck/tasks/test_tasks.yaml
intgr: Yeah that's a PITA.

For mypy there is a solution, but it doesn't work with other typecheckers

```python
from mypy_extensions import Arg
...
assert_type(test_task, Task[[Arg(int, "x")], int])
```

I think a better way is to use a callable without named arguments. Although then you can't use `@task` as a decorator

```python
test_func: Callable[[int], int] = cast("Any", lambda: ...)

test_task = task(test_func)
assert_type(test_task, Task[[int], int])
```

intgr: There's also a 3rd option, but I think it's worse -- if for some bug something becomes `Any`, this won't fail. You can use a normal assignment to verify that types are "compatible" without requiring equal types.

```python
@task
def test_task(x: int) -> int:
    return x + 1

task_downcast: Task[[int], int] = test_task
```

federicobond: Happy new year! 🎆  Thanks for taking the time to review this.

Yes, I agree that none of the solutions are optimal. Would you prefer to keep the tests as is or use `assert_type`-style tests using a Callable without named arguments?

intgr: In retrospect: all of these work-arounds have downsides or make it less obvious what is happening. Let's keep the current mypy tests.

intgr (approved):

django-stubs/tasks/backends/immediate.pyi
```diff
@@ -0,0 +1,13 @@
+from logging import Logger
+from typing import Any
+
+from django.tasks.base import Task, TaskResult
+
+from .base import BaseTaskBackend as BaseTaskBackend
```
sobolevn:
```suggestion
from .base import BaseTaskBackend
```

django-stubs/tasks/backends/base.pyi
```diff
@@ -0,0 +1,21 @@
...
+    supports_defer: bool
+    supports_async_task: bool
+    supports_get_result: bool
+    supports_priority: bool
+    alias: str
+    queues: list[str]
+    options: dict[str, Any]
+    def __init__(self, alias: str, params: dict[str, Any]) -> None: ...
+    def validate_task(self, task: Task) -> None: ...
+    def enqueue(self, task: Task, args: list[Any], kwargs: dict[str, Any]) -> TaskResult: ...
```
sobolevn, line 18: I though about this method not being abstract a little bit. I prefer this solution:
- We should define this method as abstract
- We should create `@type_check_only` type like

```python
@type_check_only
class _default_task_backend(BaseTaskBackend):
    def enqueue(self, task: Task, args: list[Any], kwargs: dict[str, Any]) -> TaskResult: ...
```

to show that it is not abstract and has a real value. 

And type `default_task_backend: _default_task_backend` in `__init__.pyi`

- Later we can use a plugin to replace this type with a real one based on settings

What do you think?

federicobond: Sounds good. I've updated the pull request with that type.

sobolevn (approved):
Great work! Thanks a lot! 

We can continue with the advanced features in the next PRs.
🎉

PR #2969, opened 2025-12-23
fix: Annotate SessionStore with a type var for the Session type (#2056)

### LINKED ISSUE #2056: `assigment` error when overriding `create_model_instance` class method of  `SessionStore`.
## What's wrong
`mypy` doesn't recognize the **type** of the returned class from `get_model_class` when overriding `SessionStore`, and because of that it gives false `assignemt` error alarm when assigning  `super().create_model_instance(data)` to a variable

## Example
```python
from django.contrib.sessions.backends.cached_db import SessionStore as CachedDbSessionStore
from django.contrib.sessions.base_session import AbstractBaseSession

# custom session model
class RealmSession(AbstractBaseSession):
   ....custom code

    @classmethod
    @override
    def get_session_store_class(cls) -> Type["SessionStore"]:
        return SessionStore

class SessionStore(CachedDbSessionStore):

    @classmethod
    @override
    def get_model_class(cls) -> Type[RealmSession]:
        return RealmSession

    @override
    def create_model_instance(self, data: Dict[str, Union[int, str]]) -> RealmSession:
        session_object: RealmSession = super().create_model_instance(data) # Error arises here
        return create_realm_session_instance(session_object, data)
```
## mypy output :
```shell
Incompatible types in assignment (expression has type "AbstractBaseSession", variable has type "RealmSession")  [assignment]
```  

## How is that should be

`RealmSession` should be the expected return type from `super().create_model_instance(data)`, so no error should arise

CharString:
# I have made things!

## Related issues

Closes #2056


sobolevn (approved):

sobolevn:
Thank you!

PR #2970, opened 2025-12-23
fix: `prefetch_related_objects` requires `Sequence` not `Iterable`

### LINKED ISSUE #2948: prefetch_related_objects cannot work with Iterables - it is expecting a Sequence
# Bug report
Typing for `prefetch_related_objects` says that it can accept `Iterable`, but code indicates that it should be a `Sequence` (I suppose)

Rel. to this typing stubs https://github.com/typeddjango/django-stubs/blob/51beeee9763f2f81dbb48327efb673341feae509/django-stubs/db/models/query.pyi#L288

Django 5.2.8

## What's wrong
Typing for `prefetch_related_objects` says that it can accept Iterable, but if we will call it with  `set`, `frozenset`, keys or values view, we will see this error

```
   2370     break
   2372 # Descend down tree
   2373 
   2374 # We assume that objects retrieved are homogeneous (which is the premise
   2375 # of prefetch_related), so what applies to first object applies to all.
-> 2376 first_obj = obj_list[0]
   2377 to_attr = lookup.get_current_to_attr(level)[0]
   2378 prefetcher, descriptor, attr_found, is_fetched = get_prefetcher(
   2379     first_obj, through_attr, to_attr
   2380 )

TypeError: 'set' object is not subscriptable
```

emmanuel-ferdman:
## PR Summary
`prefetch_related_objects` and `aprefetch_related_objects` were typed to accept `Iterable[Model]`, but Django's implementation uses `obj_list[0]` indexing internally, which requires a `Sequence`. Passing a `set` or `frozenset` would pass type checking but crash at runtime with `TypeError: 'set' object is not subscriptable`. This PR changes the parameter type to `Sequence[Model]` to match the actual runtime requirement.

Fixes #2948


tests/assert_type/db/models/test_query.py
```diff
@@ -0,0 +1,20 @@
+from collections.abc import Sequence
+
+from django.db.models import Model
+from django.db.models.query import prefetch_related_objects
+
+models_list: list[Model] = []
+prefetch_related_objects(models_list, "pk")
```
UnknownPlatypus, line 10: Can we double each line for the `aprefetch_related_objects`

tests/assert_type/db/models/test_query.py
```diff
@@ -0,0 +1,20 @@
+from collections.abc import Sequence
+
+from django.db.models import Model
+from django.db.models.query import prefetch_related_objects
```
UnknownPlatypus: I think you can ignore this one, this is not due to this PR but due to the fact that we use a bare `Prefetch` while it's a generic type

tests/assert_type/db/models/test_query.py
```diff
@@ -0,0 +1,20 @@
...
+
+models_tuple: tuple[Model, ...] = ()
+prefetch_related_objects(models_tuple, "pk")
+
+models_sequence: Sequence[Model] = []
+prefetch_related_objects(models_sequence, "pk")
+
+# failure cases
+models_set: set[Model] = set()
+prefetch_related_objects(models_set, "pk")  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]
```
UnknownPlatypus: needs the pyrefly ignore also I believe

UnknownPlatypus (commented):
Thanks for the pr, I've left a few comments

tests/assert_type/db/models/test_query.py
emmanuel-ferdman: Added async tests wrapped in async `def test_async()` with proper await calls.

tests/assert_type/db/models/test_query.py
emmanuel-ferdman: Added `# pyright: ignore[reportUnknownVariableType]` on the import.

tests/assert_type/db/models/test_query.py
emmanuel-ferdman: The `# pyright: ignore[reportArgumentType]` should cover both pyright and pyrefly. If pyrefly needs separate ignore, I can add `# pyrefly: ignore[...]` as well.

tests/assert_type/db/models/test_query.py
UnknownPlatypus: The line seem to have moved with the import, see https://github.com/typeddjango/django-stubs/actions/runs/20486836798/job/58876283221?pr=2970

tests/assert_type/db/models/test_query.py
UnknownPlatypus: > The # pyright: ignore[reportArgumentType] should cover both pyright and pyrefly

ci is failing so it seems not https://github.com/typeddjango/django-stubs/actions/runs/20486836798/job/58876283206?pr=2970

Also looking at their doc, it could respect type: ignore but it was disabled here https://github.com/typeddjango/django-stubs/pull/2914/changes#diff-0b1d04abb16afadfb97588c73fdcaf0cc5797f851db59cf88736081afbe2b89f

emmanuel-ferdman: I see, thanks for the explanation - should be good now 😄

UnknownPlatypus:
Thanks!

PR #2976, opened 2025-12-28
Django 6.0: support passing MIMEPart object to EmailMessage.attach

federicobond:

> [EmailMessage.attach()](https://docs.djangoproject.com/en/dev/topics/email/#django.core.mail.EmailMessage.attach) now accepts a [MIMEPart](https://docs.python.org/3/library/email.message.html#email.message.MIMEPart) object from Python’s modern email API.

https://docs.djangoproject.com/en/dev/releases/6.0/#email

### DIFF
--- a/django-stubs/core/mail/message.pyi
+++ b/django-stubs/core/mail/message.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Sequence
-from email.message import Message
+from email.message import Message, MIMEPart
 from email.mime.base import MIMEBase
 from email.mime.message import MIMEMessage
 from email.mime.multipart import MIMEMultipart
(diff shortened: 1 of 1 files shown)


django-stubs/core/mail/message.pyi
```diff
@@ -99,6 +99,8 @@ class EmailMessage:
     def attach(
         self, filename: str | None = None, content: _AttachmentContent | None = None, mimetype: str | None = None
     ) -> None: ...
+    @overload
+    def attach(self, mimepart: MIMEPart) -> None: ...
```
sobolevn:
```suggestion
    def attach(self, filename: MIMEPart) -> None: ...
```

We try to focus on what runtime will work with vs what is documented.

See https://github.com/django/django/blob/4afe463950453b6d71f0c98f23fd0f1f2bd341bc/django/core/mail/message.py#L372

sobolevn (approved):
Thanks!

PR #2988, opened 2026-01-08
fix: django.contrib.gis.utils.LayerMapping attribute types

kelvan:
fid_range, step and progress allow more types than bool

To allow the correct usage of LayerMapping.save I extended the type definition of the attributes, according to the django source code and docstring.

https://github.com/django/django/blob/main/django/contrib/gis/utils/layermapping.py#L559


sobolevn (approved):
lgtm https://github.com/django/django/blob/459a3d17b973df23fc45328d4e976a270f0edd7f/django/contrib/gis/utils/layermapping.py#L179

PR #2992, opened 2026-01-13
Make all system checks return a `list`, not a `Sequence`

brianhelba:
This provides a stronger bound for what we know Django will always return.

Also, make all checks accept a Sequence. This fixes one inconsistent annotation.


sobolevn (approved):
They indeed return lists, thanks!

PR #2996, opened 2026-01-15
Allow plain `Form` in `DeleteView` type parameters

### LINKED ISSUE #2777: Incorrect bound type for BaseDeleteView.form_class, should be plain Form.
# Bug report

The type-stubs for `BaseDeleteView` and `DeleteView` incorrectly require the form value to be a `ModelForm`, while Django documentation implies a plain `Form` should be enough.

## What's wrong

The Django stubs for `django.views.generic.edit` state the following:

```
class BaseDeleteView(Generic[_M, _ModelFormT], DeletionMixin[_M], FormMixin[_ModelFormT], BaseDetailView[_M]):
class DeleteView(Generic[_M, _ModelFormT], SingleObjectTemplateResponseMixin, BaseDeleteView[_M, _ModelFormT]):
```

looking at the actual implementation however,

```
class BaseDeleteView(DeletionMixin, FormMixin, BaseDetailView):
    ...
    form_class = Form
```

The Django documentation itself also states that the default form is the empty form: https://docs.djangoproject.com/en/5.2/ref/class-based-views/generic-editing/#django.views.generic.edit.DeleteView.form_class, suggesting that using a non-model-form is indeed acceptable behavior.

## How is that should be

I think the stubs should read 

```
class BaseDeleteView(Generic[_M, _FormT], DeletionMixin[_M], FormMixin[_FormT], BaseDetailView[_M]):
class DeleteView(Generic[_M, _FormT], SingleObjectTemplateResponseMixin, BaseDeleteView[_M, _FormT]):
```

## System information

(I think all of the following are irrelevant, but just in case):

- OS: Debian Linux.
- `python` version: Python 3.13.6
- `django` version: 5.2.5
- `mypy` version: 1.17.1

emmanuel-ferdman:
## PR Summary
`DeleteView` and `BaseDeleteView` were incorrectly typed to only accept `ModelForm` subclasses. Django's actual implementation sets `form_class = Form` by default (since Django 4.0) because delete views use forms for confirmation, not for saving model data. This change relaxes the type parameter from `_ModelFormT` to `_FormT`, matching Django's documented behavior while remaining backward compatible with ModelForm usage.

Closes #2777.

### DIFF
--- a/django-stubs/views/generic/edit.pyi
+++ b/django-stubs/views/generic/edit.pyi
@@ -66,9 +66,9 @@ class DeletionMixin(Generic[_M]):
     def delete(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse: ...
     def get_success_url(self) -> str: ...
 
-class BaseDeleteView(DeletionMixin[_M], FormMixin[_ModelFormT], BaseDetailView[_M], Generic[_M, _ModelFormT]):
+class BaseDeleteView(DeletionMixin[_M], FormMixin[_FormT], BaseDetailView[_M], Generic[_M, _FormT]):
     object: _M
 
-class DeleteView(SingleObjectTemplateResponseMixin, BaseDeleteView[_M, _ModelFormT], Generic[_M, _ModelFormT]):
+class DeleteView(SingleObjectTemplateResponseMixin, BaseDeleteView[_M, _FormT], Generic[_M, _FormT]):
     object: _M
     template_name_suffix: str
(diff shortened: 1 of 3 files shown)


sobolevn (commented):
Please, add a test to `test/typecheck` :)

emmanuel-ferdman:
@sobolevn Added. thanks for the review!

sobolevn (commented):
Sorry, I mislead you :)
Would https://github.com/typeddjango/django-stubs/tree/master/tests/assert_type/views work instead? Current folder is for mypy only, the one I am proposing is also for pyright and pyrefly

emmanuel-ferdman:
@sobolevn I see. Thanks for the guidance! Moved the test to `tests/assert_type/views/test_generic.py` as suggested.

sobolevn (approved):
Thanks!

PR #2997, opened 2026-01-15
Fix `FileField` and `ImageField` instance access return types

### LINKED ISSUE #2989: FileField / ImageField results in Any as instance attribute type
# Bug report

## What's wrong

When defining a model with a `FileField`, the instance attribute's static type is `Any` rather than `FieldFile`.

Snippet to reproduce:

```python

from django.db import models


class Example(models.Model):
    file = models.FileField()


instance = Example()
reveal_type(instance.file)  # Any
```

## How is that should be

I would expect the inferred type to be [`FieldFile`](https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.fields.files.FieldFile), at least by default.

## System information

- OS:
- `python` version: 3.13.3
- `django` version: 6.0.1
- `mypy` version: 1.17.1

## Note

* `ImageField` has the same problem.
* It seems like the definition of `FileField` subclasses `FileField` without defining the `_GT` and `_ST` type parameters, and explicitly specifies the return type of `__get__` of `FileField` as `Any`. Maybe this is just an oversight?

emmanuel-ferdman:
## PR Summary
Accessing `FileField` or `ImageField` on a model instance was typed as `Any`, which meant type checkers couldn't catch bugs like passing a file field to functions expecting incompatible types. Django's source explicitly documents that accessing these fields ["will always return an instance of `attr_class`"](https://github.com/django/django/blob/main/django/db/models/fields/files.py#L237-L239) (`FieldFile` or `ImageFieldFile`), so the `Any` was simply a placeholder that was never updated. This fixes the `__get__` overloads to return the correct types.

Fixes #2989.


sobolevn (approved):
Thanks a lot!

PR #3002, opened 2026-01-18
Django 6.0: add support for AnyValue aggregate

federicobond:

> The new [AnyValue](https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.AnyValue) aggregate returns an arbitrary value from the non-null input values. This is supported on SQLite, MySQL, Oracle, and PostgreSQL 16+.

See https://docs.djangoproject.com/en/dev/releases/6.0/#models

### DIFF
--- a/django-stubs/db/models/aggregates.pyi
+++ b/django-stubs/db/models/aggregates.pyi
@@ -1,10 +1,12 @@
 from collections.abc import Sequence
 from typing import Any, ClassVar
 
+from django.db.backends.base.base import BaseDatabaseWrapper
 from django.db.models.expressions import Combinable, Func
 from django.db.models.fields import IntegerField
 from django.db.models.functions.mixins import FixDurationInputMixin, NumericOutputFieldMixin
 from django.db.models.query import _OrderByFieldName
+from django.db.models.sql.compiler import SQLCompiler, _AsSqlType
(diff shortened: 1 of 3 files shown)


django-stubs/db/models/aggregates.pyi
```diff
@@ -22,6 +24,11 @@ class Aggregate(Func):
         **extra: Any,
     ) -> None: ...
 
+class AnyValue(Aggregate):
+    def as_sql(  # type: ignore[override]
```
sobolevn, line 28: I would prefer to have the same args/kwargs as the `super()` has, because it calls `super().as_sql()` inside: https://github.com/django/django/blob/0d31ca98830542088299d2078402891d08cc3a65/django/db/models/aggregates.py#L234

It just does not list all kwargs in the definition.

federicobond: I tried that but the parameters in the supertype method are positional, not keyword-only, so the signatures don't match and mypy reports an error.

sobolevn: hm. indeed 🤔 
ok, let's stick to the current solution. thanks!

sobolevn (approved):

PR #3009, opened 2026-01-21
Django 6.0: add async pagination support

emmanuel-ferdman:
## PR Summary
Django 6.0 introduced async pagination with `AsyncPaginator` and `AsyncPage` classes, plus a `BasePaginator` base class. This PR adds the missing stubs for these classes.


sobolevn (approved):

93578237:
This is a breaking change or/and bug: now you can't do this in django 5.2
```python
from django.core.paginator import Paginator as _Paginator

class MyPaginator[T](_Paginator[T]): pass
```

TypeError: type 'Paginator' is not subscriptable

sobolevn:
We forgot to register it in `django-stubs-ext` :)
PR is welcome.

PR #3021, opened 2026-01-29
Allow `formfield_callback` to return `None`

### LINKED ISSUE #3015: `formfield_callback` should accept functions returning `Field | None`
# Bug report

`fields_for_model()` specifies the type of `formfield_callback` that demands the callable to return `Field`, while a callback returning `Field | None` is both accepted and the common case.

## What's wrong

The parameter type of `formfield_callback` is `_FormFieldCallback`, which is defined as follows:

```python
_FormFieldCallback: TypeAlias = Callable[[models.Field], Field]
```

However, a typical formfield callback calls out to `models.Field.formfield()`, whose return type is specified as `forms.Field | None`. This is correct: for some model field types (e.g. `BigAutoField`), `formfield()` returns `None` to indicate that no corresponding form field should be generated. See the [Django docs](https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.Field.formfield).

Due to this mismatch, typical and correct formfield callback implementations like the one below (adapted [from the Django docs](https://docs.djangoproject.com/en/6.0/topics/forms/modelforms/#overriding-the-default-fields)) causes a type error:

```python
from typing import Any

from django import forms
from django.db import models


class ExampleModel(models.Model):
    pass


class MySlugFormField(forms.SlugField):
    pass


def formfield_for_dbfield(db_field: models.Field, **kwargs: Any) -> forms.Field | None:
    if db_field.name == "slug":
        return MySlugFormField()
    return db_field.formfield(**kwargs)


fields = forms.fields_for_model(
    ExampleModel,
    formfield_callback=formfield_for_dbfield,
)
```

## How is that should be

`_FormFieldCallback` should be defined as `Callable[[models.Field], Field | None]` to match reality.

## System information

- OS:
- `python` version: 3.14.2
- `django` version: 6.0.1
- `mypy` version: 1.19.1
- `django-stubs` version: 5.2.8

emmanuel-ferdman:
# PR Summary
The `formfield_callback` parameter in `fields_for_model` and related factory functions was typed to require returning a `Field`, but Django actually accepts `None` to skip fields. This matches `Field.formfield()` which returns `None` for fields like `BigAutoField`. The fix changes `_FormFieldCallback` to return `Field | None`

Fixes #3015


tests/assert_type/forms/test_models.py
sobolevn, line 1: I don't think we should test it :)

emmanuel-ferdman: Removed, thanks for the review)

sobolevn (approved):

Issue #2687, opened 2025-05-18
support for psycopg 3 `Range` types in `django.contrib.postgres.fields.ranges`


jorenham:
In `django.contrib.postgres.fields.ranges` the `Range` types are imported from [`psycopg2`](https://pypi.org/project/psycopg2/), which has been superseded by [`psycopg` 3](https://pypi.org/project/psycopg/):

https://github.com/typeddjango/django-stubs/blob/2446b523d5caee05dc064805216a07f03e8d828d/django-stubs/contrib/postgres/fields/ranges.pyi#L9

<sub>the [`types-psycopg2`](https://pypi.org/project/types-psycopg2/) typeshed stubs package could help with the `# type: ignore`, btw</sub>

The range fields use the psycopg2 `Range` type as upper bound for the generic type parameters of the range fields:

https://github.com/typeddjango/django-stubs/blob/2446b523d5caee05dc064805216a07f03e8d828d/django-stubs/contrib/postgres/fields/ranges.pyi#L28-L30

## Ok, so what's the problem?

So I'm not able to use the psycopg 3 `psycopg.types.range.Range` as type argument here. Note that `_RangeT` has no default, so I'm *required* to pass a type argument here.

## Working around it

The least invalid workaround would be to pass `Any` as type argument. But if you don't mind getting your hands dirty, you could also ignore the upper bound, and throw some of those `# type: ignore[misc]` and/or `# pyright: ignore[reportVerboseDescriptionOfAnErrorThatCouldAlsoBeWrittenInOneWOrd]` onto it, and it'll probably work regardless.

And I hate to say it, so please forgive me for putting it in such strong terms, but I consider both workarounds to be *suboptimal*. 

## Why it's impossible to solve, and some possible solutions.

That's, well, not possible... Or at least, not if you want to support both `psycopg2` and `psycopg` 3. Because `try: import ...` isn't a thing in stubs, unfortunately. 

However, it looks like both `Range` types from v2 and v3 are structurally equivalent. So another option would be to copy-paste the typeshed code of `Range`, swap the `Generic` for a `Protocol`, and voilà.

Another option would be to just get rid of the `_RangeT` upper bound altogether. 

## Covariance, maybe?

Since `_RangeT` only annotates `range_type: type[_RangeT]`, and assuming that it's intended as a read-only attribute, perhaps it makes sense to slap a `_co` on the `_RangeT`?

intgr:
[...]

PR #3037, opened 2026-02-04
Django 6.0: add postgres module stubs

emmanuel-ferdman:
## PR Summary
This PR adds postgres module stubs (sorry for big change). It contains the following changes:
- Add `LexemeCombinable` class for combining search lexemes with `&` and `|` operators.
- Add `CheckPostgresInstalledMixin` class for consistent postgres installation checking.
- Update `OrderableAggMixin` to reflect Django 6.0 refactoring: add `allow_order_by` class variable and `__init_subclass__`, remove methods moved to parent class.
- Add stubs for search utilities: `UTF8Dumper`, `quote_lexeme`, `normalize_spaces`, `psql_escape`, and regex patterns.
- Update `Lexeme` and `CombinedLexeme` to inherit from `LexemeCombinable` and add `__invert__` method.
- Update `ExclusionConstraint`, `SearchVectorField`, and `SearchQueryField` to inherit from `CheckPostgresInstalledMixin`.
- Add `psycopg` as test dependency (required for `UTF8Dumper` stub which inherits from `psycopg.adapt.Dumper`). Which adds more stubs todo.

Ref #2944.


tests/typecheck/test_import_all.yml
```diff
@@ -679,6 +679,8 @@
         import django.views.i18n
         import django.views.static
     mypy_config: |
+        [mypy-psycopg.*]
+        ignore_errors = true
```
sobolevn, line 683: Can you please explain why is this needed? Right here in the comments please :)

emmanuel-ferdman: @sobolevn I added the `[mypy-psycopg.*]` section because psycopg3 ships with its own inline type annotations (it has a `py.typed` marker), but those annotations have some internal inconsistencies and signature mismatches that mypy complains about. Since the new `UTF8Dumper` stub inherits from `psycopg.adapt.Dumper`, mypy will analyze psycopg's internal types when running the test suite, and any errors inside psycopg itself would fail the django-stubs tests. So I added `ignore_errors = true` to tell mypy to not report errors originating from within the psycopg package - same reason `redis.*` already has this setting in the same file.

That said, I'm not sure `UTF8Dumper` alone justifies adding psycopg3. But we do have `django.db.backends.postgresql.psycopg_any` in `allowlist_todo_django52.txt` and it re-exports many psycopg3 types (`Range`, `DateRange`, `Jsonb`, `IsolationLevel`, etc.) - see issue #2687 for context. so I think that if we ever stub that module properly we'd need psycopg3 anyway. What do you think?

sobolevn (commented):
One more question: is it safe to use `from psycopg import`? Is it even listed in our deps?

emmanuel-ferdman:
@sobolevn I believe so. we already import from psycopg2 (for example in [contrib/postgres/fields/ranges.pyi](https://github.com/typeddjango/django-stubs/blob/c8434d01c30d520729eae0fc125311ae686e6ef6/django-stubs/contrib/postgres/fields/ranges.pyi#L9)) with inline `# type: ignore [import-untyped]` (because psycopg2 has no types). psycopg3 has `py.typed` so it needs the mypy config approach like `redis`, so the PR adds `[mypy-psycopg.*] ignore_errors = true`. From what I tested, for end users mypy works whether they have psycopg3 installed or not - with psycopg3 they get full type info for inherited `Dumper` attributes (like `oid: int`), without it those fall back to `Any` but no errors occur. Also, in [Django 6.0 notes](https://docs.djangoproject.com/en/6.0/ref/databases/#postgresql-notes) we can see:
> Support for psycopg2 is likely to be deprecated and removed at some point in the future.

so I think we are in the right direction.

sobolevn (approved):

PR #3038, opened 2026-02-04
docs: Document system dependencies and clarify tool usage

rgant:
## Summary

Documents system dependencies required to run the test suite, which are currently undocumented and can cause confusing build failures for new contributors.

## Changes

- **New "System Dependencies" section** covering:
  - MySQL/MariaDB client libraries (required for `mysqlclient` to build)
  - GDAL and GEOS (optional, for `django.contrib.gis` tests)
  - Platform-specific instructions for Ubuntu/Debian, Fedora/RHEL, and macOS (Homebrew)
  - Note about macOS library path issues and workaround for GIS tests

- **Pre-commit clarification**: Added note that pre-commit must be installed separately, with link to installation options and mention of `uvx` alternative.

- **`uv run pytest`**: Changed from `pytest -n auto` to `uv run pytest -n auto` for consistency with uv-based workflow (works regardless of venv activation).

## Motivation

When setting up the dev environment on macOS, `uv sync` fails with a cryptic `mysqlclient` build error unless you have MariaDB/MySQL client libraries installed. Similarly, 2 GIS tests fail without GDAL/GEOS.
These dependencies aren't mentioned anywhere in the docs.

This PR documents what I learned while setting up the project for the first time.

## Test plan

- [x] `pre-commit run --files CONTRIBUTING.md` passes
- [x] Verified instructions work on macOS with Homebrew

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,30 @@ As a first step you will need to fork this repository and clone your fork locall
 In order to be able to continuously sync your fork with the origin repository's master branch, you will need to set up an upstream master.
 To do so follow this [official github guide](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/syncing-a-fork).
 
+### System Dependencies
+
+The test suite requires some system libraries that Django itself treats as optional.
+This project depends on `mysqlclient`, which needs MySQL/MariaDB C client libraries to build.
+Install them for your platform following the
+[mysqlclient install guide](https://github.com/PyMySQL/mysqlclient#install).
+For Debian/Ubuntu, the [django-docker-box packages list](https://github.com/django/django-docker-box/blob/main/packages.txt)
+is also a useful reference.
+
+**GDAL and GEOS** are needed to pass all tests (2 GIS-related tests require them).
+See the Django documentation on
+[installing geospatial libraries](https://docs.djangoproject.com/en/stable/ref/contrib/gis/install/geolibs/).
+If you're not working on GIS-related stubs, you can skip GDAL/GEOS —
+the 2 failing tests won't affect other contributions.
+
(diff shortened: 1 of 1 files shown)


rgant:
I'm working on a _real_ PR, but first I wanted to get this setup. So I documented my process as a fresh contributor to ensure the documentation is robust.

I normally format markdown files, but that will introduce a lot of noise so I didn't include it. If you are interested I would be happy to contribute a dprint setup to format markdown files for the project automatically.

ngnpope:
Can we not just point to the documentation already in Django to ensure that we stay in sync?

- https://docs.djangoproject.com/en/6.0/topics/install/#how-to-install-django
- https://docs.djangoproject.com/en/6.0/internals/contributing/writing-code/unit-tests/#running-all-the-tests
- https://docs.djangoproject.com/en/6.0/ref/contrib/gis/install/geolibs/#installing-geospatial-libraries

Another helpful hint as far as Debian/Ubuntu packages is this list:

- https://github.com/django/django-docker-box/blob/main/packages.txt

rgant:
Thanks for the feedback! I've updated the section to link to external docs for the install details:

- MySQL/MariaDB C libraries now link to the https://github.com/PyMySQL/mysqlclient#install (Django's own docs present this as optional, so they don't cover the system-level prerequisites)
- GDAL/GEOS now links to Django's https://docs.djangoproject.com/en/stable/ref/contrib/gis/install/geolibs/ page
- Added the https://github.com/django/django-docker-box/blob/main/packages.txt as a reference for Debian/Ubuntu

I kept the macOS GDAL/GEOS symlink workaround since that isn't covered in any upstream docs and is a real stumbling block for MacOS contributors. I also kept a brief note explaining that these dependencies are
 required here even though Django treats them as optional — a contributor coming from Django might not expect to need them. I did not.

UnknownPlatypus (approved):

UnknownPlatypus:
Thanks @rgant :raised_hands:

PR #3042, opened 2026-02-06
Add configurable `resolve_manager_on_typevars` option for .objects on type[T]

GregMFriedman:
## Summary

- Adds `objects` as a property on the `ModelBase` metaclass (consistent with `_default_manager` and `_base_manager`)
- Introduces a new `resolve_manager_on_typevars` config option (default: `false`) that controls whether the metaclass property is kept or removed
- When enabled, `type[T].objects` resolves to `Manager[T]` where `T` is a `TypeVar` bounded by a `Model` subclass
- When disabled (default), current behavior is fully preserved

## Motivation

When using `type[T]` where `T` is a `TypeVar` bounded by a Django Model subclass, accessing `.objects` on the result reports:

```
error: "type[T]" has no attribute "objects"  [attr-defined]
```

This pattern is common in generic model utilities, factory patterns, and plugin architectures where the concrete model type isn't known at the call site but is guaranteed to be a Model subclass. Meanwhile, `._default_manager` and `._base_manager` already work in this context because they are defined as properties on the `ModelBase` metaclass.

## Changes

| File | Change |
|------|--------|
| `django-stubs/db/models/base.pyi` | Add `objects` property to `ModelBase` metaclass |
| `mypy_django_plugin/config.py` | Add `resolve_manager_on_typevars` boolean config (default: `false`) |
| `mypy_django_plugin/transformers/models.py` | Guard metaclass property removal based on config |
| `tests/typecheck/managers/test_managers.yml` | 4 new test cases covering both config states |
| `tests/test_error_handling.py` | Config validation tests for INI and TOML |
| `README.md` | Document the new config option |

## Test plan

- [x] New tests: opt-in resolves `type[T].objects` to `Manager[T]`
- [x] New tests: default config preserves the existing error
- [x] New tests: `_default_manager`/`_base_manager` unaffected regardless of config
- [x] New tests: concrete models with custom managers not regressed
- [x] Existing manager tests pass (31/31)
- [x] Existing metaclass tests pass (3/3)
- [x] Error handling tests pass with updated usage strings
- [x] Ruff lint + format checks pass

Made with [Cursor](https://cursor.com)

### DIFF
--- a/django-stubs/db/models/base.pyi
+++ b/django-stubs/db/models/base.pyi
@@ -31,6 +31,11 @@ class ModelBase(type):
     def _default_manager(cls: type[_Self]) -> Manager[_Self]: ...  # type: ignore[misc]
     @property
     def _base_manager(cls: type[_Self]) -> Manager[_Self]: ...  # type: ignore[misc]
+    # This metaclass property is the fallback for 'objects' when accessed on type[T]
+    # where T is a TypeVar bounded by a Model subclass. It is removed by the plugin
+    # unless resolve_manager_on_typevars is enabled.
+    @property
+    def objects(cls: type[_Self]) -> Manager[_Self]: ...  # type: ignore[misc,unused-ignore]
 
 class Model(metaclass=ModelBase):
     # Note: these two metaclass generated attributes don't really exist on the 'Model'
--- a/README.md
+++ b/README.md
@@ -136,6 +136,16 @@ The supported settings are:
   [See here why](https://github.com/typeddjango/django-stubs?tab=readme-ov-file#how-to-use-typemodel-annotation-with-objects-attribute)
   this is dangerous to do by default.
 
+- `resolve_manager_on_typevars`, a boolean, default `false`.
+
+  Set to `true` to allow `.objects` to resolve on `type[T]` where `T` is a
+  `TypeVar` bounded by a Django `Model` subclass. This is useful for generic
+  model utilities, factory patterns, and plugin architectures where the concrete
+  model type isn't known at the call site but is guaranteed to be a `Model`
+  subclass. Without this setting, only `._default_manager` and `._base_manager`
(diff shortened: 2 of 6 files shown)


django-stubs/db/models/base.pyi
```diff
@@ -31,6 +31,11 @@ class ModelBase(type):
     def _default_manager(cls: type[_Self]) -> Manager[_Self]: ...  # type: ignore[misc]
     @property
     def _base_manager(cls: type[_Self]) -> Manager[_Self]: ...  # type: ignore[misc]
+    # This metaclass property is the fallback for 'objects' when accessed on type[T]
+    # where T is a TypeVar bounded by a Model subclass. It is removed by the plugin
+    # unless resolve_manager_on_typevars is enabled.
+    @property
+    def objects(cls: type[_Self]) -> Manager[_Self]: ...  # type: ignore[misc,unused-ignore]
```
sobolevn, line 38: This will change the default for all non-mypy type-checkers.

`.objects` does not exist for all `Model` classes :(

UnknownPlatypus: There a re also various github issues and discussion around this subject that we should probably take into account

GregMFriedman: to be honest, i drafted this whole PR and only realized after I opened it that replacing `objects` with `_default_manager` could accomplish what I was I looking to do.
Unless you think there's merit in addressing the issue @sobolevn pointed out (+ reading through the issues/discussion in github), i can just close this PR and apologize for bothering you both

UnknownPlatypus: No worries, for reference the readme talk about [the current workaround](https://github.com/typeddjango/django-stubs?tab=readme-ov-file#how-to-use-typemodel-annotation-with-objects-attribute).
Also [this discussion](https://github.com/typeddjango/django-stubs/discussions/2452) is relevant

PR #2940, opened 2025-12-09
Django 6.0: Add Model.NotUpdated & ObjectNotUpdated exceptions

intgr:
* Depends on #2939
* [x] Rebase needed after ^ is merged.
* Unblocks https://github.com/typeddjango/djangorestframework-stubs/pull/887

PR #3046, opened 2026-02-06
Django 6.0: Add plugin support for `NotUpdated` exception

ngnpope:
# I have made things!

`NotUpdated` was added in Django 6.0 and should work the same as `DoesNotExist` and `MultipleObjectsReturned`.

## Related issues

- Refs https://github.com/typeddjango/django-stubs/pull/2940

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -215,8 +215,14 @@ class AddAnnotateUtilities(ModelClassInitializer):
 
     def run(self) -> None:
         annotations = self.lookup_typeinfo_or_incomplete_defn_error("django_stubs_ext.Annotations")
-        object_does_not_exist = self.lookup_typeinfo_or_incomplete_defn_error(fullnames.OBJECT_DOES_NOT_EXIST)
-        multiple_objects_returned = self.lookup_typeinfo_or_incomplete_defn_error(fullnames.MULTIPLE_OBJECTS_RETURNED)
+        exception_bases = {
+            model_exc_name: self.lookup_typeinfo_or_incomplete_defn_error(base_exc_name)
+            for model_exc_name, base_exc_name in [
+                ("DoesNotExist", fullnames.OBJECT_DOES_NOT_EXIST),
+                ("NotUpdated", fullnames.OBJECT_NOT_UPDATED),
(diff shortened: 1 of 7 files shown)


mypy_django_plugin/transformers/models.py
```diff
@@ -235,12 +233,17 @@ def run(self) -> None:
...
-                )
-                helpers.add_new_sym_for_info(
-                    annotated_model, "MultipleObjectsReturned", TypeType(Instance(multiple_objects_returned, []))
-                )
+                for model_exc_name, base_exc_name in [
+                    ("DoesNotExist", fullnames.OBJECT_DOES_NOT_EXIST),
+                    ("NotUpdated", fullnames.OBJECT_NOT_UPDATED),
+                    ("MultipleObjectsReturned", fullnames.MULTIPLE_OBJECTS_RETURNED),
+                ]:
+                    base_exc_type = self.lookup_typeinfo_or_incomplete_defn_error(base_exc_name)
```
emmanuel-ferdman: Hey, nice work on this! I have a small question regarding this part - before this PR, the exception base lookups happened before the `if annotated_model is None:` check, so if they failed and deferred, nothing was created yet. Now they're inside that block, after the annotated model is already saved. So if a lookup fails mid-way, the model exists but without the exception symbols, and on retry the guard sees it already exists and skips the whole thing. Should we move the lookups back outside the guard? We could do even something like:
```python
     def run(self) -> None:
         annotations = self.lookup_typeinfo_or_incomplete_defn_error("django_stubs_ext.Annotations")
         exception_bases = {
            name: self.lookup_typeinfo_or_incomplete_defn_error(fullname)
            for name, fullname in [
                ("DoesNotExist", fullnames.OBJECT_DOES_NOT_EXIST),
                ("NotUpdated", fullnames.OBJECT_NOT_UPDATED),
                ("MultipleObjectsReturned", fullnames.MULTIPLE_OBJECTS_RETURNED),
            ]
        }
...
                for model_exc_name, base_exc_type in exception_bases.items():
                     helpers.add_new_sym_for_info(
                         annotated_model,
                         model_exc_name,
```

ngnpope: Yeah, I guess that's fair. It's probably unlikely to happen, but it would risk that this bit would no longer be executed post-deferral as the annotated model would have already been added. Will adjust to your suggestion.

sobolevn (approved):

PR #3055, opened 2026-02-09
Add choices property to Choices class to fix functional API error

### LINKED ISSUE #1858: "TextChoices" has no attribute "choices"
# Bug report
## What's wrong
When using an enum created from the functional enum API, mypy complains that the enum doesn't have an attribute `choices`, however it does.

## How is that should be
`choices` should be a supported attribute when the enum is instantiated.

## System information

- OS:
- `python` version: 3.10.12
- `django` version: 4.2.8
- `mypy` version: 1.7.1

## Repro
```python
from django.db.models import TextChoices

print(TextChoices("test", "A B C").choices)
```

```
repro.py:3:1: error: "TextChoices" has no attribute "choices"  [attr-defined]
```

If you run the script you see the choices are printed correctly:

```
$ python repro.py 
[('A', 'A'), ('B', 'B'), ('C', 'C')]
```

TOMAL5085:
Fixes #1858.

I have added the choices property to the base Choices class in enums.pyi. This ensures that mypy recognizes the attribute when using the functional enum API, preventing the attr-defined error.

This follows the pragmatic approach suggested by @intgr to improve support for dynamic enums.


ngnpope:
I still firmly disagree with this. The correct solution is for mypy to fix [this issue](https://github.com/python/mypy/issues/6037).

sobolevn:
I trust @ngnpope's judgement here :)

TOMAL5085:
Thank you for the detailed review and for running the Mypy Primer. I understand the decision to wait for a fix in Mypy core rather than adding a workaround here. This was a great learning experience for me.

intgr:
> I understand the decision to wait for a fix in Mypy core

Welp, it hasn't been solved in 6 years... but perhaps in the next 6? :sweat_smile:

PR #3065, opened 2026-02-11
fix: allow get_violation_error_message to return dictionaries

### LINKED ISSUE #3062: Can not return dict[str, str] from overridden get_violation_error_message
# Bug report

## What's wrong

I'm getting this issue

```
error: Return type "dict[str, str]" of "get_violation_error_message" incompatible with return type "str" in supertype "django.db.models.constraints.BaseConstraint"  [override]
```

This is the current signature:

https://github.com/typeddjango/django-stubs/blob/9e105d9e7cb789d321630af15663918e2e90c8ec/django-stubs/db/models/constraints.pyi#L39

## How is that should be

Maybe I'm abusing the spirit of the base code, but I believe I should be able to return a `dict[str, str]` when overriding  `django.db.models.BaseConstraint.get_violation_error_message`, [since it's passed directly into a `ValidationError`](https://github.com/django/django/blob/0dac3dd4a1573b3c9cef3aea6a98440decfc5460/django/db/models/constraints.py#L217).

Code:

```python
class UniqueEmailConstraint(UniqueConstraint):

        kwargs.setdefault("name", "user_email_ci_uniqueness")
        # *args omitted, otherwise it will infinitely generate migrations
        super().__init__(Lower("email"), **kwargs)

    def get_violation_error_message(self) -> dict[str, str]:
        """Override default message and specify field to ValidationError."""
        return {"email": "User with this email address already exists"}
```

## System information

- OS:
- `python` version: 3.11.13
- `django` version: 5.2.11
- `mypy` version: 1.19.1
- `django-stubs` version: 5.2.9
- `django-stubs-ext` version: 5.2.9

meunomemauricio:
# I have made things!

Updated the signature of `django.db.model.constraints.BaseConstraint.get_violation_error_message` to  `_StrOrPromise | dict[str, Any]` and wrote some tests.

I based the changes on the [signature of `message` in `django.core.exceptions.ValidationError.__init__`](https://github.com/typeddjango/django-stubs/blob/c5c108cee94f80cb0051f9273164b7a5dbbfaf26/django-stubs/core/exceptions.pyi#L41).

I wasn't sure if I should go all the way and include `ValidationError` and `list[Any]` too, though. I also haven't touched the `violation_error_message` class attribute, trying to keep the changes more punctual to my case.

But please advise and I can expand the changes to include them if it's a good idea.

## Related issues

Closes #3062


sobolevn (approved):
Thanks!

meunomemauricio:
@sobolevn out of curiosity, were there any issues with the tests?

PR #3068, opened 2026-02-12
add a test for WithAnnotations and custom querysets + managers

rtpg:
This snippet is currently not passing but I believe it _should_ pass. 

I am a bit worried that [absent an extra generic var on Manager](https://github.com/typeddjango/django-stubs/discussions/3067) we might have trouble simplifying things for this case. But there might be some simpler way forwards


sobolevn:
And tests failed 😄

PR #3077, opened 2026-02-13
fix(mypy): Move AnonymousUser up to avoid mypy defer error

andres-ins:
# I have made things!


andres-ins:
Ah, so sorry, I didn't mean to make the PR against the actual repo 😬

PR #3080, opened 2026-02-13
Fix conflated decorator and decorator factory return types

### LINKED ISSUE #2824: Incorrect Union typing of decorators in django-stubs\utils\decorators.pyi
# Bug report

## What's wrong

In django-stubs\utils\decorators.pyi, the typing of decorators is as follow:

```python
# django-stubs\utils\decorators.pyi
_DECORATOR: TypeAlias = Callable[..., Callable[..., HttpResponseBase] | Callable[..., Callable[..., HttpResponseBase]]]

# This hint is used for the decorator_from_middleware methods
def decorator_from_middleware_with_args(middleware_class: type) -> _DECORATOR: ...
def decorator_from_middleware(middleware_class: type) -> _DECORATOR: ...
```

That creates several issues when attempting to build custom middleware decorators. We show below a minimal example, and two simplified use cases where type errors are incorrectly generated because of the Union type.

```python
import functools
from typing import Any, Callable, ParamSpec

from django.http.response import HttpResponseBase
from django.utils.decorators import (
    decorator_from_middleware_with_args,
    decorator_from_middleware,
)
from django.middleware.cache import CacheMiddleware

P = ParamSpec("P")


class CustomCacheMiddleWare(CacheMiddleware): ...


# Minimal issue
# We provide the expected base example of the decoration
# The example works as expected
@decorator_from_middleware_with_args(CustomCacheMiddleWare)(page_timeout=1)
def example_function(*args: Any, **kwargs: Any) -> HttpResponseBase:
    return HttpResponseBase()


# The below is problematic. That should not be allowed
value = decorator_from_middleware_with_args(CustomCacheMiddleWare)(page_timeout=1)(1) # <--- no error
if not callable(value):
    # If called with anything, a HttpResponse is a valid result!
    # problematic for a decorator
    reveal_type(value)  # Revealed type is "django.http.response.HttpResponseBase" !


# Minimal "real life" simplified application issue: building a custom cache page
def custom_cache_page(
    test_parameter: str,
) -> Callable[[Callable[P, HttpResponseBase]], Callable[P, HttpResponseBase]]:
    print(test_parameter)

    def main_decorator(
        func: Callable[P, HttpResponseBase],
    ) -> Callable[P, HttpResponseBase]:
        decorator = decorator_from_middleware_with_args(CustomCacheMiddleWare)()
        return functools.wraps(func)(decorator(func))  # mypy error:
        # Argument 1 to "__call__" of "_Wrapper" has incompatible type
        # "HttpResponseBase | Callable[..., HttpResponseBase]";
        # expected "Callable[[VarArg(Any), KwArg(Any)], HttpResponseBase]"  [arg-type]

    return main_decorator

# Minimal "real life" simplified application issue 2: building a custom cache page with no parameter
def custom_cache_page_2(
    test_parameter: str,
) -> Callable[[Callable[P, HttpResponseBase]], Callable[P, HttpResponseBase]]:
    print(test_parameter)

    return decorator_from_middleware(CustomCacheMiddleWare)  # mypy error:
    # error: Incompatible return value type (got
    # "Callable[..., Callable[..., HttpResponseBase] | Callable[..., Callable[..., HttpResponseBase]]]",
    # expected "Callable[[Callable[P, HttpResponseBase]], Callable[P, HttpResponseBase]]")  [return-value]
```

## How is that should be

We should have a different hint for the decorators and the decorators with parameters.
In the example below, we use TypeVar and ParamSpec, but ellipsis and HttpResponseBase for the return type
also work if needed.
With those amendments, the problematic line from above does raise a mypy error, and the "real life" applications do not.

```python
# django-stubs\utils\decorators.pyi

from typing import Callable, ParamSpec, TypeVar

from django.http.response import HttpResponseBase

_HttpResponseType = TypeVar("_HttpResponseType", bound=HttpResponseBase)
_ViewParamSpec = ParamSpec("_ViewParamSpec")

# The decorator return value needs to be a function and not a decorator
# Before: _DECORATOR: TypeAlias = Callable[..., Callable[..., HttpResponseBase] | Callable[..., Callable[..., HttpResponseBase]]]
# Correction with previous notations: _DECORATOR: TypeAlias = Callable[[Callable[..., HttpResponseBase]], Callable[..., HttpResponseBase]]
_DECORATOR: TypeAlias = Callable[
    [Callable[_ViewParamSpec, _HttpResponseType]],
    Callable[_ViewParamSpec, _HttpResponseType],
]  # <--- Standard decorator definition, with return value bound to HttpResponseBase

# We introduce a _DECORATOR_WITH_PARAMETERS type, this one's return type is a decorator
_DECORATOR_WITH_PARAMETERS: TypeAlias = Callable[..., _DECORATOR]

# decorator_from_middleware_with_args is actually returning a decorator with parameters
def decorator_from_middleware_with_args(
    middleware_class: type,
) -> _DECORATOR_WITH_PARAMETERS: ...


# decorator_from_middleware typing doesn't change
def decorator_from_middleware(middleware_class: type) -> _DECORATOR: ...

```
## System information

- OS: Windows
- `python` version: 3.10
- `django` version: 4.2
- `mypy` version: 3.10
- `django-stubs-ext` version: 5.0.0

emmanuel-ferdman:
## PR Summary
The `_DECORATOR` type alias was incorrectly defined as a union that conflated decorators with decorator factories. This caused type errors when using `decorator_from_middleware` in return type annotations and failed to catch bugs like passing non-callables to decorators. This PR splits it into two distinct types: `_ViewDecorator` for actual decorators and `_DecoratorFactory` for functions that produce them.

Fixes #2824.


django-stubs/utils/decorators.pyi
```diff
@@ -7,16 +7,21 @@ from django.views.generic.base import View
...
 
 def method_decorator(
-    decorator: _DECORATOR | Iterable[_DECORATOR], name: str = ""
+    decorator: _ViewDecorator | _DecoratorFactory | Iterable[_ViewDecorator | _DecoratorFactory],
+    name: str = "",
 ) -> Callable[[_ViewType], _ViewType]: ...
-def decorator_from_middleware_with_args(middleware_class: type) -> _DECORATOR: ...
-def decorator_from_middleware(middleware_class: type) -> _DECORATOR: ...
-def make_middleware_decorator(middleware_class: type[MiddlewareMixin]) -> _DECORATOR: ...
+def decorator_from_middleware_with_args(middleware_class: type) -> _DecoratorFactory: ...
```
sobolevn: can we be more specific here? can we type `middleware_class` as a protocol maybe? can we type `_DecoratorFactory` args?

UnknownPlatypus (commented):
Can we add some assert_type tests using samples from the related issue ?

django-stubs/utils/decorators.pyi
```diff
@@ -1,22 +1,31 @@
...
 
 def method_decorator(
-    decorator: _DECORATOR | Iterable[_DECORATOR], name: str = ""
+    decorator: _ViewDecorator | _DecoratorFactory | Iterable[_ViewDecorator | _DecoratorFactory],
+    name: str = "",
 ) -> Callable[[_ViewType], _ViewType]: ...
-def decorator_from_middleware_with_args(middleware_class: type) -> _DECORATOR: ...
-def decorator_from_middleware(middleware_class: type) -> _DECORATOR: ...
-def make_middleware_decorator(middleware_class: type[MiddlewareMixin]) -> _DECORATOR: ...
+def decorator_from_middleware_with_args(middleware_class: _MiddlewareClass) -> _DecoratorFactory: ...
```
sobolevn, line 26:
```suggestion
def decorator_from_middleware_with_args(middleware_class: type[_MiddlewareClass]) -> _DecoratorFactory: ...
```

?

django-stubs/utils/decorators.pyi
emmanuel-ferdman: @sobolevn Looked into both questions. For `middleware_class` - yes, switched to a `Protocol` with `__call__` that validates the middleware signature (first arg is get_response, rest is `*args`/`**kwargs`). This catches invalid types like `str` while accepting all Django middleware. For typing the factory args precisely with ParamSpec - it's doable but adds complexity and ParamSpec support varies across checkers, so kept it as `Callable[..., _ViewDecorator]` for now. Do you think its worth it?

django-stubs/utils/decorators.pyi
emmanuel-ferdman: @sobolevn Tried `type[_MiddlewareClass]` but it doesn't work - mypy rejects all middleware classes. I think the issue is `type[Protocol]` checks if instances match the protocol, but we need to match how calling the class works (the constructor). The current `_MiddlewareClass` without `type[]` works because Protocol's `__call__` matches the class constructor signature.

emmanuel-ferdman:
@UnknownPlatypus Added two test cases to `tests/typecheck/utils/test_decorators.yml` that fail on master and pass with the fix. Let me know if that what you had in mind (because I see there's no `tests/assert_type/utils/` in the repo). btw what's the rule of thumb for adding tests to this project? I wondered that when I opened the PR.

django-stubs/utils/decorators.pyi
emmanuel-ferdman: @sobolevn Tried `type[_MiddlewareClass]` but it rejects all middleware classes. The issue is `type[Protocol]` checks if instances match the protocol, but middleware instances have `__call__(self, request)`, not `__call__(self, get_response, ...)`. The current approach without `type[]` works because it checks if the class itself is callable with `(get_response, *args, **kwargs)` - which matches the constructor.

sobolevn (approved):

UnknownPlatypus:
> Let me know if that what you had in mind (because I see there's no tests/assert_type/utils/ in the repo)

We have a `tests/assert_type` folder, feel free to create new files in sideit, it's just more recent so not completely exhaustive.
For your specific case, it's a good fit and would also test pyrefly and pyright.

> btw what's the rule of thumb for adding tests to this project? I wondered that when I opened the PR

Usually it's something like that:
- for trivial stub changes, no test really required
- for stub changes that are more involved (this pr working with decorators, changes using generics, ...), test are great to also showcase actual usages. 
- For pr fixing an issue with a clear reproducible exemple snippet, adding a regression test is great to ensure we d'ont break it again later. 

As for tests, they can either be added to `tests/typecheck` or `tests/assert_type`.
If possible it's best to use `tests/assert_type` because the test are much faster and also check pyright and pyrefly.
However, it does not currently work well if you have to use django models or alter settings. Any code rellying on the mypy plugin will also make pyright/pyrefly fail so these are usually in `tests/typecheck`

@sobolevn also probably have some valuable input on this and we should document it in the CONTRIBUTING.md file maybe ?

emmanuel-ferdman:
@UnknownPlatypus thanks for the detailed explanation. Moved the tests to `tests/assert_type` :)

UnknownPlatypus:
Thanks for the continuous work on the project, this is very cool to have new active contributors! Great work!

PR #3081, opened 2026-02-14
Django 6.0: remove deprecated stubs and add referenced_fields

emmanuel-ferdman:
## PR Summary
This PR removes stubs for methods and classes that were deprecated in Django 4.2/5.0 and finally removed in Django 6.0.

Ref #2944.


UnknownPlatypus (commented):
Looking good, just wondering if we should keep them around a bit since we declare partial support for 5.0 and this helps people migrating ?

Wdyt @sobolevn ?

sobolevn (approved):
I think we should remove them, because `deprecated` is an optional error code in mypy. So, people might use something on 6.0, get no type checker error and will get a runtime error.

While partial support means that we support only, well, a part of things :)

PR #3090, opened 2026-02-16
Fix login_required overloads and auth model stubs

emmanuel-ferdman:
## PR Summary
This PR fixes a few auth stub issues that were in the allowlist.


django-stubs/contrib/auth/decorators.pyi
```diff
@@ -12,16 +12,18 @@ def user_passes_test(
...
 ) -> Callable[[_VIEW], _VIEW]: ...
-
-# There are two ways of calling @login_required: @with(arguments) and @bare
 @overload
 def login_required(
-    redirect_field_name: str | None = ..., login_url: _StrOrPromise | None = ...
-) -> Callable[[_VIEW], _VIEW]: ...
+    function: _VIEW,
+    redirect_field_name: str | None = ...,
+    login_url: _StrOrPromise | None = ...,
```
sobolevn, line 19:
```suggestion
```

Shouldn't `function` be the only parameter here?

django-stubs/contrib/auth/models.pyi
```diff
@@ -123,6 +123,7 @@ class AnonymousUser:
     is_staff: Literal[False]
     is_active: Literal[False]
     is_superuser: Literal[False]
+    def __int__(self) -> NoReturn: ...
```
```suggestion
    def __int__(self) -> Never: ...
```

django-stubs/contrib/auth/decorators.pyi
```diff
@@ -12,16 +12,18 @@ def user_passes_test(
...
+) -> _VIEW: ...
 @overload
 def login_required(
-    function: _VIEW, redirect_field_name: str | None = ..., login_url: _StrOrPromise | None = ...
-) -> _VIEW: ...
+    function: None = None,
```
sobolevn, line 23: Can you please check: do we have tests for this function? If not, can you please add one to `assert_type`?

django-stubs/contrib/auth/decorators.pyi
emmanuel-ferdman: I believe the extra params are needed - `login_required(view_func, redirect_field_name='a')` is valid Django code and there's already a test for it in [test_decorators.yml](https://github.com/typeddjango/django-stubs/blob/master/tests/typecheck/contrib/auth/test_decorators.yml#L33). See example in Django's source code [here](https://github.com/django/django/blob/fb3a11071aae27ef869d2b029289b9f59cc41128/tests/auth_tests/test_decorators.py#L99). I also tried your suggestion and that test fails with "No overload variant matches argument types". The original stub also had these params in the second overload for the same reason.

django-stubs/contrib/auth/decorators.pyi
emmanuel-ferdman: Look like there are already tests in [test_decorators.yml](https://github.com/typeddjango/django-stubs/blob/master/tests/typecheck/contrib/auth/test_decorators.yml) covering all the patterns - bare decorator, with args, and even the `login_required(view, redirect_field_name='a')` case. Want me to add any specific case that's missing?

sobolevn (approved):
Thanks! All good :)

PR #3096, opened 2026-02-20
Add model expressions stubs

emmanuel-ferdman:
## PR Summary
This PR adds stubs that were missing from `expressions.pyi`.


sobolevn (approved):
Wow! So many things are fixed! 🤝

PR #3097, opened 2026-02-20
Handle lookups in Annotated fields 

### LINKED ISSUE #3098: Lookup of annotated fields
# Bug report

## What's wrong
Lookup parts of annotated queryset are not taken into account, so e.g. using `__isnull` on a annotated field fails.

See failing test case here: https://github.com/typeddjango/django-stubs/pull/3097/changes/b37cd889b83f188d074e0651d1a9d8b2ac80c905 , with (claude generated) fix in https://github.com/typeddjango/django-stubs/pull/3097

## How is that should be

Annotation taken into account 

## System information

- OS: linux
- `python` version: `3.13`
- `django` version: `5.2.11`
- `mypy` version: `1.19.1`
- `django-stubs` version: `5.2.9`
- `django-stubs-ext` version:  `5.2.9`

gg-mmill:
The test documents the failure to take lookups into account on annotated fields

# I have made things!

- add test documenting failure of using lookup on an annotated field
- (claude generated) fix
## Related issues

- Closes #3098

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -512,7 +512,7 @@ class DjangoContext:
 
         lookup_base_field, *annotation_lookup_parts = lookup.split("__")
 
-        if not lookup_base_field in model_instance.extra_attrs.attrs:
+        if lookup_base_field not in model_instance.extra_attrs.attrs:
             return None
 
         if annotation_lookup_parts:
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -474,9 +474,7 @@ class DjangoContext:
                         # Resolve the lookup class (e.g. "isnull" -> IsNull)
                         lookup_cls = Field().get_lookup(annotation_lookup_parts[-1])
                         if lookup_cls is not None:
-                            lookup_info = helpers.lookup_class_typeinfo(
-                                helpers.get_typechecker_api(ctx), lookup_cls
-                            )
+                            lookup_info = helpers.lookup_class_typeinfo(helpers.get_typechecker_api(ctx), lookup_cls)
                             if lookup_info is not None:
                                 for lookup_base in helpers.iter_bases(lookup_info):
                                     if lookup_base.args and isinstance(
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -459,34 +459,60 @@ class DjangoContext:
             raise LookupsAreUnsupported()
         return self._resolve_field_from_parts(field_parts, model_cls)
 
+    def _resolve_lookup_type_from_lookup_class(
+        self, ctx: MethodContext, lookup_cls: type, field: Union["Field[Any, Any]", ForeignObjectRel] | None = None
+    ) -> MypyType | None:
+        """Resolve the expected type for a lookup class.
+
+        Args:
+            ctx: The method context
+            lookup_cls: The Django lookup class (e.g., IsNull, Contains)
+            field: Optional field for resolving Field-dependent types
+
+        Returns:
+            The resolved type, or None if it couldn't be determined
+        """
+        lookup_info = helpers.lookup_class_typeinfo(helpers.get_typechecker_api(ctx), lookup_cls)
+        if lookup_info is None:
+            return None
+
+        for lookup_base in helpers.iter_bases(lookup_info):
+            if lookup_base.args and isinstance((lookup_type := get_proper_type(lookup_base.args[0])), Instance):
+                # if it's Field, consider lookup_type a __get__ of current field
+                if lookup_type.type.fullname == fullnames.FIELD_FULLNAME:
+                    if field is None:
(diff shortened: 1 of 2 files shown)


UnknownPlatypus:
What is the mypy error you get without this exactly ?

gg-mmill:
> What is the mypy error you get without this exactly ?

Sorry should have made that clearer, let's say I have an annotated model , and I use a lookup filter on the annotated field:

```
class MyAnnotations(TypedDict):
    scan_id: UUID | None

AnnotatedModel = WithAnnotations[MyModel, MyAnnotations]

def filter_annotated(qs: QuerySet[AnnotatedModel]) -> QuerySet[AnnotatedModel]:
    return qs.filter(scan_id__isnull=True)
```

the error is 
```
Incompatible type for lookup 'scan_id__isnull': (got "bool", expected "UUID | None")
```

mypy_django_plugin/django/context.py
```diff
@@ -468,8 +468,25 @@ def resolve_lookup_expected_type(
             if helpers.is_annotated_model(model_instance.type) and model_instance.extra_attrs:
                 # If the field comes from .annotate(), we assume Any for it
                 # and allow chaining any lookups.
-                lookup_base_field, *_ = lookup.split("__")
+                lookup_base_field, *annotation_lookup_parts = lookup.split("__")
```
UnknownPlatypus: We already handle correctly `isnull` resolving in other cases as we can see in [this test](https://github.com/typeddjango/django-stubs/blob/98c37f35283e9bb703d52023213d77acb1e1ef58/tests/typecheck/managers/querysets/test_filter.yml#L110) so we should be able to reuse something already existing instead.

I'm not super familiar with this part of the plugin code so I can't give you a more precise hint sadly

gg-mmill: Added a commit with some code factorisation

mypy_django_plugin/django/context.py
```diff
@@ -459,17 +459,60 @@ def resolve_lookup_into_field(
...
         except FieldError as exc:
+            # We handle annotation lookups here using _resolve_lookup_type_from_lookup_class,
+            # which is the same helper used for regular field lookups below.
             if helpers.is_annotated_model(model_instance.type) and model_instance.extra_attrs:
-                # If the field comes from .annotate(), we assume Any for it
-                # and allow chaining any lookups.
-                lookup_base_field, *_ = lookup.split("__")
+                lookup_base_field, *annotation_lookup_parts = lookup.split("__")
                 if lookup_base_field in model_instance.extra_attrs.attrs:
+                    if annotation_lookup_parts:
```
sobolevn: Can we please refactor it into a separate function. Right now it is really hard to read what it does.

gg-mmill: Done, also added another test case

mypy_django_plugin/django/context.py
```diff
@@ -493,27 +493,47 @@ def _resolve_lookup_type_from_lookup_class(
...
+            return None
+
+        lookup_base_field, *annotation_lookup_parts = lookup.split("__")
+        if lookup_base_field in model_instance.extra_attrs.attrs:
+            if annotation_lookup_parts:
+                lookup_cls = Field().get_lookup(annotation_lookup_parts[-1])
+                if lookup_cls is not None:
+                    lookup_type = self._resolve_lookup_type_from_lookup_class(ctx, lookup_cls)
+                    if lookup_type is not None:
+                        return lookup_type
```
sobolevn: Can we use early returns pattern here instead? The level of nesting is quite hard to read still :)

gg-mmill: Added a bit more early return --- I'm not convinced adding more would be beneficial, this would give 

```py
        if not annotation_lookup_parts:
            return model_instance.extra_attrs.attrs[lookup_base_field]

        lookup_cls = Field().get_lookup(annotation_lookup_parts[-1])
        if lookup_cls is None:
            return model_instance.extra_attrs.attrs[lookup_base_field]

        lookup_type = self._resolve_lookup_type_from_lookup_class(ctx, lookup_cls)
        if lookup_type not None:
            return model_instance.extra_attrs.attrs[lookup_base_field]

        return lookup_type

```

(i.e. triple `return model_instance.extra_attrs.attrs[lookup_base_field]`)

sobolevn (approved):

UnknownPlatypus (approved):

mypy_django_plugin/django/context.py
UnknownPlatypus: Could maybe be simplified with walrus ? 

```python
        if (
            not annotation_lookup_parts
            or (lookup_cls := Field().get_lookup(annotation_lookup_parts[-1])) is None
            or (lookup_type := self._resolve_lookup_type_from_lookup_class(ctx, lookup_cls)) is None
        ):
            return model_instance.extra_attrs.attrs[lookup_base_field]
        return lookup_type
```

PR #3103, opened 2026-02-21
Fix `Subquery` and `Exists` to accept subquery inputs

### LINKED ISSUE #985: Incomplete `Exists` argument annotation
# Bug report

## What's wrong

The [Exists](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/expressions.pyi#L145) class' first argument is currently type hinted as `queryset: Union[Query, QuerySet]`. I think `Subquery` or a TypeVar bound by `Subquery` should be added to the union.

This code currently fails type checking:
```python
from django.db.models import Exists, Subquery

from app.models import Project

qs = Project.objects.filter(
    Exists(
        Subquery(
            Project.objects.filter(status='created')
        )
    )
)
```
```
test.py:6:12: error: Argument 1 to "Exists" has incompatible type "Subquery"; expected "Union[Query, _QuerySet[Any, Any]]"  [arg-type]
```

but in reality, both of these snippets work and produce identical queries

```python
qs1 = Project.objects.filter(Exists(Subquery(Project.objects.filter(status='created'))))
qs2 = Project.objects.filter(Exists(Project.objects.filter(status='created')))
print(qs1.query == qs2.query)  # True
```

emmanuel-ferdman:
# PR Summary
This PR fixes the `Subquery`, `Exists`, and `ArraySubquery` constructors to correctly accept `Subquery` instances as input, not just `Query` or `QuerySet`. This matches [Django's runtime behavior](https://github.com/django/django/blob/stable/6.0.x/django/db/models/expressions.py#L1782) where any object with a `.query` attribute works, including subquery subclasses.

Fixes #985.


UnknownPlatypus (commented):
Seems good to me but it's a bit weird to do 
`Project.objects.filter(Exists(Subquery(Project.objects.filter(status='created'))))` 
in the first place when you can do 
`Project.objects.filter(Exists(Project.objects.filter(status='created')))` 
and have the same result isn't it ?

emmanuel-ferdman:
@UnknownPlatypus Yeah I agree but it looks like there are real usecases - [Saleor](https://github.com/saleor/saleor/blob/3e5172f5acd07cb8ef13bbf6c44b00bdaab80b24/saleor/checkout/tasks.py#L83-L89) hits this exact issue and has even a `# type: ignore[arg-type]` with a comment linking to the issue.

PR #3100, opened 2026-02-21
Django 6.0: Add types BaseTextDumper in postgis base 

ahmedasar00:
This PR adds missing type stubs for BaseTextDumper in
django.contrib.gis.db.backends.postgis.base.

Changes:

Added `dump(obj: Any) -> bytes` method signature

Removed `BaseTextDumper` entry from TODO/allowlist

PR #2327, opened 2024-08-07
Add a lot of runtime defaults to stubs with `stubdefaulter`

sobolevn:
You can do the same with:

```bash
# Install
pip install stubdefaulter no_implicit_optional

# Run
PYTHONPATH='PATH_TO_YOUR/site-packages/django' stubdefaulter -p django-stubs
no_implicit_optional django-stubs --use-union-or
```

PR #3105, opened 2026-02-22
Add types for register_geometry_adapters in postgis base

DeXtAr47-oss:
Hi I have added stubs for  register_geometry_adapters in DatabaseWrapper.

## Related issues
- Refs #2944 

As this is my first pr for Django a help or any comment on my pr from mentors would be much appreciated.


ahmedasar00:
Hi @DeXtAr47-oss 
**Just to let you know, I’ve already opened a PR for this part here:**
https://github.com/typeddjango/django-stubs/pull/3100
_Feel free to take a look_ 
Thanks.

DeXtAr47-oss:
Thanks for pointing out @ahmedasar00 I will make the changes.

django-stubs/contrib/gis/db/backends/postgis/base.pyi
```diff
@@ -15,3 +15,4 @@ class DatabaseWrapper(Psycopg2DatabaseWrapper):
     introspection: Any
     def __init__(self, *args: Any, **kwargs: Any) -> None: ...
     def prepare_database(self) -> None: ...
+    def register_geometry_adapters(self, pg_connection: Any, cache: Any) -> None: ...
```
UnknownPlatypus: This is incorrect and does not match the actual argument names.
See the implementation in django https://github.com/django/django/blob/158fd81ef5a5647d27eb3065063284f9ee0a3ca4/django/contrib/gis/db/backends/postgis/base.py#L148

Also as a general rule, avoid Any as much as possible because it's too loose of a type and will prevent the type checker from doing a good job

UnknownPlatypus (changes requested):

django-stubs/contrib/gis/db/backends/postgis/base.pyi
DeXtAr47-oss: hi thanks for the clarification i had made the changes could you plese check it.

UnknownPlatypus: What about pg_connection ? 
We can see in [the implementation](https://github.com/django/django/blob/158fd81ef5a5647d27eb3065063284f9ee0a3ca4/django/contrib/gis/db/backends/postgis/base.py#L133-L138) that this connection object is passed to `TypeInfo.fetch` that is type [from psycopg](https://github.com/django/django/blob/158fd81ef5a5647d27eb3065063284f9ee0a3ca4/django/contrib/gis/db/backends/postgis/base.py#L133-L138).

We should probably match the type they use for the connection

DeXtAr47-oss: OK ok i had made my changes again plz check it once.

UnknownPlatypus: You are quite close but I think it should be BaseConnection based on https://github.com/psycopg/psycopg/blob/db3c43584320ab5d97e49378e5c9dc09a560b031/psycopg/psycopg/_typeinfo.py#L62-L77 ?

django-stubs/contrib/gis/db/backends/postgis/base.pyi
```diff
@@ -15,3 +16,4 @@ class DatabaseWrapper(Psycopg2DatabaseWrapper):
     introspection: Any
     def __init__(self, *args: Any, **kwargs: Any) -> None: ...
     def prepare_database(self) -> None: ...
+    def register_geometry_adapters(self, pg_connection: BaseConnection[bytes], clear_caches: bool) -> None: ...
```
UnknownPlatypus: It's missing the fact that clear_caches is optional, see stubtest failure in ci https://github.com/typeddjango/django-stubs/actions/runs/22287057095/job/64467882584?pr=3105

DeXtAr47-oss: i had kept the `clear_caches` as given i hope there would be no failure this time

DeXtAr47-oss:
@UnknownPlatypus thanks for the help i do want to contribute more to django could you suggest me some issues that i should go on next or any suggestions would be much helpful for understanding the codebase

UnknownPlatypus:
@DeXtAr47-oss You can have a look at the other [allowlist](https://github.com/typeddjango/django-stubs/blob/master/scripts/stubtest/allowlist_todo.txt) file, it contains a bunch of known discrepancies between django runtime and the stubs we have.

Help in trimming this list is greatly appreciated. 
To better understand the codebase, check the `CONTRIBUTING.md` file and django source code.

django-stubs/contrib/gis/db/backends/postgis/base.pyi
UnknownPlatypus: i've updated to keep ther version with the default visible. We tend to prefer that to be more explicit here but both are working.
See https://github.com/typeddjango/django-stubs/pull/2327

PR #3106, opened 2026-02-22
Fix encoding functions returning Any instead of str/bytes

### LINKED ISSUE #2762: force_str / force_bytes don't force Any properly
# Bug report

```python
from typing import Any, assert_type
from django.utils.encoding import force_bytes, force_str

def f(x: Any) -> None:
    assert_type(force_bytes(x), bytes)
    assert_type(force_str(x), str)
```

## What's wrong

```console
$ mypy t2.py
t2.py:5: error: Expression is of type "Any", not "bytes"  [assert-type]
t2.py:6: error: Expression is of type "Any", not "str"  [assert-type]
Found 2 errors in 1 file (checked 1 source file)
```

## How is that should be

(no errors)

## System information

- OS:
- `python` version: 3.13.1
- `django` version: 5.2.something
- `mypy` version: 1.17.1

emmanuel-ferdman:
### PR Summary
This PR contains the following changes (tricky PR!):
- Fix `force_str`/`force_bytes`/`smart_str`/`smart_bytes` returning `Any` when given `Any` input -When you pass a variable typed as `Any` to these functions, mypy was inferring the return type as `Any` instead of `str`/`bytes`. This happened because the `TypeVar` overloads (added for `strings_only=True` support) were matching `Any` inputs and propagating the `Any` through. The fix puts an `object` overload first to catch `Any` and return the concrete type. `TypeVar` overloads now require `strings_only` as a keyword-only argument without a default, so they only match when you explicitly pass `strings_only=True`. This looks like the best approach for all what I tested - it matches the original design intent from PR #254 before stubdefaulter broke it by adding explicit defaults to `Literal[True]` overloads, fixes the `Any` bug without losing type preservation for `strings_only=True` cases, and passes all stubtest/mypy validation. But I'm open for suggestions.
- Fix `DjangoUnicodeDecodeError` stub - The stub incorrectly defined an obj attribute and custom `__init__` that don't exist at runtime. This class just inherits from `UnicodeDecodeError` without any modifications.
- Fix `uri_to_iri` overlapping overload - Changed `uri: Any` to `uri: str | bytes` to avoid overlap with the `uri: None` overload.

Fixes #2762


tests/typecheck/utils/test_encoding.yml
```diff
@@ -13,3 +13,31 @@
       reveal_type(force_str('foo'))  # N: Revealed type is "builtins.str"
       reveal_type(force_str('foo', strings_only=True))  # N: Revealed type is "builtins.str"
       reveal_type(force_str(S('foo'), strings_only=True))  # N: Revealed type is "main.S"
+
+-   case: force_bytes_or_str_with_any
```
UnknownPlatypus: Could you move this to the tests/assert_type folder so that it also tests pyright and pyrefly?

emmanuel-ferdman: moved :)

UnknownPlatypus: Thanks, i've moved the rest of the file too, it seem to work!

PR #3112, opened 2026-02-23
fix: add missing GeometryType stub in postgis base.pyi

ahmedasar00:
Add `GeometryType` class stub to `django-stubs/contrib/gis/db/backends/postgis/base.pyi` 
and remove it from `allowlist_todo_django60.txt`.

The class exists in the runtime (`base.py`) but was missing from the stub file, 
causing a stubtest error. This fix brings the stub in line with the source.

## Changes

- `django-stubs/contrib/gis/db/backends/postgis/base.pyi`: added `GeometryType` stub
- `scripts/stubtest/allowlist_todo_django60.txt`: removed related allowlist entry


sobolevn (commented):
Closing as a duplicate of https://github.com/typeddjango/django-stubs/pull/3113

PR #3113, opened 2026-02-23
fix: add missing GeographyType stub in postgis base.pyi

ahmedasar00:
Add `GeographyType` class stub to `django-stubs/contrib/gis/db/backends/postgis/base.pyi` 
and remove it from `allowlist_todo_django60.txt`.

The class exists in the runtime (`base.py`) but was missing from the stub file, 
causing a stubtest error. This fix brings the stub in line with the source.

## Changes
- `django-stubs/contrib/gis/db/backends/postgis/base.pyi`: added `GeographyType` stub
- `scripts/stubtest/allowlist_todo_django60.txt`: removed related allowlist entry


django-stubs/contrib/gis/db/backends/postgis/base.pyi
```diff
@@ -20,3 +20,6 @@ class DatabaseWrapper(Psycopg2DatabaseWrapper):
     def __init__(self, *args: Any, **kwargs: Any) -> None: ...
     def prepare_database(self) -> None: ...
     def register_geometry_adapters(self, pg_connection: BaseConnection[bytes], clear_caches: bool = False) -> None: ...
+
+class GeometryType: ...
+class GeographyType: ...
```
sobolevn, line 26: Let's also add `RasterType` and `postgis_adapters` while we are here. These are 2 only features left in the allowlist and they are in the same file.

DeXtAr47-oss: @sobolevn I had openned a pr for RasterType #3115, so should I close it then ?

sobolevn: yes, let's do it one PR to avoid merge conflicts. sorry!

ahmedasar00: Okay, I will add `RasterType` and `postgis_adapters`
in a single PR.

ahmedasar00:
Hi @sobolevn 

I updated the stubtest `allowlist_todo_django60.txt`  to include `PostGISBinaryDumper` and `PostGISTextDumper`. This fixes the runtime missing errors for these classes in `PostGIS` when running stubtests.

django-stubs/contrib/gis/db/backends/postgis/base.pyi
```diff
@@ -20,3 +21,18 @@ class DatabaseWrapper(Psycopg2DatabaseWrapper):
...
+class RasterType: ...
+class PostGISTextDumper(BaseTextDumper): ...
+
+class PostGISBinaryDumper(BaseBinaryDumper):
+    format: Format
+
+@lru_cache
```
UnknownPlatypus, line 33:
```suggestion
```
I'm not 100% sure but I think this is not really necessary in stubs as it does not influence the types ?

sobolevn: Technically it does influence the runtime type by providing `.clear_cache` method for `postgis_adapters` object.

So, it can stay :)

scripts/stubtest/allowlist_todo_django60.txt
```diff
@@ -1,2 +1,2 @@
-django.contrib.gis.db.backends.postgis.base.RasterType
-django.contrib.gis.db.backends.postgis.base.postgis_adapters
+django.contrib.gis.db.backends.postgis.base.PostGISBinaryDumper
```
sobolevn: Why is this ignored?

django-stubs/contrib/gis/db/backends/postgis/base.pyi
UnknownPlatypus: Oh you are right, it does add 3 method and mypy understand it!
https://mypy-play.net/?gist=22aba0a507aaf4fa05ed766589f7559f

```
from functools import lru_cache


@lru_cache
def foo():
    pass


foo.cache_clear()
foo.cache_info()
foo.cache_parameters()
```

scripts/stubtest/allowlist_todo_django60.txt
ahmedasar00: This is ignored because the error is raised by stubtest when running `./scripts/stubtest.sh`.
At runtime the ` PostGISBinaryDumper` and `PostGISTextDumper` classes are not present, so stubtest reports them as missing.

UnknownPlatypus: Can you move these 2 entries to the allowlist.txt file with a comment explaining they are dynamically generated classes and delete the now empty `scripts/stubtest/allowlist_todo_django60.txt` ?

ahmedasar00: I tried to move and adjust the entries as you suggested, but every time I did, the **CI** reported new errors.
Whenever I tried to ignore one error, another would appear. These errors are unrelated to the changes I made.

scripts/stubtest/allowlist.txt
```diff
@@ -273,6 +273,12 @@ django.test.selenium.SeleniumTestCase.__init_subclass__
 django.urls.ResolverMatch.__iter__
 django.urls.resolvers.ResolverMatch.__iter__
 
-# Loop leak variables - values depend on dict iteration order, not public API
-django.template.smartif.key
-django.template.smartif.op
+
```
UnknownPlatypus, line 286: This was removed by mistake I think and causes ci failure

UnknownPlatypus (changes requested):

scripts/stubtest/allowlist.txt
ahmedasar00: Thanks @UnknownPlatypus

scripts/stubtest/allowlist_todo_django60.txt
UnknownPlatypus, line 1: Can you remove this file now that you moved the ignore to the main allowlist?

Should be the last step for this mr to be ready

PR #3114, opened 2026-02-24
Improve template module stubs

emmanuel-ferdman:
## PR Summary
This PR fixes `template` module stubs to match runtime. It adds missing `FilterExpression.is_var`, corrects `json_script` and `filesizeformat` signatures, fixes `template_cache` to be the correct `get_template_cache` and adds `Loader.get_dirs`. It also removes redundant `origin`/`token` from `Node` subclasses (inherited), moves `key`/`op` loop leak variables to permanent allowlist and fixes order of a few stubs to meet the source.

Note about `smartif.pyi`: `EndToken` becomes an instance declaration since Django does [`EndToken = EndToken()` at module load](https://github.com/django/django/blob/9a8461f92250b49994aa133b564ec68617f6d78c/django/template/smartif.py#L153). Added `lbp: int` to `TokenBase` - it acutally doesn't exist on base at runtime, but all subclasses have it and `IfParser.expression()` accesses `.lbp` on `TokenBase`-typed tokens. This documents the real interface and prevents `isinstance(x, EndToken)` which fails at runtime.


sobolevn (approved):
Created https://github.com/django/django/pull/20760 to fix this.

Thanks a lot!

PR #3116, opened 2026-02-24
stubs for LogEntry  fields

DeXtAr47-oss:
Fix stubs for LogEntry fields in `contrib.admin.models.LogEntry`

 - `action_time` updated to `DateTimeField[datetime, datetime]`
 -  `user and content_type` updated to use explicit `ForeignKey` generics with related models.
 -  `object_id`, `object_repr` and `change_msg` updated to `[str, str]`.
 -  `action_flag` updated to `[int, int]`.  
 -  removed the above from `allowlist_todo.txt`
 
I don't think we need to add any stubs for the rest LogEntry fields as i couldn't find any reference to them in django.

### DIFF
--- a/django-stubs/contrib/admin/models.pyi
+++ b/django-stubs/contrib/admin/models.pyi
@@ -1,6 +1,6 @@
 from collections.abc import Iterable
 from datetime import datetime
-from typing import Any, Literal, overload
+from typing import Any, ClassVar, Literal, overload
 from uuid import UUID
 
 from django.contrib.auth.models import User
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
we had a long discussion about how we can properly annotate model fileds in stubs. I don't remember which conclusion was made there. And I am not sure we ended up with the conclusion at all :(

DeXtAr47-oss:
oh ok sorry I was unaware of the discussions, I just tried to convert the fields into generic. could you please guide me or suggest me any other issue that i can work on.

PR #3117, opened 2026-02-25
Add stubs for `AdminSeleniumTestCase.assertCountSeleniumElements`

DeXtAr47-oss:
Changes  I had made:
- added stubs for `assertCountSeleniumElements` in `tests.pyi`
- removed `django.contrib.admin.tests.AdminSeleniumTestCase.assertCountSeleniumElements` from the `allowlist_todo.txt`.

I hope that the changes works fine and any review from the maintainers are much appreciated.


django-stubs/contrib/admin/tests.pyi
```diff
@@ -20,6 +20,7 @@ class AdminSeleniumTestCase(SeleniumTestCase, StaticLiveServerTestCase):
     def admin_login(self, username: str, password: str, login_url: _StrOrPromise = ...) -> None: ...
     def select_option(self, selector: str, value: Any) -> None: ...
     def deselect_option(self, selector: str, value: Any) -> None: ...
+    def assertCountSeleniumElements(self, selector: str, count: int, root_element: Any = ...) -> None: ...
```
UnknownPlatypus: Given the usage, I think we can improve the type of root_element here to `root_element: WebElement | None = None` I think.

Can you have a look at the various usages of `assertCountSeleniumElements` to confirm ?
Also we prefer to keep the default values in the stub to be extra explicit instead of using `...`

UnknownPlatypus (changes requested):

django-stubs/contrib/admin/tests.pyi
DeXtAr47-oss: yes we can improve the type check https://github.com/django/django/blob/main/tests/admin_views/test_autocomplete_view.py#L452 here the `result_container` type must be `WebElement` which then passed to `root_element` in line 460

I would do make the changes soon

DeXtAr47-oss:
I think the tests are failing because of missing `selenium` packages. So do we need to update the dependencies then ?

UnknownPlatypus:
We probably need `types-selenium` for it to work properly but I don't know if it's worth adding given it's stubs for internal django tests. What do you think @sobolevn ?

I know t[here are discussions](https://github.com/django/new-features/issues/13) around migrating from selenium to playwright in django which makes me wonder.

sobolevn:
I don't think that we need it :(
No real user ever asked for this.

DeXtAr47-oss:
so then can i keep the `root_element` stubs as `Any` then ? @UnknownPlatypus

UnknownPlatypus:
Yes sure

PR #3121, opened 2026-02-26
Allow expressions in ordering types

### LINKED ISSUE #3109: BaseModelAdmin.get_ordering return type too strict
# Bug report

## What's wrong

I have relatively simple `ModelAdmin` that includes an override for `get_ordering`:
```
    @override
    def get_ordering(self, request):
        return [Upper('first_name')]
```

While this is correct and Django is working with this I get an error on the typing:
```
Method "get_ordering" overrides class "BaseModelAdmin" in an incompatible manner
  Return type mismatch: base method returns type "_ListOrTuple[str]", override returns type "list[Upper]"
```

## How is that should be

I believe this would be fixed nicely if the stub for `BaseModelAdmin` was updated and have the return type of `get_ordering` updated to be `_ListOrTuple[str | Func]` where `Func` would be imported as `from django.db.models.expressions import Func`

## System information

- OS: MacOS
- `python` version: 3.13
- `django` version: 5.2.11
- `mypy` version: 1.19.1
- `django-stubs` version: 5.2.9
- `django-stubs-ext` version: 5.2.9

emmanuel-ferdman:
### PR Summary
Django allows ordering with expressions like `F('date').desc()` or `Upper('name')` in `ModelAdmin`, generic views, and model Meta - but the stubs currently only accepted strings. This PR updates all ordering-related types to use `_OrderByFieldName` (which is `str | Combinable`), which matches what Django actually accepts at runtime.

Fixes #3109


sobolevn (approved):

UnknownPlatypus:
Very nice 🙌

PR #3133, opened 2026-02-27
Fix lazy function return types

### LINKED ISSUE #2840: The function `format_lazy` should return `_StrPromise`

emmanuel-ferdman:
# PR Summary 
This PR continues the work from PR #2841. `format_lazy` was returning `_StrOrPromise` but the `lazy()` wrapper always returns a proxy object, never a plain `str`. Overloads wouldn't help here since the return type is always `_StrPromise` regardless of input - overloads only add value when different inputs produce different return types. So this PR changes it to return `_StrPromise` and converts it to a function definition to match other lazy functions like `gettext_lazy` and `reverse_lazy`. Also typed `csrf_input_lazy` and `csrf_token_lazy` which were `Any` - the former returns `SafeString` (HTML input element) and the latter returns `_StrPromise` (plain token string).

Fixes #2840.

### DIFF
--- a/django-stubs/template/backends/utils.pyi
+++ b/django-stubs/template/backends/utils.pyi
@@ -1,9 +1,10 @@
-from typing import Any
+from collections.abc import Callable
 
 from django.http.request import HttpRequest
+from django.utils.functional import _StrPromise
 from django.utils.safestring import SafeString
 
 def csrf_input(request: HttpRequest) -> SafeString: ...
 
-csrf_input_lazy: Any
-csrf_token_lazy: Any
+csrf_input_lazy: Callable[[HttpRequest], SafeString]
+csrf_token_lazy: Callable[[HttpRequest], _StrPromise]
--- a/django-stubs/utils/text.pyi
+++ b/django-stubs/utils/text.pyi
@@ -1,11 +1,11 @@
-from collections.abc import AsyncIterable, AsyncIterator, Callable, Iterable, Iterator
+from collections.abc import AsyncIterable, AsyncIterator, Iterable, Iterator
 from html.parser import HTMLParser
 from io import BytesIO
 from re import Pattern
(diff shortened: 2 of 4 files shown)


UnknownPlatypus (approved):
Thanks!

emmanuel-ferdman:
@UnknownPlatypus Thanks for the review - We could also close the previous PR as part of the cleanup :)

UnknownPlatypus:
Oh i did not see the other pr, tho looking at it, they talk about the use case of passing a str and returning a str, is it normal you skipped this here ? Am i missing something?
https://github.com/typeddjango/django-stubs/pull/2841#pullrequestreview-3274175308

emmanuel-ferdman:
@UnknownPlatypus The REPL output was a bit misleading - it shows `'ab'` because `repr()` is called on the result. But check the actual type: 
```python
>>> type(format_lazy('a{0}', 'b')).__name__
'__proxy__'
>>> isinstance(format_lazy('a{0}', 'b'), Promise)
True
````
`format_lazy` uses `lazy()` not `keep_lazy()`, so it always returns a `Promise` regardless of input types. Overloads wouldn't help since the return type never varies. We could ask @sobolevn to look as well :)

UnknownPlatypus:
Thanks for the clarification, seems good to me !

PR #3134, opened 2026-02-27
Refine Site model fields and reorder SiteManager

ahmedasar00:
### Changes

- Removed outdated `Any` type hints and replaced them with concrete types:
  - `django.contrib.sites.models.Site.domain` → `str`
  - `django.contrib.sites.models.Site.name` → `str`
  - `django.contrib.sites.models.Site.id` → `int`
- Updated `SiteManager` methods typing for better accuracy.

### Notes

- Moved `SiteManager` definition **after** the `Site` class to avoid potential forward reference issues.

### DIFF
--- a/django-stubs/contrib/sites/models.pyi
+++ b/django-stubs/contrib/sites/models.pyi
@@ -3,7 +3,7 @@ from typing import Any, ClassVar
 from django.db import models
 from django.http.request import HttpRequest
 
-SITE_CACHE: Any
+SITE_CACHE: dict[int | str, Site]
 
 class SiteManager(models.Manager[Site]):
     def get_current(self, request: HttpRequest | None = ...) -> Site: ...
@@ -13,8 +13,10 @@ class SiteManager(models.Manager[Site]):
 class Site(models.Model):
     objects: ClassVar[SiteManager]
 
-    domain = models.CharField(max_length=100)
-    name = models.CharField(max_length=50)
+    id: int
+    domain: str
+    name: str
+
     def natural_key(self) -> tuple[str]: ...
 
 def clear_site_cache(sender: type[Site], **kwargs: Any) -> None: ...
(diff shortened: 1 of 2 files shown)


django-stubs/contrib/sites/models.pyi
```diff
@@ -13,8 +13,9 @@ class SiteManager(models.Manager[Site]):
 class Site(models.Model):
     objects: ClassVar[SiteManager]
 
-    domain = models.CharField(max_length=100)
-    name = models.CharField(max_length=50)
+    domain: Any
```
sobolevn: Sorry, why do we change a model field to be `Any`? What is the original error?

ahmedasar00: > Sorry, why do we change a model field to be `Any`? What is the original error?

I used `Any` while debugging a stubtest error, but I agree the concrete types are better.
Should I revert them to `str / int` ?

UnknownPlatypus: I dont think we should change anything here, I'm pretty sure the current state is correctly infered by the mypy plugin

ahmedasar00: In my opinion, I find it useful because it prevents `forward references` and using `str` and `int` helps ensure greater accuracy, although I could be wrong.

UnknownPlatypus: Still not convinced it's a good idea, I'm almost sure we loose some strict type checking with mypy when setting attribute for ex.
I don't think this is the way to go, we need to research existing issues and pr on the subject, weight the various options we have and how well they work with typechecker with no plugin (which should be goal, but not at the expense of the mypy setup)

ahmedasar00: Thanks for the feedback.
 I reopened this PR because I want to find the right solution rather than just close it without understanding the correct approach.
Could you point me to any existing issues or PRs that discussed how the mypy plugin handles Site model field types? That would help me understand the right direction before making further changes.
If you think there's no clear path forward right now, I'm also happy to close this and revisit it later with a better-informed approach.

PR #3143, opened 2026-02-28
Fix testcases module stubs

emmanuel-ferdman:
### PR Summary
This PR fixes a few stubtest errors in `django.test.testcases`. It added the missing `msg_prefix` and `count` params to `_AssertTemplateUsedContext.__init__`, added the `server_class` attribute to `LiveServerThread`, and fixed `connections_override` to accept `None`. Also it cleans some stale stubs - removed the phantom `multi_db` attribute that was dropped from Django ages ago, and replaced bare `Callable` with `Callable[..., Any]` to match the repo pattern.


sobolevn (approved):

django-stubs/test/testcases.pyi
```diff
@@ -36,7 +36,7 @@ class _AssertTemplateUsedContext:
     template_name: str
     rendered_templates: list[Template]
     context: ContextList
-    def __init__(self, test_case: Any, template_name: Any) -> None: ...
+    def __init__(self, test_case: Any, template_name: Any, msg_prefix: str = ..., count: int | None = ...) -> None: ...
```
UnknownPlatypus: I think we prefer to keep the defaults.
See:
- https://github.com/typeddjango/django-stubs/pull/2327
- https://github.com/typeddjango/django-stubs/pull/2326

emmanuel-ferdman: @UnknownPlatypus I also prefer to keep the defaults since they add useful context, but right now the codebase is inconsistent. Hopefully we can standardize this as we move forward. All methods in this file have been updated. Thanks! 🙂

sobolevn: Defaults can be added at any moment in a single big PR. It is fine.

emmanuel-ferdman: @sobolevn Let me work on it then :)

PR #3144, opened 2026-02-28
Django 6.0: Add DB and GIS backends module stubs

emmanuel-ferdman:
### PR Summary
This PR fixes stubtest errors across DB and GIS backend modules. It adds the missing `Expressions` class to `ddl_references` and the `debug_transaction`, `split_tzname_delta`, and `names_digest` functions to `utils`. Fixes `@property` vs `@cached_property` mismatches for `wrap_database_errors`, `function_names`, and `spatial_version`. Corrected `__exit__` param names to match runtime and removes phantom attributes like `operators` from `BaseDatabaseWrapper` (only exists on subclasses). For PostGIS, this PR adds the psycopg3 adapter types (`GeometryType`, `GeographyType`, `RasterType`, `postgis_adapters`) and the `convert_extent` overrides. Similar additions for SpatiaLite.

Ref #2944.


django-stubs/contrib/gis/db/backends/spatialite/operations.pyi
```diff
@@ -10,6 +10,7 @@ class SpatialiteNullCheckOperator(SpatialOperator): ...
 class SpatiaLiteOperations(BaseSpatialOperations, DatabaseOperations):
     name: str
     spatialite: bool
+    from_text: str  # type: ignore[assignment]
```
sobolevn, line 13: question: why is this needed?

emmanuel-ferdman: Well, `from_text` was missing from the stub (it was in the [allowlist](https://github.com/typeddjango/django-stubs/blob/e228e181fdb7d96d816ce8294728fb9da35ab5a8/scripts/stubtest/allowlist_todo.txt#L88)). The base class types it as [`bool`](https://github.com/typeddjango/django-stubs/blob/e228e181fdb7d96d816ce8294728fb9da35ab5a8/django-stubs/contrib/gis/db/backends/base/operations.pyi#L21), but [SpatiaLite sets it to "GeomFromText"](https://github.com/django/django/blob/202eca7795710c82cd64a248650afc64f905a17e/django/contrib/gis/db/backends/spatialite/operations.py#L36) which is a string. The `# type: ignore[assignment]` is needed because mypy doesn't allow narrowing `bool` to `str`. We have similiar pattern in [oracle/operations.pyi](https://github.com/typeddjango/django-stubs/blob/e228e181fdb7d96d816ce8294728fb9da35ab5a8/django-stubs/contrib/gis/db/backends/oracle/operations.pyi#L29).

sobolevn: Ok, I thought it would be `override`

sobolevn (approved):

sobolevn:
@emmanuel-ferdman can you please resolve a conflict?

sobolevn (approved):

PR #3151, opened 2026-03-02
Fix edge case with `Prefetch` + `to_attr` and multiple levels

UnknownPlatypus:
# I have made things!
Fixes an edge case where a `Prefetch` with `to_attr` that traverses multiple relations was annotating the attribute on the initial model instead of the intermediate one it actually ends up on at runtime.

```python
from django.db.models import Prefetch

user = (
    User.objects
    .prefetch_related(Prefetch("groups__user_set", User.objects.all(), to_attr="users"))
    .get()
)
user.users  # Revealed type was "list[User]" but is an `AttributeError` at runtime
user.groups.last().users # Mypy attribute error but it's actually present at runtime
```
In this example, `to_attr="users"` is stored on the `Group` instances you get when doing `user.groups.all()` (the intermediate model), not on `User` (the queryset's root model). Previously, the plugin incorrectly annotated it on User.

### DIFF
--- a/tests/typecheck/managers/querysets/test_basic_methods.yml
+++ b/tests/typecheck/managers/querysets/test_basic_methods.yml
@@ -343,3 +343,29 @@
 
                 class Book(models.Model):
                     author = models.ForeignKey(Person, on_delete=models.CASCADE)
+
+-   case: prefetch_related_traversing_multiple_relations
+    main: |
+        from typing_extensions import reveal_type
+        from django.db.models import Prefetch, F
+        from django.contrib.auth.models import User, Group
+
+        # When prefetching a relation that is traversing multiple relations
+        # The `to_attr` arg is set on the last item of the chain, not the initial one.
+        # We should NOT annotate the root model (User) with the to_attr.
+        user = (
+            User.objects
+            .prefetch_related(Prefetch("groups__user_set", User.objects.all(), to_attr="users"))
+            .get()
+        )
+        reveal_type(user) # N: Revealed type is "django.contrib.auth.models.User"
+        reveal_type(user.groups) # N: Revealed type is "django.contrib.auth.models.Group_ManyRelatedManager[django.contrib.auth.models.User_groups]"
+        last_group = user.groups.get()
+        reveal_type(last_group) # N: Revealed type is "django.contrib.auth.models.Group"
+
+        # Ideally we would like the type to be inferred as list[User] here
+        last_group.users # E: "Group" has no attribute "users"; maybe "user_set"?  [attr-defined]
+
+    installed_apps:
+        - django.contrib.auth
+        - django.contrib.contenttypes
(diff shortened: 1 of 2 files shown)


boldrack:
`

> ```python
> # The `to_attr` arg is set on the last item of the chain, not the initial one.
> ```

Please correct me if i'm wrong. when you make a lookup query like in your example `groups__user_set`,  isn't the result the last type in query .. in this case `user_set` which is list[User] . isn't that correct already ? I presume your PR is suggesting it ought to be list[Group] according to this statement you made 

> ```python
> not the initial one.
> ```

I'm  trying to get myself accustomed to be internals of the project to enable me able to contribute for the Django 6.0 as called for by the last release. so these inferences of mine above might be wrong. I'm well open to corrections.

boldrack:
so I just found out from this Django issue here [https://github.com/django/new-features/issues/23#issuecomment-3759955809](https://github.com/django/new-features/issues/23#issuecomment-3759955809) that @UnknownPlatypus is a maintainer of this project 😁 . No doubt, with that new knowledge, Looks like i'm the one mis-interpreting the PR. Nevertheless, I'd still like a help with clarity. Thank you.

UnknownPlatypus:
> Looks like i'm the one mis-interpreting the PR. Nevertheless, I'd still like a help with clarity. Thank you

I make my fair share of mistake too ahah, there is no silver bullet. My wording is not clear, I'll update the description

boldrack:
> > Looks like i'm the one mis-interpreting the PR. Nevertheless, I'd still like a help with clarity. Thank you
> 
> I make my fair share of mistake too ahah, there is no silver bullet. My wording is not clear, I'll update the description

Thanks for the inline comment. It made it clearer. Also, the purpose of the prefetch in this particular example almost threw me off but it makes sense now. users of each group user belongs in. 
Thank you.

UnknownPlatypus:
Yes it's a pattern I would strongly advise against, it's quite confusing

boldrack:
@UnknownPlatypus  Whilst you're here i should mention that the todo file for Django 6.0 [scripts/stubtest/allowlist_todo_django60.txt](https://github.com/typeddjango/django-stubs/blob/8f14025965a27bf7eaca7e81faedf8a24bb66bf7/scripts/stubtest/allowlist_todo_django60.txt) even though mentioned in the contributing guide, It doesn't exist.

UnknownPlatypus:
We just finished the work on this, I've updated the CONTRIBUTING.md file [just now](https://github.com/typeddjango/django-stubs/commit/9bc31f4f1d69a5ea09e4bb2051507ec572b721e2). You can find more info on the tracking issue

tests/typecheck/managers/querysets/test_basic_methods.yml
```diff
@@ -343,3 +343,29 @@
...
+            .prefetch_related(Prefetch("groups__user_set", User.objects.all(), to_attr="users"))
+            .get()
+        )
+        reveal_type(user) # N: Revealed type is "django.contrib.auth.models.User"
+        reveal_type(user.groups) # N: Revealed type is "django.contrib.auth.models.Group_ManyRelatedManager[django.contrib.auth.models.User_groups]"
+        last_group = user.groups.get()
+        reveal_type(last_group) # N: Revealed type is "django.contrib.auth.models.Group"
+
+        # Ideally we would like the type to be inferred as list[User] here
+        last_group.users # E: "Group" has no attribute "users"; maybe "user_set"?  [attr-defined]
```
sobolevn, line 367: Yes, this looks like a missing feature :)

sobolevn (approved):
Thanks! Good find!

PR #3155, opened 2026-03-03
Improve stubs for `django.utils.datastructures.ImmutableList`

ngnpope:
There was a comment about using `NoReturn` in `allowlist.txt`, so let's do just that. Suppress a couple of [`PYI034`](https://docs.astral.sh/ruff/rules/non-self-return-type/) that want the return type to be `Self` and also ignore a type error which wants the signatures of `__add__` and `__iadd__` to be compatible. Interestingly, [this is also ignored in typeshed for the built-in `list`](https://github.com/python/typeshed/blob/0e16ea31d2e188fdc126cb31e7c4fcc6b5a8da96/stdlib/builtins.pyi#L1161). Looks like it might be a [bug](https://github.com/python/mypy/issues/6225) in mypy.


django-stubs/utils/datastructures.pyi
```diff
@@ -93,8 +93,21 @@ class MultiValueDict(dict[_K, _V]):
...
 class ImmutableList(tuple[_V, ...]):
     warning: str
-    def __new__(cls, *args: Any, warning: str = "ImmutableList object is immutable.", **kwargs: Any) -> Self: ...
-    def complain(self, *args: Any, **kwargs: Any) -> NoReturn: ...
+    def __new__(cls, *args: object, warning: str = "ImmutableList object is immutable.", **kwargs: object) -> Self: ...
+    def complain(self, *args: object, **kwargs: object) -> NoReturn: ...
+    def __delitem__(self, *args: object, **kwargs: object) -> NoReturn: ...
+    def __delslice__(self, *args: object, **kwargs: object) -> NoReturn: ...
+    def __iadd__(self, *args: object, **kwargs: object) -> NoReturn: ...  # type: ignore[misc] # noqa: PYI034
+    def __imul__(self, *args: object, **kwargs: object) -> NoReturn: ...  # noqa: PYI034
```
UnknownPlatypus, line 101: What are the ruff / mypy error here with th ignores ?
And should we have `__add__` too ?

ngnpope: > What are the ruff / mypy error here with th ignores ?

See the description above in this pull request.
I'd already explained there, but have expanded a little.

> And should we have `__add__` too ?

No. Django doesn't override this because it cannot mutate.

UnknownPlatypus: Thanks for added detail in the description, makes sense to me!

PR #3150, opened 2026-03-02
Validate the `update_fields` and `unique_fields` of `bulk_create`

UnknownPlatypus:
# I have made things!

Followup to #2808 with a bit of refactoring because I plan to add similar validation to other qs methods later

PR #3158, opened 2026-03-03
Validate `earliest` / `latest` / `orderby`

UnknownPlatypus:
# I have made things!

Followup to #3150 adding more strict validation on allowed items in `order_by` and variants


mypy_django_plugin/transformers/querysets.py
emmanuel-ferdman, line 1: There is another one on line 665 that we can replace with `LOOKUP_SEP`

emmanuel-ferdman (approved):

PR #3164, opened 2026-03-06
Add missing field stub members and fix type annotations

emmanuel-ferdman:
### PR Summary
This PR fixes a bunch of stub issues across model fields, related descriptors, and utilities. A few notes (Sorry for big PR!):
- The `RelatedManager.clear/aclear` methods had a param named `clear` instead of `bulk`.
- `OneToOneField.__init__` accepted `Any` for `on_delete` instead of `Callable`.
- `ForeignObjectRel.target_field` was typed as `AutoField` when it returns `Field`.
- `get_extra_descriptor_filter` was missing `Q` from its return type.
- `get_extra_restriction` had a stale `where_class` parameter removed in Django.
- `Tuple.__iter__` now returns `Iterator[Expression]` instead of `Any`.
- `GeneratedField` gets proper generic params.
- `DatabaseErrorWrapper` gets its missing `__call__`.
-  `GenericForeignKey` removed from `_AnyField` type alias because it inherits from `Field` at runtime (MRO: `GenericForeignKey` -> `FieldCacheMixin` -> `Field`), so listing it separately in `Field[Any, Any] | ForeignObjectRel |  GenericForeignKey` was redundant. This also removes the stale `GenericForeignKey` from `Options.parents`, `add_field`, and `setup_pk` signatures. 
- Fixed` get_expected_types` in the mypy plugin to check for `GenericForeignKey` before the generic `Field` `isinstance` check. Since GFK is now a `Field` subclass, the old `elif` branch for GFK was dead code - GFK would enter the `Field` branch and get incorrect type inference. The GFK check now runs first with `continue` so it gets its proper `_pyi_private_set_type` with `is_nullable=True`.


django-stubs/contrib/gis/db/models/fields.pyi
```diff
@@ -180,8 +180,8 @@ class GeometryCollectionField(GeometryField[_ST, _GT]):
     geom_class: type[GeometryCollection]
     form_class: type[forms.GeometryCollectionField]
 
-class ExtentField(Field):
-    def get_internal_type(self) -> Any: ...
+class ExtentField(Field[Any, Any]):
```
sobolevn, line 183: How many errors will we have if we change the config to disallow missing type args?

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -191,59 +189,82 @@ class Field(RegisterLookupMixin, Generic[_ST, _GT]):
...
     def flatchoices(self) -> list[_Choice]: ...
-    def has_default(self) -> bool: ...
-    def has_db_default(self) -> bool: ...
-    def get_default(self) -> Any: ...
-    def check(self, **kwargs: Any) -> list[CheckMessage]: ...
-    def get_col(self, alias: str, output_field: Field | None = None) -> Col: ...
-    @cached_property
-    def cached_col(self) -> Col: ...
+    def save_form_data(self, instance: Model, data: Any) -> None: ...
+    # TODO: plugin support
```
sobolevn: maybe we should remove this? I have no idea what this comment is about :)

django-stubs/db/models/fields/related.pyi
```diff
@@ -188,7 +209,12 @@ class ForeignKey(ForeignObject[_ST, _GT]):
         error_messages: _ErrorMessagesMapping | None = ...,
         db_comment: str | None = ...,
     ) -> None: ...
+    @classmethod
+    def __class_getitem__(cls, *args: Any, **kwargs: Any) -> type[Self]: ...
```
sobolevn, line 213:
```suggestion
    def __class_getitem__(cls, *args: Any, **kwargs: Any) -> Any: ...
```

`Self` is too specific for no reason here :)

emmanuel-ferdman: I went with `type[Self]` here since [Django's `__class_getitem__`](https://github.com/django/django/blob/b33c31d992591bc8e8d20ac156809e4ae5b45375/django/db/models/fields/related.py#L1049) does `return cls` (returning the class itself rather than a `types.GenericAlias`), and it's the same convention used in [djangorestframework-stubs](https://github.com/typeddjango/djangorestframework-stubs/blob/44d99e3/rest_framework-stubs/fields.pyi#L107). That said, it's a meaningless distinction in practice since type checkers handle `Class[args]` syntax internally - happy to switch to `Any` if you prefer :)

django-stubs/db/models/fields/__init__.pyi
emmanuel-ferdman: Nice catch. I traced it back to a 2019 commit when `formfield` returned `Any` - it's been well-typed since then and no plugin support was ever added. Removed

django-stubs/contrib/gis/db/models/fields.pyi
emmanuel-ferdman: 651 errors across ~90 generic types - I tried to fix the ones in the modules I was already touching, but enabling it repo-wide would be a separate effort which I can work on :)

UnknownPlatypus: [649](https://github.com/typeddjango/django-stubs/issues/64#issuecomment-4016421203)

But also shouldn't we proxy type vars here instead of using `Any` ?

emmanuel-ferdman: @UnknownPlatypus I think `Field[Any, Any]` is the right fit here - `ExtentField` is an internal aggregate output field (only created by the `Extent` aggregate as its `output_field`), never placed on a model by users. It doesn't have `_pyi_private_set_type`/`_pyi_private_get_type` overrides, so there's nothing meaningful to parameterize. This also matches the pattern used by other other fields. Proxying `[_ST, _GT]` would imply users should parameterize it like `ExtentField[SomeType, SomeType]`, which I not sure if it makes sense for a field that only exists as a database aggregate result container. But I might missing some usecase?

UnknownPlatypus: Any is probably fine here yes

django-stubs/db/models/fields/related.pyi
UnknownPlatypus: type[Self] matches runtime but anyway, it probably does not matter too much ?

sobolevn (approved):
Thanks!

PR #3167, opened 2026-03-07
Replaced IO[bytes] with _PostDataProtocol to fix HttpRequest compatibility

### LINKED ISSUE #2717: Inconsistency between `HttpRequest` and `parse_file_upload` method introduced in 5.2.1
# Bug report

## What's wrong

#2654 removed `BytesIO` inheritance from `HttpRequest` which clashes with `parse_file_upload` signature.

```python
    def parse_file_upload(
        self, META: Mapping[str, Any], post_data: BinaryIO
    )
```

A `HttpRequest` instance used to be allowed as a `post_data` value. But not anymore:

```shell
error: Argument 2 to "parse_file_upload" of "HttpRequest" has incompatible type "HttpRequest"; expected "BinaryIO"  [arg-type]
```

## How is that should be

A `HttpRequest` instance should be a legitimate `post_data` value, as seen in `HttpRequest._load_and_post_files` method:

```python
            if hasattr(self, "_body"):
                # Use already read data
                data = BytesIO(self._body)
            else:
                data = self
            try:
                self._post, self._files = self.parse_file_upload(self.META, data)
            except (MultiPartParserError, TooManyFilesSent):
                ...
```

## System information

- OS: debian
- `python` 3.12
- `django` 5.2
- `mypy` 1.16.1
- `django-stubs` 5.2.1
- `django-stubs-ext` 5.2.1

Skyiesac:
# I have made things!

Closes #2717

PR #2654 removed BytesIO inheritance from HttpRequest, which broke passing an HttpRequest instance as post_data to parse_file_upload. 

Changes to fix it :
- Introduce _PostDataProtocol in request.pyi: a minimal structural Protocol with only read, readline, readlines, and __iter__
- Replace IO[bytes] / BinaryIO with _PostDataProtocol in:
  - HttpRequest.parse_file_upload
  - MultiPartParser.__init__
  - ChunkIter.flo attribute and __init__

Same issue was present in MultiPartParser. Both HttpRequest and BytesIO satisfy _PostDataProtocol, so both branches of _load_and_post_files type-check correctly without any runtime impact.

Alternatively , union of `BinaryIO | HttpRequest` would have worked , but it will fail on other  `BytesIO(self._body)` data for some edge cases,  that's why I have introduced a new Protocol and its more future proof .


UnknownPlatypus (approved):

django-stubs/http/request.pyi
```diff
@@ -15,6 +15,12 @@ from typing_extensions import Self
 RAISE_ERROR: object
 host_validation_re: Pattern[str]
 
+class _PostDataProtocol(Protocol):
+    def read(self, num_bytes: int | None = ...) -> bytes: ...
+    def readline(self, limit: int | None = ...) -> bytes: ...
```
emmanuel-ferdman: `HttpRequest.read` and `HttpRequest.readline` use positional-only params in the stub, so I believe the protocol needs them too:
```suggestion
    def read(self, num_bytes: int | None = ..., /) -> bytes: ...
    def readline(self, limit: int | None = ..., /) -> bytes: ...
```

Skyiesac: Thankyou for noticing this . I have fixed it now .

emmanuel-ferdman (approved):
Thank you!

PR #3170, opened 2026-03-09
Fix stubs for `django.db.models.expressions.Window`

ngnpope:
The fixes a number of issues:

- `Window.order_by` should be `OrderByList | None` as that is the type explicitly set in `Window.__init__()`. It's true that `OrderByList` is a subclass of `ExpressionList`, but better to type explicitly.
- The `partition_by` argument to `Window.__init__()` didn't accept a `list[str]` which should be supported - they're converted to `F()` inside `BaseExpression._parse_expressions()` which is called by `Func.__init__()`.
- The `partition_by` argument to `Window.__init__()` also accepted `Iterable` which isn't valid as there is a check on `partition_by` matching `list | tuple` otherwise being inserted into a 1-tuple. While we possibly ought to `_ListOrTuple` this, I wonder whether we should just keep this simple and just use `Sequence` which makes `partition_by` and `order_by` identical. As they share the same base `ExpressionList` type, this seems reasonable. (Although interestingly, `order_by` could have used `Iterable` instead of `Sequence` now that I look at its implementation.


django-stubs/db/models/expressions.pyi
```diff
@@ -326,11 +326,11 @@ class Window(SQLiteNumericMixin, Expression):
     contains_aggregate: bool
     contains_over_clause: bool
     partition_by: ExpressionList | None
-    order_by: ExpressionList | None
+    order_by: OrderByList | None
     def __init__(
         self,
         expression: BaseExpression,
-        partition_by: str | Iterable[BaseExpression | F] | F | BaseExpression | None = None,
+        partition_by: Sequence[BaseExpression | F | str] | BaseExpression | F | str | None = None,
```
sobolevn: Let's maybe add an alias? This is used at several places.

sobolevn (approved):

PR #3172, opened 2026-03-09
Enable `explicit-override` and `--disallow-subclassing-any`

ngnpope:
For a long time, there have been a number of issues that I now realise all stem from the same root issue. Recently I encountered it again when attempting to enable `explicit-override`.

Other related issues include:

- Not being able to enable `--disallow-subclassing-any` because of errors in the stdlib.
- Needing to ignore errors for some dependencies, e.g. `psycopg`, `redis`, `pyyaml`.

I did some digging around and found this issue:

- https://github.com/typeddjango/pytest-mypy-plugins/issues/134

While the `--mypy-only-local-stub` flag helped reduce the number of errors somewhat by ensuring that mypy wasn't passed the `--no-silence-site-packages` flags, it was clear that this broke those tests that relied on identifying type errors that are not in the main test file.

On further inspection, I realised that the main test file is passed to mypy as an absolute path already and the temporary test path is added to `MYPY_PATH`, but the alteration of `PYTHONPATH` was breaking what mypy perceived as a first-party path. Removing this made the error checking in additional files work correctly again.

(Note that it was important to ensure that `/tmp/.mypy_cache*` was removed between each run when I was testing this as the path changes seem to still be affected by stale cache otherwise.)

For the changes to `pytest-mypy-plugins`, see https://github.com/typeddjango/pytest-mypy-plugins/pull/183. This will need some sort of tidying up - do we still provide an inverted version of the `--mypy-only-local-stub` flag to allow re-enabling the `--no-silence-site-packages` flag if people should need this? Do we want to keep the ability to set `PYTHONPATH` behind a flag in case someone still needs that?


mypy.ini
```diff
@@ -7,8 +7,6 @@ incremental = true
 # Strictness:
 # TODO: add type args to all generics
 disallow_any_generics = false
-# TODO: fix `Any` subclassing in `typeshed/builtins.pyi`
-disallow_subclassing_any = false
```
ngnpope, line 11: Finally solved! 🎉

tests/typecheck/test_import_all.yml
```diff
@@ -689,11 +689,3 @@
...
         import django.views.i18n
         import django.views.static
-    mypy_config: |
-        [mypy-psycopg.*]
-        ignore_errors = true
-        [mypy-redis.*]
-        ignore_errors = true
-        [mypy-yaml.*]
-        disallow_untyped_defs = false
-        disallow_incomplete_defs = false
```
ngnpope, line 699: Also nice to not need this slowly expanding list of exceptions... 🥳

ngnpope:
Don't worry about the failures right now... That's just differences in Python versions.

We should either:
- Add some version-specific branches for these few cases
- Drop support for Python < 3.12 for 6.0 [^1]

[^1]: For 5.2, django-stubs dropped Python < 3.10, so there's precedent here.

ngnpope:
Ping @intgr: Thought you'd be interested in this as you raised the issue mentioned above.

sobolevn (commented):
I would prefer to adjust the test cases, supporting <python3.12 is critical for 5.2 partial support.

Other than that - this looks impressive! Did you use any LLMs / tools to fix this? I would be intersted to learn the workflow.

ngnpope:
> I would prefer to adjust the test cases, supporting <python3.12 is critical for 5.2 partial support.

No problem. Will do that.

> Other than that - this looks impressive! Did you use any LLMs / tools to fix this? I would be intersted to learn the workflow.

I didn't use any LLM, no. Just a combination of plopping in a breakpoint and manually running the subprocess that the plugin runs. When I understood how it ran, I could see that the way paths were being set up was affecting mypy's understanding of what was first-party vs site-packages.

I'll whip up a pull request for https://github.com/ngnpope/pytest-mypy-plugins/commit/c0e56248ac6e63edd621eebdf5490e37c6b5d686 later which this depends on - it'll be a breaking change there, but will aim to keep some inverted flags to make it possible to re-enable the old behaviour.

intgr:
Great that our long-running infrastructure issues are getting solved. :)

> We should either:
>
> * Add some version-specific branches for these few cases
> * Drop support for Python < 3.12 for 6.0

Actually no need for that. If you get mypy issues in one Python version and not another, you can suppress it using: `# type: ignore[override,unused-ignore]`

ngnpope:
Good shout. I forgot that we could ignore unused ignores 😂

ngnpope:
Ok. Now that https://github.com/typeddjango/pytest-mypy-plugins/pull/183 is merged we just need a release of `pytest-mypy-plugins`.

sobolevn:
Done: https://github.com/typeddjango/pytest-mypy-plugins/releases/tag/4.0.0

sobolevn (approved):
🎉 

This is really impressive! 
I want to merge it as soon as possible not to create merge conflicts.

PR #3174, opened 2026-03-10
adding warnings for __isnull (#3033)

### LINKED ISSUE #3033: Feature request : warning when incorrectly using `__isnull` filters
# Bug report

## What's wrong

This is more of a feature request (for which I did not find a template).
I would like to have a warning or an error when incorrectly using `__isnull` filters on a field that cannot be null.

## How is that should be

For example, with the following model

```python
class MyModel(models.Model):
  name = models.CharField(null=False)
```
I would like to have warnings in the following cases :

```python
MyModel.objects.filter(name__isnull=False) # Always true, this is a noop that should be removed
MyModel.objects.filter(name__isnull=True) # Filters out the entire queryset, probably a mistake
```

## System information

- OS: Ubuntu 22.04.5 LTS
- `python` version: Python 3.11.14
- `django` version: 4.2.26
- `mypy` version: 1.16.1

DeXtAr47-oss:
This PR adds the feature to generate warnings while incorrectly using __isnull feature, 
 - Detect lookups ending with `__isnull`
 - Check whether the provided value is `Literal[True]`
 - Remove the `__isnull` suffix to obtain the field lookup path.
 - Resolve the lookup using `django_context.resolve_lookup_into_field`.
 - Inspect the resolved Django `Field.null` attribute.
 - If the field has `null=False`, emit a mypy error.

## Related issues
- Closes #3033

### DIFF
--- a/mypy_django_plugin/transformers/orm_lookups.py
+++ b/mypy_django_plugin/transformers/orm_lookups.py
@@ -33,7 +33,7 @@ def typecheck_queryset_filter(ctx: MethodContext, django_context: DjangoContext)
             except LookupsAreUnsupported:
                 pass
             else:
-                if field.null is False:
+                if getattr(field, "null", None) is False:
                     ctx.api.fail(
                         f'Field "{field.name}" does not allow NULL;'
                         f'using "__isnull=True" will always return an empty queryset.',
--- a/mypy_django_plugin/transformers/orm_lookups.py
+++ b/mypy_django_plugin/transformers/orm_lookups.py
@@ -33,7 +33,7 @@ def typecheck_queryset_filter(ctx: MethodContext, django_context: DjangoContext)
             except LookupsAreUnsupported:
                 pass
             else:
-                if getattr(field, "null", None) is False:
+                if field is not None and getattr(field, "null", None) is False:
                     ctx.api.fail(
                         f'Field "{field.name}" does not allow NULL;'
                         f'using "__isnull=True" will always return an empty queryset.',
--- a/mypy_django_plugin/transformers/orm_lookups.py
+++ b/mypy_django_plugin/transformers/orm_lookups.py
@@ -1,8 +1,8 @@
 from mypy.plugin import MethodContext
-from mypy.types import AnyType, Instance, ProperType, TypeOfAny, get_proper_type
+from mypy.types import AnyType, Instance, LiteralType, ProperType, TypeOfAny, get_proper_type
 from mypy.types import Type as MypyType
 
-from mypy_django_plugin.django.context import DjangoContext
+from mypy_django_plugin.django.context import DjangoContext, LookupsAreUnsupported
 from mypy_django_plugin.exceptions import UnregisteredModelError
 from mypy_django_plugin.lib import fullnames, helpers
 
@@ -11,6 +11,7 @@ def typecheck_queryset_filter(ctx: MethodContext, django_context: DjangoContext)
     django_model = helpers.get_model_info_from_qs_ctx(ctx, django_context)
     if django_model is None:
         return ctx.default_return_type
+    model_cls = django_model.cls
 
     # Expected formal arguments for filter methods are `*args` and `**kwargs`. We'll only typecheck
     # `**kwargs`, which means that `arg_names[1]` is what we're interested in.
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
Please, add tests

mypy_django_plugin/transformers/orm_lookups.py
```diff
@@ -21,6 +22,24 @@ def typecheck_queryset_filter(ctx: MethodContext, django_context: DjangoContext)
...
+
+        if lookup_kwarg.endswith("__isnull"):
+            is_true_literal = isinstance(provided_type, LiteralType)
+
+            if is_true_literal:
+                lookup = lookup_kwarg[:-8]
+
+                try:
+                    field, _ = django_context.resolve_lookup_into_field(model_cls, lookup)
+                except (LookupsAreUnsupported, Exception):
```
ngnpope: Why does this need `Exception`?

mypy_django_plugin/transformers/orm_lookups.py
```diff
@@ -21,6 +22,24 @@ def typecheck_queryset_filter(ctx: MethodContext, django_context: DjangoContext)
         if lookup_kwarg is None:
             continue
         provided_type = get_proper_type(provided_type)
+
+        if lookup_kwarg.endswith("__isnull"):
+            is_true_literal = isinstance(provided_type, LiteralType)
```
ngnpope: This checks that it's a literal type, but not that it's `True`.

mypy_django_plugin/transformers/orm_lookups.py
```diff
@@ -21,6 +22,24 @@ def typecheck_queryset_filter(ctx: MethodContext, django_context: DjangoContext)
...
+                    field, _ = django_context.resolve_lookup_into_field(model_cls, lookup)
+                except (LookupsAreUnsupported, Exception):
+                    field = None
+
+                if field is not None and not getattr(field, "null", False):
+                    ctx.api.fail(
+                        f'Filed "{field.name}" does not allow NULL;'
+                        f'using "__isnull=True" will always return an empty queryset.',
+                        ctx.context,
+                    )
```
ngnpope: Instead of the fragility of slicing 8 characters from the end of the string, this could be made more legible, e.g.

```python
        lookup_path, sep, lookup_name = lookup_kwarg.rpartition(LOOKUP_SEP)
        is_true_literal = isinstance(provided_type, LiteralType) and provided_type.value is True

        if lookup_name == "isnull" and is_true_literal:
            try:
                field, _ = django_context.resolve_lookup_into_field(model_cls, lookup_path)
            except LookupsAreUnsupported:
                pass
            else:
                if field is not None and field.null is False:
                    ctx.api.fail(
                        f'Field "{field.name}" does not allow NULL;'
                        f'using "__isnull=True" will always return an empty queryset.',
                        ctx.context,
                    )

```

`LOOKUP_SEP` can be imported via `from django.db.models.constants import LOOKUP_SEP`.

ngnpope (commented):
A few quick drive-by comments...

mypy_django_plugin/transformers/orm_lookups.py
DeXtAr47-oss: to skip for invalid fields or unsupported lookup chains instead of raising an error

mypy_django_plugin/transformers/orm_lookups.py
DeXtAr47-oss: thanks for pointing that out I would make the changes by checking the `provided_type.value is True`

mypy_django_plugin/transformers/orm_lookups.py
DeXtAr47-oss: ok sure I will do that

mypy_django_plugin/transformers/orm_lookups.py
UnknownPlatypus: We should use more narrow exception types then, this catches too many things

UnknownPlatypus:
@DeXtAr47-oss Thanks for the work on this! 
There are a lot of ci failure, could you assess them to see if they are true or false positives ?

UnknownPlatypus:
This will need a small rebase too, still on my radar once the ci is green :eyes:

PR #3175, opened 2026-03-10
Remove stale stub attributes that no longer exist at runtime

emmanuel-ferdman:
### PR Summary
This PR removes stub attributes/classes/methods that no longer exist at runtime in Django - things like renamed functions (`_get_new_csrf_token` -> `_get_new_csrf_string`), removed class attributes (`Field.widget`, `SecurityMiddleware.xss_filter`, `URLValidator.ul`), and phantom classes that were never importable. Also this PR adds missing `@type_check_only` to `_QueryDictMixin` and `_Converter` since they're stub-only constructs. 

The `PostGISTextDumper`/`PostGISBinaryDumper` classes were removed because they are not at [module level ](https://github.com/django/django/blob/3483bfc0920b0ef0b28563aabe8ff546699b6ece/django/contrib/gis/db/backends/postgis/base.py#L85), making them unimportable at runtime. The `postgis_adapters` return type now uses `tuple[type[Dumper], type[Dumper]]`, which matches the actual MRO (`PostGISTextDumper` -> `BaseDumper` → `Dumper`) and is directly compatible with `register_dumper`'s `type[Dumper]` parameter. This follows the same pattern being used for other factory functions like `modelform_factory` -> `type[ModelForm]` and `create_many_to_many_intermediary_model` -> `type[Model]`.


sobolevn (approved):
Awesome! How did you find these items? :)

emmanuel-ferdman:
@sobolevn I wrote a quick script that basically checks every attribute and method in the stub files against the actual Django 6.0+ runtime (each version in different venv), flagging anything that doesn't exist. Then I manually verified each flag against the Django source and git history to confirm they were truly removed or renamed :)

sobolevn:
Can you please share the script for the history? This is surely not the last time we remove older things from stubs.

Btw, you can merge now :)

emmanuel-ferdman:
@sobolevn its pretty messy and ugly now., but I'll work on refactoring it and share it :)

UnknownPlatypus:
Wondering if these could be found by grepping stubtest output also

PR #3178, opened 2026-03-12
Fix order_by false positives with field transforms

federicobond:
The new `order_by` validation introduced in https://github.com/typeddjango/django-stubs/pull/3158 generated some false-positives in my project due to the use of field transforms.

I'm hesitant to add a new dependency on the Django runtime by querying the field lookups, but let me know if it makes sense in this case.

*Full disclosure:* this PR was partly written with AI, though read carefully after that.

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -12,6 +12,7 @@ from django.db.models.fields.related_descriptors import (
     ReverseOneToOneDescriptor,
 )
 from django.db.models.fields.reverse_related import ForeignObjectRel
+from django.db.models.lookups import Transform
 from django.db.models.sql.query import Query
 from mypy.checker import TypeChecker
 from mypy.errorcodes import NO_REDEF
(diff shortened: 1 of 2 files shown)


mypy_django_plugin/transformers/querysets.py
```diff
@@ -914,6 +915,16 @@ def _validate_order_by_lookup(ctx: MethodContext, model_cls: type[Model], parts:
     try:
         Query(model_cls).names_to_path(parts, model_cls._meta, fail_on_missing=True)
     except FieldError as exc:
+        # Check if the trailing part is a valid transform (e.g. __year, __month) on the field.
+        # Transforms are allowed in order_by, but lookups (e.g. __exact) are not.
+        if len(parts) >= 2:
+            try:
+                field = model_cls._meta.get_field(parts[0])
+                lookup_cls = field.get_lookups().get(parts[-1])
+                if lookup_cls is not None and issubclass(lookup_cls, Transform):
```
sobolevn: Do we need all three lines to catch `FieldDoesNotExist`?
Or does it only happen for `field = model_cls._meta.get_field(parts[0])`?

tests/typecheck/managers/querysets/test_order_by.yml
```diff
@@ -134,3 +134,21 @@
...
+-   case: order_by_with_field_transforms
+    installed_apps:
+        - myapp
+    main: |
+        from myapp.models import Article
+
+        Article.objects.order_by("created__year", "created__month")
+        Article.objects.order_by("-created__year")
+        Article.objects.values("created__month", "created__year").order_by("created__year")
+    files:
```
UnknownPlatypus, line 148: We can probably add some test cases from #3184 once @ngnpope find some time for other issues with transformers ('endswith', ...)

federicobond:
Reworked the commit to make better use of the `Query.names_to_path` return value and expanted the test.

federicobond:
I could not reproduce the crash described in https://github.com/typeddjango/django-stubs/issues/3184 because mypy catches the abstract model `objects` manager as an `attr-defined` error before the order_by validation ever runs in my case.

sobolevn (commented):
LGTM, let's wait for others :)

UnknownPlatypus (approved):
Thanks, seems reasonable to me.

ngnpope:
> I could not reproduce the crash described in #3184 because mypy catches the abstract model `objects` manager as an `attr-defined` error before the order_by validation ever runs in my case.

Yeah. I'll have to have a deeper dive when I have time. There's some fairly complex stuff going on where things are passed through multiple layers which makes it a pain to diagnose 😑 But it's definitely an abstract model, or at least mypy sees it as such sometimes. And we don't seem to be using `strict_model_abstract_attrs = false`, so not sure why we don't see the `attr-defined` unless that's suppressed at a deeper level somewhere I haven't looked.

This looks good though!

ngnpope (approved):

PR #3182, opened 2026-03-12
Remove references to stale stubtest allowlists

ngnpope:
Removes stale references to stubtest allowlists that no longer exist.


sobolevn (commented):
Please, also fix https://github.com/typeddjango/django-stubs/blob/master/CONTRIBUTING.md#testing-stubs-with-stubtest

sobolevn (approved):

Issue #3184, opened 2026-03-12
Crash when validating `QuerySet.order_by()`


ngnpope:
# Bug report

I'm testing out the master branch on a very large repo at work to check that everything is working before we make a release as there are a significant number of changes.

## What's wrong

It looks like we have a crash in the new validation for `Queryset.order_by()`.

I've not managed to create a minimal reproducer yet, but here's the traceback:

```pytb
Please report a bug at https://github.com/python/mypy/issues
version: 1.19.1
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "mypy/checkexpr.py", line 6103, in accept
  File "mypy/checkexpr.py", line 6138, in accept_maybe_cache
  File "mypy/nodes.py", line 2289, in accept
  File "mypy/checkexpr.py", line 508, in visit_call_expr
  File "mypy/checkexpr.py", line 645, in visit_call_expr_inner
  File "mypy/checkexpr.py", line 1496, in check_call_expr_with_callee_type
  File "mypy/checkexpr.py", line 1589, in check_call
  File "mypy/checkexpr.py", line 1832, in check_callable_call
  File "mypy/checkexpr.py", line 1287, in apply_function_plugin
  File ".../.venv/lib/python3.13/site-packages/mypy_django_plugin/transformers/querysets.py", line 930, in validate_order_by
    _validate_order_by_lookup(ctx, django_model.cls, parts)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.13/site-packages/mypy_django_plugin/transformers/querysets.py", line 915, in _validate_order_by_lookup
    Query(model_cls).names_to_path(parts, model_cls._meta, fail_on_missing=True)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.13/site-packages/django/db/models/sql/query.py", line 1789, in names_to_path
    name = opts.pk.name
           ^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'name'
```

## System information

- OS:
- `python` version: 3.13
- `django` version: 6.0.3
- `mypy` version: 1.19.1
[...]

PR #3185, opened 2026-03-13
Fix order_by false positive with values() expression aliases

federicobond:
Skip order_by validation for fields that were introduced as keyword arguments in values()/values_list() calls (e.g. `values(name=F(...))`).

Should take care of https://github.com/typeddjango/django-stubs/issues/3184#issuecomment-4050402587


ngnpope (approved):
LGTM

UnknownPlatypus:
Thanks!

PR #3186, opened 2026-03-13
Fix stubs for `allowed_default` in expression classes

ahmedasar00:
## Summary

In Django's source, `BaseExpression.allowed_default` is a plain class
variable (`= False`), but several subclasses override it as a
`cached_property` that dynamically computes whether the expression can be
used as a database-level column default, based on their sub-expressions:

- `CombinedExpression`: `True` if both `lhs` and `rhs` allow it
- `Func`: `True` if all source expressions allow it
- `ExpressionWrapper`: delegates to the wrapped expression
- `When`: `True` if both `condition` and `result` allow it
- `Case`: `True` if `default` and all `cases` allow it

## Changes

- `BaseExpression.allowed_default`: changed from `bool` class variable to
  a read-only `@property`
- Added `@property` + `@override` for `CombinedExpression`, `Func`,
  `ExpressionWrapper`, `When`, and `Case`
- Removed 13 entries from `allowlist_todo.txt` (9 for `django.db.models.*`
  and 4 for `django.contrib.gis.db.models.*`)

### DIFF
--- a/scripts/stubtest/allowlist_todo.txt
+++ b/scripts/stubtest/allowlist_todo.txt
@@ -77,12 +77,8 @@ django.contrib.flatpages.models.FlatPage.title
 django.contrib.flatpages.models.FlatPage.url
 django.contrib.gis.admin.site
 django.contrib.gis.db.backends.spatialite.base.DatabaseWrapper.ops
-django.contrib.gis.db.models.Case.allowed_default
 django.contrib.gis.db.models.CharField.description
-django.contrib.gis.db.models.ExpressionWrapper.allowed_default
 django.contrib.gis.db.models.Field.description
-django.contrib.gis.db.models.Func.allowed_default
-django.contrib.gis.db.models.When.allowed_default
 django.contrib.gis.db.models.functions.Area.as_sql
 django.contrib.gis.db.models.functions.GeoFuncMixin.as_sql
 django.contrib.gis.db.models.functions.Length.as_sql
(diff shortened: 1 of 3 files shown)


scripts/stubtest/allowlist_todo.txt
```diff
@@ -151,7 +138,7 @@ django.db.models.expressions.rhs
 django.db.models.expressions.result
 django.db.models.expressions.lhs
 
-# These functions exist in Django 5.2 but are missing from stubs (need to be added):
+# Django 6.0.3 security fix new APIs:
```
ngnpope, line 141: This comment change is unrelated. But it looks like 24d3be085ec2854d3b57187a88e6fa3d0c5e5bf9 accidentally changed it, so we can let it stand to fix it back.

ngnpope (approved):
This seems reasonable in the absence of [PEP 767](https://peps.python.org/pep-0767/).

emmanuel-ferdman (commented):
I suggest adding `@cached_property` overrides (and `@property` for `ExpressionWrapper`) on the subclasses with `# type: ignore[override]`, while keeping `BaseExpression.allowed_default: bool` with `@property`. This matches Django's runtime exactly - the base class has a plain bool, subclasses override it with properties. The ignore comments are needed because type checkers don't like property overriding a class variable, but that's what Django actually does. This follows the same pattern already used for `Lookup.allowed_default`.

scripts/stubtest/allowlist_todo.txt
ngnpope: Sorry, I was trying to say this change back to what it was is correct. The previous change was wrong. You didn't need to add the wrong change back in addition.

ahmedasar00:
> I suggest adding `@cached_property` overrides (and `@property` for `ExpressionWrapper`) on the subclasses with `# type: ignore[override]`, while keeping `BaseExpression.allowed_default: bool` with `@property`. This matches Django's runtime exactly - the base class has a plain bool, subclasses override it with properties. The ignore comments are needed because type checkers don't like property overriding a class variable, but that's what Django actually does. This follows the same pattern already used for `Lookup.allowed_default`.


I implemented your suggestion, but when I ran the CI, mypy reported: 

on all 6 lines where I added `# type: ignore[override]`.

It seems that with `@cached_property` + `@override` decorators, mypy doesn't actually 
complain about the override, so the ignore comments aren't needed.

I've updated the PR to remove the unnecessary `# type: ignore[override]` comments. 
The current implementation is:

- `BaseExpression`: `@cached_property def allowed_default(self) -> bool: ...`
- `Case`, `Func`, `When`: `@cached_property @override def allowed_default(self) -> bool: ...`
- `CombinedExpression`, `ExpressionWrapper`: `@property @override def allowed_default(self) -> bool: ...`

This follows the `Lookup.allowed_default` pattern in `lookups.pyi` which also uses 
`@cached_property @override` without `# type: ignore`.

Let me know if you'd like any changes!

emmanuel-ferdman:
@ahmedasar00 Please see my changes for what I meant - `BaseExpression.allowed_default` is a plain `bool` at runtime, so I kept it as `allowed_default: bool` and add `# type: ignore[override]` on subclasses that override it with properties. That way we don't have additional pyright warnings :)

PR #3193, opened 2026-03-16
Fix stubtest: annotate leaked loop variables in django.db.models.expressions

ahmedasar00:
## Problem
stubtest reported 6 unresolved variables in `django.db.models.expressions`:
`connector`, `d`, `field_types`, `lhs`, `rhs`, `result`.

These are leaked loop variables from the module-level loop in Django's
`expressions.py` that registers combinable field types:

    for d in _connector_combinations:
        for connector, field_types in d.items():
            for lhs, rhs, result in field_types:

## Fix
Add proper type annotations for these variables in the stub file instead
of silencing them in `allowlist_todo.txt`.

### DIFF
--- a/django-stubs/db/models/expressions.pyi
+++ b/django-stubs/db/models/expressions.pyi
@@ -127,6 +127,16 @@ def register_combinable_fields(
     result: type[Field],
 ) -> None: ...
 
+# These are leaked loop variables from the module-level loop that registers
+# combinable field types (`for d in _connector_combinations`). They are not
+# part of the public API but are visible as module attributes at runtime.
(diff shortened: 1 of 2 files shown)


emmanuel-ferdman:
@ahmedasar00 Actually @sobolevn and I fixed this issue in the django source code - see https://github.com/django/django/pull/20782. So I don't think we should stub them, those will be removed from allowlist in next Django release. So I'd keep them where they are at the moment.

ahmedasar00:
> @ahmedasar00 Actually @sobolevn and I fixed this issue in the django source code - see [django/django#20782](https://github.com/django/django/pull/20782). So I don't think we should stub them, those will be removed from allowlist in next Django release. So I'd keep them where they are at the moment.


I'll leave them as they are.

PR #3202, opened 2026-03-18
Run ty over test cases

ngnpope:
# I have made things!

While it's still relatively early days for [ty](https://github.com/astral-sh/ty), we've added [support for pyrefly](https://github.com/typeddjango/django-stubs/pull/2914) already and it's probably a good idea to see the state of different type checkers - particularly as we try to work out how we can make some things work without the plugin.

Currently there isn't any specific support for Django in ty, but [something is planned](https://github.com/astral-sh/ty/issues/1018). Having testing set up here may help that effort later.

Notes:
* I've configured ty to ignore `# type: ignore` comments and added `# ty: ignore` comments where needed, following what was done for pyright and pyrefly.
* I've ignored the `empty-body` rule which we also ignore for mypy.

## Related issues

Refs https://github.com/typeddjango/django-stubs/issues/579


sobolevn (commented):
It feels like a little bit too many errors for me just yet :(

What do others think?

ngnpope:
> It feels like a little bit too many errors for me just yet :(

There's a few more than the others, yes, but it's not _that_ much more...

```console
$ git grep --count -h 'ty: ignore' tests/assert_type/ | paste --delimiters=+ --serial - | bc
115

$ git grep --count -h 'pyright: ignore' tests/assert_type/ | paste --delimiters=+ --serial - | bc
87

$ git grep --count -h 'pyrefly: ignore' tests/assert_type/ | paste --delimiters=+ --serial - | bc
36

$ git grep --count -h 'type: ignore' tests/assert_type/ | paste --delimiters=+ --serial - | bc
28
```

emmanuel-ferdman (approved):
I think its good. We could also add `continue-on-error` if needed :)

ngnpope:
> I think its good. We could also add `continue-on-error` if needed :)

Yeah. I did consider that, but let's see how stable it is to begin with 🙂

UnknownPlatypus:
Im OK with adding it too, would be nice to track their progress on django-stubs

UnknownPlatypus (approved):

sobolevn (approved):
Ok then :)
Let's try it out.

PR #3203, opened 2026-03-18
Use newer syntax and enable more mypy configuration options

ngnpope:
# I have made things!

- Remove use of `Union[X, Y]` in favour of `X | Y`
  - We're targeting Python 3.10+, so the old syntax is no longer necessary
- Use `from __future__ import annotations` everywhere
  - Aligns closer to the behaviour for Python 3.14 onwards
  - Allows not needing to quote annotations in some cases
  - Allows moving non-runtime imports into type checking block
- Enable `strict_equality_for_none`
- Enable `allow_redefinition_new` (planned to be default after mypy v2.0)
- Enable `fixed_format_cache` (to become default in mypy v1.20)
- Remove obsolete `reportShadowedImports` from pyright configuration

### DIFF
--- a/ext/django_stubs_ext/aliases.py
+++ b/ext/django_stubs_ext/aliases.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import typing
 
 if typing.TYPE_CHECKING:
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -1,7 +1,8 @@
+from __future__ import annotations
+
 import os
 import sys
 from collections import defaultdict
-from collections.abc import Iterable, Iterator, Mapping, Sequence
 from contextlib import contextmanager
 from functools import cached_property
 from typing import TYPE_CHECKING, Any, Literal
@@ -10,15 +11,11 @@ from django.core.exceptions import FieldDoesNotExist, FieldError
 from django.db import models
 from django.db.models.base import Model
 from django.db.models.constants import LOOKUP_SEP
-from django.db.models.expressions import Expression
 from django.db.models.fields import AutoField, CharField, Field
 from django.db.models.fields.related import ForeignKey, RelatedField
 from django.db.models.fields.reverse_related import ForeignObjectRel
 from django.db.models.lookups import Exact
 from django.db.models.sql.query import Query
-from mypy.checker import TypeChecker
-from mypy.nodes import TypeInfo
-from mypy.plugin import MethodContext
 from mypy.typeanal import make_optional_type
 from mypy.types import AnyType, Instance, ProperType, TypeOfAny, UnionType, get_proper_type
 from mypy.types import Type as MypyType
@@ -36,9 +33,15 @@ except ImportError:
 
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable, Iterator, Mapping, Sequence
+
     from django.apps.registry import Apps
     from django.conf import LazySettings
+    from django.db.models.expressions import Expression
     from django.db.models.options import _AnyField
+    from mypy.checker import TypeChecker
+    from mypy.nodes import TypeInfo
+    from mypy.plugin import MethodContext
 
 
 @contextmanager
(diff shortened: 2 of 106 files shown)


mypy_django_plugin/django/context.py
```diff
@@ -420,7 +418,7 @@ def solve_lookup_type(
         self, model_cls: type[Model], lookup: str
     ) -> tuple[Sequence[str], Sequence[str], Expression | Literal[False]] | None:
         query = Query(model_cls)
-        if (lookup == "pk" or lookup.startswith("pk__")) and query.get_meta().pk is None:
+        if (lookup == "pk" or lookup.startswith("pk__")) and query.get_meta().pk is None:  # type: ignore[comparison-overlap]
             # Primary key lookup when no primary key field is found, model is presumably
             # abstract and we can't say anything about 'pk'.
             return None
```
ngnpope, line 427: This one is a little tricky because of the change in e3f5b3d60fbc9d4e5812d281e3cd60d739f01562. I decided to punt on removing this block for now - we can follow up later.

mypy_django_plugin/transformers/managers.py
ngnpope, line 388: Again, not entirely sure here, but likely this block can go too - the existing suppressed unreachable is an extra hint. Looking in [`mypy/nodes.py`](https://github.com/python/mypy/blob/master/mypy/nodes.py), the only `fullname` that is `str | None` is on `SymbolTableNode`.

mypy_django_plugin/transformers/models.py
```diff
@@ -317,7 +323,7 @@ class AddPrimaryKeyAlias(AddDefaultPrimaryKey):
     def run_with_model_cls(self, model_cls: type[Model]) -> None:
         # We also need to override existing `pk` definition from `stubs`:
         auto_field = model_cls._meta.pk
-        if auto_field is not None:
+        if auto_field is not None:  # type: ignore[comparison-overlap]
```
ngnpope, line 326: This is another case of the `Options.pk` one.

ext/django_stubs_ext/aliases.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
sobolevn, line 1: Since python 3.14 using `from __future__ import annotations` is not recommended. It is considered "soft deprecated". So, I don't think that we should do it.

Let's instead use explicit string annotations.
As the diff shows - there are not too many of them.

ngnpope: The point of this is that wrapping annotations in strings is obsolete in Python 3.14 and evaluation of annotations is deferred. By using this future import we opt into that behaviour earlier - well, similar behaviour that's closer than the original eager evaluation.

By doing so, it keeps the code more straightforward, removes overhead of evaluating annotations at runtime when not needed, allows moving imports into a type checking block of not needed at runtime, etc. 

It also means that, when 3.14 becomes the minimum sorted version, only the simple import line needs to be removed everywhere - the work of removing string annotations is already done.

So this future import might be soft-deprecated **from** 3.14, but think of it as a tool to help manage the transition until we don't need to support older versions.

sobolevn: Oh, I see. Good point! Let's do it then :)

mypy_django_plugin/django/context.py
```diff
@@ -365,15 +366,12 @@ def get_field_get_type(
...
         return helpers.get_private_descriptor_type(field_info, "_pyi_private_get_type", is_nullable=is_nullable)
 
-    def get_field_related_model_cls(self, field: Union["RelatedField[Any, Any]", ForeignObjectRel]) -> type[Model]:
+    def get_field_related_model_cls(self, field: RelatedField[Any, Any] | ForeignObjectRel) -> type[Model]:
         if isinstance(field, RelatedField):
             related_model_cls = field.remote_field.model
         else:
             related_model_cls = field.field.model
 
-        if related_model_cls is None:
```
emmanuel-ferdman, line 374: This case was added in PR #1956 to fix issue #1953 - shouldn't it be kept with `# type: ignore[comparison-overlap]` instead of removed?

ngnpope: Yeah. Probably. This likely indicates that the stubs are wrong somewhere.

ngnpope: Have reverted this and added the ignore.

sobolevn (approved):

PR #3204, opened 2026-03-18
Add missing `swappable` parameter to ForeignKey and OneToOneField

### LINKED ISSUE #298: Swappable keyword argument missing on ForeignKey, OneToOneField
# Bug report

## What's wrong

```python
    user = models.ForeignKey(
        myusermodel,
        on_delete=models.CASCADE,
        swappable=True,
        blank=False,
        null=False
    )
```

produces:

```
tiaki/audit_log/models.py:155: error: Unexpected keyword argument "swappable" for "ForeignKey"
/venv/lib/python3.6/site-packages/django-stubs/db/models/fields/related.pyi:95: note: "ForeignKey" defined here
```

## System information

- OS: docker: bionic
- `python` version: 3.6.9
- `django` version: 2.2.9
- `mypy` version: 0.761

emmanuel-ferdman:
### PR Summary
This PR adds the `swappable` keyword argument to `ForeignKey` and `OneToOneField` stubs. This param is defined in `ForeignObject` and passed through via `**kwargs` at runtime, but was missing from the child class stubs. Without it, using `swappable=False` (common for non-swappable model references) raised an unexpected keyword argument error.

Fixes #298


tests/typecheck/fields/test_related.yml
```diff
@@ -1657,3 +1657,13 @@
 
         p: Person
         reveal_type(p.friends)  # N: Revealed type is "django.db.models.fields.related_descriptors.ManyRelatedManager[main.Person, main.Person_friends]"
+
+-   case: test_foreign_key_swappable_param
```
UnknownPlatypus: I think this could be an `assert_type` test instead ? (it would be a little more lightweight)

emmanuel-ferdman: Good point, moved :)

sobolevn (commented):
Why does it pass stubtest check?

emmanuel-ferdman:
@sobolevn  Looking in the [stubtest sourcecode](https://github.com/python/mypy/blob/v1.19.1/mypy/stubtest.py#L1133-L1138), stubs can list explicit params that runtime catches via `**kwargs`, which is the case for `swappable` here

sobolevn (approved):

PR #3206, opened 2026-03-20
Fix `in` operator false positive with union types

emmanuel-ferdman:
### PR Summary
This PR fixes false positives when using the `in` operator with union types on container classes. Python falls back to `__iter__` for membership tests at runtime, but mypy requires an explicit `__contains__` to type-check `x in container` when `x` is a union like `Model | None`. Added a `_SupportsMembership` protocol in `query.pyi` with `__contains__(self, item: object) -> bool` and applied it to `QuerySet`, `RawQuerySet`, `BaseForm`, `BaseFormSet`, `CaptureQueriesContext`, and `BaseConnectionHandler`.

A few notes about my thoughts during implementation:
- Runtime: The `__contains__` does not exist in runtime, but Python's in operator falls back to `__iter__` automatically - the stub just makes mypy aware of this.
- Why protocol over alternatives: Adding `__contains__` directly to each class works but duplicates the full signature everywhere. A protocol lets us define it once and mix it in, keeping things DRY while documenting the fix.
- `item: object`: Standard PEP 484 signature for `__contains__`.

Example:
```python

class User(Model):
    class Meta:
        app_label = "myapp"

def check_user(user: User | None, qs: QuerySet[User]) -> bool:
    return user in qs
```
Error without fix:
> error: Unsupported operand types for in ("User | None" and "QuerySet[User, User]")  [operator]

Ref PR #1925 and issue #1924.

### DIFF
--- a/django-stubs/db/models/query.pyi
+++ b/django-stubs/db/models/query.pyi
@@ -3,7 +3,6 @@ import sys
 from collections.abc import AsyncIterator, Collection, Iterable, Iterator, Mapping, Sequence, Sized
 from typing import Any, Generic, Literal, NamedTuple, TypeAlias, overload, type_check_only
 
-from django.contrib.auth.models import AnonymousUser
 from django.db.backends.utils import _ExecuteQuery
 from django.db.models import Manager
 from django.db.models.base import Model
@@ -13,6 +12,7 @@ from django.utils.functional import cached_property
 from typing_extensions import Self, TypeVar, override
 
 _T = TypeVar("_T", covariant=True)
+_ContainsT = TypeVar("_ContainsT")
 _Model = TypeVar("_Model", bound=Model, covariant=True)
 _Row = TypeVar("_Row", covariant=True, default=_Model)  # ONLY use together with _Model
 _TupleT = TypeVar("_TupleT", bound=tuple[Any, ...], covariant=True)
@@ -27,8 +27,6 @@ _PrefetchedQuerySetT = TypeVar("_PrefetchedQuerySetT", bound=QuerySet[Model], co
 _ToAttrT = TypeVar("_ToAttrT", bound=str, covariant=True, default=str)
 
 _OrderByFieldName: TypeAlias = str | Combinable
-
-_ModelOrAnon: TypeAlias = Model | AnonymousUser
 
 MAX_GET_RESULTS: int
 REPR_OUTPUT_SIZE: int
@@ -61,10 +59,10 @@ class FlatValuesListIterable(BaseIterable[_T]):
     def __iter__(self) -> Iterator[_T]: ...
 
 @type_check_only
-class _QuerySetContainsMixin:
-    def __contains__(self, item: _ModelOrAnon, /) -> bool: ...
+class _SupportsContains(Generic[_ContainsT]):
+    def __contains__(self, item: _ContainsT, /) -> bool: ...
 
-class QuerySet(_QuerySetContainsMixin, Iterable[_Row], Sized, Generic[_Model, _Row]):
+class QuerySet(_SupportsContains[object], Iterable[_Row], Sized, Generic[_Model, _Row]):
     model: type[_Model]
     query: Query
     _iterable_class: type[BaseIterable]
@@ -241,7 +239,7 @@ class QuerySet(_QuerySetContainsMixin, Iterable[_Row], Sized, Generic[_Model, _R
     def _fetch_all(self) -> None: ...
     def resolve_expression(self, *args: Any, **kwargs: Any) -> Any: ...
 
-class RawQuerySet(_QuerySetContainsMixin, Iterable[_Model], Sized):
+class RawQuerySet(_SupportsContains[object], Iterable[_Model], Sized):
     raw_query: RawQuery | str
     model: type[_Model] | None
     query: RawQuery
--- a/django-stubs/db/models/query.pyi
+++ b/django-stubs/db/models/query.pyi
@@ -12,7 +12,7 @@ from django.utils.functional import cached_property
 from typing_extensions import Self, TypeVar, override
 
 _T = TypeVar("_T", covariant=True)
-_ContainsT_co = TypeVar("_ContainsT_co", covariant=True)
+_ContainsT = TypeVar("_ContainsT")
 _Model = TypeVar("_Model", bound=Model, covariant=True)
 _Row = TypeVar("_Row", covariant=True, default=_Model)  # ONLY use together with _Model
 _TupleT = TypeVar("_TupleT", bound=tuple[Any, ...], covariant=True)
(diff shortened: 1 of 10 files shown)


django-stubs/forms/forms.pyi
```diff
@@ -12,7 +12,12 @@ from typing_extensions import override
 
 class DeclarativeFieldsMetaclass(MediaDefiningClass): ...
 
-class BaseForm(RenderableFormMixin):
+# Explicit __contains__ for `in` operator with union types
+@type_check_only
+class _SupportsMembership(Protocol):
```
sobolevn: You can reuse the class from `query.pyi`, it is fine to import protected types here

tests/assert_type/db/models/test_query.py
```diff
@@ -31,3 +35,12 @@ async def test_async() -> None:
     # failure cases
     await aprefetch_related_objects(models_set, "pk")  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]  # pyrefly: ignore[bad-argument-type]
     await aprefetch_related_objects(models_frozenset, "pk")  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]  # pyrefly: ignore[bad-argument-type]
+
+
+# `in` operator with union types (tests _SupportsMembership protocol)
+qs = cast("QuerySet[Model]", None)
```
sobolevn:
```suggestion
def test_qs_in(qs: QuerySet[Model]) -> None: ...
```

it better to do it this way.

tests/assert_type/db/models/test_query.py
```diff
@@ -31,3 +34,8 @@ async def test_async() -> None:
     # failure cases
     await aprefetch_related_objects(models_set, "pk")  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]  # pyrefly: ignore[bad-argument-type]  # ty: ignore[invalid-argument-type]
     await aprefetch_related_objects(models_frozenset, "pk")  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]  # pyrefly: ignore[bad-argument-type]  # ty: ignore[invalid-argument-type]
+
+
+def test_in_operator(qs: QuerySet[Model], raw_qs: RawQuerySet[Model], obj: Model | None) -> None:
+    assert_type(obj in qs, bool)
```
sobolevn, line 47: what about `'a' in qs`? This should be an error, right?
Right now we allow `object` in `__contains__`, it might hide some bugs. On the other hand, some other type might be too strict.

What do you think?

django-stubs/forms/forms.pyi
```diff
@@ -13,7 +12,11 @@ from typing_extensions import override
 
 class DeclarativeFieldsMetaclass(MediaDefiningClass): ...
 
-class BaseForm(_SupportsMembership, RenderableFormMixin):
+@type_check_only
+class _FormContainsMixin:
```
sobolevn: We can just create one generic `_SupportsContains[T]`

tests/assert_type/forms/test_forms.py
```diff
@@ -0,0 +1,6 @@
+from django.forms import Form
+from typing_extensions import assert_type
+
+
+def test_in_operator(form: Form, field: str) -> None:
+    assert_type(field in form, bool)
```
sobolevn, line 7: Let's assert that negative cases raise type errors here

tests/assert_type/db/models/test_query.py
UnknownPlatypus: Changing from object seem to cause new issue in mypy_primer

django-stubs/db/models/query.pyi
```diff
@@ -12,6 +12,7 @@ from django.utils.functional import cached_property
 from typing_extensions import Self, TypeVar, override
 
 _T = TypeVar("_T", covariant=True)
+_ContainsT = TypeVar("_ContainsT")
```
sobolevn, line 15: Let's maybe try this?

```suggestion
_ContainsT_co = TypeVar("_ContainsT_co", covariant=True)
```

django-stubs/db/models/query.pyi
```diff
@@ -57,7 +58,11 @@ class NamedValuesListIterable(ValuesListIterable[NamedTuple]):
 class FlatValuesListIterable(BaseIterable[_T]):
     def __iter__(self) -> Iterator[_T]: ...
 
-class QuerySet(Iterable[_Row], Sized, Generic[_Model, _Row]):
+@type_check_only
+class _SupportsContains(Generic[_ContainsT]):
+    def __contains__(self, item: _ContainsT, /) -> bool: ...
+
+class QuerySet(_SupportsContains[object], Iterable[_Row], Sized, Generic[_Model, _Row]):
```
sobolevn, line 66: question: how will `_SupportsContains[_Row | None]` behave here?

emmanuel-ferdman: From what I see, with `_SupportsContains[_Row | None]`, `Model | None in qs` works and `"hello" in qs` correctly errors, but it breaks `Model | AnonymousUser in qs` and breaks Collection protocol compatibility - same tradeoff as `Model | AnonymousUser` just with different scope. Looks like `object` is the only option that satisfies both Collection protocol and all valid union patterns.

django-stubs/db/models/query.pyi
emmanuel-ferdman: Looks like covariant doesn't work here because `_ContainsT` is used in a parameter position, which is contravariant. mypy rejects it with "Cannot use a covariant type variable as a parameter [misc]". Typeshed's `Container` uses a contravariant TypeVar for the same reason.

django-stubs/db/models/query.pyi
sobolevn: Let's add a comment about that to the source code :)
We will forget about it very soon.

tests/assert_type/db/models/test_query.py
```diff
@@ -31,3 +35,17 @@ async def test_async() -> None:
     # failure cases
     await aprefetch_related_objects(models_set, "pk")  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]  # pyrefly: ignore[bad-argument-type]  # ty: ignore[invalid-argument-type]
     await aprefetch_related_objects(models_frozenset, "pk")  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]  # pyrefly: ignore[bad-argument-type]  # ty: ignore[invalid-argument-type]
+
+
+def test_in_operator(qs: QuerySet[Model], raw_qs: RawQuerySet[Model], obj: Model) -> None:
+    assert_type(obj in qs, bool)
+    assert_type(obj in raw_qs, bool)
```
sobolevn, line 48: let's test that different `in` checks are allowed, not just `Model`.

sobolevn (approved):

emmanuel-ferdman:
@UnknownPlatypus Could you please check the latest changes as well? :)

UnknownPlatypus:
> @UnknownPlatypus Could you please check the latest changes as well? :)

I'm trying to but github is having a stroke
image

PR #3208, opened 2026-03-20
Support WithAnnotations TypedDict fields in queryset validation

### LINKED ISSUE #3207: WithAnnotations and order_by
# Bug report

## What's wrong

When I annotate a query mypy complains when I try to order_by the annotated field:

```python
# app/models.py
from django.db import models


class Item(models.Model):

# repro.py
from typing import TYPE_CHECKING, TypedDict, reveal_type

from django.db.models import IntegerField, QuerySet, Value
from django_stubs_ext import WithAnnotations

from app.models import Item

class ExtraAnnotation(TypedDict):
    extra_id: int | None

if TYPE_CHECKING:
    ItemWithExtraAnnotation = WithAnnotations[Item, ExtraAnnotation]
else:
    ItemWithExtraAnnotation = Item

ItemWithExtraAnnotationQuerySet = QuerySet[Item, ItemWithExtraAnnotation]

def annotate_extra(qs: QuerySet[Item]) -> ItemWithExtraAnnotationQuerySet:
    return qs.annotate(extra_id=Value(None, output_field=IntegerField()))


qs = annotate_extra(Item.objects.all())

reveal_type(qs)
_ = qs.order_by("extra_id")
```
```bash
$ mypy
repro.py:31: note: Revealed type is "django.db.models.query.QuerySet[app.models.Item, app.models.Item@AnnotatedWith[TypedDict('repro.ExtraAnnotation', {'extra_id': builtins.int | None})]]"
repro.py:32: error: Cannot resolve keyword 'extra_id' into field. Choices are: id, name  [misc]
Found 1 error in 1 file (checked 4 source files)
```

## How is that should be

Mypy should not raise an error when I `order_by` with an annotated field.  

## System information

- OS:
- `python` version: 3.13.12 
- `django` version: 6.0.3
- `mypy` version: 1.19.1
- `django-stubs` version: 6.0.0
- `django-stubs-ext` version: 6.0.0

emmanuel-ferdman:
### PR Summary
When using `WithAnnotations[Model, TypedDict]` to declare annotated fields, `order_by()` was incorrectly reporting "Cannot resolve keyword" errors for fields defined in the TypedDict. The plugin was only checking `args[0].extra_attrs` (set by `.annotate()`) but `WithAnnotations` places its fields in `args[1].extra_attrs`. This PR also fixes `prefetch_related` conflict detection for `WithAnnotations` fields and preserves declared fields when chaining `.annotate()` on a `WithAnnotations` queryset.

Fixes #3207

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -250,12 +250,20 @@ def extract_proper_type_queryset_annotate(ctx: MethodContext, django_context: Dj
         if check_valid_attr_value(ctx, django_context, django_model, attr_name)
     }
 
+    existing_row_extra_attrs: dict[str, MypyType] = {}
+    if len(default_return_type.args) > 1:
+        original_row_type = get_proper_type(default_return_type.args[1])
+        if isinstance(original_row_type, Instance) and original_row_type.extra_attrs:
+            existing_row_extra_attrs = dict(original_row_type.extra_attrs.attrs)
+
+    all_fields = {**existing_row_extra_attrs, **expression_types}
+
     annotated_type: ProperType = django_model.typ
-    if expression_types:
+    if all_fields:
         fields_dict = helpers.make_typeddict(
             api,
-            fields=expression_types,
-            required_keys=set(expression_types.keys()),
+            fields=all_fields,
+            required_keys=set(all_fields.keys()),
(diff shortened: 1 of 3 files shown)


UnknownPlatypus (approved):
Awesome, thanks!

Issue #3148, opened 2026-03-02
stubtest: Inconsistencies in `TextChoices`, `IntegerChoices` and missing members in `django.utils.inspect`


ahmedasar00:
Description:

Running `./scripts/stubtest.sh` on the current version of `django-stubs` reveals several inconsistencies between the stubs and the runtime environment.
1. Choices Classes Inconsistency (`__init__` vs `__new__`)

Multiple errors indicate that `IntegerChoices` and `TextChoices` define `__init__` with parameters in the stubs, but at runtime, these parameters do not exist in `__init__`. The error message suggests moving the logic to `__new__`.

Affected modules:

   - `django.db.models.enums`

   - `django.contrib.gis.db.models`

   - `django.tasks.base`

Example Error:
```
error: django.db.models.enums.IntegerChoices.__init__ is inconsistent, runtime does not have parameter "x". 
You may need to write stubs for __new__ instead of __init__.
Stub: Overload(def (self: ..., x: ...), def (self: ..., x: ..., label: ...))
Runtime: def (self)
```

2. Missing members in `django.utils.inspect`

The runtime has `lock` and `safe_signature_from_callable` which are completely missing from the stubs.

Errors:
```
error: django.utils.inspect.lock is not present in stub
error: django.utils.inspect.safe_signature_from_callable is not present in stub
```

Environment:

    Django version: [Write your Django version here, e.g., 5.0]

    Python version: [Write your Python version here, e.g., 3.12]

    Command run: `./scripts/stubtest.sh`

Question:

Is this expected behavior for the current state of the stubs, or should these be addressed by moving the signatures to `__new__ `and adding the missing members to `django.utils.inspect`?

ngnpope:
Yeah, this is a [known issue](https://github.com/typeddjango/django-stubs/pull/2866) with Python 3.14 - it's just that nobody has had a chance to properly look into this yet. Assistance is welcome.

ngnpope:
> ...adding the missing members to `django.utils.inspect`?

This part was completed in #3152. I'll need some time to think about the enum issues.

ahmedasar00:
> This part was completed in [#3152](https://github.com/typeddjango/django-stubs/pull/3152)

Thank you.

PR #3215, opened 2026-03-23
Fix enum stubs: move __init__ to __new__ for IntegerChoices/TextChoices

ahmedasar00:
Refs #3148

## Summary

Partially addresses stubtest inconsistencies in `IntegerChoices`
and `TextChoices` caused by Python 3.14 enum changes.

## Changes

- Moved overloaded signatures from `__init__` to `__new__` in
  `enums.pyi`
- Renamed parameters to match runtime (`x` → `value`,
  `object` → `value`)
- Updated the comment above `IntegerChoices` to reference `__new__`

## Results

- **Before:** 14 stubtest errors
- **After:** 10 stubtest errors (only `label` parameter remains)

## Remaining issue

The runtime `__new__` signature is just `(cls, value)` — no
`label` parameter. The label is handled by the metaclass
`ChoicesType.__new__` during class creation.

Keeping `label` in the stubs is useful for type checking but
doesn't match the runtime. Looking for guidance on whether to:
1. Remove `label` to match runtime exactly
2. Add these entries to the stubtest allowlist


scripts/stubtest/allowlist.txt
```diff
@@ -231,6 +231,21 @@ django.contrib.auth.views.UserModel
 # Limited stubtest support for enums (https://github.com/python/mypy/issues/16806)
 django.db.migrations.operations.base.OperationCategory.__new__
 
+# The `label` parameter in __new__ is a deliberate stub-only addition:
+# Django's ChoicesType metaclass processes (value, label) tuples at class definition time,
+# but the runtime __new__ (from enum.Enum) only accepts (cls, value).
+# We keep the label overload so that defining enum members type-checks correctly.
+django.db.models.enums.IntegerChoices.__new__
```
UnknownPlatypus, line 238: I have absolutely no clue if any of that is correct. Do you have any ideas @ngnpope ?

ngnpope: Not sure I'll have time to look into it this week

PR #3217, opened 2026-03-24
Reparametrize implicit generic QuerySet subclasses

federicobond:
When a custom QuerySet subclass is defined without explicit type parameters (e.g. `class MyQS(QuerySet): ...`), make it implicitly generic by copying the parent QuerySet's type variables, analogous to the existing `reparametrize_any_manager_hook` for Manager subclasses.

This enables the annotate plugin hook to propagate annotation type information through custom querysets via `copy_modified(args=...)`, which requires the queryset class to have type variables.

*Disclosure: I've used Claude Opus 4.6 to assist in writing this changeset, but I have carefully iterated and manually reviewed the results.*


sobolevn (commented):
Please, test this with explicit `disallow_any_generics = True`, since your docstring mentions this.

sobolevn (approved):

sobolevn:
Thanks a lot!

federicobond:
Thank you. Was just about to comment on the new test, got sidetracked.

UnknownPlatypus:
You are too fast ahah, I also did that in #2776 + a small refactor, I'll extract the refactor, it was rather nice.
Currently both functions are exact duplicates of 35+ lines with only variable names changing

This is what I ended up with:
```python
def reparametrize_any_manager_hook(ctx: ClassDefContext) -> None:
    """
    Add implicit generics to manager classes that are defined without generic.

    Eg.

        class MyManager(models.Manager): ...

    is interpreted as:

        _T = TypeVar("_T", bound=Model, covariant=True)
        _QS = TypeVar("_QS", bound=QuerySet[Any], covariant=True, default=QuerySet[_T])
        class MyManager(models.Manager[_T, _QS]): ...
    """
    reparametrize_generic_class(ctx, fullnames.BASE_MANAGER_CLASS_FULLNAME)


def reparametrize_any_queryset_hook(ctx: ClassDefContext) -> None:
    """
    Add implicit generics to queryset classes that are defined without generic.

    Eg.

        class MyQuerySet(models.QuerySet): ...

    is interpreted as:

        _T = TypeVar("_T", bound=Model, covariant=True)
        class MyQuerySet(models.QuerySet[_T]): ...
    """
    reparametrize_generic_class(ctx, fullnames.QUERYSET_CLASS_FULLNAME)
```

federicobond:
Makes sense, I decided not to prematurely abstract without a better understanding of the project, but you have more context than I do.

UnknownPlatypus:
No worries, thanks for the contribution, this is very useful !

PR #3219, opened 2026-03-24
Refactor auto reparametrize to reduce code duplication

UnknownPlatypus:
Followup to #3217 with a small refactor extracted from #2776

_I had to create this using the cli because github UI was returning 500_


sobolevn (approved):
Thanks! Nice one :)

PR #3221, opened 2026-03-25
Resolve queryset annotate types for expressions with static ClassVar

federicobond:
Instead of always typing annotated fields as Any, resolve the concrete Python type when the expression has a static ClassVar output_field (e.g. Count → int, Exists → bool, Length → int, Now → datetime).

The current implementation introduces some unsoundness, since functions like `Length` can return `None` for nullable fields. This is not the case for `Count` or `Exists`, for example. Setting `is_nullable=True` here is theoretically safer but would make the types wrong in the other direction, and is more annoying in practice.

The real solution is to query the Django runtime for the type of the requested field argument, but it introduces quite a bit of complexity, and I'm not sure if it's worth pursuing before agreeing this is a desirable feature.

For now, I'm happy to discuss and gather more feedback.

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -16,7 +16,7 @@ from django.db.models.lookups import Transform
 from django.db.models.sql.query import Query
 from mypy.checker import TypeChecker
 from mypy.errorcodes import NO_REDEF
-from mypy.nodes import ARG_NAMED, ARG_NAMED_OPT, ARG_STAR, CallExpr, Expression, ListExpr, SetExpr, TupleExpr
+from mypy.nodes import ARG_NAMED, ARG_NAMED_OPT, ARG_STAR, CallExpr, Expression, ListExpr, SetExpr, TupleExpr, Var
 from mypy.plugin import FunctionContext, MethodContext
 from mypy.types import AnyType, Instance, LiteralType, ProperType, TupleType, TypedDictType, TypeOfAny, get_proper_type
 from mypy.types import Type as MypyType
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
This looks super cool! cc @UnknownPlatypus

UnknownPlatypus:
I'll check this this week, it looks promising !

tests/typecheck/managers/querysets/test_annotate.yml
```diff
@@ -584,3 +584,40 @@
...
+        from typing_extensions import reveal_type
+        from myapp.models import User
+        from django.db.models import Count, Exists, OuterRef
+        from django.db.models.functions import Length, Now
+
+        # Count has output_field: ClassVar[IntegerField] → int
+        qs = User.objects.annotate(user_count=Count('id'))
+        annotated = qs.get()
+        reveal_type(annotated.user_count) # N: Revealed type is "builtins.int"
+
```
UnknownPlatypus:
```suggestion
        user_count = User.objects.annotate(user_count=Count('id')).get().user_count
        reveal_type(user_count) # N: Revealed type is "builtins.int"

```
SO that we avoid having to increment the annotated intermediary variables

tests/typecheck/managers/querysets/test_annotate.yml
```diff
@@ -584,3 +584,40 @@
         class MyModel(models.Model):
             name = models.CharField(max_length=100)
             objects = MyManager()
+
+-   case: annotate_resolves_output_field_type
```
UnknownPlatypus: Can we test more django functions ? Maybe using some exemple from django / django docs here -> https://github.com/search?q=repo%3Adjango%2Fdjango+output_field&type=code

mypy_django_plugin/transformers/querysets.py
```diff
@@ -198,12 +198,39 @@ def gather_kwargs(ctx: MethodContext) -> dict[str, MypyType] | None:
...
+    if output_field_sym is None or output_field_sym.node is None:
+        return None
+
+    node = output_field_sym.node
+    if not isinstance(node, Var) or node.type is None:
+        return None
+
+    field_type = get_proper_type(node.type)
+    if not isinstance(field_type, Instance):
+        return None
```
UnknownPlatypus, line 247:
```suggestion
    if not (
        isinstance(proper := get_proper_type(expr_type), Instance)
        and (output_field_sym := proper.type.get("output_field")) is not None
        and output_field_sym.node is not None
        and isinstance(node := output_field_sym.node, Var)
        and node.type is not None
        and isinstance(field_type := get_proper_type(node.type), Instance)
    ):
        return None
```
Maybe a version like this is more readable ? Not sure, what do you think?

mypy_django_plugin/transformers/querysets.py
```diff
@@ -198,12 +198,39 @@ def gather_kwargs(ctx: MethodContext) -> dict[str, MypyType] | None:
...
+
+
 def gather_expression_types(ctx: MethodContext) -> dict[str, MypyType]:
     kwargs = gather_kwargs(ctx)
     if not kwargs:
         return {}
 
-    # For now, we don't try to resolve the output_field of the field would be, but use Any.
+    # Try to resolve the output_field type for each expression. For expressions
+    # with a static ClassVar output_field (e.g. Count → IntegerField → int),
```
UnknownPlatypus, line 258: Could we use the return type of the property/cached_property when applicable ?

mypy_django_plugin/transformers/querysets.py
federicobond: I find the original a bit easier in the eyes to be honest, but happy to change it if the second style is more conventional across the project.

mypy_django_plugin/transformers/querysets.py
federicobond: Not sure what you mean, which property?

federicobond: Nvm, got it!

tests/typecheck/managers/querysets/test_annotate.yml
federicobond: Done!

tests/typecheck/managers/querysets/test_annotate.yml
federicobond: Done!

mypy_django_plugin/transformers/querysets.py
UnknownPlatypus: no need, we don't have strong opinion on that

mypy_django_plugin/transformers/querysets.py
federicobond: I've added resolution via property return type in a separate commit.

PR #3223, opened 2026-03-26
Allow missing config section when `DJANGO_SETTINGS_MODULE` is set

### LINKED ISSUE #2912: no section [tool.django-stubs] found
# Bug report

## What's wrong

Running mypy with `DJANGO_SETTINGS_MODULE=mysite.settings` and `[tool.django-stubs]` section missing in `pyproject.toml` fails:
```
(config)
...
[tool.django-stubs]
django_settings_module = str (default: `os.getenv("DJANGO_SETTINGS_MODULE")`)
strict_settings = bool (default: true)
strict_model_abstract_attrs = bool (default: true)
...

(django-stubs) mypy: error: no section [tool.django-stubs] found
```

To fix this, an empty `[tool.django-stubs] ` sections needs to be created.
```

[tool.django-stubs]
# intentionally left empty, `django_settings_module` will be read from DJANGO_SETTINGS_MODULE env var
```
## How is that should be
User shouldn't need to set a useless empty section. Only fail if `[tool.django-stubs]` is missing and `DJANGO_SETTINGS_MODULE` is not set.

## System information

- OS: n/a
- `python` version: 3.14.0
- `django` version: 5.2.8
- `mypy` version: 1.18.2
- `django-stubs-ext` version: 5.2.7

emmanuel-ferdman:
### PR Summary
When `DJANGO_SETTINGS_MODULE` is set as an env var, the plugin currently still requires an empty `[tool.django-stubs]` (or `[mypy.plugins.django-stubs]`) section to exist in the config file. This is annoying for setups that rely on the env var (Docker, CI, multi-environment deploys). PR #2021 made `django_settings_module` optional within the section, but the section itself remained mandatory - so the env var fallback was unreachable when the section was missing. This PR just adds an env var check before erroring on a missing section, so if you've got `DJANGO_SETTINGS_MODULE` set, it uses that with defaults and moves on.

Fixes #2912

### DIFF
--- a/mypy_django_plugin/config.py
+++ b/mypy_django_plugin/config.py
@@ -89,7 +89,9 @@ class DjangoPluginConfig:
         try:
             config: dict[str, Any] = data["tool"]["django-stubs"]
         except KeyError:
-            toml_exit(MISSING_SECTION.format(section="tool.django-stubs"))
+            if not os.getenv(DJANGO_SETTINGS_ENV_VAR):
+                toml_exit(MISSING_SECTION.format(section="tool.django-stubs"))
+            config = {}
 
         django_settings_module = config.get("django_settings_module") or os.getenv(DJANGO_SETTINGS_ENV_VAR)
         if not django_settings_module:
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Great, thanks!

PR #3226, opened 2026-03-26
Fix `NestedObjects.nested` return type to `list[Any]`

### LINKED ISSUE #1985: NestedObjects.nested should return a more complex type
# Bug report

## What's wrong

https://github.com/typeddjango/django-stubs/blob/9880684a36f70e61124115a296e34cd0df15e3f1/django-stubs/contrib/admin/utils.pyi#L57
In the code above, the `nested` method returns a simple `list[Model]`, while it is a bit more complex than that.

## How is that should be

It should return something like `list[RecursiveModelList]` with RecursiveModelList being:
```
RecursiveModelList: TypeAlias = Model | list['RecursiveModelList']
```

### How to reproduce
You can verify by running the following and checking what's in `collected_instances`:
```
class Author(Model):
    name = CharField(max_length=100)


class BooksCollection(Model):
    name = CharField(max_length=100)
    author = ForeignKey(Author, on_delete=CASCADE)


class Book(Model):
    name = CharField(max_length=100)
    collection = ForeignKey(BooksCollection, on_delete=CASCADE)


author = Author.objects.create(name="Author")
collection = BooksCollection.objects.create(name="Author's life adventures", author=author)
book = Book.objects.create(name="Author's childhood", collection=collection)

collector = NestedObjects(using="default")
collector.collect([author])
collected_instances = collector.nested()
```

## System information

- OS: MacOS
- `python` version: 2.11.2
- `django` version: 4.2.10
- `mypy` version: 1.8.0

emmanuel-ferdman:
### PR Summary
`NestedObjects.nested()` returns a recursive tree structure like `[author, [collection, [book]]]`, but the stub typed it as a flat `list[Model]`. Changed the return type to `list[Any]` since the nested structure can't be expressed without recursive type aliases, which are slow in mypy.

Fixes #1985


django-stubs/contrib/admin/utils.pyi
```diff
@@ -21,6 +21,8 @@ from typing_extensions import TypedDict, override
 
 _T = TypeVar("_T")
 
+_NestedList: TypeAlias = _T | list[_NestedList[_T]]
```
sobolevn: Can you please test if this slows down the type-checking process? What will the difference be between this version and `list[Any]`?

emmanuel-ferdman: @sobolevn benchmarked 10 runs each on a file that uses `nested()` with narrowing:

| | Min | Max | Avg |
|---|---|---|---|
| `_NestedList[_T]` | 1.54s | 1.67s | 1.58s |
| `list[Any]` | 1.49s | 1.54s | 1.52s |

~40ms difference (~3%), well within noise for a single-method type alias. 

I think the recursive alias is worth it because `list[Any]` silently discards the `format_callback` return type:
```python
def fmt(m: Model) -> str:
    return str(m)

# With list[Any]:
result = collector.nested(format_callback=fmt)
reveal_type(result)  # list[Any] - lost that fmt returns str

# With _NestedList[_T]:
result = collector.nested(format_callback=fmt)
reveal_type(result)  # list[str | list[...]] - str preserved
```
It also enables `isinstance` narrowing at every depth:

```python
# With list[Any]: no help from mypy, item is Any
# With _NestedList: item is Model, inner is Model | list[...]
for item in collector.nested():
    if isinstance(item, Model):
        item.pk  # narrowed to Model
    elif isinstance(item, list):
        for inner in item:
            if isinstance(inner, Model):
                inner.pk  # narrowed again
```
That said, I didn't actually find any real-world project on GH that type-checks code around `nested()` - every repo I checked uses untyped flatten helpers or passes the result straight to templates. So happy to switch to `list[Any]` if you prefer keeping it simple.

sobolevn: I would prefer `list[Any]`, because recursive type aliases are very slow in mypy and has a lot of potential problems :(

Since there's no real gain from this, so let's keep it simple.

emmanuel-ferdman: Done :)

sobolevn (approved):

PR #3228, opened 2026-03-27
Use `Model` instead of `Self` in parms and add `AltersData` inheritance

### LINKED ISSUE #2605: `Self` used in contravariant positions
In e.g. `django.db.models.Model.[a]refresh_from_db`, `Self` is used in a contravariant position:
https://github.com/typeddjango/django-stubs/blob/0be73ebfbc405d0b9faaa898660b2e51e7d6086a/django-stubs/db/models/base.pyi#L96-L107

When I override this in my own model as e.g.

```py
    @override
    def refresh_from_db(
        self,
        using: str | None = None,
        fields: Iterable[str] | None = None,
        from_queryset: QuerySet[Self] | None = None,
    ) -> None:
```

it violates the LSP, as explained in https://discuss.python.org/t/unsoundness-of-contravariant-self-type/86338. On `pyright>=1.1.399`, it will `reportIncompatibleMethodOverride`.

The workaround is to instead use `QuerySet[Any]` in the code above.

Perhaps a `self: T` with `QuerySet[T]` could work here? I'm not sure whether that'd make it theoretically sound, but it might be worth a shot 🤷🏻.

## System information

- OS: irrelevant
- `python` version: 3.13
- `django` version:  5.2
- `mypy` version: `pyright==1.1.399` or `basedpyright==1.28.5`

emmanuel-ferdman:
### PR Summary
`refresh_from_db`, `arefresh_from_db`, `_do_update`, `_get_pk_val`, and `unique_error_message` used `Self` in their parameter types, which is unsound in contravariant positions - overriding these methods in a subclass with the same `Self` pattern triggers `reportIncompatibleMethodOverride` in pyright/ty/pyrefly because `QuerySet[Model]` is not assignable to `QuerySet[MyModel]` (LSP violation). This PR changes these to use `Model` instead, which is the widest valid type and what the runtime actually accepts. Also this PR adds the missing `AltersData` to `Model` and `QuerySet` base classes to match the runtime MRO.

Fixes #2605.

### DIFF
--- a/tests/assert_type/db/models/test_base.py
+++ b/tests/assert_type/db/models/test_base.py
@@ -1,0 +1,18 @@
+from collections.abc import Iterable
+
+from django.db import models
+from django.db.models.query import QuerySet
+from typing_extensions import override
+
+
+class MyModel(models.Model):
+    @override
+    def refresh_from_db(
+        self,
+        using: str | None = None,
+        fields: Iterable[str] | None = None,
+        from_queryset: QuerySet[models.Model] | None = None,
+    ) -> None: ...
+
+    class Meta:
+        app_label = "myapp"
--- a/tests/assert_type/db/models/test_base.py
+++ b/tests/assert_type/db/models/test_base.py
@@ -2,10 +2,21 @@ from collections.abc import Iterable
 
 from django.db import models
 from django.db.models.query import QuerySet
-from typing_extensions import override
+from typing_extensions import assert_type, override
 
 
 class MyModel(models.Model):
+    class Meta:
+        app_label = "myapp"
+
+
+class OtherModel(models.Model):
+    class Meta:
+        app_label = "myapp"
+
+
+# Override with QuerySet[Model] — valid
+class OverrideWithModel(models.Model):
     @override
     def refresh_from_db(
         self,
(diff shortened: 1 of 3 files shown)


sobolevn (commented):
Let's please add a `assert_type` test for this, it should be possible to override `Model.refresh_from_db` method with the correct annotation.

emmanuel-ferdman:
@sobolevn had it ready in my local tests :)

tests/assert_type/db/models/test_base.py
```diff
@@ -0,0 +1,18 @@
...
+from typing_extensions import override
+
+
+class MyModel(models.Model):
+    @override
+    def refresh_from_db(
+        self,
+        using: str | None = None,
+        fields: Iterable[str] | None = None,
+        from_queryset: QuerySet[models.Model] | None = None,
```
sobolevn, line 25: `MyModel` won't work here, right?

emmanuel-ferdman: Right, `QuerySet[MyModel]` errors on both mypy and pyright since it narrows the parameter type. Should we add another test case for it?

sobolevn: Yes, please. But, it seems kinda right to have it typed as `QuerySet[MyModel]`?

Because passing `from_queryset=QuerySet[SomeOther]` with no error does seem like a false negative to me.

Do you have any ideas on how we can make this right?

emmanuel-ferdman: We could try `self: _Self` with `QuerySet[_Self]` - this ties the queryset to the actual model type. So `user.refresh_from_db(from_queryset=article_qs)` would correctly error. It's stricter than runtime since `QuerySet[Model]` gets rejected for typed subclass instances, but I think passing a wrong model's queryset is a bug waiting to happen anyway :)

sobolevn: Let's try this :)

emmanuel-ferdman: Code updated :)

sobolevn (approved):

PR #3231, opened 2026-03-27
Migrate typecheck tests to assert type

UnknownPlatypus:
This made a ~25% speed improvement for local run of tests and is also great so that these test are checked by other typecheckers.

I've explicitly left tests that are too tied to the mypy plugin for obvious reasons.
Other that that, this was mostly just moving code around


sobolevn (approved):
I fully support this!

PR #3232, opened 2026-03-28
Resolve .values() field types on annotated querysets

federicobond:
Previously, `extract_proper_type_queryset_values` bailed out early for annotated models, returning `dict[str, Any]` instead of a precise TypedDict. Now it resolves annotation fields from the model's TypedDict type arg and model fields from Django's field metadata.

I've been working on `values_list` too but this changeset ended up quite neat that I didn't want to introduce more friction.

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -361,6 +361,15 @@ def resolve_field_lookups(lookup_exprs: Sequence[Expression], django_context: Dj
     return field_lookups
 
 
+def _get_annotation_field_types(model_type: Instance) -> dict[str, MypyType]:
+    """Extract annotation field types from an annotated model's TypedDict type argument."""
+    if model_type.args:
+        annotations = get_proper_type(model_type.args[0])
+        if isinstance(annotations, TypedDictType):
+            return dict(annotations.items)
+    return {}
+
+
 def extract_proper_type_queryset_values(ctx: MethodContext, django_context: DjangoContext) -> MypyType:
     """
     Extract proper return type for QuerySet.values(*fields, **expressions) method calls.
@@ -368,7 +377,7 @@ def extract_proper_type_queryset_values(ctx: MethodContext, django_context: Djan
     See https://docs.djangoproject.com/en/stable/ref/models/querysets/#values
     """
     django_model = helpers.get_model_info_from_qs_ctx(ctx, django_context)
-    if django_model is None or django_model.is_annotated:
+    if django_model is None:
         return ctx.default_return_type
 
     default_return_type = get_proper_type(ctx.default_return_type)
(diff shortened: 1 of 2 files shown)


tests/typecheck/managers/querysets/test_annotate.yml
```diff
@@ -300,7 +300,7 @@
         reveal_type(values_list_flat_unknown)  # N: Revealed type is "Any"
 
         values_no_params = Blog.objects.annotate(foo=F('id')).values().get()
-        reveal_type(values_no_params)  # N: Revealed type is "builtins.dict[builtins.str, Any]"
+        reveal_type(values_no_params)  # N: Revealed type is "TypedDict({'id': builtins.int, 'num_posts': builtins.int, 'text': builtins.str, 'foo': Any})"
```
federicobond: Should I remove these lines since type broadening no longer applies to `.values()` calls?

mypy_django_plugin/transformers/querysets.py
```diff
@@ -379,22 +390,32 @@ def extract_proper_type_queryset_values(ctx: MethodContext, django_context: Djan
...
+
         field_lookup_type = get_field_type_from_lookup(
-            ctx, django_context, django_model.cls, lookup=field_lookup, method="values"
+            ctx, django_context, django_model.cls, lookup=field_lookup, method="values", silent_on_error=is_annotated
         )
         if field_lookup_type is None:
-            return default_return_type.copy_modified(args=[django_model.typ, AnyType(TypeOfAny.from_error)])
-
-        column_types[field_lookup] = field_lookup_type
+            if is_annotated:
```
```suggestion
            if django_model.is_annotated:
```

this seems more readable to me.

mypy_django_plugin/transformers/querysets.py
```diff
@@ -361,16 +361,27 @@ def resolve_field_lookups(lookup_exprs: Sequence[Expression], django_context: Dj
...
     Extract proper return type for QuerySet.values(*fields, **expressions) method calls.
 
     See https://docs.djangoproject.com/en/stable/ref/models/querysets/#values
     """
     django_model = helpers.get_model_info_from_qs_ctx(ctx, django_context)
-    if django_model is None or django_model.is_annotated:
+    if django_model is None:
         return ctx.default_return_type
 
+    is_annotated = django_model.is_annotated
```
```suggestion
```

I prefer not to have read-only vars that are really far away from their usage

tests/typecheck/managers/querysets/test_annotate.yml
sobolevn: Sorry, I don't understand your question :)

federicobond: This test name is `annotate_values_or_values_list_before_or_after_annotate_broadens_type`, but with the latest changes, annotate no longer causes the type of `.values()` to widen to `dict[builtins.str, Any]`.

Now this code is covered by the new `test_values_on_with_annotations_queryset` test.

UnknownPlatypus: we can probably remove it ye

federicobond: Done.

mypy_django_plugin/transformers/querysets.py
```diff
@@ -379,22 +388,37 @@ def extract_proper_type_queryset_values(ctx: MethodContext, django_context: Djan
     if field_lookups is None:
         return AnyType(TypeOfAny.from_error)
 
+    annotation_types = _get_annotation_field_types(django_model.typ) if django_model.is_annotated else {}
+
```
UnknownPlatypus, line 392: I'm wondering if we should replace this `is_annotated` attribute with `annotation_types`, populating it in `get_model_info_from_qs_ctx`.

We had a few issues lately due to not tracking these annotation well enough and it might solve it elegantly.

We can defer this to a later pr if needed tho

federicobond: Makes sense. There is a similar case I found when making the changes to resolve values in `values_list` calls. I would prefer to review all changes to `is_annotated` together in that follow up pr as we will have more context to decide.

UnknownPlatypus: Looking forward to review that 🙌

PR #3235, opened 2026-03-28
Make Lookup type argument optional by defaulting _T to Any

### LINKED ISSUE #2649: Unexpected type argument required for django.db.models.lookups.Lookup
# Bug report

## What's wrong

For some reason [`django.db.models.lookups.Lookup` requires a type argument](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/lookups.pyi#L15) and I do not see anywhere in the file where the `_T` argument is actually used other than as a type argument for subclasses.

## How is that should be

The type argument should not be required.

## System information

- OS: Linux
- `python` version: 3.12
- `django` version: 5.1
- `mypy` version: Latest stable
- `django-stubs` version: 5.2
- `django-stubs-ext` version: Latest stable

UnknownPlatypus:
Fixes #2649.

The type argument is still required because: 
  1. Subclasses specialize `_T` with concrete types, for ex `PatternLookup(BuiltinLookup[str])`, `IsNull(BuiltinLookup[bool])`, `Exact[int | float]`, etc.                     
  2. The plugin code in `_resolve_lookup_type_from_lookup_class` extracts that type argument to determine what types are valid for lookup values. This powers error messages like: `User.objects.filter(username__contains=1) # E: Incompatible type for lookup 'username__contains': (got "int", expected "str")`  

Having a default to Any is a good middle ground I think, people can still have more precise types if needed

### DIFF
--- a/django-stubs/db/models/lookups.pyi
+++ b/django-stubs/db/models/lookups.pyi
@@ -10,7 +10,7 @@ from django.utils.datastructures import OrderedSet
 from django.utils.functional import cached_property
 from typing_extensions import override
 
-_T = TypeVar("_T")
+_T = TypeVar("_T", default=Any)
 
 class Lookup(Expression, Generic[_T]):
     lookup_name: str | None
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
Let's add a test case for this :)

sobolevn (approved):

PR #3236, opened 2026-03-28
Include django and django-stubs versions in plugin config data

### LINKED ISSUE #2338: Include current `django-stubs` version in reporting of config data
I'm not 100% sure but I think that if you change version of `django-stubs` but keep the same version of `mypy` after a cached run,  `mypy` reuses the cache and there could possibly be crashes or all sorts of, strange, errors emitted for the second run.

But if we're able to include the current `django-stubs` version when we report config data in

https://github.com/typeddjango/django-stubs/blob/f04323aaec46118125484248fbd9c00966f0ed0e/mypy_django_plugin/main.py#L307-L314

We should get "automatic" cache tainting whenever switching between `django-stubs` versions.

I must admit that I haven't gotten around to try out if this is indeed a problem, it's just a suspicion so far. Gonna have to see if I can find 2 `django-stubs` versions and a snippet of code that backs up my suspicion here. Think I've recently run in to this in one of the projects I'm working on though.

## System information

- OS:
- `python` version:
- `django` version:
- `mypy` version:
- `django-stubs` version:
- `django-stubs-ext` version:

UnknownPlatypus:
This ensures mypy automatically invalidates its cache whenever either package version changes, preventing stale cache issues when upgrading django-stubs or django.

It would crash if django or django-stubs is not installed with importlib.metadata.PackageNotFoundError, should we handle it ?

Closes #2338

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -1,3 +1,4 @@
+import importlib.metadata
 import itertools
 import sys
 from collections.abc import Callable
@@ -327,6 +328,8 @@ class NewSemanalDjangoPlugin(Plugin):
         # Cache would be cleared if any settings do change.
         extra_data = {
             "AUTH_USER_MODEL": self.django_context.settings.AUTH_USER_MODEL,
+            "django_version": importlib.metadata.version("django"),
+            "django_stubs_version": importlib.metadata.version("django-stubs"),
         }
         return self.plugin_config.to_json(extra_data)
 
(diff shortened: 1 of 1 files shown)


mypy_django_plugin/main.py
```diff
@@ -327,6 +328,8 @@ def report_config_data(self, ctx: ReportConfigContext) -> dict[str, Any]:
         # Cache would be cleared if any settings do change.
         extra_data = {
             "AUTH_USER_MODEL": self.django_context.settings.AUTH_USER_MODEL,
+            "django_version": importlib.metadata.version("django"),
+            "django_stubs_version": importlib.metadata.version("django-stubs"),
```
sobolevn, line 332: Let's include `django-stubs-ext` if it is present, maybe?

It can also influence the type-checking process.

sobolevn (approved):

PR #3237, opened 2026-03-28
Expose django admin fieldset types in `django_stubs_ext`

### LINKED ISSUE #1960: Add `FieldOpts` type alias to `django_stubs_ext`
# Bug report

## What's wrong

Similar to the existing type aliases defined within `django_stubs_ext` (E.g. `StrPromise`), it would be useful to have the `FieldOpts` type alias exposed for type checking the `get_fieldsets()` method of a `ModelAdmin`

## How it should work

```python
from typing import TypeAlias, TypeVar

from django.contrib.admin import ModelAdmin
from django_stubs_ext import FieldOpts

T_model = TypeVar("T_model", bound=Model)
Fieldsets: TypeAlias = list[tuple[str | StrPromise | None, FieldOpts]] | tuple[tuple[str | StrPromise | None, FieldOpts], ...] | tuple[()]

class CustomModelAdmin(ModelAdmin[T_model]):
    @override
    def get_fieldsets(self, request: HttpRequest, obj: T_model | None = None) -> Fieldsets:
        ...
```

## System information

- OS: Windows 10 Home 22H2 19045.4046
- `python` version: 3.12.1
- `django` version: 4.2.10
- `mypy` version: 1.7.1

UnknownPlatypus:
Fixes #1960 

Some open quiestions:
- should we use a namespace ? ie `from django_stubs_ext.admin import FieldsetSpec`
- should I move `FieldsetSpec` in django_stub_ext and import it in the stubs ?

### DIFF
--- a/ext/django_stubs_ext/__init__.py
+++ b/ext/django_stubs_ext/__init__.py
@@ -1,3 +1,5 @@
+from .aliases import FieldOpts as FieldOpts
+from .aliases import FieldsetSpec as FieldsetSpec
 from .aliases import QuerySetAny as QuerySetAny
 from .aliases import StrOrPromise, StrPromise
 from .aliases import ValuesQuerySet as ValuesQuerySet
@@ -9,6 +11,8 @@ from .types import AnyAttrAllowed as AnyAttrAllowed
 __all__ = [
     "Annotations",
     "AnyAttrAllowed",
+    "FieldOpts",
+    "FieldsetSpec",
     "QuerySetAny",
     "StrOrPromise",
     "StrPromise",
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
I think for now we are good! These are just aliases.

UnknownPlatypus:
> should I move FieldsetSpec in django_stub_ext and import it in the stubs ?

Thinking about it a bit more I think we should because I ide will autocomplete by copying the stub signature so it’s better to have the new public apis

Will submit a pr for this

UnknownPlatypus:
> Thinking about it a bit more I think we should because I ide will autocomplete by copying the stub signature so it’s better to have the new public apis. Will submit a pr for this

Nevermind, this is more complicated than I thought, especially in a backward compatible way

PR #3239, opened 2026-03-28
Validate `Field(validators=[...])` more precisely

### LINKED ISSUE #264: Model field validators type is incorrect

### LINKED ISSUE #1264: Feature: Use PEP 696 defaults for fields

UnknownPlatypus:
# I have made things!

Make `_ValidatorCallable` more strict by adding a typevar and binding it to the getter type of the various fields:

```diff
-_ValidatorCallable: TypeAlias = Callable[[Any], None] # noqa: PYI047
+_VT = TypeVar("_VT", contravariant=True)
+_ValidatorCallable: TypeAlias = Callable[[_VT], None] 
```

Fixes #264

### DIFF
--- a/django-stubs/contrib/gis/db/models/fields.pyi
+++ b/django-stubs/contrib/gis/db/models/fields.pyi
@@ -66,7 +66,7 @@ class BaseSpatialField(Field[_ST, _GT]):
         db_column: str | None = ...,
         db_comment: str | None = ...,
         db_tablespace: str | None = ...,
-        validators: Iterable[_ValidatorCallable] = ...,
+        validators: Iterable[_ValidatorCallable[_GT]] = ...,
         error_messages: _ErrorMessagesMapping | None = ...,
     ) -> None: ...
     @override
@@ -115,7 +115,7 @@ class GeometryField(BaseSpatialField[_ST, _GT]):
         db_column: str | None = ...,
         db_comment: str | None = ...,
         db_tablespace: str | None = ...,
-        validators: Iterable[_ValidatorCallable] = ...,
+        validators: Iterable[_ValidatorCallable[_GT]] = ...,
         error_messages: _ErrorMessagesMapping | None = ...,
     ) -> None: ...
     @override
--- a/django-stubs/contrib/postgres/fields/array.pyi
+++ b/django-stubs/contrib/postgres/fields/array.pyi
@@ -28,7 +28,7 @@ class ArrayField(CheckPostgresInstalledMixin, CheckFieldDefaultMixin, Field[_ST,
     default_error_messages: ClassVar[_ErrorMessagesDict]
     base_field: Field
     size: int | None
-    default_validators: Sequence[_ValidatorCallable]
+    default_validators: Sequence[_ValidatorCallable[_GT]]
     from_db_value: Any
     def __init__(
         self,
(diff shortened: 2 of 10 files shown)


UnknownPlatypus:
> + zerver/models/realm_emoji.py:34: error: Argument "validators" to "TextField" has incompatible type "list[_Deconstructible]"; expected "Iterable[Callable[[Never], None]]"  [arg-type]

This is very interesting, because we now have `_GT` in the init signature, mypy tries to resolve the typevar from it and seem to fail and pick  `Never`.

I think it would work if we managed to remove the custom `_pyi_private_get_type` and replace `_GT` with typevar + defaults like `TypeVar("_StrFieldGT", covariant=True, default=str)` -> cf #1264

UnknownPlatypus:
@sobolevn Do you have any context on the `_pyi_private_get_type` special variable and if you think we could get rid of them ? I have a feeling we should ?

sobolevn:
Yes, we should totally try. We implemented `_pyi_private_{g,s}et_type` when there were no type var defaults.

PR #1273, opened 2022-11-26
Project script fixes, Justfile

benspaulding:
As I was working on the bits in #1272 I found some issues with stub generation, permissions on scripts, and other project tooling that was a little cumbersome.

- If you want a separate PR for each commit, please let me know!
- If you want to cherry-pick all or some of these commits yourself, please do!

Most commits are quite small, but the one with the `Justfile` made working on the project much simpler as I did not have to remember various scripts for listing, testing, building, etc. I added the following to the README file:

> You can now follow instructions in subsequent sections of this document to manually install dependencies, run tests, etc., _or_ you can [install `just`](https://just.systems/man/en/) to run our `Justfile`. For example:

```console
> just
Available recipes:
    ls              # list available recipes
    help            # print help info & list available recipes
    full            # refresh setup, run checks & builds
    clean           # remove development artifacts
    setup           # setup up project development environment
    draft-stubs     # generate new draft stub files
    lint            # run all linters & checkers
    test            # run all tests
    build           # build project packages
    release         # build & release project packages

> just --show full
# refresh setup, run checks & builds
full: clean setup lint test build

> just full
... # proceeds to create virtualenv, install dependencies, run tests, etc.
```

PR #3240, opened 2026-03-28
Add justfile and update CONTRIBUTING.md

### LINKED ISSUE #320: Create Makefile or tox.ini
While working on #319 I found our setup very complicated.
We need to have a single command to test things locally.

It might exclude some long-running parts like `python ./scripts/typecheck_tests.py --django_version=3.0` by default.

UnknownPlatypus:
Fixes #320
Fixes #1273

I personally mostly use `just pre-mr-check` before submitting a PR to avoid unnecessary ci errors

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,7 +80,8 @@ Available recipes:
     ty             # Run ty on test cases
 
     [test]
-    test *args='tests' # Run pytest test suite
+    test +args     # Run pytest on specific files or with custom args (no xdist)
+    all-test       # Run full pytest test suite with parallel workers
     stubtest *args # Run stubtest to check stubs match runtime
     ext-test       # Run django-stubs-ext tests
 
@@ -100,7 +101,8 @@ Or run individual checks:
 ```bash
 just lint          # pre-commit hooks (ruff, codespell, ...)
 just mypy          # mypy on plugin, ext, scripts, stubs and tests
-just test          # pytest test suite
+just test tests/test_xyz.yml  # run specific tests (no xdist)
+just all-test      # full pytest suite (parallel)
 just stubtest      # stubtest: check stubs match runtime
 just pyright       # pyright on test cases
 just ty            # ty on test cases
@@ -110,7 +112,7 @@ just pyrefly       # pyrefly on test cases
 Extra arguments can be passed to `test` and `stubtest`:
 
 ```bash
-just test -k test_name
+just test tests -k test_name
 just stubtest --allowlist extra.txt
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,46 +38,87 @@ To do so follow this [official github guide](https://docs.github.com/en/free-pro
 We use [uv](https://github.com/astral-sh/uv) to manage our dev dependencies.
 To install it, see their [installation guide](https://docs.astral.sh/uv/getting-started/installation/)
 
-Once it's done, simply run the following command to automatically setup a virtual environment and install dev dependencies:
+We use [just](https://github.com/casey/just) as a command runner. Install it with:
+
+```bash
+uv tool install rust-just
+```
+
+Then bootstrap your environment (installs pre-commit hooks and syncs dependencies):
+
+```bash
+just bootstrap
+```
+
+<details>
+<summary>Manual setup (without <code>just</code>)</summary>
 
 ```bash
 uv sync
 source .venv/bin/activate
+pre-commit install --install-hooks
 ```
-
-Finally, install the pre-commit hooks:
-
-```bash
-pre-commit install
-```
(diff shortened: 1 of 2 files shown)


ahmedasar00:
Hi @UnknownPlatypus , I sent you an email regarding the GSoC proposal,
just wanted to make sure it didn't get lost. No rush, whenever you have time..

UnknownPlatypus:
@sobolevn @emmanuel-ferdman Could you try checking this branch out and testing the just commands ? 
To get a feel of it

justfile
```diff
@@ -0,0 +1,73 @@
...
+    uv sync
+
+# Run pre-commit hooks on all files
+[group('dev')]
+lint:
+    pre-commit run --all-files
+
+# Run all checks before submitting a PR
+[group('dev')]
+pre-mr-check: lint ty pyrefly mypy stubtest pyright test
```
emmanuel-ferdman: should it also run `ext-test`?
```suggestion
pre-mr-check: lint ty pyrefly mypy stubtest pyright test ext-test
```

justfile
```diff
@@ -0,0 +1,73 @@
+# List all available recipes
+_default:
+    @just --list --unsorted
+
+# Bootstrap dev environment: install pre-commit hooks and sync dependencies
+[group('dev')]
+bootstrap:
+    uv tool install pre-commit
+    pre-commit install --install-hooks
+    uv sync
```
emmanuel-ferdman: I think we should change the order of bootstrap - `uv sync` should run before `pre-commit install --install-hooks` since the hooks may depend on packages from the venv.
```suggestion
    uv sync
    pre-commit install --install-hooks
```

justfile
```diff
@@ -0,0 +1,73 @@
...
+
+# Run ty on test cases
+[group('typecheck')]
+ty:
+    uv run ty check tests/assert_type
+
+# Run pytest test suite
+[group('test')]
+test *args:
+    uv run pytest -n auto tests {{ args }}
```
emmanuel-ferdman: Should we do something like the following?
```suggestion
test *args='tests':
    uv run pytest -n auto {{ args }}
```
This way `just test` still runs all tests, but `just test path/to/file.yml` only runs that file instead of the whole suite.

CONTRIBUTING.md
```diff
@@ -38,46 +38,87 @@ To do so follow this [official github guide](https://docs.github.com/en/free-pro
...
+just pyrefly       # pyrefly on test cases
 ```
 
-If you get some unexpected results or want to be sure that tests run is not affected by previous one, remove `mypy` cache:
+Extra arguments can be passed to `test` and `stubtest`:
 
 ```bash
-rm -r .mypy_cache
+just test -- -k test_name
+just stubtest -- --allowlist extra.txt
```
emmanuel-ferdman: From what I tried it does not need the `--` seperator.

emmanuel-ferdman (commented):
Really great work, thanks for putting this together! Personally, I'm more familiar with `make` but this is was really straightforward and well-structured. Tested locally and works well. I have a few suggestions :)

justfile
UnknownPlatypus: yes totally, but before test to fail fast

justfile
UnknownPlatypus: > since the hooks may depend on packages from the venv.

This is bad practice to do so tho. `pre-commit` is supposed to handle the tools install independently unless you use the `system` language which have been renamed to [unsupported](https://pre-commit.com/#unsupported) because it's a classic footgun

But anyway, cost not much so I changed the order

justfile
UnknownPlatypus: actually `just test path/to/file.yml` is a bit wasteful currently because it spawn a lot of worker with pytest-xdist.

I've changed to have 2 different receipes:

```just
# Run pytest on specific files or with custom args
[group('test')]
test +args:
    uv run pytest {{ args }}

# Run full pytest test suite with parallel workers
[group('test')]
all-test:
    uv run pytest -n auto tests
```

justfile
```diff
@@ -0,0 +1,78 @@
...
+
+# Run pytest on specific files or with custom args
+[group('test')]
+test +args:
+    uv run pytest {{ args }}
+
+# Run full pytest test suite with parallel workers
+[group('test')]
+all-test:
+    uv run pytest -n auto tests
```
sobolevn:
```suggestion
# Run pytest tests
[group('test')]
test +args="-n auto tests":
    uv run pytest {{ args }}
```

Will this work?

UnknownPlatypus: most likely! Let me try

UnknownPlatypus: It does, neat. Thanks :raised_hands:

UnknownPlatypus:
Thanks for the feedback, I think I've addressed every suggestions. I'm going to merge!

PR #3247, opened 2026-03-30
Add an ai policy

### LINKED ISSUE #3238: Add an AI policy
We see more and more automated PR from agents so I think it would be nice to have a policy on what we want and expect from people using such tools.

astral added an [AI policy](https://github.com/astral-sh/.github/blob/main/AI_POLICY.md) recently and I'm very aligned with the content and would like to add something similar here.

I personally have no issues with people using AI, I'm using it myself and it definitely help me do more stuff, especially the tedious tasks sometimes required here. But I don't like interacting with MR that are 100% generated because when users don't own or understand the code, it takes a lot of time to review and the maintenance burden fall on us.
I also do open source to interact with humans and I'm not interested in giving advices to bots.

Would love to get maintainers stance on this before proceeding with anything (cc @sobolevn @intgr @ngnpope @emmanuel-ferdman ...)

UnknownPlatypus:
Fixes #3238

### DIFF
--- a/.github/AI_POLICY.md
+++ b/.github/AI_POLICY.md
@@ -1,0 +1,31 @@
+We support using AI (i.e., LLMs) as tools for coding. However, you remain
+responsible for any code you publish and we are responsible for any code we
+merge and release. We hold a high bar for all contributions to our projects.
+
+**AI should not be used to generate comments when communicating with
+maintainers**. We expect comments on our projects to be written by humans. We
+may hide any comments that we believe are AI generated.
+
+If you are opening an issue, we expect you to describe the problem in your own
+words.
+
+If you are opening a pull request, we expect you to be able to explain the
+proposed changes in your own words. This includes the pull request body and
+responses to questions. **Do not copy responses from the AI when replying to
+questions from maintainers.**
+
+Due to the foundational nature of our projects, we require a human in the loop
+who understands the work produced by AI. **We do not allow autonomous agents to
+be used for contributing to our projects**. We will close any pull requests that
+we believe were created autonomously.
+
+If you wish to include context from an interaction with AI in your comments, it
+must be in a quote block (e.g., using `>`) and disclosed as such. It must be
+accompanied by human commentary explaining the relevance and implications of the
+context. Do not share long snippets.
+
+We understand that AI is useful when communicating as a non-native English
+speaker. If you are using AI to edit your comments for this purpose, please take
+the time to ensure it reflects your own voice and ideas. If using AI for
+translation, we recommend writing in your native language and including the AI
+translation in a quote block.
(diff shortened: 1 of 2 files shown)


.github/AI_POLICY.md
```diff
@@ -0,0 +1,31 @@
+We support using AI (i.e., LLMs) as tools for coding. However, you remain
```
sobolevn, line 11: Let's add a comment with the original authors attribution, like here: https://github.com/wemake-services/django-modern-rest/blob/ef2df2e44b3e07c2250ba2f362585e69feca555c/.github/AI_POLICY.md#L1-L9

.github/pull_request_template.md
```diff
@@ -27,3 +27,25 @@ If you have any feedback, just write it here.
 
 It can be whatever you want!
```
intgr: Not necessarily for this PR, but the PR template was already way too long. I think everything above could be summed up in 2 sentences.

So even to get to the AI policy one would have to scroll the edit box a lot.

Image

.github/pull_request_template.md
```diff
@@ -27,3 +27,25 @@ If you have any feedback, just write it here.
...
+   authors of frequent low quality PRs may be blocked not to waste maintainers' time
+
+-->
+
+- [ ] I have read the [AI Policy](https://github.com/typeddjango/django-stubs/blob/master/.github/AI_POLICY.md)
+      and I agree to follow it
+- [ ] I have disclosed AI usage in this PR,
+      if any AI / LLM / coding agent was participating in the making of this PR
+- [ ] I have removed any "Co-Authored-By" lines that attribute coding agents
+      and I have carefully manually reviewed the final result produced by the AI
```
intgr: Let's avoid duplicating the instructions in comment & checkboxes. Either drop the `<!-- comment -->` or drop/squash the checkboxes, or both.

Personally I'd prefer no more than 1 checkbox per PR for the AI policy.

intgr: Instructions unclear if AI was **not** involved. Should this be unchecked? I can't say "[x] I've removed AI Co-Authored-By lines" because there were none to begin with. Also somewhat applies to 2nd checkbox.

intgr (commented):
Thanks, the AI policy is great, but I think PR template needs changes.

.github/pull_request_template.md
UnknownPlatypus: Yes agree

.github/pull_request_template.md
UnknownPlatypus: I've tried a shorter version of the template that fit on a screen, let me know what you think

sobolevn (approved):

intgr (approved):
Perfect, thanks!

PR #3252, opened 2026-03-31
Update dependency mypy to >=1.13,<1.21 & fix tests

intgr:
I took PR #3250 as base, did a massive search & replace, then fixed few remaining cases by hand.

* Closes #3250

AI disclosure:
* `allowlist` maintenance & few `.pyi` stubs changes made by AI


sobolevn (approved):
Thanks a lot! I propose to release 6.0.2 with this change.

tests/typecheck/models/test_related_fields.yml
```diff
@@ -176,7 +176,6 @@
 
 
 - case: test_related_fields_with_one_generic_parameter
-  expect_fail: True
```
intgr, line 179: @flaeppe Do I understand discussion at https://github.com/typeddjango/django-stubs/pull/2261#pullrequestreview-2179254958 correctly that this was just a work-around for mypy's bug?

So `expect_fail: True` can be removed now that this test no longer fails?

django-stubs/contrib/gis/db/backends/postgis/base.pyi
```diff
@@ -30,7 +29,7 @@ class GeometryType: ...
 class GeographyType: ...
 class RasterType: ...
 
-@lru_cache
+# Uses @functools.lru_cache
 def postgis_adapters(
```
intgr: The `@lru_cache` seems to cause this stub to become incorrectly typed.

Could it be bug a bug in `typeshed`?

intgr: stubtest output:

```
error: django.contrib.gis.db.backends.postgis.base.postgis_adapters is inconsistent, stub does not have parameter "geo_oid"
error: django.contrib.gis.db.backends.postgis.base.postgis_adapters is inconsistent, stub does not have parameter "geog_oid"
error: django.contrib.gis.db.backends.postgis.base.postgis_adapters is inconsistent, stub does not have parameter "raster_oid"
error: django.contrib.gis.db.backends.postgis.base.postgis_adapters is inconsistent, runtime does not have *args parameter "args"
error: django.contrib.gis.db.backends.postgis.base.postgis_adapters is inconsistent, runtime does not have **kwargs parameter "kwargs"
Stub: in file .../django-stubs/contrib/gis/db/backends/postgis/base.pyi:-1
def (*args: typing.Hashable, **kwargs: typing.Hashable) -> tuple[type[psycopg.adapt.Dumper], type[psycopg.adapt.Dumper]]
Runtime: at line 48
def (geo_oid, geog_oid, raster_oid)
```

intgr: Oh actually it's not a new problem. mypy 1.20 just started warning about this.

`@lru_cache` causes type erasure even as far back as mypy 1.0.0 & Python 3.8: https://mypy-play.net/?mypy=1.0.0&python=3.8&gist=4390954c6728af653fdcb9853332041d

intgr: ~~Reported to typeshed https://github.com/python/typeshed/issues/15580~~

Known issue: https://github.com/python/typeshed/issues/11280 - will add a code comment & keep `@lru_cache` commented out.

sobolevn:
@intgr should we make a new release?

intgr:
Yes. I ran out of time yesterday.

sobolevn:
Done :)

PR #3253, opened 2026-03-31
Resolve .values_list() field types on annotated querysets

federicobond:
Extend annotation type resolution to values_list(), matching the values() support added in #3232.

For named=True with no params, annotation fields are now merged into the NamedTuple instead of relying on the AnyAttrAllowed fallback base.

*AI Disclosure: I've used AI to write these changes, but I've carefully guided and reviewed the output.*

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -107,7 +107,7 @@ def get_values_list_row_type(
     django_context: DjangoContext,
     model_cls: type[Model],
     *,
-    is_annotated: bool,
+    annotation_types: dict[str, MypyType],
     flat: bool,
     named: bool,
 ) -> MypyType:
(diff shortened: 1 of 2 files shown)


mypy_django_plugin/transformers/querysets.py
```diff
@@ -154,11 +147,20 @@ def get_values_list_row_type(
 
     column_types = {}
     for field_lookup in field_lookups:
+        if field_lookup in annotation_types:
+            column_types[field_lookup] = annotation_types[field_lookup]
```
sobolevn: We do a double lookup here: first for `in` then for `[field_lookup]`

federicobond: Done.

federicobond:
Rebased against latest `master` so that it runs against mypy 1.20

sobolevn (approved):

PR #3255, opened 2026-04-01
Remove patching of `reveal_type()` and `reveal_locals()`

ngnpope:
This was originally added to avoid a release of django-stubs-ext going out with these left in mistakenly which would cause projects using the stubs downstream to break. As we have enabled the `unimported-reveal` error code for mypy in 2602ed8d487ac9fdab6cf5aabb8a96175c954d7c, errors will be caught in CI and we should remove this.

While there is a risk that downstream projects may not do their own checking to ensure that they are not using these unimported, that's not something for django-stubs to guard against.

There are [very few uses](https://github.com/search?q=monkeypatch+include_builtins%3D&type=code) of the `include_builtins` argument in the wild - most are just references to forks of this repo - so it seems fine to just remove it.


sobolevn (commented):
Related idea: we have a single example of `reveal_type` without a `from typing import reveal_type` in the `README.md`

Can you please fix it as well?

UnknownPlatypus (approved):

PR #3261, opened 2026-04-03
Preserve NamedTuple row type when annotating values_list(named=True) querysets

federicobond:
Previously, chaining `.annotate()` after `.values_list(named=True)` collapsed the row type to Any, losing type information about the original fields. Now we rebuild the NamedTuple with the existing fields plus annotation fields, so attribute access on the result is properly typed.

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -351,9 +351,14 @@ def extract_proper_type_queryset_annotate(ctx: MethodContext, django_context: Dj
             )
         elif isinstance(original_row_type, TupleType):
             if original_row_type.partial_fallback.type.has_base("typing.NamedTuple"):
-                # TODO: Use a NamedTuple which contains the known fields, but also
-                #  falls back to allowing any attribute access.
-                row_type = AnyType(TypeOfAny.implementation_artifact)
+                # Rebuild the NamedTuple with existing fields + annotation fields.
+                existing_fields = {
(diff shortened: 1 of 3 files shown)


mypy_django_plugin/transformers/querysets.py
```diff
@@ -351,9 +351,14 @@ def extract_proper_type_queryset_annotate(ctx: MethodContext, django_context: Dj
...
-                #  falls back to allowing any attribute access.
-                row_type = AnyType(TypeOfAny.implementation_artifact)
+                # Rebuild the NamedTuple with existing fields + annotation fields.
+                existing_fields = {
+                    name: sym.node.type or AnyType(TypeOfAny.special_form)
+                    for name, sym in original_row_type.partial_fallback.type.names.items()
+                    if sym.plugin_generated and isinstance(sym.node, Var)
+                }
+                annotation_fields = {name: AnyType(TypeOfAny.from_omitted_generics) for name in expression_types}
+                row_type = helpers.make_oneoff_named_tuple(api, "Row", {**existing_fields, **annotation_fields})
```
federicobond: These lines might be a good candidate for a helper function? Something like `extend_oneoff_named_tuple`?

sobolevn: Yes, let's do this, please :)

federicobond: Done!

sobolevn (approved):

PR #3263, opened 2026-04-05
Prefetch related queryset incompatible type

steele:
## Observed issue
When calling `prefetch_related` on a manager created via `from_queryset()`, passing a `Prefetch(...)` argument produced a spurious "[arg-type]" error.

### Example
```python
from django.db import models

class ObjectQuerySet(models.QuerySet["Object"]): ...
class ObjectManager(models.Manager["Object"]): ...

class Object(models.Model):
    objects = ObjectManager.from_queryset(ObjectQuerySet)()

class RelatedObject(models.Model):
    object = models.ForeignKey(Object, on_delete=models.CASCADE, related_name="related_objects")

from django.db.models import Prefetch

# Both of these incorrectly produce arg-type errors:
Object.objects.prefetch_related(Prefetch("related_objects"))
Object.objects.prefetch_related(
    Prefetch("related_objects", queryset=RelatedObject.objects.all())
)
```
Generates errors like these:
```
error: Argument 1 to "prefetch_related" of "QuerySet" has incompatible type "Prefetch[Literal['related_objects'], _PrefetchedQuerySetT, _ToAttrT]"; expected "str | Prefetch[_LookupT, _PrefetchedQuerySetT, _ToAttrT]"  [arg-type]

error: Argument 1 to "Prefetch" has incompatible type "str"; expected "_LookupT"  [arg-type]
```

## Root cause

It appears that the loss of all type variables, when the return type for the `prefetch_related` is dynamically updated to the concrete `QuerySet` type, prevents the correct argument matching.

More generally, in `_process_dynamic_method`, when resolving methods on managers created via `from_queryset()`, all method-level type variables were unconditionally cleared by setting `variables = ()` for `MANAGER_METHODS_RETURNING_QUERYSET` methods. I believe this was originally correct when the stubs used `Any` for all other argument types except `self`. Once more precise typing/overloads for the remaining arguments were added to the stubs, simply clearing all variables was no longer correct. For example, `_LookupT`, `_PrefetchedQuerySetT`, and `_ToAttrT` from `prefetch_related` is stripped, meaning mypy is unable to infer types for `Prefetch(...)` arguments.

## Fix

Replace the blanket `variables = ()` with a targeted filter that removes only the `Self` type variable, thus preserving other method-level type variables. Additionally relocate this logic to a more general location not specific to "queryset returning methods" but to where references to the `Self` type variable is replaced.

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -140,10 +140,6 @@ def _process_dynamic_method(
     # used by the typing stubs.
     if method_name in MANAGER_METHODS_RETURNING_QUERYSET:
         ret_type = queryset_instance
-        # Keep method-level type variables (e.g. _LookupT in prefetch_related)
-        # but remove Self since we've replaced the return type.
-        if base_that_has_method.self_type:
-            variables = tuple(v for v in variables if v.id != base_that_has_method.self_type.id)
     args_types = method_type.arg_types[1:]
     if _has_compatible_type_vars(base_that_has_method):
         typed_var = manager_instance.args or queryset_info.bases[0].args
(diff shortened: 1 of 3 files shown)


UnknownPlatypus (approved):
Nice catch, good to see one liner fixes like that

PR #3264, opened 2026-04-05
Fix `databases` type in `SimpleTestCase` to use `Literal["__all__"]`

ahmedasar00:
## Summary

The `databases` attribute in `SimpleTestCase` was typed as `set[str] | str`,
but Django only accepts the string `"__all__"` — not any arbitrary string.

This PR narrows the type to `set[str] | Literal["__all__"]` for better
type safety and accuracy.

Fixes the TODO comment that was left in the code:
`# TODO: str -> Literal['__all__']`

## Changes

- `django-stubs/test/testcases.pyi`: added `Literal` to imports, updated
  `databases` annotation, removed TODO comment

### DIFF
--- a/django-stubs/test/testcases.pyi
+++ b/django-stubs/test/testcases.pyi
@@ -3,7 +3,7 @@ import unittest
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from contextlib import AbstractContextManager
 from types import TracebackType
-from typing import Any, overload
+from typing import Any, Literal, overload
 
 from django.core.exceptions import ImproperlyConfigured
 from django.core.handlers.wsgi import WSGIHandler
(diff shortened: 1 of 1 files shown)


django-stubs/test/testcases.pyi
```diff
@@ -62,8 +62,7 @@ class SimpleTestCase(unittest.TestCase):
     client: Client
     async_client_class: type[AsyncClient]
     async_client: AsyncClient
-    # TODO: str -> Literal['__all__']
-    databases: set[str] | str
+    databases: set[str] | Literal["__all__"]
```
UnknownPlatypus, line 65: Did you check the git blame for the todo? For such a minimal change, I'm wondering why it was not fixed at the point, maybe it hides another issue ?

ahmedasar00: I checked the git blame the TODO was added in commit` 7ce5692a (PR #1575) by Gabriel Dugny in June 2023.
`
In the Django source, databases is only ever compared to "__all__" as a string:
if cls.databases == "__all__":

So any other string value would be silently ignored, which confirms Literal["__all__"] is the correct typ

sobolevn (approved):

PR #3266, opened 2026-04-05
Propagate `WithAnnotations` through custom `QuerySet` methods

### LINKED ISSUE #2681: How to correctly type manager functions that adds annotations?

UnknownPlatypus:
# I have made things!
Add a method hook that fires for any method call on a `QuerySet`/`Manager` subclass that returns a `WithAnnotation[...]` in order to merge the annotations

Also does some minor short circuting in `get_method_hook` and `get_attribute_hook

## Related issues
Fixes #2681

## AI Policy

### DIFF
--- a/tests/typecheck/managers/querysets/test_annotate.yml
+++ b/tests/typecheck/managers/querysets/test_annotate.yml
@@ -886,10 +886,10 @@
 
         class FooQuerySet(models.QuerySet[_Model]):
             def with_bar(self) -> "FooQuerySet[WithAnnotations[_Model, BarDict]]":
-                return self.annotate(bar=models.Value("bar"))  # type: ignore[no-any-return]
+                return self.annotate(bar=models.Value("bar"))
 
-            def with_baz(self) -> "FooQuerySet[WithAnnotations[FooModel, BazDict]]":
-                return self.annotate(baz=models.Value("baz"))  # type: ignore[no-any-return]
+            def with_baz(self) -> "FooQuerySet[WithAnnotations[_Model, BazDict]]":
+                return self.annotate(baz=models.Value("baz"))
 
         FooManager = Manager.from_queryset(FooQuerySet)
 
--- a/tests/typecheck/managers/querysets/test_annotate.yml
+++ b/tests/typecheck/managers/querysets/test_annotate.yml
@@ -745,3 +745,153 @@
 
         class SomeQuerySet(models.QuerySet["Accommodation"]):
             pass
+
+- case: custom_queryset_method_with_annotations
+  main: |
+    from typing_extensions import reveal_type
+    from myapp.models import FooModel
+    from django.db.models.expressions import F
+
+    # Simple method
+    qs = FooModel.objects.with_bar()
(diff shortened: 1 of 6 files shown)


tests/typecheck/managers/querysets/test_annotate.yml
```diff
@@ -651,6 +651,206 @@
...
+
+        class BarIntDict(TypedDict):
+            bar: int
+
+        class FooQuerySet(models.QuerySet):
+            def with_bar(self) -> "FooQuerySet[WithAnnotations[FooModel, BarDict]]":
+                return self.annotate(bar=models.Value("x"))  # type: ignore[no-any-return]
+
+            def with_baz(self) -> "FooQuerySet[WithAnnotations[FooModel, BazDict]]":
+                return self.annotate(baz=models.Value(1))  # type: ignore[no-any-return]
```
UnknownPlatypus: Ideally this should not be needed, still need to investigate why we lose type in this annotate

UnknownPlatypus: Okay so I understood the issue, we fall into this case here, not exactly sure why

https://github.com/typeddjango/django-stubs/blob/fab231f0c94b30b00f0c611addc60907e1061928/mypy_django_plugin/transformers/querysets.py#L338-L341

sobolevn: Yes, let's add an issue for that.

sobolevn (approved):

emmanuel-ferdman (approved):
Great stuff!

UnknownPlatypus:
Thanks for the review both! I'll add en entry to the Q&A in the readme so that we can properly document how to make this work (with the currently required workaround) and wrap this up.

I think it was my last blocker before the Manager generic on queryset refacto, funny things coming up. Stay tuned

tests/typecheck/managers/querysets/test_annotate.yml
UnknownPlatypus: Did some changes in the last commits and it's working now, can you check @sobolevn ?
I've added an exemple in the README 

It's behaving well I think, I even had to adjust a test with a valid inaccurate return type

sobolevn (approved):

PR #3268, opened 2026-04-05
Cache apt packages for GDAL in CI

UnknownPlatypus:
# I have made things!
Noticed It was taking a noticeable time in every ci run (~20-25s) so probably worth caching
It improves slightly here (~12-17s)

PS: I also tried https://github.com/marketplace/actions/cache-apt-packages but it's not working properly for us due to gdal triggering a bunch of post install scripts

## AI Policy

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,10 +61,15 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
+      - name: Cache apt packages
+        uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7  # v5
+        with:
+          path: /var/cache/apt/archives
+          key: apt-gdal-${{ runner.os }}
       - name: Setup system dependencies
         run: |
-          sudo apt-get update
-          sudo apt-get install binutils libproj-dev gdal-bin
+          sudo apt-get update || true
+          sudo apt-get install -y --no-install-recommends libgdal-dev libproj-dev
       - name: Install uv
         uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78  # v7
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


UnknownPlatypus:
I think this might help we the flakyness we often have recently when installing gdal via apt for ex [here](https://github.com/typeddjango/django-stubs/actions/runs/24106461156/job/70330801229?pr=3275)

emmanuel-ferdman (commented):
Thanks for working on this! My only concern is pulling in an unknown third-party action (repo with 4 stars) that nobody in the broader ecosystem seems to use. Is there a more "official" caching action we can use here? I'm with `actions/cache` but it might require additional commands since it generic. If there isn't, maybe we should convert to the `ghcr.io/osgeo/gdal` container for the GIS-needing jobs (upstream-maintained, no apt at all). Also, are we sure we need GDAL in all 4 cases? And if none of that works out, we could always just wrap the install in a bash retry loop - ugly but works :)

sobolevn (commented):
I agree with @emmanuel-ferdman 
The effect is not quite worth the potential risk.

However, thanks a lot for the idea and your work! 👍

UnknownPlatypus:
> My only concern is pulling in an unknown third-party action (repo with 4 stars) that nobody in the broader ecosystem seems to use. Is there a more "official" caching action we can use here?

We could just use the official cache action with something like that. I found a [bunch of exemple](https://sourcegraph.com/search?q=context:global+lang:YAML+%22Cache+apt+packages%22&patternType=keyword&sm=0):

```yaml
  - name: Cache apt packages                                                                                                                                    
    uses: actions/cache@v5                                   
    with:                                                    
      path: /var/cache/apt/archives                       
      key: apt-gdal-${{ runner.os }}                                                                                                                            
  - name: Setup system dependencies
    run: |                                                                                                                                                      
      sudo apt-get update || true                        
      sudo apt-get install -y --no-install-recommends libgdal-dev libproj-dev
```

> Also, are we sure we need GDAL in all 4 cases?

I think stubtest and test-ext don't need it indeed, I can try that

sobolevn:
This is much better! 👍

sobolevn (approved):

PR #3269, opened 2026-04-05
Treat ty warnings as errors and migrate typechecker configs to pyproject.toml tool sections

UnknownPlatypus:
# I have made things!

These warnings were passing ci silently, I added a config to escalate them to actual error + fixed the error (unsued-ignoes)

## AI Policy


pyproject.toml
```diff
@@ -99,6 +99,16 @@ build-backend = "uv_build"
 module-name = ["django-stubs", "mypy_django_plugin"]
 module-root = ""
 
+[tool.ty.terminal]
```
sobolevn, line 125: I thought that we are using different files for different tools for better isolation or something 🤔

UnknownPlatypus: Oh really? Looking at the pr introducing ty I don't see mention of why it used a ty.toml

I don't have strong feelings about that but tend to prefer to keep toml based co fig in one place when possible, the [tool.ty] table is enough separation I think

UnknownPlatypus: Looking at it a bit more, I don"t think there was a real reason to having multiple files.

So I've migrated every config to the root pyproject, following modern python practices.

I've also removed the pyright job [we ignored errors from](https://github.com/typeddjango/django-stubs/actions/runs/24007623408/job/70013859685?pr=3266)  to only keep the run on the `tests/assert_type` directory

sobolevn (approved):

PR #3271, opened 2026-04-06
Fix row type lost when using `.annotate(...)` on custom bound `QuerySet`

### LINKED ISSUE #1049: QuerySet annotations are lost when getting an instance in the queryset (bug with WithAnnotations)
# Bug report

## What's wrong

Even though a queryset is correctly typed with `WithAnnotations`, this annotation is lost when accessing an instance in the queryset, which means we can't use this typing feature properly. Here is a minimal example:

```py
class SomeQuerySet(QuerySet[Accommodation]):
    pass


queryset = SomeQuerySet().annotate(hello=Value(True)).filter(is_on_shelf=True).annotate(hi=Value(False))
reveal_type(queryset) # CORRECT: Revealed type is "SomeQuerySet[WithAnnotations[Accommodation, TypedDict({'hello': Any, 'hi': Any})]]"
reveal_type(queryset.first()) # INCORRECT: Revealed type is "Union[Accommodation, None]"
reveal_type(queryset.get(pk=1)) # INCORRECT: Revealed type is "Accommodation"
reveal_type(queryset[0]) # INCORRECT: Revealed type is "Accommodation"
```
## How should it be instead:

Revealed types should be:

```py
reveal_type(queryset.first()) # Revealed type is "Union[WithAnnotations[Accommodation, TypedDict({'hello': Any, 'hi': Any})], None]"
reveal_type(queryset.get(pk=1)) # Revealed type is "WithAnnotations[Accommodation, TypedDict({'hello': Any, 'hi': Any})]"
reveal_type(queryset[0]) # Revealed type is "WithAnnotations[Accommodation, TypedDict({'hello': Any, 'hi': Any})]"

```

## System information

- OS: Mac Os Monterey
- `python` version: 10.0.3
- `django` version: 4.0.4
- `mypy` version: 0.961
- `django-stubs` version: current `master` (just after 1.12.0)

UnknownPlatypus:
# I have made thing

The annotated types are lost when we use `.annotate` on a queryset like this

```python
class SomeQuerySet(QuerySet[Accommodation]): # SomeQuerySet is not generic, and bind the queryset typevar
    pass
```

The root issue is exactly here in `extract_proper_type_queryset_annotate` because in our case, `default_return_type` is `SomeQuerySet(models.QuerySet["Accommodation"])` which is not generic. So we try to add updated typevars to `SomeQuerySet` instead of the child `QuerySet`.
https://github.com/typeddjango/django-stubs/blob/61722fdcf4dab387819db406920f099c4464b33f/mypy_django_plugin/transformers/querysets.py#L358 

The fix here is to ensure we patch the correct TypeVar bound `Instance`. 

## Related issues
Fixes #1049
Blocking #3266

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -58,7 +58,7 @@ if TYPE_CHECKING:
 def determine_proper_manager_type(ctx: FunctionContext) -> MypyType:
     """
     Fill the manager TypeVar on instantiation using the enclosing class.
-    For ex:
+    For example:
 
         class MyModel(Model):
             objects = MyManager()
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -332,7 +332,7 @@ def reparametrize_queryset(instance: Instance, args: list[MypyType]) -> Instance
         if base.type.has_base(fullnames.QUERYSET_CLASS_FULLNAME):
             instance.type.bases[i] = base.copy_modified(args=args)
 
-    return instance.copy_modified(args=args)
+    return instance
 
 
 def extract_proper_type_queryset_annotate(ctx: MethodContext, django_context: DjangoContext) -> MypyType:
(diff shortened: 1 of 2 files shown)


mypy_django_plugin/transformers/querysets.py
```diff
@@ -56,6 +56,18 @@
 
 
 def determine_proper_manager_type(ctx: FunctionContext) -> MypyType:
+    """
+    Fill the manager TypeVar on instantiation using the enclosing class.
+    For ex:
```
```suggestion
    For example:
```

sobolevn (approved):

PR #3275, opened 2026-04-07
Make LazyObject generic to improve __deepcopy__ return type

ahmedasar00:
`LazyObject.__deepcopy__` returned `Any` because `LazyObject` wasn't generic.

Django's `__deepcopy__` can return either a new `LazyObject` (if uninitialized)
or the wrapped object — so the correct type is `Self | _T`.

This PR makes `LazyObject` generic and updates all subclasses accordingly.

Resolves the TODO added in #1575.


UnknownPlatypus (commented):
Looking good overall but could you add a test with real world usage in the tests/assert_type folder ?
You can probably find exemple in Django itself, at least in there test suite

ahmedasar00:
Thanks for the feedback! I've updated the tests to use the Django `User` model for a more realistic scenario, including type assertions for `copy, `deepcopy`, and properties. Let me know what you think

tests/assert_type/utils/test_functional.py
```diff
@@ -71,7 +72,27 @@ def bar(content: Promise) -> None: ...
 bar(s)
 
 
-# classproperty usage
+# SimpleLazyObject usage (Real-world example with Django User)
+
+
+def get_user() -> User:
+    return User()
+
```
UnknownPlatypus: This does not test the feature at stake and does pretty much nothing

tests/assert_type/utils/test_functional.py
```diff
@@ -71,7 +72,27 @@ def bar(content: Promise) -> None: ...
...
+lazy_user = SimpleLazyObject(get_user)
+assert_type(lazy_user, SimpleLazyObject[User])
+
+# Test that copying returns either the lazy proxy or the resolved User
+copied = lazy_user.__copy__()
+assert_type(copied, SimpleLazyObject[User] | User)
+
+# Test deepcopy behavior for lazy objects
+deepcopied = lazy_user.__deepcopy__({})
+assert_type(deepcopied, SimpleLazyObject[User] | User)
```
UnknownPlatypus: this is usually not how we use these dunder method, they are invoked automatically when using `copy.copy` and `copy.deepcopy`

tests/assert_type/utils/test_functional.py
```diff
@@ -71,7 +72,27 @@ def bar(content: Promise) -> None: ...
...
+
+
+# classproperty usage (Keep this as is, it's perfect)
```
UnknownPlatypus: This comment does not help clarifying incertainty or complex behavior so it's better to just remove it

tests/assert_type/utils/test_functional.py
```diff
@@ -71,7 +72,27 @@ def bar(content: Promise) -> None: ...
...
+
+def get_user() -> User:
+    return User()
+
+
+# This simulates how request.user is handled in Django
+lazy_user = SimpleLazyObject(get_user)
+assert_type(lazy_user, SimpleLazyObject[User])
+
+# Test that copying returns either the lazy proxy or the resolved User
```
UnknownPlatypus:
```suggestion

```
Same this one is not needed

tests/assert_type/utils/test_functional.py
```diff
@@ -71,7 +72,27 @@ def bar(content: Promise) -> None: ...
...
+copied = lazy_user.__copy__()
+assert_type(copied, SimpleLazyObject[User] | User)
+
+# Test deepcopy behavior for lazy objects
```
UnknownPlatypus:
```suggestion
```
And this one

UnknownPlatypus (changes requested):

ext/django_stubs_ext/patch.py
```diff
@@ -94,6 +94,8 @@ def __repr__(self) -> str:
     MPGeneric(ForeignKey, (4, 1)),
     MPGeneric(RawQuerySet),
     MPGeneric(classproperty),
+    MPGeneric(LazyObject),
+    MPGeneric(SimpleLazyObject),
```
UnknownPlatypus: We probably only need LazyObject since this one inherit from it? Or am i missing something ?

PR #3280, opened 2026-04-08
Use `ignoreDeps` to prevent Renovate from updating Python versions

UnknownPlatypus:
## Summary

The `allowedVersions: "3.10"` rule was not working — Renovate still created PRs to bump Python (see #3279). Use `ignoreDeps: ["python"]` instead, as recommended by Renovate [here](https://github.com/typeddjango/django-stubs/pull/3279#issuecomment-4208779615).


sobolevn (approved):
Let's try it!

PR #3296, opened 2026-04-12
Type admin site as DefaultAdminSite inheriting from Adminsite

ahmedasar00:
# I have made things!

## Related issues
Fixes stubtest error: `django.contrib.admin.site  variable differs from runtime type django.contrib.admin.sites.DefaultAdminSite`

## AI Policy
*  [x] I have read and agree to the [AI Policy](https://github.com/typeddjango/django-stubs/blob/master/.github/AI_POLICY.md), removed any "Co-Authored-By" lines attributing coding agents, and manually reviewed the final result

### DIFF
--- a/django-stubs/contrib/admin/sites.pyi
+++ b/django-stubs/contrib/admin/sites.pyi
@@ -13,7 +13,7 @@ from django.http.request import HttpRequest
 from django.http.response import HttpResponse, HttpResponseBase
 from django.template.response import TemplateResponse
 from django.urls import URLPattern, URLResolver
-from django.utils.functional import LazyObject, _StrOrPromise
+from django.utils.functional import _StrOrPromise
 
 all_sites: WeakSet[AdminSite]
 
@@ -81,6 +81,6 @@ class AdminSite:
     def catch_all_view(self, request: HttpRequest, url: str) -> HttpResponse: ...
     def get_log_entries(self, request: HttpRequest) -> QuerySet[LogEntry]: ...
 
-class DefaultAdminSite(LazyObject[AdminSite]): ...
+class DefaultAdminSite(AdminSite): ...
 
-site: AdminSite
+site: DefaultAdminSite
--- a/django-stubs/contrib/admin/sites.pyi
+++ b/django-stubs/contrib/admin/sites.pyi
@@ -13,7 +13,7 @@ from django.http.request import HttpRequest
 from django.http.response import HttpResponse, HttpResponseBase
 from django.template.response import TemplateResponse
 from django.urls import URLPattern, URLResolver
-from django.utils.functional import _StrOrPromise
+from django.utils.functional import LazyObject, _StrOrPromise
 
 all_sites: WeakSet[AdminSite]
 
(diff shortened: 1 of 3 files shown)


django-stubs/contrib/admin/sites.pyi
```diff
@@ -81,6 +81,6 @@ class AdminSite:
     def catch_all_view(self, request: HttpRequest, url: str) -> HttpResponse: ...
     def get_log_entries(self, request: HttpRequest) -> QuerySet[LogEntry]: ...
 
-class DefaultAdminSite(LazyObject[AdminSite]): ...
+class DefaultAdminSite(AdminSite): ...
 
-site: AdminSite
```
UnknownPlatypus, line 86: This one is tricky but I don't think this is a correct fix.

The LazyObject make this `site` behaves like an `AdminSite` object. You can try it yourself 

```
>>> from django.contrib.admin.sites import site
>>> site
AdminSite(name='admin')
```

ahmedasar00: Thanks for the review! I agree it's tricky,  I understand that `DefaultAdminSite`  proxies to `DefaultAdminSite`  via `LazyObject`, not through actual inheritance.!

**What would be the correct approach here?**

UnknownPlatypus: What is the stubtest error exactly ? 

We probably will just have to mive this allowlist entry to the main allowlist.txt with a comment, possibly next to similar entries if there are any

ahmedasar00: error: `django.contrib.admin.site`  variable differs from runtime type 
`django.contrib.admin.sites.DefaultAdminSite`
Stub: in file ` django-stubs/contrib/admin/__init__.pyi:86`
`django.contrib.admin.sites.AdminSite`
Runtime:
`AdminSite(name='admin')`

And the same for `django.contrib.gis.admin.site.`

I can update this PR to move the entry to `allowlist.txt` with a comment instead, if you'd prefer that approach.

UnknownPlatypus: My preference would be to get to the bottom of this, it's either a bug in stubtest or something wrong with how we type LazyFunc or use it here

Should it be that maybe ?

```python
class DefaultAdminSite(AdminSite): ...
site: LazyObject[AdminSite]
```

but `site: DefaultAdminSite` will cause false positives for pretty much every attribute access that are expected on AdminSite but not present here no ?

ahmedasar00: Interesting finding: `type(site)` returns `DefaultAdminSite`, but `site.__class__` returns `AdminSite` because `LazyObject.__getattr__` proxies `__class__` to the wrapped object. That's why `isinstance(site, AdminSite)` is `True` even though `issubclass(DefaultAdminSite, AdminSite)` is `False`.

 So the issue seems to be that stubtest uses `type(site)` and sees `DefaultAdminSite`, while the stub says `AdminSite`. This might be a stubtest limitation, it doesn't account for `LazyObject`'s `__class__` proxying behavior.
 
 Should I investigate how `LazyObject` is typed to see if there's a way to make this work properly?

UnknownPlatypus: > So the issue seems to be that stubtest uses type(site) and sees DefaultAdminSite, while the stub says AdminSite. This might be a stubtest limitation, it doesn't account for LazyObject's __class__ proxying behavior.

@sobolevn Shoul we do something about it ? Is it a stubtest limitation with thes proxy objects ?

UnknownPlatypus: @ahmedasar00 I made changes so that the typing reflect the runtime better, even tho there is a proxy involved.
i took inspiration from `DefaultStorage` that does exactly the same pattern

```
# default_storage is actually an instance of DefaultStorage, but it proxies through to a Storage
django.core.files.storage.default_storage
```

PR #3297, opened 2026-04-12
Propagate Subquery output field type through generics

federicobond:
Make Subquery generic over its output field so that passing an explicit output_field preserves the type through .annotate().

```python
# Before: val is Any
# After:  val is int                                                                                          
sub = Subquery(qs.values("id")[:1], output_field=IntegerField())
val = MyModel.objects.annotate(val=sub).get().val 
```

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/fields.py
+++ b/mypy_django_plugin/transformers/fields.py
@@ -5,7 +5,6 @@ from typing import TYPE_CHECKING, Any, NamedTuple, cast
 from django.core.exceptions import FieldDoesNotExist
 from django.db.models.fields import AutoField, Field
 from django.db.models.fields.related import RelatedField
-from mypy.maptype import map_instance_to_supertype
 from mypy.nodes import AssignmentStmt, NameExpr, TypeInfo
 from mypy.types import AnyType, Instance, NoneType, ProperType, TypeOfAny, UninhabitedType, UnionType, get_proper_type
 from mypy.types import Type as MypyType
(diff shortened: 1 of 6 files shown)


sobolevn (approved):
Looks solid to me, maybe someone else wants to comment as well? :)

@emmanuel-ferdman @UnknownPlatypus

mypy_django_plugin/transformers/fields.py
```diff
@@ -156,9 +155,9 @@ def set_descriptor_types_for_field(
     )
 
     # reconcile set and get types with the base field class
-    base_field_type = next(base for base in default_return_type.type.mro if base.fullname == fullnames.FIELD_FULLNAME)
-    mapped_instance = map_instance_to_supertype(default_return_type, base_field_type)
-    mapped_set_type, mapped_get_type = tuple(get_proper_type(arg) for arg in mapped_instance.args)
+    mapped_types = helpers.get_field_type_args(default_return_type)
+    assert mapped_types is not None
+    mapped_set_type, mapped_get_type = mapped_types
```
UnknownPlatypus: Not a big fan of function returning a tuple or None because we need this kind of destructuring.

Can we maybe introduce a NamedTuple ?

```python
class FieldArgs(NamedTuple):
    get: ProperType
    set: PropertType
```

federicobond: Done! Thanks for the suggestion.

PR #3299, opened 2026-04-12
Fix cross-test mypy cache contamination

UnknownPlatypus:
# I have made things!

Prior to this change, I was having the `models_triple_circular_reference` test always failing when running the full test suite with `pytest -n auto` while it was passing in isolation.

The issue was with pytest-mypy-plugin that was not handling correctly new mypy cache format.
The fix is to pull https://github.com/typeddjango/pytest-mypy-plugins/pull/186

This is slightly slower but is expected since previously we were not doing any cache cleanup because of the bug.
**I can confirm that it fixes `pytest -n auto locally`**

<details><summary>Outdated investigation</summary>
<p>

Use a unique cache directory per test instead of per worker to prevent stale module entries from contaminating subsequent tests that reuse module names like "myapp".

~Marking as draft for now to check if the performance impact is ok or not~
The performance hit is too high (2x slower), I can fix the offending test by renaming the app but it seem brittle. Should something be patched in pytest-mypy-plugin @sobolevn ?
```
  ┌──────────────────────────────┬────────────────┬───────────────┬──────────────┐                                                                              
  │             Job              │ #3296 (before) │ #3299 (after) │    Change    │                                                                              
  ├──────────────────────────────┼────────────────┼───────────────┼──────────────┤                                                                              
  │ test (avg across all shards) │ ~1m55s         │ ~3m35s        │ ~1.9x slower │                   
  ├──────────────────────────────┼────────────────┼───────────────┼──────────────┤                   
  │ test-older-django (avg)      │ ~2m00s         │ ~3m41s        │ ~1.8x slower │                                                                              
  ├──────────────────────────────┼────────────────┼───────────────┼──────────────┤
```

</p>
</details> 

## AI Policy

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ redis = ["redis", "types-redis"]
 tests = [
   # Dev tools:
   "pytest==9.0.3",
-  "pytest-mypy-plugins==4.0.0",
+  "pytest-mypy-plugins==4.0.1",
   "pytest-shard==0.1.2",
   "pytest-xdist==3.8.0",
   # Django deps:
--- a/uv.lock
+++ b/uv.lock
@@ -331,7 +331,7 @@ dev = [
     { name = "pyrefly", specifier = "==0.60.2" },
     { name = "pyright", specifier = "==1.1.408" },
     { name = "pytest", specifier = "==9.0.3" },
-    { name = "pytest-mypy-plugins", specifier = "==4.0.0" },
+    { name = "pytest-mypy-plugins", specifier = "==4.0.1" },
     { name = "pytest-shard", specifier = "==0.1.2" },
     { name = "pytest-xdist", specifier = "==3.8.0" },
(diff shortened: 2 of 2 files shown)


sobolevn:
> Should something be patched in pytest-mypy-plugin?

Do you have any ideas?

UnknownPlatypus:
> Do you have any ideas?

I don't know where it appens but we would need to make sure the temporary fiels crated in tests contain a hash or the test case name

sobolevn:
Using test case name as a folder for temp files - sounds like a good idea! PR is welcome :)

UnknownPlatypus:
The issue was with the new style of mypy cache actually, cf https://github.com/typeddjango/pytest-mypy-plugins/pull/186

UnknownPlatypus:
@sobolevn I've updated the pr description with details, seem to work nicely for me

sobolevn (approved):

PR #3306, opened 2026-04-14
refactor: use f-strings in config.py

01xnikhil:
# I have made things!

Replaced some old .format() calls with f-strings in config.py. Makes the error messages a bit easier to read. Everything else stays the same.

## AI Policy


01xnikhil:
Closing this in favor of #3307 which has a cleaner history.

PR #3307, opened 2026-04-14
refactor: use f-string in config.py

01xnikhil:
# I have made things!
I have updated the django_settings_module check in config.py to use an f-string for the error message.

Why this change is made?
Just a small refactor to make the code a bit more modern and readable. I only changed this specific line and kept the other .format() calls as they are, since those use constants and I didn't want to mess with the existing test strings or the project structure.

## Related issues
Supersedes #3306

## AI Policy

### DIFF
--- a/mypy_django_plugin/config.py
+++ b/mypy_django_plugin/config.py
@@ -102,7 +102,8 @@ class DjangoPluginConfig:
         self.django_settings_module = django_settings_module
 
         if not isinstance(self.django_settings_module, str):
-            toml_exit("invalid 'django_settings_module': the setting must be a string")
+            # Refactored to f-string for better readability
+            toml_exit(f"invalid '{"django_settings_module"}': the setting must be a string")
 
         self.strict_settings = config.get("strict_settings", True)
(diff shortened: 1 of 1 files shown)


sobolevn:
This is not an fstring, this is a regular constant string :)

mypy_django_plugin/config.py
```diff
@@ -102,7 +102,8 @@ def parse_toml_file(self, filepath: Path) -> None:
         self.django_settings_module = django_settings_module
 
         if not isinstance(self.django_settings_module, str):
-            toml_exit("invalid 'django_settings_module': the setting must be a string")
+            # Refactored to f-string for better readability
+            toml_exit(f"invalid '{'django_settings_module'}': the setting must be a string")
```
UnknownPlatypus, line 106: This is not an improvement, the string is static. There is no good reason to use an fstring

01xnikhil:
Thanks for the feedback! I understand your point about the f-string being unnecessary for a static constant. I will focus on more impactful functional changes for my next contributions.

Issue #3224, opened 2026-03-26
Signature of InlineModelAdmin.get_readonly_fields is incompatible with supertype BaseModelAdmin


pai-danielkleinridder:
# Bug report

## What's wrong

The signature of the method `get_readonly_fields` for `InlineModelAdmin` has the wrong type for its second parameter `obj`.

The django documentation also has the following [note](https://docs.djangoproject.com/en/dev/ref/contrib/admin/#django.contrib.admin.InlineModelAdmin.has_delete_permission)

> The `obj` argument passed to `InlineModelAdmin` methods is the parent object being edited or `None` when adding a new parent.

In this example I set the correct type for the parameter `obj`:

```python
class BookInline(admin.StackedInline[Book, Author]):
    model = Book

    def get_readonly_fields(self, request: HttpRequest, obj: Author | None = None):
        return super().get_readonly_fields(request, obj)
```

Running mypy give this error:

```
app/admin.py:10: error: Argument 2 of "get_readonly_fields" is incompatible with supertype "django.contrib.admin.options.BaseModelAdmin"; supertype defines the argument type as "Book | None"  [override]
app/admin.py:10: note: This violates the Liskov substitution principle
app/admin.py:10: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
app/admin.py:12: error: Argument 2 to "get_readonly_fields" of "BaseModelAdmin" has incompatible type "Author | None"; expected "Book | None"  [arg-type]
```

This is also related to #469

## How is that should be

The `obj` parameter of methods for `InlineModelAdmin` should have the type of the parent object instead of the child object.

## System information

- OS: MacOS 26.4
- `python` version: 3.13.5
- `django` version: 6.0.3
- `mypy` version: 1.19.1
- `django-stubs` version: 6.0.1
- `django-stubs-ext` version: 6.0.1

UnknownPlatypus:
Pr is welcome!

PR #3308, opened 2026-04-14
Fix InlineModelAdmin parent-object method signatures

MyGodIsHe:
PR Summary
I fixed the InlineModelAdmin.get_readonly_fields typing issue from #3224 and added regression tests for it. The new test covers StackedInline[Child, Parent] with obj: Parent | None, and I also added a negative case to make sure wrong obj types are still rejected.

I ended up using BaseModelAdmin[Any] in InlineModelAdmin because I couldn’t find a cleaner way to satisfy mypy override checks while keeping compatibility and preserving Django’s inline parent-object semantics. I’m not fully happy with this approach, and I’m still not 100% sure Any can’t leak into some edge paths and affect typing somewhere unexpected. I tried to reduce that risk by explicitly typing the important inline methods and adding tests, but I’d definitely welcome a better alternative.


django-stubs/contrib/admin/options.pyi
```diff
@@ -324,15 +325,36 @@ class InlineModelAdmin(BaseModelAdmin[_ChildModelT], Generic[_ChildModelT, _Pare
...
+    def get_prepopulated_fields(
+        self, request: HttpRequest, obj: _ParentModelT | None = ...
+    ) -> dict[str, Sequence[str]]: ...
+    @override
+    def has_view_or_change_permission(self, request: HttpRequest, obj: _ParentModelT | None = ...) -> bool: ...
+    @override
+    def get_view_on_site_url(self, obj: _ParentModelT | None = ...) -> str | None: ...
+    @property
+    @override
+    def view_on_site(self) -> Callable[[_ParentModelT], str | None] | bool: ...
```
UnknownPlatypus: Are class level `view_on_site = False` still accepted ?

This can be a boolena or a method https://docs.djangoproject.com/en/6.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.view_on_site

django-stubs/contrib/admin/options.pyi
```diff
@@ -296,8 +296,9 @@ class ModelAdmin(BaseModelAdmin[_ModelT]):
 _ChildModelT = TypeVar("_ChildModelT", bound=Model)
 _ParentModelT = TypeVar("_ParentModelT", bound=Model)
 
-class InlineModelAdmin(BaseModelAdmin[_ChildModelT], Generic[_ChildModelT, _ParentModelT]):
+class InlineModelAdmin(BaseModelAdmin[Any], Generic[_ChildModelT, _ParentModelT]):
```
UnknownPlatypus, line 298: Why the change from `_ChildModelT` to `Any` ? 

Can it cause issue with `get_sortable_by` ?

django-stubs/contrib/admin/options.pyi
UnknownPlatypus: I think these two take the child instance, see https://github.com/django/django/blob/50bbf71bbd51616e2ce48785336ca3746fbe5f24/django/contrib/admin/helpers.py#L346-L359

tests/typecheck/contrib/admin/test_options.yml
```diff
@@ -114,6 +114,47 @@
           @override
           def view_on_site(self, obj: "A") -> str:
               return "http://example.org"
+-   case: test_inline_get_readonly_fields_parent_obj
```
UnknownPlatypus: [I've just migrated](https://github.com/typeddjango/django-stubs/pull/3353) the rest of this file test to the `tests/assert_type` folder so that it also checks ty/pyrefly/pyright. 

Could you rebase on master and put these two test case in the file `tests/assert_type/contrib/admin/test_options.py` ?

Thanks !

django-stubs/contrib/admin/options.pyi
MyGodIsHe: Because with `_ChildModelT`, this case is incompatible with the base class signature.

If we keep `class InlineModelAdmin(BaseModelAdmin[_ChildModelT], ...)` then in `BaseModelAdmin`, the method `get_readonly_fields` is typed as `obj: _ChildModelT | None`.

However, for a real Django inline, it needs to be `obj: _ParentModelT | None`.

----

There is another option: introduce an additional type parameter.

```python
_ObjT = TypeVar("_ObjT", bound=Model, default=_ModelT)

class BaseModelAdmin(Generic[_ModelT, _ObjT]): ...

class InlineModelAdmin(
    BaseModelAdmin[_ChildModelT, _ParentModelT],
    Generic[_ChildModelT, _ParentModelT],
): ...
```

This approach is cleaner, but it changes the public API. What do you think?

UnknownPlatypus: > This approach is cleaner

I'm not sure it is, the notion of Child/Parent leak to BaseModelAdmin which does not have to know about it and I think it might be confusing for users to have 2 TypeVar on `ModelAdmin` while it make sense on `InlineModelAdmin`, maybe your approach with Any is correct but it does mean we have to redefine every method that would have used `_ChildModelT` in `BaseModelAdmin` right ?

MyGodIsHe: Yes, that's right.

MyGodIsHe: With the second approach, `_ObjT `has a default `(_ObjT = _ModelT)`, so existing one-parameter usages should continue to work.
The model becomes more complex, but it matches Django’s real semantics: `InlineModelAdmin` is child-model based, while `obj` in many hooks is parent-model based.

UnknownPlatypus: still feels off to me to declare a typevar in `BaseModelAdmin` that we don't use in the class directly.
I think the version with `Any` is better, just need to make sure we redefine every method with the new typevars

MyGodIsHe:
Rebased on master, here's what changed:
1. Migrated test cases from YAML to `tests/assert_type/contrib/admin/test_options.py`.
2. Changed `view_on_site` from `@property` to a plain class attribute.
3. Fixed `get_view_on_site_url` and `view_on_site` in `InlineModelAdmin` to use `_ChildModelT` instead of `_ParentModelT`.
4. Verified all `BaseModelAdmin` methods with `obj: _ModelT` are properly overridden in `InlineModelAdmin`.

UnknownPlatypus (approved):

UnknownPlatypus (approved):

PR #3311, opened 2026-04-16
Make Func generic over output_field

### LINKED ISSUE #3310: Incorrect inferred type for queryset annotations using django.db.models.functions.Substr
# Bug report

## What's wrong

I have a model with a BinaryField. In a queryset, I use `.annotate(fragment=Substr(...))` to retrieve just a fragment of the BinaryField. mypy infers the type of this field as `str`, but in reality it is `bytes`. It seems that Substr is assumed to only work on CharField fields, but in practice it works on both CharField and BinaryField fields.

Demo:

```python
from django.db import models
from django.db.models.functions import Substr


class File(models.Model):
    contents = models.BinaryField()


def experiment() -> None:
    q = File.objects.annotate(preview=Substr("contents", 1, 100))

    for f in q:
        reveal_type(f.preview)
        print(type(f.preview))
```

* mypy says: `Revealed type is "str"`
* if I comment out the `reveal_type` line and run the function, it prints: 

```
>>> experiment()
<class 'bytes'>
```

## How is that should be

Ideally the inferred type should depend on whether Substr is applied on CharField or BinaryField, but not sure if that is feasible.

## System information

- OS: Ubuntu 24.04
- `python` version: 3.12.3
- `django` version: 6.0.4
- `mypy` version: 1.20.1
- `django-stubs` version: 6.0.2
- `django-stubs-ext` version: 6.0.2

federicobond:
Substr, Left, and Right declare `output_field = CharField()` but return bytes when applied to a `BinaryField`. This change makes Func generic over _OutputField so that passing `output_field=BinaryField()` correctly infers the annotation type as bytes instead of str.

Fixes https://github.com/typeddjango/django-stubs/issues/3310 as long as users explicitly declare `output_field=BinaryField()` in the Substr call.

## AI Policy

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -158,6 +158,9 @@ class NewSemanalDjangoPlugin(Plugin):
             if info.has_base(fullnames.PREFETCH_CLASS_FULLNAME):
                 return partial(querysets.specialize_prefetch_type, django_context=self.django_context)
 
+            if info.has_base(fullnames.FUNC_EXPRESSION_FULLNAME):
+                return querysets.reparameterize_func_output_field
+
         return None
 
     @cached_property
--- a/django-stubs/db/models/expressions.pyi
+++ b/django-stubs/db/models/expressions.pyi
@@ -187,7 +187,7 @@ class OuterRef(F):
     def __init__(self, name: str | OuterRef) -> None: ...
     def relabeled_clone(self, relabels: Any) -> Self: ...
 
-class Func(SQLiteNumericMixin, Expression):
+class Func(SQLiteNumericMixin, Expression, Generic[_OutputField]):
     @cached_property
     @override
     def allowed_default(self) -> bool: ...  # type: ignore[override]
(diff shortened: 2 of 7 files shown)


mypy_django_plugin/main.py
```diff
@@ -158,6 +158,9 @@ def get_function_hook(self, fullname: str) -> Callable[[FunctionContext], MypyTy
             if info.has_base(fullnames.PREFETCH_CLASS_FULLNAME):
                 return partial(querysets.specialize_prefetch_type, django_context=self.django_context)
 
+            if info.has_base(fullnames.FUNC_EXPRESSION_FULLNAME):
+                return querysets.reparameterize_func_output_field
```
federicobond, line 162: It's not exactly the cleanest solution but I could not find a better way to fix the missing parametrization.

UnknownPlatypus: could `kwargs: object` solve the unconstraint issue ?

federicobond: `object` doesn't help either :(

UnknownPlatypus: This is a bummer to have so much work to do for that :'(. But I don’t have a lot of ideas right now. Maybe some llm could explore other ideas?

Just to be sure, this is required especially for the annotate use case right ?

federicobond: Yeah, this is required because Mypy apparently avoids resolving the typevar when the target is a kwarg typed as Any (the inline call case).

If the function call is defined in a variable and then used in the annotate call Mypy correctly resolves the full type of the Func instance.

UnknownPlatypus: Alright, thanks for the deep investigation!

PR #3314, opened 2026-04-17
Fix type checking of `__in` lookup

### LINKED ISSUE #3313: Type for the `__in` lookup doesn't wrap with `Iterable[...]`
# Bug report

## What's wrong

```python
Calendar.objects.filter(month__in=[1, 2, 3])
```

```
error: Incompatible type for lookup 'month__in': (got "list[int]", expected "int")  [misc]
```

This also seems to be incorrect for filtering on annotated fields.

## How is that should be

The type inferred for a field for the `__in` lookup should wrap the field or annotation type in `Iterable[...]`.

## System information

- OS: Linux
- `python` version: 3.14.4
- `django` version: 6.0.4
- `mypy` version: 1.20.1
- `django-stubs` version: 6.0.1
- `django-stubs-ext` version: 6.0.1

ngnpope:
# I have made things!

The base type was being used instead of expecting an iterable, leading to errors such as:

```
error: Incompatible type for lookup 'month__in': (got "list[int]", expected "int")  [misc]
```

This fixes for lookups on both standard fields and annotations.

## Related issues

Fixes #3313

## AI Policy


django-stubs/db/models/lookups.pyi
```diff
@@ -105,7 +105,7 @@ class IntegerGreaterThanOrEqual(IntegerFieldOverflow, IntegerFieldFloatRounding,
 class IntegerLessThan(IntegerFieldOverflow, IntegerFieldFloatRounding, LessThan[int | float]): ...
 class IntegerLessThanOrEqual(IntegerFieldOverflow, LessThanOrEqual[int | float]): ...
 
-class In(FieldGetDbPrepValueIterableMixin, BuiltinLookup[_T]):
+class In(FieldGetDbPrepValueIterableMixin, BuiltinLookup[Iterable[_T]]):
```
ngnpope, line 108: I wasn't sure whether we wanted to do this... 🤔

I also wonder whether this should be something more like the following as `__in` will accept a `QuerySet`:


```suggestion
class In(FieldGetDbPrepValueIterableMixin, BuiltinLookup[Iterable[_T] | QuerySet[_M]]):
```

Not sure whether we could implement anything in the plugin to narrow `_M` beyond `models.Model`...

See https://docs.djangoproject.com/en/6.0/ref/models/querysets/#in

mypy_django_plugin/django/context.py
```diff
@@ -563,9 +565,13 @@ def resolve_lookup_expected_type(
                 # unknown lookup
                 return AnyType(TypeOfAny.explicit)
 
-        if lookup_cls is None or isinstance(lookup_cls, Exact):
+        if lookup_cls is None or issubclass(lookup_cls, Exact):
             return self.get_field_lookup_exact_type(helpers.get_typechecker_api(ctx), field)
 
+        if issubclass(lookup_cls, In):
+            exact_type = self.get_field_lookup_exact_type(helpers.get_typechecker_api(ctx), field)
+            return ctx.api.named_generic_type("typing.Iterable", [exact_type])
```
ngnpope, line 573: It's a bit annoying to duplicate the call to `get_field_lookup_exact_type()`[^1], but the alternatives seemed to be a lot more messy 🙁

I'm using `typing.Iterable` rather than `collections.abc.Iterable` as I'm mirroring what I've seen in mypy's source.

[^1]: Maybe this also needs a better name now we're using it for something else?

mypy_django_plugin/django/context.py
```diff
@@ -514,19 +514,21 @@ def _resolve_annotated_field_lookup(
...
-                lookup_type = self._resolve_lookup_type_from_lookup_class(ctx, lookup_cls)
-                if lookup_type is not None:
-                    return lookup_type
+        if lookup_cls is None:
+            return annotation_type
 
-        # No lookup suffix or Field-dependent lookup: fall back to annotation type
-        return model_instance.extra_attrs.attrs[lookup_base_field]
+        if annotation_type is not None and issubclass(lookup_cls, In):
+            return ctx.api.named_generic_type("typing.Iterable", [annotation_type])
```
ngnpope, line 528: This is a bit more interesting. We don't want to use the value returned from `_resolve_lookup_type_from_lookup_class()` because it doesn't give good results. For `User.objects.filter(age__in=(1, 2, 3))` we'll get `tuple[Literal[1]?, Literal[2]?, Literal[3]?]` for the RHS, but for `[1, 2, 3]` we'll always get `list[Any]` even though it's passed as a literal which results in the false positive being replaced by false negatives.

tests/typecheck/managers/querysets/test_filter.yml
```diff
@@ -88,8 +88,21 @@
...
+        # The `__in` lookup expects an iterable of the field type.
+        # Regression test for https://github.com/typeddjango/django-stubs/issue/xxxx
+        User.objects.filter(age__in=[1, 2, 3])
+        User.objects.filter(age__in=(1, 2, 3))
+        User.objects.filter(age__in=range(4))
+        User.objects.filter(age__in=123)  # E: Incompatible type for lookup 'age__in': (got "int", expected "Iterable[str | int]")  [misc]
+        User.objects.filter(age__in=None)  # E: Incompatible type for lookup 'age__in': (got "None", expected "Iterable[str | int]")  [misc]
+        User.objects.filter(age__in="abc")  # Doesn't error because Django allows `Iterable[str]` here so `str` is valid.
+        User.objects.filter(age__in=[User()])  # Doesn't error because Django turns this into a list of PKs.
+        User.objects.filter(age__in=(User(),))  # E: Incompatible type for lookup 'age__in': (got "tuple[User]", expected "Iterable[str | int]")  [misc]
```
sobolevn, line 104: question: why do not tuples work?

ngnpope: This is correct - the issue is actually that `list[User]` shouldn't work [^1]. I've rephrased the comments for `[User()]` and added another example with an explicitly annotated variable that works as expected. The issue is that mypy seems to resolve the provided type to `list[Any]` in the `MethodContext` accessed [here](https://github.com/typeddjango/django-stubs/blob/5a3a4cd768b40d77cb9ac1e814484aef5ed9c7b7/mypy_django_plugin/transformers/orm_lookups.py#L27).

Not sure if this is a bug in mypy given we're passing `[User()]` directly and not by reference - surely it'd be reasonable to assume this is `list[User]`? 🤔 

[^1]: I believe that Django sort of supports this by automatically extracting the `pk`, but users can just `[user.pk for user in users]` if we're more strict here.

sobolevn (approved):
Great stuff!

PR #3317, opened 2026-04-18
Use PEP 696 for model fields

### LINKED ISSUE #766: `MyField(models.CharField)` not handled as `str` (but `Any`)
Given `t_derived_model.py`:
```python
from django.db import models


class MyField(models.CharField):
    pass


class MyModel(models.Model):
    myfield = MyField()
    field = models.CharField()


m = MyModel()
reveal_type(m.field)
reveal_type(m.myfield)
```

`mypy t/t_derived_modelfield.py` results in:
```
t/t_derived_modelfield.py:14: note: Revealed type is 'builtins.str*'
t/t_derived_modelfield.py:15: note: Revealed type is 'Any'
```

I.e. the type of `MyField` is not handled like when using `modfels.CharField` directly.

## System information

- OS: Arch Linux
- `python` version: 3.9.9
- `django` version: 3.2
- `mypy` version: 0.910
- `django-stubs` version: 1.9.0-22-ge5361f1 (current master)
- `django-stubs-ext` version: -

### LINKED ISSUE #1264: Feature: Use PEP 696 defaults for fields

### LINKED ISSUE #2043: Custom fields with overwritten passthrough constructors don't change types based on attributes

### LINKED ISSUE #2724: error: DateTimeField is nullable but its generic get type parameter is not optional

UnknownPlatypus:
# I have made things!

Replace `_pyi_private_set_type` / `_pyi_private_get_type` attributes on field stubs with [PEP 696](https://peps.python.org/pep-0696/) `TypeVar` defaults. This makes field set/get types visible to all type checkers, not just mypy with the plugin.

This is a big PR, sry for that but the plugin code is so entangled I did not manage to make a smaller chunk.
I've rebased to make it a bit more digestible, hopefully that helps.

I'm pretty convinced we can improve the stubs of `ForeignKey`/`ManyToMany` and friends using trick similar to what is done to `ArrayField` in this PR and remove even more plugin code. 
Most of the code in `transform_into_proper_return_type` should go I think.

## Related issues
Fixes #766
Fixes #1264
Fixes #1900
Fixes #2043
Fixes #2590
Fixes #2724

Improves #579

<details><summary><b>Handling <code>null=True</code></b></summary>
<p>

The important part is to make `null` working (and later `primary_key`). Two approches have emerged in the other issue discussion, I've investigated further with various type checker and **was able to make the `_NT` typevar approach work with every typechecker I know off!**

### Using `__new__` overloads (`django-types` is doing that)

<details><summary>Code snippet</summary>
<p>

```python
from typing import TypeVar, Literal, Generic, overload, reveal_type


_ST = TypeVar("_ST")
_GT = TypeVar("_GT")

class Model:
    pass


class Field(Generic[_ST, _GT]):
    def __init__(self, *, null: bool = False) -> None:
        super().__init__()


_ST_1 = TypeVar("_ST_1", default=int)
_GT_1 = TypeVar("_GT_1", default=int)


class IntegerFieldSimple(Field[_ST_1, _GT_1]): ...

class IntegerFieldWithOverloads(Field[_ST_1, _GT_1]):
    @overload
    def __new__(
        cls, *, null: Literal[True]
    ) -> IntegerFieldWithOverloads[_ST_1 | None, _GT_1 | None]: ...

    @overload
    def __new__(
        cls, *, null: Literal[False] = False
    ) -> IntegerFieldWithOverloads[_ST_1, _GT_1]: ...

    def __new__(   # type: ignore[misc]
        cls, *, null: bool = False
    ) -> (
        IntegerFieldWithOverloads[_ST_1, _GT_1]
        | IntegerFieldWithOverloads[_ST_1 | None, _GT_1 | None]
    ):
        return super().__new__(cls)


class A(Model):
    field1 = IntegerFieldSimple(null=False)
    field2 = IntegerFieldSimple(null=True)
    field3 = IntegerFieldWithOverloads(null=False)
    field4 = IntegerFieldWithOverloads(null=True)

a = A()
reveal_type(a.field1) # Revealed type is "builtins.int"
reveal_type(a.field2) # Revealed type is "Union[builtins.int, None]"
reveal_type(a.field3) # Revealed type is "builtins.int"
reveal_type(a.field4) # Revealed type is "Union[builtins.int, None]"
```

</p>
</details> 

Works with:
- [ty](https://play.ty.dev/24a0c6aa-6080-4c84-aee7-b0095cbbefff) ->  🟠  Work with explicit `__new__` overrides on every custom field (and need to be quoted)
- [pyright](https://pyright-play.net/?strict=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgGoCGIANFADJIzEjkA2NA4sSk0gMY1gAbkxZhyAExohiw1gH14CYgFgAUGrVyAygSgBeQiQogAFACJtBMwEpN7XQaJLj5ufZsbVvFuQDOvqABZMHFiFgAuNShoqAQ-X081b3ioADEkMPETTm4QPgBtSxo3AgBda0jVGKhQ4Cg5OVQGBpNfMOAaACoaFABXFgioACMwMBZ9NNY26ygAWgA%2BKAA5MG5K6urfXqVTawA6BqaYFtt1M8s5AEYJpzJKVx0rsxpa8n6YPVQYU5Krm6N7hZ7E8XsRgG8WB8vqdPMl-FAAJIoRhoJjpTJaLAIFjEEzoljiQqPS7FYGXcrhKB7amwnzwpEotEZAkAdQYAAsAPLCECiCS%2BPHMwkXEn1MkUqIxAACQhEYnEkuitXqcm4AHcWoqNt5fF0ev1BvRGMwWPkCCBesRSlqZgtEcjiKiQPjxGyYFyeXzxL4iQQ-gAfZarYikv3XQMrbilSnUvaeaoyz3yrXKhrqzVVDbRHV6qB9AaUo1MVj5VJTK0TMssNo2uaLBmOpmZN0euX831XUNXaNUmlnaqp1XEDVyEwxADEsBIlKQaBQuGI%2BQgSF8vGtmazOag3TzBspIzGlfLtbtJi11QbTpdLe5be9HdFv3J55igcvTdZHNvvPlPpFUAjYMu3DIMoxtdYsygaQYF6EAUCgLYdhMfY02HFodRhM44QCABBExglCFgKi1YAhWuAx32dIVMWwHETHzFg9CraYSKFAAmCZKJdGjsVxBi9HNS1TmqUjMgAZk4h0ryFG8k35eiDSY8thJiUSCQAFkkxkqObL85O9BSBgEi1iEw8gJjw05pFkFgFBIExyD2NTxEuGZJwAJRkYhWGIcQpyUTAAjMIZeiQSFUF8PYvjMNRrO82zFFxRznLYtyoE8mzfP84hAqgMwAFUUCQVZ8hCsKYAiqLkRoSMrRi1Q4vkRKHKcoUxLSjL4qyxLcuC0LwpQSLotirymvs5KhXUjrRpxPyepXPLCuKlBSv6irBqqmAauDUozCAA) -> 🟠  Work with explicit `__new__` overrides on every custom field
Not with:
- [pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVZjABqqSgBo6AGQgMYlVFCkBxGOgUQAxlNwA3BVFypMUyjANKA%2Bk2IwAOukeOrAZQF0AvIOFjKACnsQNwEggEoXFQ9vITs-QOCo8Od0LShUODg6AFlcTBgoREc6ErpiDLgUxzSKugAxCALMfzUNSm0AbRCpKyiAXTCi9FK6fLA6KytWOUn-OAKwKQAqKXQAVyhCujxcKC96pXmwugBaAD46ADlcDSGRkbg1uwCwwknphlmIpx%2BQqwBGfaxUTiBJ-f5BKRjVAbBieVgMb69AQAoG%2BUFBZEAyGjGBgGFQOEI74pGqZOgASXQ8jYCgaTVcXGIsH89KgmC67gBPSiAIGiDohCFpPS5KpNLpjXZAHU5AALADyBkoRhMcFZUo54J5KP%2B-OKpQAAvpDMZMAaSmMJlYNAB3WYW%2B5pODLVYbLayeSKKAdASUNYwPqO47nSnUmC0yhszCyhiK5WqzBwTm6ugAHyuNxgOtRGeuGj6AqFhBSI2NCbNjqtkztDuG9xKztddHWmwFnoUSg6dUOgf2Pag82Dpwu4ojkqasfjprVKe5E15eqLwp%2BI2rNpg9qs-lKAGJGMIBRA2OheDAOjQIHAtEH6w2m3QVi33QKdntvAOh3e6CGLoFvyMY6RtGU5KjOSZzv8OZ6o6IwZkBE4yvKYEqmaybgummYaNBmH5oGwZ3A2dDmAwayUMMjzPP4rw1puszOiSPxklkACC-i5PkUCDI6kBNIC3gIVGmqMtwLKtlAnifjA3wjLx7IAEz7IJ0YicyMD%2BOJnh%2BgGMmlHJmAAMxKeGwGaqBFZqhp7qSb2uklPpAAsxkSkJk7IRZSZWZsWn%2BtJKSoPsbHfOYlhQDYwj%2BKghD6f8xz7gAShYMBKDAmAHnYnBZEE2BrNADCsHAhAIkEjghclYW2OpUX6fJcV0IloWpelMCZXQQQAKroBANwdDleUFUV1JSHhQYgKVSXWJVkXRZqBl1Q15VNZVrXZblhIDcVY3oGVk0RdVmoOfNE2wGly1Xm1ICdd16C9Wt%2BXoIVCLDVmo0gBIIBkOYYBQKQhAMLQUAUPuAAKpBfT9dBoFgeD4HQWg3JAbBkag%2BU3CW6D7q4MAtXKDAMMQcCIAA9ETn14j9hC8GwRPqETmC4FocBE-D6CI8jqPoETdBgLwdCoHoqDQKg2CwHDCPHuz110LgxAc4VjhkHGNwnMqcBS94QQGYQ-zySVN0KNQlBwLeaxddwvDyJgJyYBA5haPlBj7AA5Mep7mE7jingwJzmAAjrl5hWwA1jApAnKgWhaPAWTeE7triOgHvoCAAC%2B70Rw7MANLAFBQzgBAkOQKdAA) -> 🔴  Not working
- [mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=1c5922c64b2c9d66cec7945924849082)  -> 🟠 Work with explicit `__new__` overrides on every custom field
- zuban -> 🟠 Work with explicit `__new__` overrides on every custom field, (but a false positive error on the `__new__` definition)

### Using third type variable `_NT` to `Field` encoding nullability as `Literal[True]` or `Literal[False]`

<details><summary>Code snippet</summary>
<p>

```python
from typing import TypeVar, Literal, Generic, overload, Any, reveal_type, Self, Never

_NT = TypeVar("_NT", Literal[True], Literal[False], default=Literal[False])

class Model:
    pass

class Field(Generic[_ST, _GT, _NT]):
    def __init__(self, null: _NT) -> None: ...
    @overload
    def __get__(self: Field[Any, Any, Literal[False]], instance: Any, owner: Any) -> _GT: ...
    @overload
    def __get__(self: Field[Any, Any, Literal[True]], instance: Any, owner: Any) -> _GT | None: ...
    def __get__(self, instance: Any, owner: Any) -> _GT | None:
        pass

_ST_1 = TypeVar("_ST_1", bound=int, default=int)
_GT_1 = TypeVar("_GT_1", bound=int, default=int)

class IntegerFieldSimple(Field[_ST_1, _GT_1, _NT]): ...

class IntegerFieldWithOverload(Field[_ST_1, _GT_1, _NT]):
    @overload
    def __get__(self: IntegerFieldWithOverload[Any, Any, Literal[False]], instance: Any, owner: Any) -> _GT_1: ...
    @overload
    def __get__(self: IntegerFieldWithOverload[Any, Any, Literal[True]], instance: Any, owner: Any) -> _GT_1 | None: ...
    def __get__(self, instance: Any, owner: Any) -> _GT_1 | None:
        pass

class A(Model):
    field1 = IntegerFieldSimple(null=False)
    field2 = IntegerFieldSimple(null=True)
    field3 = IntegerFieldWithOverload(null=False)
    field4 = IntegerFieldWithOverload(null=True)

a = A()
reveal_type(a.field1) # Revealed type is "int"
reveal_type(a.field2) # Revealed type is "int | None"
reveal_type(a.field3) # Revealed type is "int"
reveal_type(a.field4) # Revealed type is "int | None"
```

</p>
</details> 

Works with all type-checkers:
- [pyright](https://pyright-play.net/?pyrightVersion=1.1.353&pythonVersion=3.13&pythonPlatform=Linux&strict=true&locale=en&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgGoCGIANFADJIzEjkA2NA4sSk0gMY1gAbkxZhyAExoBBFHBohiw1gH14CYgCgNygMoEoAXkIkKIABQAiXQQs1eYFDGaDKSco4MEQAV2IBKbXZ9IyJ1U0tlINsoexcQNw8vXwDlADlg4zDKCPTo%2BkZmFgBtJOIAXRp8plYigDFWAGdymnFiYHJvFhgDKsK6xvKAjV4WcgaGqABZMFaWAC4NKCWoBDGGrRG1qFqkYhZxM05ueN4i6xpIggv0sr8F5ahW4ChlZVQGV7MmlmAaFE75i90n4oABaAB8UFSDmIcygADpEYtlgABIQiMTiZFLJ4vZRoYgwT7fYBwnZ7cRFGRyKDUyoMarFeosJplCqYFANGDuXiw2myAQAd2OcOpIIhLyCcMR8OxUDRwhAogkctxrwJROUXz2pO2u32VIF-JpvRqpTZNFQXJ5fLpUDAwqYotk4shlygAB8oTDpUiHmr8YTiTrLZzuShec6aQ6RcbXZL9F7odx7g9lqtxlprMoAIyGTJkbJWPS56IAIzA3hQ4gMqBgLTaHS6tccKSCufzoUL5is7Zz5cr1Zb9cejc63TrQ024ygAElHMQCSByfsdFgECxiGYV5TszmLn3rgRbr7ZcNRjP54wlzuAOoMAAWAHlFcqDjuziX9wnc0eT3KFQxFV-TaPENWDH44SvRcmDvR8XyAyk7TtU0mQGC0OWtCNbSNGMnTjME3T7U8APRJVMVVUD1SDLUSSghcb31cR7xgZ9X0xQ0aRQhk%2BnNdkrXDSNjSFWMxUIn88yTH0ET9ZYA3A2iQ0wwScOjR0QCjeNLg7KSUzlB4M3Wc8tikMxplmO45WAJi8yMaDGIpNdsE3Mx-hYFgDGZJoAgeayKQAJnzezYKYpyNy3NyPNKHzlj8-YAGYgoYkKKRYtjENcgFPIGGKlji8QABYkuvFL9jShDyIkTL3M8Hx-C0ch81MgIFCUFhVBIMxyHhfKcxBABiKAACVFGIVhiHEWASEwCYLDrCwNFasb2rULduvy-yBuG0bxsm1aZqgObHE9b1uAWpaVFWrqeqY%2BKtpGtqJqm9QDqOmBzp2lbOvWpiCvuz6nv2pBZrrE7k2ICwgA) -> 🟢  Works (but false positive on the explicit override case)
---> todo
- [mypy 1.16+](https://mypy-play.net/?mypy=1.16.0&python=3.13&gist=4ca24614e75f284a3e32e3457598b4c0) -> 🟢 Works
- [pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVZjABqqSgBo6AGQgMYlVFCkBxGOgUQAxlNwA3BVFypMUgILpSUyjANKA%2Bk2IwpAZRhQwUgHK2FAHXRA%2B1cBOgBeQWExSgAKfxAQgQSpLVx0BkU9cQgMBnCBSgBXGABKYJUwyKFnGPjEypS6NOzKXIyC4rLg7yqo2vF6%2B16m2XlFKABtQpKAXSkxhSVJgDElOBh5ukwYMFQiqHzFidX1zfKg9C0oVDg4OgBZXB2oREC6D7piW7hAwOufnQVhAPJhYmoNG0tJMklJ7JU4b1ZqU3uhPttdnR7PZWHJsbENp4pOgDq8sb1SnQALQAPjo3nSMEQdEIrPenwAAvpDMZMOyPjswFj7GwYAx8YSwMzgaDJhYrHR5Qs5EspmsoBtZltWHAGBgtEzFZZdAB3SHM%2BWU2lYyrM1mEfl0LkGShGEyOwXC0Xi%2BwEjxSoEgqCYOXGo0K47LGabbXoXX6w1Kui4M0KC2WK10%2BFhAA%2B9MZdrZaM%2Bnux3ol-qkOr1V0TYZT5vDmZtufzGlR6PR3zuf3QSXsAEYIv1RIMEv2B008EUsOFWAwpIL9oc5xkLtnB8OaqO4uPKoOp7gZ5hVwuMXsDvl5xd-jc7nQAJIZGCiygy4OuLjEWCxd8hidwvuA6IgIyKFg6lwAveT7yK%2Bf4AOpyAAFgA8i6bpgn%2BMKhIOgECLh5KgSijrOjy7rFgKmJlmKFaeMyMEvgoCHIWhZEhkmSaRmqZxalWcY1ga6YKg2aZNtSWZAeBJHcq6vIelRIo0b6kr0c%2BcFBpgiEMKh6G8qGCqcSqJzRrxnD8QmQmmo2lriS2m55gy7YskW6KlopPp%2BkSZnxrWlnJqmlBCc2G5Dg5BaOl2Py9lB9xmLETwvMRFF0JAoJDpEDHqaCn7cD%2BJJQFA4TqhsFzoqlwYAEzDplTEaTl34wLE%2BWFdGpWfOVmAAMzVWptWglpOlsU1pJFWcbUfB1AAsPWwX1wYDaxskmMNBWdCUN7oKgw5xRcNh2FAjjCLEqCEB1A6UgAxHQABKfhKDAmCMMInD3Ak84JIEe0wA4TiNSdHUVZdN13bAj2-S9dBvRkdBhRoH3oF9P1Hf9GmdUDt37Q9T3OBDUMMPDiMHb9x2nRpk3oyDWPgxAr0gPOMNtjACQgBIIBkDYYBQKQhAMLQUAUFdAAKpAc1zdBoFgeD4M06SQGwRSKAwEDpBBV3uDAdBIQwDDEHAiAAPT6%2Bzuxc4QvBsPr6j65guBaHA%2BtpOgcsK6gSvpPrKW8HQqDZNAqDYLAMtOxA8uK8raK4MQbtxhBZDaekVIunA4fDgknWEAOFXw5MCjUJQcCzIEM5frw8iYFSmAQDYWhKwYw4AOQh%2BgvAwPXgTNwwVI2AAjkUVcPVSADWMCkFSqBaAa96RPXJriOgbfoCAAC%2BrPj7XMDArAFASzgBAkOQS9AA) -> 🟢 Works
- [ty](https://play.ty.dev/a5f0948d-8613-47b2-a0bf-690be7458f04) -> 🟢 Works (but need to be quoted)
- zuban -> 🟢 Works

</p>
</details> 

<details><summary><b>Custom field typing</b></summary>

<p>

Currently, bare custom field (`class HTMLField(models.TextField): ...`) are an issue because the TypeVar default machinery makes `body_nullable = HTMLField(null=True)` resolve to `TextField[str, str, Literal[False]]`, causing this kind of errors

> Argument "null" to "HTMLField" has incompatible type "Literal[True]"; expected "Literal[False]"

The mypy plugin currently makes this work by reparametrizing HTMLField to remain generic in `_ST_Text`, `_GT_Text`, `_NT`, so mypy's natural inference picks up `_NT = Literal[True]` from the call. But other typecheckers cannot do that so it means that for them, they have to explicitely fix their custom fields
following the recommendation in the README ie

```python
from typing_extensions import TypeVar, Literal
from django.db import models
from django.db.models.expressions import Combinable

_ST_Text = TypeVar("_ST_Text", contravariant=True, default=str | int | Combinable)
_GT_Text = TypeVar("_GT_Text", covariant=True, default=str)
_NT = TypeVar("_NT", Literal[True], Literal[False], default=Literal[False])

class HtmlField(models.TextField[_ST_Text, _GT_Text, _NT]):
    pass
```

<b>That is probably something we can live with because the workaround is documented (and I say workaround, it's more of a regular typing pattern, omitting generics is often an issue and flagged in strict modes) </b>

Alternative I considered

####  Drop `default=Literal[False]` from `_NT`
his Violates PEP 696 ordering (no-default TypeVar can't follow defaulted ones) so we have to either:
    -  :x:  reorder typevars to `Generic[_NT, _ST, _GT]` which would completely change the meaning of existing `Field[X, Y]` annotations, it feels bad 
    - drop the `_ST`/`_GT` defaults we just added and inline the defaults. This would make custom field subclassing anything other than `Field` very difficult as the `_GT`/`_ST` are not reachable anymore. People would have to redeclare the whole `__get__` / `__set__` machinery which seems bad. It also requires that we do `null: _NT = False` to not regress Bare instantiation ( `models.IntegerField()` -> `_NT=Literal[False]`) but this only works with pyright and pyrefly. 
Passing a default value to a TypeVar is not supported by mypy:
- [mypy](https://github.com/python/mypy/issues/3737), we can add a type ignore and the inferrence still works
- [ty](https://github.com/astral-sh/ty/issues/592), even with a atype ignore, the inference break and gives `Unkown`

####  _NT = TypeVar("_NT", bound=bool, default=Any) 

But this breaks narrowing to `_GT` or `_GT | None` based on `null=...` because `Literal[False]` / `Literal[True]` become bool which is a blocker

</p>
</details> 

## Followups
- [ ] Upstream issue for remaining false positives
- [ ] upstream ty issue for `# TODO: ty should reject that too`
- [ ] Support Fk/OneToOne/m2m with minimal plugin work

## TODOS:
- `__init__` overload branch and compare
- Add a test for allowing setting None for field with `db_default`, should still block for only `default` because it crashes at runtime with `django.db.utils.IntegrityError: (1048, "Column 'type' cannot be null")`
```
class Thing(models.Model):
    type = models.CharField(max_length=255, blank=True, default="")

Thing.objects.update_or_create(
    defaults={"type": thing.type},  # type:ignore[misc] # Should be allowed to set none because the field declare `default=`
            )
```


django-stubs/contrib/gis/db/models/fields.pyi
```diff
@@ -18,13 +18,28 @@ from django.db.models.expressions import Combinable, Expression
...
 from django.utils.functional import _StrOrPromise
-from typing_extensions import override
+from typing_extensions import TypeVar, override
 
 # __set__ value type
 _ST = TypeVar("_ST")
 # __get__ return type
 _GT = TypeVar("_GT")
 
+_ST_POINT = TypeVar("_ST_POINT", default=Point | Combinable)
```
sobolevn: Nit: let's fix the naming. It should be `_ST_Point`, otherwise this looks like a constant now.

ahmedasar00:
Great work on this! It’s really impressive.!
 Are there any **specific tasks**  I can help with to move this forward?

UnknownPlatypus:
>  Are there any specific tasks I can help with to move this forward?

Still trying to remove unnecessary plugin code and need to test on my work repo (1.5M Loc) for regressions. I think it breaks some parts that are not tested well enough for now.

Not much to do to help I think but thanks for the proposal. When I'll make this ready for review, testing on external codebases would be valuable 

[edit]: actually, checking [the mypy_primer output](https://github.com/typeddjango/django-stubs/actions/runs/24604288870?pr=3317) would be awesome, the `Need type annotation for` droping are good but there are false positives at the end I think

UnknownPlatypus:
@sobolevn A bunch of issues in `mypy_primer` are because it does not load the plugin, hence the nullable part is bogus and we got false positives we did not have before when it was `Any`. 

I'm gonna be expanding the scope of this pr to do the `Using third type variable _NT to Field encoding nullability as Literal[True] or Literal[False]` I think

ahmedasar00:
> [edit]: actually, checking [the mypy_primer output](https://github.com/typeddjango/django-stubs/actions/runs/24604288870?pr=3317) would be awesome, the `Need type annotation for` droping are good but there are false positives at the end I think

Hello ,
I’ve looked into the `mypy_primer`  output to analyze the new assignment errors in Zulip.
A clear example occurs in `zerver/actions/realm_linkifiers.py` at lines [144](https://github.com/ahmedasar00/zulip/blob/11bf985d1f38c4e65e1b795de7496ede60e0ba8a/zerver/actions/realm_linkifiers.py#L144) and [145](https://github.com/ahmedasar00/zulip/blob/11bf985d1f38c4e65e1b795de7496ede60e0ba8a/zerver/actions/realm_linkifiers.py#L145):
```python
linkifier.example_input = normalize_optional_text(example_input)
linkifier.reverse_template = normalize_optional_text(reverse_template)
```
The error Incompatible types in assignment (expression has type `str | None` , variable has type `str | Combinable)` is happening because `normalize_optional_text() returns` `str | None`. Since my current implementation of PEP 696 defaults to a strict non-nullable `str` for these fields, it flags the assignment.

This confirms that while the base `TypeVar`  logic is working, we definitely need the next step of integrating `null=True`  and `default` logic to allow the type to dynamically include `None` when the field configuration permits it.

---
Since the `mypy_primer` output shows quite a few errors that need thorough investigation, I was wondering where you’d prefer me to share the rest of my analysis?

I can continue posting my findings **here on GitHub for transparency or I can send a detailed breakdown via email** if you’d prefer to keep the PR conversation more focused. Let me know what works best for you!

ngnpope:
> Still trying to remove unnecessary plugin code and need to test on my work repo (1.5M Loc) for regressions. I think it breaks some parts that are not tested well enough for now.
> 
> Not much to do to help I think but thanks for the proposal. When I'll make this ready for review, testing on external codebases would be valuable

We will be able to test out on a 16M LoC codebase at work when you're ready. (FYI @delfick)

delfick:
well, we only run mypy on 5.9 million lines of that (excluding blank lines and comments), but yes, our codebase is often a rich source of edge cases!

ngnpope:
Shh! You make it sound less impressive 😂

ahmedasar00:
> > [edit]: actually, checking [the mypy_primer output](https://github.com/typeddjango/django-stubs/actions/runs/24604288870?pr=3317) would be awesome, the `Need type annotation for` droping are good but there are false positives at the end I think
> 
> Hello , I’ve looked into the `mypy_primer` output to analyze the new assignment errors in Zulip. A clear example occurs in `zerver/actions/realm_linkifiers.py` at lines [144](https://github.com/ahmedasar00/zulip/blob/11bf985d1f38c4e65e1b795de7496ede60e0ba8a/zerver/actions/realm_linkifiers.py#L144) and [145](https://github.com/ahmedasar00/zulip/blob/11bf985d1f38c4e65e1b795de7496ede60e0ba8a/zerver/actions/realm_linkifiers.py#L145):
> 
> ```python
> linkifier.example_input = normalize_optional_text(example_input)
> linkifier.reverse_template = normalize_optional_text(reverse_template)
> ```
> 
> The error Incompatible types in assignment (expression has type `str | None` , variable has type `str | Combinable)` is happening because `normalize_optional_text() returns` `str | None`. Since my current implementation of PEP 696 defaults to a strict non-nullable `str` for these fields, it flags the assignment.
> 
> This confirms that while the base `TypeVar` logic is working, we definitely need the next step of integrating `null=True` and `default` logic to allow the type to dynamically include `None` when the field configuration permits it.
> 
> Since the `mypy_primer` output shows quite a few errors that need thorough investigation, I was wondering where you’d prefer me to share the rest of my analysis?
> 
> I can continue posting my findings **here on GitHub for transparency or I can send a detailed breakdown via email** if you’d prefer to keep the PR conversation more focused. Let me know what works best for you!

Hey @UnknownPlatypus, hope you're doing well! Just wanted to follow up on my offer to share more mypy_primer findings. I've gone through the rest of the output and categorized the false positives,  happy to post the full breakdown here whenever it's useful for you. No rush though, totally understand if you're focused on the implementation right now.

UnknownPlatypus:
cc @delfick, this should be testable now 

thanks @ahmedasar00 for the help on mypy_primer review!

tests/assert_type/db/models/fields/test_nullable.py
```diff
@@ -0,0 +1,19 @@
...
+from typing import Any
+
+from django.db import models
+
+
+def field_null_true_expression_does_not_trigger_nullability_check() -> None:
+    """
+    Field[Any, Any, Any] as function type arg should accept both nullable and non-nullable fields
+
+    TODO: False positive pyrefly
```
sobolevn: Can we please report this issue to the `pyrefly` team and link the issue here?

tests/assert_type/db/models/fields/test_related.py
```diff
@@ -8,8 +8,8 @@ class Author(models.Model):
 
 
 class Book(models.Model):
-    author = models.ForeignKey(Author, on_delete=models.CASCADE, swappable=False)  # pyright: ignore[reportUnknownVariableType]
+    author = models.ForeignKey(Author, on_delete=models.CASCADE, swappable=False)
 
 
 class Profile(models.Model):
-    user = models.OneToOneField(Author, on_delete=models.CASCADE, swappable=False)  # pyright: ignore[reportUnknownVariableType]
+    user = models.OneToOneField(Author, on_delete=models.CASCADE, swappable=False)
```
sobolevn, line 16: Let's add `assert_type` calls here :)

django-stubs/db/models/fields/related.pyi
```diff
@@ -134,14 +140,27 @@ class ForeignObject(RelatedField[_ST, _GT]):
         error_messages: _ErrorMessagesMapping | None = ...,
         db_comment: str | None = ...,
     ) -> None: ...
+    @overload
+    @override
+    def __set__(self: ForeignObject[Any, Any, Literal[False]], instance: Any, value: _ST | Combinable) -> None: ...
```
sobolevn, line 139: Can we add `@type_checking_only` decorator here?

tests/assert_type/db/models/fields/test_nullable.py
UnknownPlatypus: Was planning on doing that as a followup once this lands so that it's easier for them to check using master.

I'll add a followup list to the PR description

tests/assert_type/db/models/fields/test_base.py
```diff
@@ -0,0 +1,292 @@
...
+    """
+
+    class HTMLField(models.TextField): ...
+
+    class IntWrap(models.IntegerField): ...
+
+    class Article(models.Model):
+        body = HTMLField()
+        body_nullable = HTMLField(null=True)  # pyright: ignore[reportArgumentType]  # pyrefly: ignore[bad-argument-type]  # ty: ignore[invalid-argument-type]
+        count_nullable = IntWrap(null=True)  # pyright: ignore[reportArgumentType]  # pyrefly: ignore[bad-argument-type]  # ty: ignore[invalid-argument-type]
```
UnknownPlatypus: Currently, bare custom field are an issue because the TypeVar default machinery makes `body_nullable = HTMLField(null=True)` resolve to `TextField[str, str, Literal[False]]`, causing this kind of errors

> Argument "null" to "HTMLField" has incompatible type "Literal[True]"; expected "Literal[False]"

The mypy plugin currently makes this work by reparametrizing HTMLField to remain generic in `_ST_Text`, `_GT_Text`, `_NT`, so mypy's natural inference picks up `_NT = Literal[True]` from the call. 

But other typecheckers cannot do that so it means that for them, they have to explicitely declare the typevar for their custom fields like in the test above (`nullable_subclass_inherits_null_overload`)

<b>That is probably something we can live with because the workaround is documented (and I say workaround, it's more of a regular typing pattern, omitting generics is often an issue and flagged in strict modes) </b>

I've updated the README to reflect that and guide users in this specifc case.

<details>
<summary>Alternative I considered:</summary>

####  Drop `default=Literal[False]` from `_NT`
This Violates PEP 696 ordering (no-default TypeVar can't follow defaulted ones) so we have to either:
    -  :x:  reorder typevars to `Generic[_NT, _ST, _GT]` which would completely change the meaning of existing `Field[X, Y]` annotations, it feels bad 
    - drop the `_ST`/`_GT` defaults we just added and inline the defaults. This would make custom field subclassing anything other than `Field` very difficult as the `_GT`/`_ST` are not reachable anymore. People would have to redeclare the whole `__get__` / `__set__` machinery which seems bad. It also requires that we do `null: _NT = False` to not regress Bare instantiation ( `models.IntegerField()` -> `_NT=Literal[False]`) but this only works with pyright and pyrefly. 
Passing a default value to a TypeVar is not supported by:
- [mypy](https://github.com/python/mypy/issues/3737)
- [ty](https://github.com/astral-sh/ty/issues/592)

However, mypy still infer the correct thing on usage so a type ignore on `null: _NT = False` does the trick, but it would still be an error for ty.
We can maybe ne consider it later if support is added in ty ? 

####  _NT = TypeVar("_NT", bound=bool, default=Any) 

But this breaks narrowing to `_GT` or `_GT | None` based on `null=...` because `Literal[False]` / `Literal[True]` become bool which is a blocker

</details>

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -46,11 +46,13 @@ _ErrorMessagesMapping: TypeAlias = Mapping[str, _StrOrPromise]
...
-_ST = TypeVar("_ST", contravariant=True)
+_ST = TypeVar("_ST", contravariant=True, default=Any)
 # __get__ return type
-_GT = TypeVar("_GT", covariant=True)
+_GT = TypeVar("_GT", covariant=True, default=Any)
+# null flag type
+_NT = TypeVar("_NT", Literal[True], Literal[False], default=Literal[False])
 
-class Field(RegisterLookupMixin, Generic[_ST, _GT]):
+class Field(RegisterLookupMixin, Generic[_ST, _GT, _NT]):
```
sobolevn, line 56: Do we have a test that `Field[_ST, _GT]` is still allowed? We must be sure that people who used it before this change will have the same experience.

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -46,11 +46,13 @@ _ErrorMessagesMapping: TypeAlias = Mapping[str, _StrOrPromise]
 _ErrorMessagesDict: TypeAlias = dict[str, _StrOrPromise]
 
 # __set__ value type
-_ST = TypeVar("_ST", contravariant=True)
+_ST = TypeVar("_ST", contravariant=True, default=Any)
 # __get__ return type
-_GT = TypeVar("_GT", covariant=True)
+_GT = TypeVar("_GT", covariant=True, default=Any)
+# null flag type
+_NT = TypeVar("_NT", Literal[True], Literal[False], default=Literal[False])
```
sobolevn, line 54: Do we have a test that `Field[_ST, _GT, bool]` is not allowed?

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -700,13 +730,14 @@ class AutoFieldMixin:
 
 class AutoFieldMeta(type): ...
 
-class AutoField(AutoFieldMixin, IntegerField[_ST, _GT], metaclass=AutoFieldMeta):  # type: ignore[misc]
-    _pyi_private_set_type: Combinable | int | str
-    _pyi_private_get_type: int
+_ST_Auto = TypeVar("_ST_Auto", contravariant=True, default=int | str)
```
sobolevn, line 707: Let's please add a bunch of variance tests for different fields. We should ensure that some type casts are allowed, while some are not, like (brief example):

```python
field: AutoField[int | float, str | int] = AutoField()
auto: AutoField[int, str] = field
```

django-stubs/db/models/fields/json.pyi
```diff
@@ -1,25 +1,20 @@
...
+from django.utils.choices import _ChoicesInput
 from django.utils.functional import _StrOrPromise
-from typing_extensions import Self, TypeVar, override
+from typing_extensions import Self, override
 
-from . import Field
-from .mixins import CheckFieldDefaultMixin
-
-# __set__ value type
-_ST = TypeVar("_ST", contravariant=True, default=Any)
```
sobolevn, line 18: I would prefer to redefine the type vars locally, it is a common practice and it is easier to read in a single file.

django-stubs/db/models/fields/related.pyi
```diff
@@ -33,11 +41,9 @@ def lazy_related_operation(
 ) -> None: ...
 
 # __set__ value type
-_ST = TypeVar("_ST", contravariant=True)
-# __get__ return type
-_GT = TypeVar("_GT", covariant=True, default=_ST)
+_ST = TypeVar("_ST", contravariant=True, default=Any)
```
sobolevn, line 37: Especially in cases, we already redefine parts of type vars, but not all required ones. Maybe it is fine to import `_NT`, since it is rather complex one.

sobolevn (commented):
Great work! I really want to be sure that old `Field[_ST, _GT]` annotations will continue to work as before, plus several small nitpicks :)

UnknownPlatypus:
mypy primer diff look mostly great, just noticed a pre-existing [issue with db_default](https://github.com/typeddjango/django-stubs/actions/runs/25291748500/job/74144492111?pr=3317#step:6:790)

```py
class AbstractMessage(models.Model):
    class MessageType(models.IntegerChoices):
        NORMAL = 1
        RESOLVE_TOPIC_NOTIFICATION = 2

    type = models.PositiveSmallIntegerField(
        choices=MessageType.choices,
        default=MessageType.NORMAL,
        # Note: db_default is a new feature in Django 5.0, so we don't use
        # it across the codebase yet. It's useful here to simplify the
        # associated database migration, so we're making use of it.
        db_default=MessageType.NORMAL,
    )
```
```
- zerver/actions/message_send.py:1872: error: Incompatible types in assignment (expression has type "int", variable has type "MessageType")  [assignment]
+ zerver/actions/message_send.py:1872: error: No overload variant of "__set__" of "Field" matches argument types "Message", "int"  [call-overload]
+ zerver/actions/message_send.py:1872: note: Possible overload variants:
+ zerver/actions/message_send.py:1872: note:     def __set__(self, instance: Any, value: MessageType | Combinable) -> None
```
The problem is that because _ST is contravariant and part of the `db_default` type, the instantiation narrow `_ST=MessageType` which then reject `int` while it was a valid choice before

Not sure how to handle this but I don't think it should be a blocker for this, this is a preexisting issue I believe. 
I can open a new issue for it, maybe just duplicating the type union we use in `_ST` default and using it in `db_default` would be ok ?

Similarly, this causes issue now:

```python
  class M(models.Model):                                                                                                                                        
      a = models.JSONField(default=dict, db_default={})  # error: Need type annotation for "a" 
```
mypy tries to bind `_ST_JSON` from the empty-dict literal in the union, fails to commit to a type, and the field is left unparameterized. I've changed this one to `Any` since we don't loose type strictness versus before but doing so in other places would be a regression.

django-stubs/db/models/fields/related.pyi
UnknownPlatypus: > Maybe it is fine to import _NT, since it is rather complex one.

That was my rationale for `_ST` and `_GT` too but that maybe make less sense.
I think it great to import `_NT` tho because it also make it easier to test changes on it

django-stubs/db/models/fields/__init__.pyi
UnknownPlatypus: You mean defining a `class CustomField(Field[_ST, _GT, bool]): ...` ?

sobolevn: Yes :)
And just using `Field[_ST, _GT, bool]` as a type form.

django-stubs/db/models/fields/__init__.pyi
UnknownPlatypus: I've added a test for this and need to fix it for mypy in the plugin. However for the other type checkers it's the same story [as above](https://github.com/typeddjango/django-stubs/pull/3317#discussion_r3178830015), it cannot work without writing all the TypeVar explicitely I think

sobolevn: What about existing annotations for some custom fields: `field: Field[A, B] = CustomField()`?

delfick:
I got around to giving this a shot on our codebase. I'm clearly gonna need to make it work against 6.0.3 first before I can be specific about this branch (might take me up to a few weeks), but that aside there's 84 unused type ignores with this branch (🥳 ) and seems to throw a bunch of new errors that indicate it actually understands what types things should be so that's nice. I think this will be great :)

UnknownPlatypus:
Awesome, I'll try to finish this later this week, would be awesome if you manage to check if some new issues are false positives or not 🙏

delfick:
yeah, I imagine you'll have merged something by the time I get a chance to run our code on a version of django-stubs that doesn't give me several hundred errors before adding this change, but don't let that stop you. Once I catch up I'm sure we'll be able to point out oversights our code reveals :)

django-stubs/db/models/fields/__init__.pyi
UnknownPlatypus: Added the test `field_two_typevar_form_in_user_annotation` to check this.

- `field: Field[A, B] = CustomField()` works out of the box
- `field: Field[A, B] = CustomField(null=True)` does not. We could add plugin code for `Field[...]` calls to re-inject `_NT` but imo people should just remove `: Field[A, B]` and it will work (and with a more precise type!). 

To me, one of the goals of this pr is to completely eliminate the `field: ...` annotations that have only be used as a workaround until now.

UnknownPlatypus:
Just fixed 2 edge cases I found running this on my work code base in the last two commits.

Other than that it is looking good mostly with ~100 unused type ignore and new correct errors surfacing in places that where previously not checked.

However, it's a bit anoying for custom field to have to redeclare the full typevars with defaults like so:

```python
from typing import Literal, reveal_type
from typing_extensions import TypeVar  # for `default=` (PEP 696)
from django.db import models
from django.db.models.expressions import Combinable

_ST = TypeVar("_ST", contravariant=True, default=float | int | str | Combinable)
_GT = TypeVar("_GT", covariant=True, default=int)
_NT = TypeVar("_NT", Literal[True], Literal[False], default=Literal[False])

class MyIntegerField(models.IntegerField[_ST, _GT, _NT]):
    ...

```

I'm wondering if we should expose our typevars so that people can do something like that instead (if we go down that path, we probably need better names too)

```python
from django.db import models
from django_stubs_ext import ST_Int, GT_Int, NT

class MyIntegerField(models.IntegerField[ST_Int, GT_Int, NT]):
    ...

```

sobolevn (commented):
This is an impressive work!

My main concern right now is that existing `Field[A, B]` annotations can change. For example, we use `default=Literal[False]` in the third type var, which can be rather limiting? Would it make sence to use `Any` as the default?

tests/assert_type/contrib/postgres/fields/test_array.py
ngnpope, line 1: Should this come under `tests/assert_type/contrib/postgres/fields/test_array.py`?

tests/assert_type/http/contrib/postgres/fields/test_ranges.py
ngnpope, line 1: Should this come under `tests/assert_type/contrib/postgres/fields/test_ranges.py`?

ngnpope (commented):
A couple of comments about test placement below.

A quick note that we've tried this on our large repo and things mostly seem good, but there are some things not quite right. We currently have 235 new errors with the following breakdown:

```
     10 [arg-type]
     12 [assignment]
    110 [attr-defined]
      5 [call-overload]
     44 [misc]
     14 [operator]
      1 [return-value]
      3 [type-var]
     12 [union-attr]
      2 [unreachable]
     10 [unused-ignore]
     12 [var-annotated]
```

The first thing I want to look into are the `attr-defined`. The majority of these look to be an issue with `GenericRelation` where I think that it's not seeing this as a `QuerySet`. We're typically seeing something like:

```
error: "MyModel" has no attribute "aggregate"  [attr-defined]
error: "MyModel" has no attribute "exists"  [attr-defined]
error: "MyModel" has no attribute "filter"  [attr-defined]
error: "MyModel" has no attribute "get"  [attr-defined]
...
```

I'm not currently sure whether this is something that would have been `Any` before and is now resolving to something else, and have yet to come up with a reproducer. But I wanted to highlight it now in case anyone else is able to get around to working something out before I can.

Once that's been resolved we can sift the rest. I suspect some of them will be merely due to needing to specify defaults for type variables of field subclasses. But starting with the biggest issue makes sense.

UnknownPlatypus:
> My main concern right now is that existing Field[A, B] annotations can change. For example, we use default=Literal[False] in the third type var, which can be rather limiting? Would it make sence to use Any as the default?

This is my concern too, and I'm also concerned the current approach will be hard to scale when we want to add `primary_key` / `default` / `db_default` handling. I'm starting to think the `_NT` typevar is not the proper solution.

I'm trying another approch with `__init__` / `__new__` overloads to see if it's better 

> The first thing I want to look into are the attr-defined. The majority of these look to be an issue with GenericRelation where I think that it's not seeing this as a QuerySet. We're typically seeing something like:

Do you have a sample snippet of one of these so I can add a test case for it ? 
I've probably deleted a bit too much mypy plugin in that area

ngnpope:
> Do you have a sample snippet of one of these so I can add a test case for it ?
> I've probably deleted a bit too much mypy plugin in that area

I'll have to try and run things again and isolate a particular example. Didn't have much success the first time I tried 🙁

django-stubs/db/models/fields/__init__.pyi
```diff
@@ -403,12 +421,13 @@ class SlugField(CharField[_ST, _GT]):
     @override
     def formfield(self, **kwargs: Any) -> forms.Field | None: ...  # type: ignore[override]
 
-class EmailField(CharField[_ST, _GT]):
-    _pyi_private_set_type: str | Combinable
+_ST_Email = TypeVar("_ST_Email", contravariant=True, default=str)
+
+class EmailField(CharField[_ST_Email, _GT_Char, _NT]):
```
ngnpope: Does it make sense to have `_GT_Email` even though it's the same as `_GT_Char` to avoid confusion?

```suggestion
_ST_Email = TypeVar("_ST_Email", contravariant=True, default=str)
_GT_Email = TypeVar("_GT_Email", covariant=True, default=str)

class EmailField(CharField[_ST_Email, _GT_Email, _NT]):
```

After all `_ST_Text` is also the same as `_ST_Char` and we have that.

ngnpope:
> > Do you have a sample snippet of one of these so I can add a test case for it ?
> > I've probably deleted a bit too much mypy plugin in that area
> 
> I'll have to try and run things again and isolate a particular example. Didn't have much success the first time I tried 🙁

Ok. Not come up with an example yet, but I did slap a `reveal_type()` in and it looks like before if we had something like:

```python
class Book(Model):
    ...

class Author(Model):
    ...

    books = GenericRelation(
        "myapp.Book",
        content_type_field="...",
        object_id_field="...",
        related_query_name="...",
    )
```

Then we'd get:

```python
reveal_type(author.books)  # N: Revealed type is "Any"
reveal_type(author.books.all())  # N: Revealed type is "Any"
reveal_type(author.books.first())  # N: Revealed type is "Any"
```

With this branch we're getting:

```python
reveal_type(author.books)  # N: Revealed type is "myapp.models.Book"
reveal_type(author.books.all())  # N: Revealed type is "Any"
reveal_type(author.books.first())  # N: Revealed type is "Any"
```

So the issue is that we just had `Any` before and now we're getting the `Book` model.

That might be because we've changed this?

```diff
-class GenericRelation(ForeignObject[Any, Any]):
+class GenericRelation(ForeignObject[_ST, _GT, _NT]):
```

Perhaps we should try to do something better here, but I think we might need to handle `GenericForeignKey`, `GenericRelation`, `ForeignObject`, etc. a little more specially?

Hope that's something to begin with. Will try to follow up with more when I get a chance to dig some more.

UnknownPlatypus:
I should probably ho back to Any in the first place or the scope of this mr will explode

sobolevn:
@UnknownPlatypus I think that this PR is really important for the future of the project.
My question is: what can we do to help getting this merged?

FallenDeity:
I just wanna chime in here

appreciate the thoroughness in surveying the alternatives, including the `__new__`-overload approach from #2590. I want to push back a bit on how that comparison is being weighed, though.

The `__new__`/`__init__`-overload pattern for steering constructor return types isn't a workaround, it's explicitly specified in the [official typing spec's Constructors section](https://typing.python.org/en/latest/spec/constructors.html#new-method), including a worked example (`self`-annotation overloads influencing the return type of `__init__`) that's essentially identical in shape to what #2590 does for `null=`. So when we found this didn't work cleanly on mypy, that's tracked as open mypy bugs ([python/mypy#15220](https://github.com/python/mypy/issues/15220), [python/mypy#17251](https://github.com/python/mypy/issues/17251)) around `__new__`-vs-`__init__` precedence when both are overloaded not a shortcoming of the approach itself. If a type checker mishandles a spec-sanctioned pattern, I'd argue that's a bug to raise upstream with that checker, not a reason to route around it by changing our stub architecture. We shouldn't be designing stubs defensively around one checker's implementation gaps, especially when the alternative changes so much surface area.

On the `_NT` side, it's worth noting the trade-off isn't entirely free either. Adding a third TypeVar to `Field[_ST, _GT, _NT]` changes the arity of the base class and everything that inherits it. The `__new__`-overload approach doesn't touch `Field`'s base arity, so it avoids that particular ripple effect although it does trade that for some repeated boilerplate per field.

It's also worth flagging the downstream impact on library authors here. Any custom field subclass that doesn't explicitly redeclare `_NT` (e.g. a bare `class HTMLField(models.TextField): ...`) will silently resolve nullability to `Literal[False]` regardless of the `null=` actually passed unless the mypy plugin is installed, which reparametrizes it correctly under the hood. For pyright/ty/pyrefly users, there's no such bridge, so every downstream library maintaining custom fields would need to manually redeclare `_NT` per the README's recommendation just to get correct types.

A lot of projects are actively moving toward plugin-free checkers like pyright and ty rather than staying tied to mypy and its plugin ecosystem along with a class of other benefits they bring in terms of conformance, coverage, speed and ecosystem like astral's for example. This might become an issue down the line too a refactor then would be more costly.

Regarding the concerns regarding `__new__` with other type checkers thats just because pyrefly, zuban etc are newer typecheckers which are becoming compliant over time so no reason to not follow the official typing conformance when we are developing a solution for example pyrefly recently added support for it perhaps because it does seem to work they had some frequent releases, quoting is not an issue either `from __future__ import annotations` should solve most of those forward reference problems but thats standard python

here is a working pyrefly example:

- [pyrefly](https://pyrefly.org/sandbox/?project=v2.pVZNa4NAEP0rS3LQFFHMMRBoLgk9FArJLYrYZLEFUfGj6c_vm_3SVRNoexFcZ9_svDfz3F9YpnErY3_GJO-Y0bzvkoTJ8YRyFKILtY-naAHVksNo_SDX9ezuyWpclfeMXR5DSKzbdmaYnzxWdHm-Ye9wG2DvUzTx7FB3Fa_dlW_2i7Q9asZ7UD05ejjBwA1HsiYd5xqg-76v6k7CceVJGC08ypJ2ebtFD0sixoG0Mg3UzLwULc94LQkST6InCQVBSWgoetZaDWsr-A21XXKYueFLqXymEYpFRdFimETCK1tSWdRbjKGVFf8po5AotrSaS60rs5LdR5_oP8IUNwI8CG9ign1zGGhwL5l_La88tzylV2Xniq-afPGnJFktteh8W3muQdh6NozUEKqn-LyjHh1MnJv6MsOKsSXj3xX9HtCs7WzUehyl1Jsa25K9yQWmPAK_dDhi1uHiAKcD9UvcGTj7aNuq2QSB3l7WWcCL4FpemsDaETDy8PQr_RTXERuOlcKIG8D-x1EdaajOIz91yE4R8dAifwA)

typing conformance results are here: https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html

Worth noting the downstream impact here too: based on testing across mypy, pyright, pyrefly, and ty, subclasses only need to redefine `__new__` when they add new constructor parameters completely normal overload maintenance, not specific to this pattern and on pyright/pyrefly/ty a missing redefinition fails loudly with a clear "no matching overload" error rather than silently producing wrong types. The one exception is mypy, which currently infers incorrectly and silently even on trivial subclasses that change nothing (tracked in the issues linked above), so the "boilerplate per field" concern isn't really inherent to the approach it's mostly a workaround for one checker's current gap.

There's also a difference in scope worth naming. The `__new__`-overload approach only adds annotations to methods already being defined on each field class `Field`'s own generic shape (`Field[_ST, _GT]`) is untouched, so nothing that already subclasses or references `Field` needs any changes. `_NT` does change `Field`'s generic arity itself, and because `Field` is inherited so widely across the stubs, that change propagates to code that has nothing to do with nullability.

UnknownPlatypus:
@FallenDeity Thanks and I mostly agree which is why I made this PR draft, I don't thing the NT approach is the way to go. The initial comparison is not accurate anymore and the shortcomings you mention are too big blockers in my opinion

I have another branch going using `__init__` overloads that is more promissing but personal life is very chaotic these day and I cannot find time to get to it again.

I'll open a draft pr with my initial thoughts and work because I think I'll need support to finish it.

I very much agree with you that we should aim at minimal plugin work and should use constructs documented here https://typing.python.org/en/latest/spec/constructors.html#init-method

FallenDeity:
ah I see would be nice to see the init approach too perhaps as a work in progress branch i can help out as well if possible, because I really do keep coming back to django-stubs but missing field typehints is a big blocker for me with almost all type checkers screaming red and unknown var types with all the django models

so would love to see this feature eventually added to the stubs

PR #3326, opened 2026-04-21
Type-check defaults and create_defaults dict literals

federicobond:
Extends the existing kwargs validation for (a)get_or_create and (a)update_or_create to check values passed via defaults= and create_defaults=. Unknown keys and incompatible value types are reported against the model's create setter types.

Only literal DictExpr arguments with string-literal keys are checked; any other shape (e.g. a variable of type dict[str, object]) is skipped to avoid false positives.

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/orm_lookups.py
+++ b/mypy_django_plugin/transformers/orm_lookups.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from mypy.nodes import DictExpr, StrExpr
+from mypy.nodes import ARG_NAMED, DictExpr, StrExpr
 from mypy.types import AnyType, Instance, ProperType, TypeOfAny, get_proper_type
 from mypy.types import Type as MypyType
 
@@ -81,6 +81,8 @@ def _typecheck_defaults_kwarg(
 
     for idx in defaults_positions:
         if not ctx.args[idx]:
+            continue
+        if ctx.arg_kinds[idx][0] != ARG_NAMED:
             continue
         dict_expr = ctx.args[idx][0]
         if not isinstance(dict_expr, DictExpr):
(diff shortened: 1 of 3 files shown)


mypy_django_plugin/transformers/orm_lookups.py
federicobond, line 18: We might have to reconsider the name of this function.

sobolevn (commented):
Please, test `**kwargs` unpacking

federicobond:
@sobolevn good intuition! There was a bug hiding there. Pushed a fix as a separate commit.

tests/typecheck/managers/querysets/test_get_or_create.yml
```diff
@@ -50,6 +50,14 @@
         some_defaults: dict[str, object] = {"unknown_field": True}
         MyModel.objects.get_or_create(num=1, defaults=some_defaults)
 
+        # `**kwargs` variable unpacking is silently accepted
+        kwargs = {"defaults": {"unknown_field": True}}
+        MyModel.objects.get_or_create(num=1, **kwargs)
```
sobolevn, line 55: Please, also add `MyModel.objects.get_or_create(num=1, {"defaults": {"unknown_field": True, **other_fields}})`

federicobond: Did you mean `MyModel.objects.get_or_create(num=1, defaults={"unknown_field": True, **other_fields})`?

sobolevn: yes, sorry :)

federicobond: I've expanded the test to include that case. Literal string keys are validated against known fields, anything else is ignored.

sobolevn (approved):
thanks!

UnknownPlatypus (approved):

Issue #602, opened 2021-04-25
QuerySet's `.annotate()` returns an incorrect type when chained with `.values()`


kracekumar:
# Bug report

## What's wrong

Iterating over the `annotate` result and accessing the aggregated item throws `Model` is not indexable.

```python
class Author(models.Model):
    author_name = models.CharField(max_length=255)
    created_at = models.DateTimeField(auto_now_add=True)
    updated_at = models.DateTimeField(auto_now=True)

    class Meta:
        db_table = "authors"
        verbose_name = "Author"
        verbose_name_plural = "Authors"

    def __str__(self):
        return self.author_name

class Book(models.Model):
    BOOK_STATUS=(
        ('PUBLISHED', 'Published'),
        ('ON_HOLD', 'On Hold'),
    )
    book_name = models.CharField(max_length=255)
    author = models.ForeignKey('Author',on_delete=models.CASCADE,
                               related_name='author')
    status = models.CharField(max_length=255, choices = BOOK_STATUS,
                              default=BOOK_STATUS[0][0])
    created_at = models.DateTimeField(auto_now_add=True)
    updated_at = models.DateTimeField(auto_now=True)

    class Meta:
        db_table="books"
        verbose_name="Book"
        verbose_name_plural="Books"

    def __str__(self):
        return self.book_name

class AuthorCount(TypedDict):
    authors: int

def get_book_counts_for_author() -> QuerySet[Book]:
    qs = Book.objects.values('author')
    qs1 = qs.annotate(
        authors=Count('author'))
    return qs1


def consume() -> list[int]:
    counts = []
    for item in get_book_counts_for_author():
        counts.append(item['authors']) # error happens here
    return counts
```

Mypy error 

```
...
polls/models.py:79: error: Value of type "Book" is not indexable
```

What should be the proper way to annotate the code here other than `type: ignore`?

Failed Approaches
- `get_book_counts_for_author() -> QuerySet[AuthorCount]:`

Output:

`polls/models.py:69: error: Type argument "TypedDict('polls.models.AuthorCount', {'author': builtins.int})" of "QuerySet" must be a subtype of "django.db.models.base.Model"`

- `get_book_counts_for_author() -> Union[QuerySet[Book], AuthorCount]:`

Output:
```
polls/models.py:80: error: Value of type "Union[Book, str]" is not indexable
polls/models.py:80: error: Invalid index type "str" for "Union[Book, str]"; expected type "Union[int, slice]"
[...]

PR #3330, opened 2026-04-22
Fix annotate leak across functions

### LINKED ISSUE #3324: Weird caching issue with annotated querysets
# Bug report

## What's wrong

Hello, I'm unsure how to replicate this, but I'm going through the many hundreds of new errors after upgrading to 6.0.3 (I'm loving all the new insight latest that django-stubs has into queryset methods btw) and I keep coming across situations where it appears that django-stubs is getting confused and thinking that inputs to functions that are normal querysets of a model without any annotations has items in it that are annotated.

It feels like some kind of caching issue perhaps?

Is this a known problem?

## System information

- OS:
- `python` version: 3.13
- `django` version: 6.0.4
- `mypy` version: 1.20.1
- `django-stubs` version: 6.0.3
- `django-stubs-ext` version: 6.0.3

UnknownPlatypus:
# I have made things!

This tries this approach https://github.com/typeddjango/django-stubs/issues/3324#issuecomment-4295539146

I think this is more correct.
cc @delfick , it fixes your bug (See test`annotate_does_not_leak_across_functions`) and also the one #3271 was trying to address (see `custom_queryset_with_annotations_issue_1049`)

A bunch of tests are failing, I expect it's mostly snapshots changing but I need to check more.
Opening to get initial feedback

## Related issues
Fixes #3324

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -606,9 +606,24 @@ def add_as_manager_to_queryset_class(ctx: ClassDefContext) -> None:
     )
 
 
-def reparametrize_generic_class(ctx: ClassDefContext, base_class_fullname: str) -> None:
+def _is_omitted_generic(arg: MypyType) -> bool:
+    """True if ``arg`` is the ``Any`` mypy substitutes for an unparametrized generic."""
+    proper_arg = get_proper_type(arg)
+    return isinstance(proper_arg, AnyType) and proper_arg.type_of_any is TypeOfAny.from_omitted_generics
+
+
+def reparametrize_generic_class(
+    ctx: ClassDefContext,
+    base_class_fullname: str,
+    *,
+    bind_explicit_args: bool = False,
+) -> None:
     """
     Add implicit generics to classes that are defined without generic.
+
+    When ``bind_explicit_args`` is True, also reparametrize partially or
+    fully-explicit parametrizations by binding each explicit arg as both
+    the bound and the PEP 696 default of a synthesized covariant TypeVar.
 
     Note that this does not happen if mypy is run with disallow_any_generics = True,
     as not specifying the generic type is then considered an error.
@@ -633,10 +648,31 @@ def reparametrize_generic_class(ctx: ClassDefContext, base_class_fullname: str) 
     if parent_class is None or not parent_class.args:
         return
 
-    type_vars = [
-        type_var.copy_modified(id=type_var.id, default=parent_type_var)
-        for type_var, parent_type_var in zip(parent_class.type.defn.type_vars, parent_class.args, strict=True)
-    ]
+    # Bind explicit args only when the direct parent is the canonical base class
+    is_direct_parent = parent_class.type.fullname == base_class_fullname
+    if not (bind_explicit_args and is_direct_parent):
+        if not _is_omitted_generic(parent_class.args[0]):
+            return
+        type_vars = list(parent_class.type.defn.type_vars)
+    else:
+        type_vars = []
+        for type_var, parent_type_var in zip(parent_class.type.defn.type_vars, parent_class.args, strict=True):
+            if _is_omitted_generic(parent_type_var):
+                # Arg was omitted -- reuse the parent's TypeVar so its existing
+                # PEP 696 default (e.g. ``_Row = TypeVar(default=_Model)``) is preserved.
+                type_vars.append(type_var)
+            else:
+                # Arg was supplied explicitly -- synthesize a TypeVar bounded by the
+                # user's arg so the subclass stays effectively concrete in user-facing
+                # contexts (defaults + bound + covariance) while still being generic
+                # enough for the plugin to flow annotation row types through.
+                type_vars.append(
+                    type_var.copy_modified(
+                        id=type_var.id,
+                        upper_bound=parent_type_var,
+                        default=parent_type_var,
+                    )
+                )
 
     # If we end up with placeholders we need to defer so the placeholders are
     # resolved in a future iteration
@@ -654,6 +690,7 @@ def reparametrize_generic_class(ctx: ClassDefContext, base_class_fullname: str) 
 def reparametrize_any_queryset_hook(ctx: ClassDefContext) -> None:
     """
     Add implicit generics to QuerySet subclasses that are defined without generic.
+    This is required for `.annotate` / `.values` call to update the typevars accordingly.
 
     Eg.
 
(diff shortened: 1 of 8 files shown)


delfick:
@UnknownPlatypus yeah, this feels better than trying to make copies of a TypeInfo.

I will note that I've realised that I needed to remove the mypy_cache for this code to actually run, and now I've realised the solution I gave yesterday actually fails (`AttributeError: type object 'TypeInfo' has no attribute '__slots__'`). And I'm a little confused.

So with this PR specifically, compared to without these changes

```

before: 650
after: 601
Same: 101, Added: 500, Removed: 549
```

which is at least removing more than it adds :p

Seems the new errors is a lot of `error: Variance of TypeVar "T_MyTypeVar" incompatible with variance in parent type  [type-var]` errors, which appear to be where we are creating a subclass of a queryset

so

```
class _MyConcreteQuerySet(MyAbstractQuerySet[MyConcreteModel]):
    pass
```

And also there seems to be a problem where it doesn't realise `.values` returns a Values QuerySet.

UnknownPlatypus:
Should be solvable, I think it’s still the coreect direction. Not sure I'll have some time this week to finish this, feel free to poke around if you have time.

delfick:
> if you have time.

I'd certainly love to have more time! I might have time next week to have a deeper dive into this if you don't get there first :)

delfick:
this week has been a lot for me, so I haven't had much time to spend on it. But I've had a quick look now and I think the problem is that we end up with types that think classes are generic when they are not.

For example

```
    pass
```

Is not a generic class, but with this change we end up with mypy providing types like `MyQuerySet[MyModel@AnnotateWith[...]], ...]`

The queryset needs to be changed in the code to be like

```
class MyQuerySet[T_Model: MyModel = MyModel](models.QuerySet[T_Model]):
   pass
```

For that to be valid.

federicobond:
Yes, this indeed looks like the correct type for any custom QuerySet moving forward:

```python
class MyQuerySet[T_Model: MyModel = MyModel](models.QuerySet[T_Model]):
   pass
```

Any queryset instance must be able to reparametrize in the presence of .annotate calls.

delfick:
yeah, we've found for quite a while doing that fixes problems where mypy doesn't see an annotate. But we have over 2000 django models and sometimes doing that also requires other refactoring to deal with new problems mypy sees so we haven't had the space to do that refactor pro-actively

UnknownPlatypus:
> I think the problem is that we end up with types that think classes are generic when they are not.

Actually, we have a plugin hook that injects the missing typevar (see reparametrize_any_*) to ensure any custom queryset is generic.

I dont think it ads the typevar default to the model however

delfick:
> to ensure any custom queryset is generic.

oh, right. Perhaps that's not actually something that's possible with mypy. Cause we should probs only do that on Instance objects and that ends up requiring the ability to make a copy of the type but with different bases. 

I think the reality is django-stubs would require mypy to have a copy_modified on TypeInfo for this to be possible and the fix is instead to make `reparametrize_queryset` not pretend querysets are generic when they are not:

```
def reparametrize_queryset(instance: Instance, args: list[MypyType]) -> Instance:
    """Reparametrize a QuerySet instance with new type arguments.

    Note this only works when the queryset is already defined as generic
    """
    if instance.type.is_generic():
        return instance.copy_modified(args=args)

    return instance
```

also, I see now why deleting the mypy cache made a difference to what mypy could see (and I imagine is part of why TypeInfo doesn't already have a copy_modified)

Seems with that change I still have a number of models where I do have to go change the queryset to be generic now to avoid "cannot resolve keyword into field" errors, but that's fine. Seems also django-stubs doesn't realise `.values` returns a ValuesQuerySet, but I assume that's a different issue I should raise a github issue about.

UnknownPlatypus:
> oh, right. Perhaps that's not actually something that's possible with mypy. Cause we should probs only do that on Instance objects and that ends up requiring the ability to make a copy of the type but with different bases.

I don't think so, but I think we do it wrong somehow now. 
The idea was to modify the type to mimic a manual `I do have to go change the queryset to be generic`, like a convenience to limit mass changes. We don't need a copy of the base type, we want to alter it to look like if it was written as generic in the first place so that we can then do regular `Instance.copy_modified` to fix the types around
[edit] I've just push something that seem to work and do that, let me know how it goes for you


> the fix is instead to make reparametrize_queryset not pretend querysets are generic when they are not:

This would bring us to square one with the types being completely lost when using non-generic custom queryset. Maybe we have to live with that but that's a lot of churn. I got kinda stuck with #2776 because of that, turn out it revealed 300+ false positives in my work codebase because custom queryset started getting typechecked in most cases.

If the user can write this which works with annotates etc:
```python
class MyQuerySet[T_Model: MyModel = MyModel](models.QuerySet[T_Model]):
   pass
```
We should be able to mimic that in the plugin by fixing the `TypeInfo` mypy processed when looking at
```python
   pass
```
or
```python
class MyQuerySet(models.QuerySet[MyModel]):
   pass
```

>  Seems also django-stubs doesn't realise .values returns a ValuesQuerySet, but I assume that's a different issue I should raise a github issue about.

I think we removed `ValuesQuerySet` in  https://github.com/typeddjango/django-stubs/pull/2104

delfick:
> We don't need a copy of the base type, we want to alter it to look like if it was written as generic in the first place so that we can then do regular Instance.copy_modified to fix the types around

right, yeah, that would work. I imagine we'll wanna introduce a `__version__` variable next to https://github.com/typeddjango/django-stubs/blob/9048fdcdf9d02e5ffd32a8923e1a2183620a84a0/mypy_django_plugin/main.py#L363 so that mypy knows to invalidate the cache ([mypy looks for such a module level variable to determine the version of a plugin](https://github.com/python/mypy/blob/145a062651b5f9996b75ef32b7040bd2e885ed82/mypy/build.py#L710))

> I've just push something that seem to work and do that, let me know how it goes for you

Nice! I'll try it out in a bit :) 

>  turn out it revealed 300+ errors

I got several hundred errors too. Has kicked us into gear of actually fixing them :)

> I think we removed ValuesQuerySet

correct, I'm being lazy with what I mean. It seems mypy is sad about dictionary access when it shouldn't be.

delfick:
> I imagine we'll wanna introduce a __version__ variable next to

though, if `mypy_django_plugin/main.py` is changed betwen 6.0.3 and 6.0.4, then it'll already consider the plugin different and invalidate the cache

>  I'll try it out in a bit :)

I think what you've done here works (it's a little hard to tell cause I still do have 336 errors but seems it's mostly other errors (certainly not the errors that made me create that github issue in the first place!) and a lot of "is not indexable" errors (what I'm referring to above with a values queryset), which I imagine is indeed a separate issue that I'll end up making a new github issue for when I get to it)

Thanks!

UnknownPlatypus:
Awesome, let me know if you still find issue related to this pr, otherwise I'll merge it during the day.

As for the cache I think it will be invalidated with the new version but Ill try to verify that

Looking forward for the other issues, real world exemple from large codebase are really valuable here

UnknownPlatypus:
Can confirm it also fixes the same pollution error I had in my work codebase.
Awesome, merging now, thanks again for the report and help on getting to the bottom of this @delfick

delfick:
Thanks!

delfick:
@UnknownPlatypus , for papertrail, I finally posted something about my problem with querysets that should be seen as indexable, https://github.com/typeddjango/django-stubs/issues/602#issuecomment-4427909661 (though my time at the moment is very constrained and so it's only a test case to reproduce the problem)

PR #3331, opened 2026-04-22
Just a quick fix

bitmooze:
# I have made things!

## AI Policy

### DIFF
--- a/django-stubs/contrib/admin/sites.pyi
+++ b/django-stubs/contrib/admin/sites.pyi
@@ -47,7 +47,7 @@ class AdminSite:
     def register(
         self,
         model_or_iterable: type[Model] | Iterable[type[Model]],
-        admin_class: type[ModelAdmin] | None = ...,
+        admin_class: type[ModelAdmin[Model]] | None = ...,
         **options: Any,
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Ideally this can later be made generic :)

cuu508:
In my project, I use a custom Admin class for Django's User objects like so:

```python
from django.contrib.auth.admin import UserAdmin


class CustomUserAdmin(UserAdmin[User]):
    ...
    
    
admin.site.unregister(User)
admin.site.register(User, CustomUserAdmin)    
```

Now with with django-stubs 6.0.4, I get this warning:

```
error: Argument 2 to "register" of "AdminSite" has incompatible type "type[CustomUserAdmin]"; expected "type[ModelAdmin[Model]] | None"  [arg-type]
```

Is my code wrong, and how can I fix it?

bitmooze:
@cuu508 you can do this for now

```python
admin.site.unregister(User)
@admin.register(User)
class CustomUserAdmin(UserAdmin[User]):
    pass
```

cuu508:
Thanks @bitmooze, that works!

PR #3346, opened 2026-04-27
Remove explicit `--strict` (we set `strict=true` in mypy.ini already)

UnknownPlatypus:
Also adds a just receipe to run every typechekers on the `tests/assert_type` folder

## AI Policy

### DIFF
--- a/justfile
+++ b/justfile
@@ -16,7 +16,7 @@ lint:
 
 # Run all checks before submitting a PR
 [group('dev')]
-pre-mr-check: lint ty pyrefly mypy stubtest pyright ext-test test
+pre-mr-check: lint typecheck-all stubtest ext-test test
 
 # Remove mypy cache
 [group('dev')]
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -26,11 +26,11 @@ clean:
...
 mypy:
-    uv run mypy --strict ext
-    uv run mypy --strict scripts
-    uv run mypy --strict mypy_django_plugin
+    uv run mypy ext
+    uv run mypy scripts
+    uv run mypy mypy_django_plugin
     uv run mypy --cache-dir=/dev/null --no-incremental django-stubs
-    uv run mypy --strict tests
+    uv run mypy tests
```
sobolevn:
```suggestion
    uv run mypy ext scripts mypy_django_plugin tests
    uv run mypy --cache-dir=/dev/null --no-incremental django-stubs
```

PR #3349, opened 2026-04-28
Just another quick fix

bitmooze:
# I have made things!

## Related issues

```
Type of "check_password" is partially unknown
  Type of "check_password" is "(password: str | None, encoded: str, setter: ((...) -> Unknown) | None = ..., preferred: str = ...) -> bool"
```
## Fix

Callable[..., Any]

## AI Policy


django-stubs/contrib/auth/hashers.pyi
```diff
@@ -9,7 +9,9 @@ UNUSABLE_PASSWORD_SUFFIX_LENGTH: int
 
 def is_password_usable(encoded: str | None) -> bool: ...
 def verify_password(password: str | None, encoded: str, preferred: str = ...) -> tuple[bool, bool]: ...
-def check_password(password: str | None, encoded: str, setter: Callable | None = ..., preferred: str = ...) -> bool: ...
+def check_password(
+    password: str | None, encoded: str, setter: Callable[..., Any] | None = ..., preferred: str = ...
```
sobolevn: Can we type `setter` better?

bitmooze: Well yes we can but that makes it way more complicated so I go with easy to understand, if you want I can improve it but what I have right now does fix the problem.

UnknownPlatypus: > it but what I have right now does fix the problem.

Which problem, if it solve a real world issue for you, could you add a test in the `tests/assert_type` folder.




And I agree we should be able to do better here, looking at the Django implementation (which I highly recommend to do when submitting stub changes), It seem to be a simple function taking  `password` and expected to return anything (the return type is ignored))

so probably 
```python
def check_password(
    password: str | None, encoded: str, setter: Callable[[str | None], Any] | None = ..., preferred: str = ...
) -> bool: ...
```

django-stubs/core/files/storage/base.pyi
```diff
@@ -1,11 +1,12 @@
+from collections import abc
 from datetime import datetime
 from typing import IO, Any
 
 from django.core.files.base import File
 from django.utils._os import _PathCompatible
 
 class Storage:
-    def open(self, name: str, mode: str = "rb") -> File: ...
+    def open(self, name: str, mode: str = "rb") -> File[abc.Iterable[object] | abc.AsyncIterable[object]]: ...
```
UnknownPlatypus: See https://github.com/typeddjango/django-stubs/actions/runs/25067146401/job/73437435626?pr=3349

File type arg is currently bound to `str, bytes` so this is not valid.
Not exactly sure what we are trying to solve here, a test case to showcase the issue would be great

UnknownPlatypus (changes requested):

django-stubs/contrib/auth/hashers.pyi
UnknownPlatypus: I've pushed a change with above suggestion, thanks!

django-stubs/core/files/storage/base.pyi
UnknownPlatypus: i've reverted to keep this pr scope clear, feel free to reopen a new one with this changed (fixed) and a test. Thanks

PR #3350, opened 2026-05-01
Ban `typing.TypeVar` in favor of `typing_extensions.TypeVar`

UnknownPlatypus:
# I have made things!

We declare a dependency on [typing_extensions](https://typing-extensions.readthedocs.io/en/latest/#) so this ensure `default=...` works with all supported python.
I've added a banned-api ruff rule to ensure if cannot regress

## Related issues
Had issue in [ci](https://github.com/typeddjango/django-stubs/actions/runs/25234008376/job/73996032755?pr=3317) for #3317 which made me think of this 

## AI Policy


sobolevn (approved):
Technically, this is only needed for `default=` argument, but we can use it for all cases with no problems.

PR #3353, opened 2026-05-02
Migrate `test_options.yml` to py

UnknownPlatypus:
# I have made things!

So that #3308 will be able to add an `assert_type` test instead

## AI Policy


tests/assert_type/contrib/admin/test_options.py
```diff
@@ -0,0 +1,163 @@
...
+    class MyModel(models.Model): ...
+
+    def an_action(modeladmin: None) -> None:
+        pass
+
+    class A(admin.ModelAdmin[MyModel]):  # pyright: ignore[reportUnusedClass]
+        actions = [an_action]  # type: ignore[list-item]  # pyright: ignore[reportAssignmentType]  # pyrefly: ignore[bad-assignment]
+
+
+def errors_for_invalid_model_admin_generic() -> None:
```
sobolevn: Can we move classes from functions and just name them correctly? This will help us to remove `# pyright: ignore[reportUnusedClass]` errors and be more in line with the potential usage: the func / module scopes are a bit different and might affect the test :)

UnknownPlatypus: Ye good idea, thanks

opq0pqo (approved):

PR #3357, opened 2026-05-06
Accept single str/Combinable for Aggregate.order_by

### LINKED ISSUE #3356: OrderableAggMixin.order_by annotation should be broader
# Bug report

## What's wrong

According to [Django docs](https://docs.djangoproject.com/en/5.2/ref/contrib/postgres/aggregates/#django.contrib.postgres.aggregates.ArrayAgg.order_by), ArrayAgg's order_by (by way of OrderableAggMixin) supports

> An optional string of a field name (with an optional "-" prefix which indicates descending order) or an expression (or a tuple or list of strings and/or expressions) 

but the type annotation in django-stubs is `Sequence[str]`

## How is that should be

Based on the docs, and a cursory look over the [Django implementation,](https://github.com/django/django/blob/stable/5.2.x/django/contrib/postgres/aggregates/mixins.py#L8) I think the correct annotation is something like:

```
Sequence[str | Expression | F] | str | Expression | F
```
?

This appears to be superseded by [OrderByList](https://github.com/django/django/blob/stable/6.0.x/django/db/models/expressions.py#L1472C19-L1472C33) in Django 6.x, though

## System information 

- OS:
- `python` version: 3.13.9
- `django` version: 5.2
- `mypy` version: 1.18.2
- `django-stubs` version: 5.2.5

federicobond:
# I have made things!

Django's `OrderByList.from_param` accepts a single string, a single expression, or a list/tuple of either. The stubs only typed the sequence form, so passing a single expression like ArrayAgg("title", order_by=F("title")) was rejected by mypy even though it is valid Django code. Fixed it by widening the accepted types accordingly.

## Related issues
Fixes https://github.com/typeddjango/django-stubs/issues/3356

## AI Policy

### DIFF
--- a/tests/assert_type/db/models/test_aggregates.py
+++ b/tests/assert_type/db/models/test_aggregates.py
@@ -1,0 +1,12 @@
+from __future__ import annotations
+
+from django.db.models import F
+from django.db.models.aggregates import Aggregate
+
+Aggregate("title", order_by="title")
+Aggregate("title", order_by="-title")
+Aggregate("title", order_by=F("title"))
+Aggregate("title", order_by=F("title").desc())
+Aggregate("title", order_by=["-title", F("title")])
+Aggregate("title", order_by=("-title", F("title").desc()))
+Aggregate("title", order_by=None)
--- a/django-stubs/contrib/postgres/aggregates/mixins.pyi
+++ b/django-stubs/contrib/postgres/aggregates/mixins.pyi
@@ -10,8 +10,8 @@ class OrderableAggMixin:
     def __init__(
         self,
         *expressions: BaseExpression | Combinable | str,
-        ordering: Sequence[_OrderByFieldName] = ...,
-        order_by: Sequence[_OrderByFieldName] = ...,
+        ordering: _OrderByFieldName | Sequence[_OrderByFieldName] = ...,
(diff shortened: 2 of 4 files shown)


tests/assert_type/db/models/test_aggregates.py
```diff
@@ -0,0 +1,12 @@
...
+from django.db.models import F
+from django.db.models.aggregates import Aggregate
+
+Aggregate("title", order_by="title")
+Aggregate("title", order_by="-title")
+Aggregate("title", order_by=F("title"))
+Aggregate("title", order_by=F("title").desc())
+Aggregate("title", order_by=["-title", F("title")])
+Aggregate("title", order_by=("-title", F("title").desc()))
+Aggregate("title", order_by=None)
```
federicobond, line 12: I wanted to use `StringAgg` here but the `**extra: Any` in the constructor signature swallows the type of `order_by`. Should we update it to explicitly list `order_by` and `orderable`?

sobolevn: If some kwargs are known, we should extract them into known params

federicobond: I'll open a separate PR for it if that's ok.

UnknownPlatypus (approved):

PR #3359, opened 2026-05-07
Expose Aggregate kwargs on subclass __init__ stubs

federicobond:
See https://github.com/typeddjango/django-stubs/pull/3357#discussion_r3194693967

## AI Policy

### DIFF
--- a/tests/assert_type/db/models/test_aggregates.py
+++ b/tests/assert_type/db/models/test_aggregates.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from django.db.models import F
+from django.db.models import Count, F, Q, StdDev, StringAgg, Variance
 from django.db.models.aggregates import Aggregate
 
 Aggregate("title", order_by="title")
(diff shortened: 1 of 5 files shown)


tests/assert_type/db/models/test_aggregates.py
```diff
@@ -1,12 +1,28 @@
 from __future__ import annotations
 
-from django.db.models import F
-from django.db.models.aggregates import Aggregate
-
-Aggregate("title", order_by="title")
```
sobolevn, line 6: Question: why do we remove these examples?

sobolevn (commented):
Thanks!

tests/assert_type/db/models/test_aggregates.py
federicobond: Because the ones down below are more interesting for actual usage, as `Aggregate` is effectively abstract. I only used `Aggregate` in my previous PR because `StringAgg` swallowed kwargs as Any. I can leave them if you prefer though.

sobolevn: Let's leave them as well, more tests the better :)

federicobond: Done!

sobolevn (approved):
Thank you!

UnknownPlatypus (approved):
This is great, thanks!

PR #3360, opened 2026-05-07
Replace **flags with explicit kwargs on Library.filter

federicobond:
is_safe, needs_autoescape, and expects_localtime are the only flags Django's Library.filter consumes, so type them explicitly.

## AI Policy


sobolevn (approved):
Thanks!

PR #3361, opened 2026-05-07
Accept datetime/timedelta on reverse ops of Combinable

### LINKED ISSUE #1548: No overload variant matches argument type "F" [operator]
# Bug report

```python3
now = timezone.now()

MyModel.objects.annotate(
    time_since_creation=now - F("created_at")
)
```

## What's wrong

```
error: No overload variant of "__sub__"
of "datetime" matches argument type "F"  [operator]
    time_since_creation=now - F("created_at")
                        ^~~~~~~~~~~~~~~~~~~~~~~
note: Possible overload variants:
note:     def __sub__(self, timedelta, /) -> datetime
note:     def __sub__(self, datetime, /) -> timedelta

```

## How is that should be

No error? Maybe I'm holding it wrong.

## System information

- OS: N/A
- `python` version: 3.11
- `django` version: 4.1
- `mypy` version: 1.3

federicobond:
## Related issues

Fixes https://github.com/typeddjango/django-stubs/issues/1548

## AI Policy


django-stubs/db/models/expressions.pyi
```diff
@@ -52,8 +52,10 @@ class Combinable:
     def bitxor(self, other: int) -> CombinedExpression: ...
     def __or__(self, other: Combinable | Q) -> Q: ...
     def bitor(self, other: int) -> CombinedExpression: ...
-    def __radd__(self, other: datetime.datetime | _Numeric | Combinable | None) -> CombinedExpression: ...
-    def __rsub__(self, other: _Numeric | Combinable) -> CombinedExpression: ...
+    def __radd__(
```
UnknownPlatypus: what about `__add__` and `__sub__`, should they also align ? See https://github.com/typeddjango/django-stubs/issues/1548#issuecomment-2047555086

Also we might want a type alias to avoid discrepancies in the future

UnknownPlatypus: not sure about the `str` in `__add__` btw

tests/assert_type/db/models/test_expressions.py
```diff
@@ -0,0 +1,27 @@
...
+
+now = datetime.datetime.now()
+delta = datetime.timedelta(seconds=10)
+
+# datetime - F(...) — Python falls back to F.__rsub__ since datetime.__sub__
+# does not accept F.
+assert_type(now - F("created_at"), CombinedExpression)
+
+# timedelta - F(...) — same fallback path.
+assert_type(delta - F("duration"), CombinedExpression)
```
UnknownPlatypus:
```suggestion

# datetime - F(...) — Python falls back to F.__rsub__ since datetime.__sub__
# does not accept F. Same applies for other supported builtin types
assert_type(now - F("created_at"), CombinedExpression)
assert_type(delta - F("duration"), CombinedExpression)
```
Maybe add cases for `_Numeric` too

tests/assert_type/db/models/test_expressions.py
```diff
@@ -0,0 +1,27 @@
...
+
+# F(...) - datetime/timedelta uses F.__sub__.
```
UnknownPlatypus: test `F.__add__` too maybe

tests/assert_type/db/models/test_expressions.py
```diff
@@ -0,0 +1,27 @@
...
+assert_type(F("created_at") - delta, CombinedExpression)
+
+# datetime/timedelta + F(...) — symmetrical fallback to F.__radd__.
+assert_type(now + F("delay"), CombinedExpression)
+assert_type(delta + F("delay"), CombinedExpression)
+
+# F(...) - F(...) — Combinable on both sides.
```
UnknownPlatypus:
```suggestion
```
I mean, that's pretty self explanatory

django-stubs/db/models/expressions.pyi
federicobond: Added type aliases and aligned the operands. Kept `str` for now, but I don't see a strong case for keeping it.

UnknownPlatypus: > but I don't see a strong case for keeping it.

Does it work at all at runtime ?

federicobond: I haven't tested it, but I *think* it might work in SQL Server, where you can do string concat with the `+` operator.

UnknownPlatypus: > but I think it might work in SQL Server

My question was mroe about the python layer and theses:
```python
In [11]: F("id") + "dd"
Out[11]: <CombinedExpression: F(id) + Value('dd')>

In [12]: "dd" + F("dd")
Out[12]: <CombinedExpression: Value('dd') + F(dd)>

In [13]: "dd" - F("dd")
Out[13]: <CombinedExpression: Value('dd') - F(dd)>

In [14]: F("id") - "dd"
Out[14]: <CombinedExpression: F(id) - Value('dd')>

````

They seem to work wich makes sense given the implementation

```python
    def _combine(self, other, connector, reversed):
        if not hasattr(other, "resolve_expression"):
            # everything must be resolvable to an expression
            other = Value(other)

        if reversed:
            return CombinedExpression(other, connector, self)
        return CombinedExpression(self, connector, other)
```

So I think `_AddOperand` and `_SubOperand` should be merged and we keep the `str` in the union ?

federicobond: We can merge them to keep stubs simpler, but technically the Python layer will accept any object and convert it to a `Value`, so by that logic the operand should be typed as `Any`.

If we want the types to restrict values to more semantically valid choices, I think keeping `str` away from `sub` makes more sense.

PR #3362, opened 2026-05-08
Resolve default_alias for positional Aggregate in annotate()

### LINKED ISSUE #2382: .annotate(...) does not seem to work with reverse-related foreign key lookupu
# Bug report

wrote a little testcase -- will see if I can figure out a quick fix for it

```yaml
- case: test_annotate_reverse_related
  main: |
    from myapp.models import MB
    from django.db.models import Count

    o = MB.objects.annotate(Count("mas")).get()
    reveal_type(o.mas__count)  # N: Revealed type is "builtins.int"
  installed_apps:
    - myapp
  files:
    - path: myapp/__init__.py
    - path: myapp/models.py
      content: |
        from django.db import models

        class MA(models.Model):
            mb = models.ForeignKey("MB", related_name="mas", on_delete=models.CASCADE)

        class MB(models.Model): pass
```

## What's wrong

```
_______________________________________ test_annotate_reverse_related _______________________________________
/Users/asottile/workspace/sentry-forked-django-stubs/tests/typecheck/managers/querysets/test_annotate.yml:405: 
E   Actual:
E     main:5: error: "MB" has no attribute "mas__count"  [attr-defined] (diff)
E     main:5: note: Revealed type is "Any"          (diff)
E   Expected:
E     main:5: note: Revealed type is "builtins.int" (diff)
E     E: main:5: note: Revealed type is "builtins.int"
E     A: main:5: error: "MB" has no attribute "mas__count"  [attr-defined]
E                ^
```

## How is that should be

should resolve correctly

## System information

- OS:
- `python` version: 3.12.6
- `django` version: 5.1.1
- `mypy` version:  1.11.
- `django-stubs` version: 5.1.0 
- `django-stubs-ext` version: 5.1.0

federicobond:
`QuerySet.annotate(Count("field"))` lost its annotation in the plugin because `gather_kwargs` only collected named arguments. Mirror Django's Aggregate.default_alias (`<source>__<name_lower>`) for positional aggregates with a single string source so the resulting attribute is typed.

## Related issues

Fixes https://github.com/typeddjango/django-stubs/issues/2382

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -18,12 +18,14 @@ from mypy.errorcodes import NO_REDEF
 from mypy.nodes import (
     ARG_NAMED,
     ARG_NAMED_OPT,
+    ARG_POS,
     ARG_STAR,
     CallExpr,
     Decorator,
     Expression,
     ListExpr,
     SetExpr,
+    StrExpr,
     TupleExpr,
     Var,
 )
@@ -227,20 +229,48 @@ def extract_proper_type_queryset_values_list(ctx: MethodContext, django_context:
     return ret
 
 
+def _aggregate_default_alias(expr: Expression, expr_type: MypyType) -> str | None:
+    """Mirror `Aggregate.default_alias`: ``<source_expression>__<aggregate_name_lower>``.
+
+    Django generates a default alias only for aggregates with a single source
+    expression that has a `name`; complex expressions raise at runtime, so the
+    plugin matches that and returns ``None`` for anything else.
+
+    See https://docs.djangoproject.com/en/dev/topics/db/aggregation/#generating-aggregates-for-each-item-in-a-queryset
+    """
+    if not (
+        isinstance(expr, CallExpr)
+        and isinstance(proper := get_proper_type(expr_type), Instance)
+        and proper.type.has_base(fullnames.AGGREGATE_CLASS_FULLNAME)
+    ):
+        return None
+    source: Expression | None = None
+    for arg, kind in zip(expr.args, expr.arg_kinds, strict=False):
+        if kind != ARG_POS:
+            continue
+        if source is not None:
+            return None
+        source = arg
+    # TODO: consider switching to `helpers.resolve_string_attribute_value` once
+    # `gather_kwargs` threads `django_context` through — that would also handle
+    # `Count(NAME_CONST)` and `Count(settings.X)`.
+    if not isinstance(source, StrExpr):
+        return None
+    return f"{source.value}__{proper.type.name.lower()}"
+
+
 def gather_kwargs(ctx: MethodContext) -> dict[str, MypyType] | None:
-    num_args = len(ctx.arg_kinds)
-    kwargs = {}
+    kwargs: dict[str, MypyType] = {}
     named = (ARG_NAMED, ARG_NAMED_OPT)
-    for i in range(num_args):
-        if not ctx.arg_kinds[i]:
-            continue
-        if any(kind not in named for kind in ctx.arg_kinds[i]):
-            # Only named arguments supported
-            continue
-        for j in range(len(ctx.arg_names[i])):
-            name = ctx.arg_names[i][j]
-            assert name is not None
-            kwargs[name] = ctx.arg_types[i][j]
+    for slot in zip(ctx.arg_kinds, ctx.arg_types, ctx.arg_names, ctx.args, strict=False):
+        for kind, arg_type, name, arg in zip(*slot, strict=False):
+            if kind in named:
+                assert name is not None
+                kwargs[name] = arg_type
+            elif kind == ARG_POS:
+                alias = _aggregate_default_alias(arg, arg_type)
+                if alias is not None:
+                    kwargs[alias] = arg_type
     return kwargs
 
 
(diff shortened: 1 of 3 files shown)


tests/typecheck/managers/querysets/test_annotate.yml
```diff
@@ -152,14 +152,15 @@
                     username = models.CharField(max_length=100)
 
 
--   case: annotate_no_field_name
+-   case: annotate_positional_aggregate_uses_default_alias
```
federicobond, line 155: I should merge the new test case into this one actually.

mypy_django_plugin/transformers/querysets.py
```diff
@@ -227,20 +229,52 @@ def extract_proper_type_queryset_values_list(ctx: MethodContext, django_context:
...
         if not ctx.arg_kinds[i]:
             continue
-        if any(kind not in named for kind in ctx.arg_kinds[i]):
-            # Only named arguments supported
-            continue
-        for j in range(len(ctx.arg_names[i])):
-            name = ctx.arg_names[i][j]
-            assert name is not None
-            kwargs[name] = ctx.arg_types[i][j]
+        for j, kind in enumerate(ctx.arg_kinds[i]):
```
federicobond: Thought about iterating with zip(ctx.arg_kinds, ctx.arg_types, ctx.arg_names, ctx.args) instead of the double indices, what do you think?

```python
for slot in zip(ctx.arg_kinds, ctx.arg_types, ctx.arg_names, ctx.args):
    for kind, arg_type, name, arg in zip(*slot):
        ...
```

UnknownPlatypus: Could be just that no ? 

```python
for kind, arg_type, name, arg in zip(
    ctx.arg_kinds, ctx.arg_types, ctx.arg_names, ctx.args, strict=True
):
    ...
```

But yeah I'm all for it, should simplify greatly

mypy_django_plugin/transformers/querysets.py
```diff
@@ -227,20 +229,52 @@ def extract_proper_type_queryset_values_list(ctx: MethodContext, django_context:
     return ret
 
 
+def _aggregate_default_alias(arg: Expression, arg_type: MypyType) -> str | None:
+    """Mirror `Aggregate.default_alias`: ``<source_expression>__<aggregate_name_lower>``.
+
+    Django generates a default alias only for aggregates with a single source
+    expression that has a `name`; complex expressions raise at runtime, so the
+    plugin matches that and returns ``None`` for anything else.
```
UnknownPlatypus, line 237: I did not know this existed 🤯 , maybe link to the documentation that showcases this ?

https://docs.djangoproject.com/en/dev/topics/db/aggregation/#generating-aggregates-for-each-item-in-a-queryset

mypy_django_plugin/transformers/querysets.py
```diff
@@ -227,20 +229,52 @@ def extract_proper_type_queryset_values_list(ctx: MethodContext, django_context:
...
+    """
+    if not isinstance(arg, CallExpr):
+        return None
+    proper = get_proper_type(arg_type)
+    if not isinstance(proper, Instance):
+        return None
+    if not proper.type.has_base(fullnames.AGGREGATE_CLASS_FULLNAME):
+        return None
```
UnknownPlatypus:
```suggestion
    if (
        not isinstance(arg, CallExpr)
        or not isinstance(proper_arg := get_proper_type(arg_type), Instance)
        or not proper_arg.type.has_base(fullnames.AGGREGATE_CLASS_FULLNAME)
    ):
        return None
```
or 
```suggestion
    if not (
        isinstance(arg, CallExpr)
        and isinstance(proper_arg := get_proper_type(arg_type), Instance)
        and proper_arg.type.has_base(fullnames.AGGREGATE_CLASS_FULLNAME)
    ):
        return None
```
wdyt ?

mypy_django_plugin/transformers/querysets.py
```diff
@@ -227,20 +229,52 @@ def extract_proper_type_queryset_values_list(ctx: MethodContext, django_context:
...
+    source: Expression | None = None
+    for a, k in zip(arg.args, arg.arg_kinds, strict=False):
+        if k != ARG_POS:
+            continue
+        if source is not None:
+            return None
+        source = a
```
UnknownPlatypus:
```suggestion
    for arg, kind in zip(arg.args, arg.arg_kinds, strict=False):
        if kind != ARG_POS:
            continue
        if source is not None:
            return None
        source = arg
```

mypy_django_plugin/transformers/querysets.py
```diff
@@ -227,20 +229,52 @@ def extract_proper_type_queryset_values_list(ctx: MethodContext, django_context:
...
+    # TODO: consider switching to `helpers.resolve_string_attribute_value` once
```
UnknownPlatypus, line 254: would be nice 👍 Can be done as a followup tho

UnknownPlatypus (commented):
Looking great, a few suggestions

mypy_django_plugin/transformers/querysets.py
federicobond: There is a nested loop there because the AST shape forces you to iterate over the parameter slots first (`*args`, then `**kwargs`), and for each parameter slot it contains the actual caller arguments bound to that slot).

Issue #1384, opened 2023-03-01
Hygienic mypy plugin: don't import Django into the mypy process


intgr:
We have no immediate plans to implement this, but I decided to open this issue for discussions, to split it out from #1337.

Original post by @intgr https://github.com/typeddjango/django-stubs/issues/1337#issuecomment-1401668445

> The mypy plugin should be "hygienic" with regards to Django code being checked. It should not directly import project code or Django into the mypy runtime, so it doesn't get entangled with the code it's checking. Instead it could for example invoke subprocesses to discover Django internals.
>
> The current implementation is problematic for many reasons. For instance mypy shouldn't crash with traceback when there's a syntax error in Django settings. And when using mypy daemon dmypy, it should be capable of re-discovering changed models files, instead of using obsolete state or randomly crashing, as it does now.

intgr:
Comment by @adamchainz https://github.com/typeddjango/django-stubs/issues/1337#issuecomment-1414395472

> Probably a bit of a brain twister but would be nice to see. I think it would also make it a bit easier to separate out what needs discovering and open the door for other type checkers.

intgr:
Original comment by @sobolevn https://github.com/typeddjango/django-stubs/issues/1337#issuecomment-1449789551

> Since we discussing a lot of things in this issue, I will add my 2c about "Same with the hygienic plugin redesign".
>
> I think it is a really bad idea. When we were originally designing the current state of the plugin, we already had static-analysis-only previous version of the plugin. It was a nightmare.
> 
> Since django has so many magic things happening in the model definition, it is almost impossible to get right.
> 
> And running django during type checking is not a bad thing. For example, django-check do this as well. So, this is rather standard.
> 
> So, I highly recommend not to change this.

intgr:
Original comment by @flaeppe https://github.com/typeddjango/django-stubs/issues/1337#issuecomment-1449863734

> I've touched the plugin code a bit and I'd would like to agree with both of your opinions regarding the hygienic plugin redesign. I've seen a bit of "shortcutting", in regards to choosing to use Django (runtime) internals to resolve types in some ends. What I mean is that it would/could be possible to use mypy's internals to figure out correct types, but Django's has been used instead, probably because it's more straight forward and simple. It could very well be possible what mypy offers is/was (currently) insufficient too.
> 
[...]

PR #3376, opened 2026-05-10
Hygienic mypy plugin Phase 0: route direct Django access through DjangoContext

federicobond:
First step toward a hygienic mypy plugin, as discussed in #1384. Every direct access to `_meta`, `apps_registry`, and `settings` outside `DjangoContext` is collapsed into wrapper methods on the context. Return types still expose live Django objects; the goal is a single chokepoint we can iterate on without touching every transformer again.

### What I'd like review on

Don't focus too much on the method implementations for now. I'd like to validate the general direction first, and make sure we're happy with these method signatures at least as a transitional API while we figure out the next step.

If this is too big to safely merge in one go, I can split it up by method group, but I think the overall direction is easier to review when seen all together.

## Related issues

Refs https://github.com/typeddjango/django-stubs/issues/1384

## AI Policy

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -116,14 +116,117 @@ class DjangoContext:
         self.django_settings_module = django_settings_module
 
         apps, settings = initialize_django(self.django_settings_module)
-        self.apps_registry = apps
-        self.settings = settings
+        self._apps_registry = apps
+        self._settings = settings
+
+    # ---- Registry wrappers ----
+
+    def get_model_class_by_label(self, label: str) -> type[Model] | None:
+        try:
+            return self._apps_registry.get_model(label)
+        except LookupError:
+            return None
+
+    def is_app_installed(self, dotted_or_label: str) -> bool:
+        return self._apps_registry.is_installed(dotted_or_label)
+
+    # ---- Settings wrappers ----
+
+    def get_setting(self, name: str, default: Any = None) -> Any:
+        return getattr(self._settings, name, default)
+
+    def has_setting(self, name: str) -> bool:
+        return hasattr(self._settings, name)
+
+    @cached_property
+    def auth_user_model_label(self) -> str:
+        return self._settings.AUTH_USER_MODEL
+
+    @cached_property
+    def default_auto_field_path(self) -> str:
+        return self._settings.DEFAULT_AUTO_FIELD
+
+    # ---- Single-model meta wrappers ----
+
+    def get_field_on_model(self, model_cls: type[Model], field_name: str) -> Field[Any, Any] | ForeignObjectRel | None:
+        try:
+            return model_cls._meta.get_field(field_name)
+        except FieldDoesNotExist:
+            return None
+
+    def is_model_abstract(self, model_cls: type[Model]) -> bool:
+        return model_cls._meta.abstract
+
+    def get_proxy_target(self, model_cls: type[Model]) -> type[Model] | None:
+        return model_cls._meta.proxy_for_model
+
+    def get_auto_field(self, model_cls: type[Model]) -> Field[Any, Any] | None:
+        return model_cls._meta.auto_field
+
+    def get_pk_field_or_none(self, model_cls: type[Model]) -> Field[Any, Any] | None:
+        # Non-raising parallel of `get_primary_key_field`.
+        return model_cls._meta.pk
+
+    def get_pk_fields(self, model_cls: type[Model]) -> tuple[Field[Any, Any], ...]:
+        # Composite-PK ready (Django 5.2+); falls back to the single PK on older Django.
+        pk_fields = getattr(model_cls._meta, "pk_fields", None)
+        if pk_fields is not None:
+            return tuple(pk_fields)
+        pk = self.get_pk_field_or_none(model_cls)
+        return (pk,) if pk is not None else ()
+
+    def get_model_parents(self, model_cls: type[Model]) -> Sequence[type[Model]]:
+        # Wraps `_meta.all_parents` (Django 5.2+) / `_meta.get_parent_list()` for older Django.
+        opts = model_cls._meta
+        parents = getattr(opts, "all_parents", None)
+        if parents is not None:
+            return list(parents)
+        return opts.get_parent_list()
+
+    def iter_reverse_relations(self, model_cls: type[Model]) -> Iterator[ForeignObjectRel]:
+        # Public-API equivalent of `_meta.related_objects` (private API).
+        # Matches `include_hidden=True` to preserve behaviour for hidden M2M etc.
+        for field in model_cls._meta.get_fields(include_hidden=True):
+            if isinstance(field, ForeignObjectRel):
+                yield field
+
+    # ---- Manager wrappers ----
+
+    def iter_managers(self, model_cls: type[Model]) -> Iterator[tuple[str, models.Manager[Any]]]:
+        yield from model_cls._meta.managers_map.items()
+
+    def get_default_manager_class(self, model_cls: type[Model]) -> type[models.Manager[Any]]:
+        return model_cls._meta.default_manager.__class__  # type: ignore[return-value]
+
+    # ---- Query helpers ----
+
+    def resolve_names_to_path(self, model_cls: type[Model], parts: Sequence[str]) -> tuple[_AnyField, Sequence[str]]:
+        """Wrap `Query(model).names_to_path(parts, _meta)` so callers don't touch `_meta`.
+
+        Returns `(final_field, remainder)`. Raises Django's `FieldError` for the caller
+        to translate into a user-facing diagnostic.
+        """
+        _, final_field, _, remainder = Query(model_cls).names_to_path(parts, model_cls._meta)
+        return final_field, remainder
+
+    def iter_select_related_choices(self, model_cls: type[Model]) -> Iterator[str]:
+        """Names valid for `QuerySet.select_related(...)` on this model.
+
+        Mirrors Django's `SQLCompiler.get_related_selections._get_field_choices`:
+        forward FK/O2O field names plus reverse-unique relation query names.
+        """
+        for field in model_cls._meta.fields:
+            if field.is_relation:
+                yield field.name
+        for rel in model_cls._meta.related_objects:
+            if rel.field.unique:
+                yield rel.field.related_query_name()
 
     @cached_property
     def model_modules(self) -> dict[str, dict[str, type[Model]]]:
         """All modules that contain Django models."""
         modules: dict[str, dict[str, type[Model]]] = defaultdict(dict)
-        for concrete_model_cls in self.apps_registry.get_models(include_auto_created=True, include_swapped=True):
+        for concrete_model_cls in self._apps_registry.get_models(include_auto_created=True, include_swapped=True):
             modules[concrete_model_cls.__module__][concrete_model_cls.__name__] = concrete_model_cls
             # collect abstract=True models
             for model_cls in concrete_model_cls.mro()[1:]:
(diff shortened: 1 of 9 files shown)


mypy_django_plugin/django/context.py
```diff
@@ -116,14 +116,117 @@ def __init__(self, django_settings_module: str) -> None:
...
+    def iter_reverse_relations(self, model_cls: type[Model]) -> Iterator[ForeignObjectRel]:
+        # Public-API equivalent of `_meta.related_objects` (private API).
+        # Matches `include_hidden=True` to preserve behaviour for hidden M2M etc.
+        for field in model_cls._meta.get_fields(include_hidden=True):
+            if isinstance(field, ForeignObjectRel):
+                yield field
+
+    # ---- Manager wrappers ----
+
+    def iter_managers(self, model_cls: type[Model]) -> Iterator[tuple[str, models.Manager[Any]]]:
```
sobolevn, line 195: Design question: is it fine for a separate mypy plugin to return `models.Manager`? Can we theoretically construct a manager with importing Django and app models?

federicobond: Most likely not, but the intent of this phase is to make those leaks more visible via the single `DjangoContext` chokepoint, so we can later refactor the API to avoid returning live Django objects.

mypy_django_plugin/django/context.py
```diff
@@ -116,14 +116,117 @@ def __init__(self, django_settings_module: str) -> None:
...
+    # ---- Single-model meta wrappers ----
+
+    def get_field_on_model(self, model_cls: type[Model], field_name: str) -> Field[Any, Any] | ForeignObjectRel | None:
+        try:
+            return model_cls._meta.get_field(field_name)
+        except FieldDoesNotExist:
+            return None
+
+    def is_model_abstract(self, model_cls: type[Model]) -> bool:
+        return model_cls._meta.abstract
```
UnknownPlatypus, line 158: A lot of these are static (they don't use self) maybe add `@staticmethod` to these ?

You can enable the ruff rule [`plr6301`](https://docs.astral.sh/ruff/rules/no-self-use/#no-self-use-plr6301) to catch them

mypy_django_plugin/django/context.py
```diff
@@ -116,14 +116,117 @@ def __init__(self, django_settings_module: str) -> None:
...
+
+    def get_proxy_target(self, model_cls: type[Model]) -> type[Model] | None:
+        return model_cls._meta.proxy_for_model
+
+    def get_auto_field(self, model_cls: type[Model]) -> Field[Any, Any] | None:
+        return model_cls._meta.auto_field
+
+    def get_pk_field_or_none(self, model_cls: type[Model]) -> Field[Any, Any] | None:
+        # Non-raising parallel of `get_primary_key_field`.
```
UnknownPlatypus, line 167: This is not correct, `get_primary_key_field` is trying to the same thing in a different way and the `raise ValueError` seem to be to satisfy the typechecker mostly. Otherwise it's a crash, we don't seem to catch the possible valueerror anywhere.

The typing also suggest this can only return Field, never None
```python
class Options(Generic[_M]):
    pk: Field
```
We should probably investigate django runtime for this and remove one of the two helpers

mypy_django_plugin/django/context.py
```diff
@@ -116,14 +116,117 @@ def __init__(self, django_settings_module: str) -> None:
...
+    def resolve_names_to_path(self, model_cls: type[Model], parts: Sequence[str]) -> tuple[_AnyField, Sequence[str]]:
+        """Wrap `Query(model).names_to_path(parts, _meta)` so callers don't touch `_meta`.
+
+        Returns `(final_field, remainder)`. Raises Django's `FieldError` for the caller
+        to translate into a user-facing diagnostic.
+        """
+        _, final_field, _, remainder = Query(model_cls).names_to_path(parts, model_cls._meta)
+        return final_field, remainder
+
+    def iter_select_related_choices(self, model_cls: type[Model]) -> Iterator[str]:
```
UnknownPlatypus, line 212: We only use it in `_get_select_related_field_choices`, does it really make sense to have this intermediary version ? 

What about just having a `DjagnoContext.get_select_related_choices` hook that replaces both ?

mypy_django_plugin/lib/helpers.py
```diff
@@ -606,8 +606,8 @@ def resolve_string_attribute_value(attr_expr: Expression, django_context: Django
     if isinstance(attr_expr, MemberExpr):
         member_name = attr_expr.name
         if isinstance(attr_expr.expr, NameExpr) and attr_expr.expr.fullname == "django.conf.settings":
-            if hasattr(django_context.settings, member_name):
-                return getattr(django_context.settings, member_name)  # type: ignore[no-any-return]
+            if django_context.has_setting(member_name):
+                return django_context.get_setting(member_name)  # type: ignore[no-any-return]
```
UnknownPlatypus, line 610:
```suggestion
            if (setting_value := django_context.get_setting(member_name)):
                return setting_value # type: ignore[no-any-return]
```

mypy_django_plugin/lib/helpers.py
```diff
@@ -736,7 +736,7 @@ def get_model_from_expression(
         and isinstance(expr.expr, NameExpr)
         and f"{expr.expr.fullname}.{expr.name}" == fullnames.AUTH_USER_MODEL_FULLNAME
     ):
-        lazy_reference = django_context.settings.AUTH_USER_MODEL
+        lazy_reference = django_context.auth_user_model_label
```
UnknownPlatypus, line 739: Feels weird to me to stuff everything on `django_context`, `django_context.settings.AUTH_USER_MODEL` is a lot more clear to me about what we are fetching than this random attribute

mypy_django_plugin/django/context.py
```diff
@@ -116,14 +116,117 @@ def __init__(self, django_settings_module: str) -> None:
...
+    def has_setting(self, name: str) -> bool:
+        return hasattr(self._settings, name)
+
+    @cached_property
+    def auth_user_model_label(self) -> str:
+        return self._settings.AUTH_USER_MODEL
+
+    @cached_property
+    def default_auto_field_path(self) -> str:
+        return self._settings.DEFAULT_AUTO_FIELD
```
UnknownPlatypus, line 147: This is odd:
- why only these two have a `cached_property` (they did not have any before)
- why are the cached property not using the `get_settings` helper
- I find these properties a lot less explicit now, see https://github.com/typeddjango/django-stubs/pull/3376/changes#r3218639715

I think we should remove the `cached_property`, looks like premature optimization.
If settings lookup are expensive, then we should be doing caching at the `get_setting` layer, and if they are not, we should not do caching

mypy_django_plugin/django/context.py
```diff
@@ -116,14 +116,117 @@ def __init__(self, django_settings_module: str) -> None:
...
+
+    # ---- Single-model meta wrappers ----
+
+    def get_field_on_model(self, model_cls: type[Model], field_name: str) -> Field[Any, Any] | ForeignObjectRel | None:
```
UnknownPlatypus, line 151: naming is not super consistant, `get_field_on_model` but `get_auto_field` `get_pk_field` later.

Again this is a bit confusing on usage, maybe use somthing like:
- `get_model_field`
- `get_model_auto_field`
- `get_model_pk_field`
- ...

mypy_django_plugin/django/context.py
```diff
@@ -116,14 +116,117 @@ def __init__(self, django_settings_module: str) -> None:
...
+        if parents is not None:
+            return list(parents)
+        return opts.get_parent_list()
+
+    def iter_reverse_relations(self, model_cls: type[Model]) -> Iterator[ForeignObjectRel]:
+        # Public-API equivalent of `_meta.related_objects` (private API).
+        # Matches `include_hidden=True` to preserve behaviour for hidden M2M etc.
+        for field in model_cls._meta.get_fields(include_hidden=True):
+            if isinstance(field, ForeignObjectRel):
+                yield field
```
UnknownPlatypus, line 191: The implementation changed, is it expected ? 
I think it's better to use an internal api than duplicating one that will drift silently.

mypy_django_plugin/transformers/querysets.py
```diff
@@ -973,15 +973,22 @@ def extract_prefetch_related_annotations(ctx: MethodContext, django_context: Dja
...
-        return None
+    if resolve_pk and field_name == "pk":
+        pk = django_context.get_pk_field_or_none(model_cls)
+        resolved_name = pk.name if pk is not None else field_name
+    else:
+        resolved_name = field_name
+    field = django_context.get_field_on_model(model_cls, resolved_name)
+    if field is None:
+        ctx.api.fail(f"{model_cls.__name__} has no field named {resolved_name!r}", ctx.context)
+    return field
```
UnknownPlatypus, line 991: The implementation changed, is it expected ? 

Also should we instead add a `suppress_errors = True` flag to `get_field_on_model` that would be switch to False here ? Similarly to what mypy does with [`analyze_member_access`](https://github.com/python/mypy/blob/a9008315ef56aa3843b957d372989f455acdd7bb/mypy/checkmember.py#L166)

mypy_django_plugin/transformers/querysets.py
```diff
@@ -993,17 +1000,17 @@ def _check_field_concrete(ctx: MethodContext, field: _AnyField, field_name: str,
...
     *,
     attr_name: str | None = None,
 ) -> bool:
-    opts = model_cls._meta
-    all_pk_fields = set(getattr(opts, "pk_fields", [opts.pk]))
-    for parent in getattr(opts, "all_parents", opts.get_parent_list()):
-        all_pk_fields.update(getattr(parent._meta, "pk_fields", [parent._meta.pk]))
+    all_pk_fields: set[_AnyField] = set(django_context.get_pk_fields(model_cls))
+    for parent in django_context.get_model_parents(model_cls):
+        all_pk_fields.update(django_context.get_pk_fields(parent))
```
UnknownPlatypus, line 1013: I was going to ask why we don't use `django_context.get_pk_fields` here but `django_context.get_pk_fields` seem to only return one pk (?)

Can we unify the implementations ?

UnknownPlatypus (commented):
Overall I think this is a great idea but claude did some weird stuff I think (or maybe you, idk these days, but it looks a lot like claude in some places)

mypy_django_plugin/django/context.py
sobolevn: Let's please model our APIs without `@staticmethod` :)
This way we have more flexibility to use `self`, when we would need it. Moreover, we don't want to provide `DjangoContext.is_model_abstract(model)` API.

federicobond: Agree with @sobolevn here, the fact that they don't use self is just accidental here.

UnknownPlatypus: But they will never need self, they don't need coupling with django_context.
I'm not sure I really see the advantage versus regular helper functions, or is it that you explicitely want to make them less easy to use in other contexts ?

federicobond:
@UnknownPlatypus thanks for the thorough look, and sorry for the noise. I mentioned in the description that I was mainly after a sanity check on the direction before polishing the implementations, but I should have been more explicit about expectations.

I'll tighten the implementations (naming, the changed behaviors you flagged, etc.) and ping you again once it's in better shape.

mypy_django_plugin/django/context.py
federicobond: Not if all they do is access `_meta` on the model class, but if we move forward with the hygienic plugin roadmap as I see it, they will probably read from an internal cache or maybe even rpc to a separate process. They will also not take a `type[Model]` but a pure mypy type but that's a refactor for later.

federicobond: The main point of this phase is to surface what queries we are asking from the Django runtime and consolidating them in a single place so we can design a better abstraction.

UnknownPlatypus:
Thanks, sorry for the tone. 

Having to review my claude and my coworker claude all day is getting on my nerves.

But thanks for the work on this, will be much needed for recent mypy parallel work

PR #3379, opened 2026-05-12
Preserve row type when .annotate() is called on a generic QuerySet

federicobond:
The row type was not being preserved for the TypeVar branch, so I extracted a `_resolve_annotate_row_type` helper to reuse on both branches.

Refs https://github.com/typeddjango/django-stubs/issues/602#issuecomment-4427909661 (I think it may solve your issue @delfick but I haven't confirmed).

## AI Policy

### DIFF
--- a/tests/typecheck/managers/querysets/test_annotate.yml
+++ b/tests/typecheck/managers/querysets/test_annotate.yml
@@ -1012,3 +1012,16 @@
 
         class FooModel(models.Model):
             objects = FooManager()
+
+- case: values_then_annotate_on_generic_queryset_typevar
+  main: |
+    from typing import Any
+    from typing_extensions import reveal_type, TypeVar
+    from django.db import models
+
+    _Model = TypeVar("_Model", bound=models.Model)
+
+    def latest_max(qs: models.QuerySet[_Model]) -> dict[str, Any] | None:
+        annotated = qs.values("id").annotate(foo=models.Max("id"))
+        reveal_type(annotated)  # N: Revealed type is "django.db.models.query.QuerySet[django.db.models.base.Model, dict[str, Any]]"
+        return annotated.first()
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -408,6 +408,31 @@ def _extract_model_type_var_upper_bound(ctx: MethodContext) -> Instance | None:
     return None
 
 
+def _resolve_annotate_row_type(
+    api: TypeChecker,
+    default_return_type: Instance,
+    annotated_model: ProperType,
+    expression_types: dict[str, MypyType],
+) -> MypyType:
+    if len(default_return_type.args) <= 1:
+        return annotated_model
+    original_row_type = get_proper_type(default_return_type.args[1])
+    if isinstance(original_row_type, TypedDictType):
+        return api.named_generic_type(
+            "builtins.dict",
+            [api.named_generic_type("builtins.str", []), AnyType(TypeOfAny.from_omitted_generics)],
+        )
+    if isinstance(original_row_type, TupleType):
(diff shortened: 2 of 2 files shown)


tests/typecheck/managers/querysets/test_annotate.yml
```diff
@@ -1012,3 +1012,14 @@
...
+- case: values_then_annotate_on_generic_queryset_typevar
+  main: |
+    from typing import Any
+    from typing_extensions import TypeVar
+    from django.db import models
+
+    _Model = TypeVar("_Model", bound=models.Model)
+
+    def latest_max(qs: models.QuerySet[_Model]) -> dict[str, Any] | None:
+        return qs.values("id").annotate(foo=models.Max("id")).first()
```
sobolevn: Can we please add `reveal_type(qs.values("id").annotate(foo=models.Max("id")))` before the return?

Right now it is really hard to follow, why the type is `dict[str, Any]`

federicobond: Sure!

sobolevn (approved):
Thank you!

UnknownPlatypus:
@federicobond have you tried the test case in the issue ?

federicobond:
Yes, I tried and they should both work now, in some cases with the right adjustments to the annotation types.

sobolevn:
@federicobond you can contribute these tests as well, more tests in the complex areas - the better.

delfick:
oh cool, thanks @federicobond !!!

Seems this removes 7 of the errors for me, I still have 56 left, I hopefully should be able to do some more detective work on them :)

tests/typecheck/managers/querysets/test_annotate.yml
```diff
@@ -1012,3 +1012,16 @@
 
         class FooModel(models.Model):
             objects = FooManager()
+
+- case: values_then_annotate_on_generic_queryset_typevar
```
delfick, line 1016: Seems here is another case that shows a problem

```
- case: values_are_carried_through
  main: |
    from typing import Any
    from typing_extensions import reveal_type, TypeVar
    from django.db import models
    import myapp.models

    with_annotate = myapp.models.Thing.objects.all().annotate(foo=models.Max("id"))

    reveal_type(with_annotate) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.Thing@AnnotatedWith[TypedDict({'foo': Any})], myapp.models.Thing@AnnotatedWith[TypedDict({'foo': Any})]]"

    with_values = with_annotate.values("id").distinct()

    reveal_type(with_values) # N: Revealed type is "django.db.models.query.QuerySet[myapp.models.Thing@AnnotatedWith[TypedDict({'foo': Any})], TypedDict({'id': int})]"
  installed_apps:
    - myapp
  files:
    - path: myapp/__init__.py
    - path: myapp/models.py
      content: |
        from django.db import models
        from typing import Self

        class Thing(models.Model):
            pass

```

If I change `with_values = with_annotate.values("id").distinct()` to not have the `distinct()` then it passes.

edit: and seems similar for doing a slice `qs[:max_sample_results]`

delfick: also I've noticed that the .annotate will erase the custom queryset type if there is one.

delfick: I've run out of time to investigate, but seems for that new test case above, the fix will be something to do with https://github.com/typeddjango/django-stubs/blob/18413d5dbec6e2d2a1cf436ceee981b9306c1dc1/mypy_django_plugin/transformers/querysets.py#L508

federicobond: Yes, both cases should be fixed by https://github.com/typeddjango/django-stubs/pull/3383

PR #3383, opened 2026-05-13
Preserve row type in merge_annotations_from_custom_method

### LINKED ISSUE #3381: [6.0.3 regression] Wrong type for slice of annotated `values_list` query
This worked in 6.0.2 and fails in 6.0.3.

```python
from typing import Iterable
from django.db.models import Model, F

class Foo(Model):
    pass

def get_foo_ids() -> Iterable[int]:
    return Foo.objects.all().annotate(a=F("id")).values_list("id", flat=True)[:]
```

6.0.2 correctly types the slice as `QuerySet[Foo@AnnotatedWith[…], int]`, but 6.0.3 incorrectly types it as `QuerySet[Foo@AnnotatedWith[…], Foo@AnnotatedWith[…]]`. Full error from 6.0.3:

```
my_app/models.py:8: error: Incompatible return value type (got "QuerySet[Foo@AnnotatedWith[TypedDict({'a': Any})], Foo@AnnotatedWith[TypedDict({'a': Any})]]", expected "Iterable[int]")  [return-value]
Found 1 error in 1 file (checked 3 source files)
```

I bisected the problem to

- 7da3047575469282c2536b8248eff89a423288d5 (#3266, cc @UnknownPlatypus)

federicobond:
In the test example, before slicing the queryset type is `QuerySet[Foo@AnnotatedWith[...], int]` (model arg is the annotated Foo, row arg is int). 

The hook ran `copy_modified(args=[annotated_type, annotated_type])`, which threw away the int and replaced it with the annotated model. With this change, the hook swaps it to the annotated model only if it was previously a model too.

## Related issues

Fixes https://github.com/typeddjango/django-stubs/issues/3381

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -505,7 +505,16 @@ def merge_annotations_from_custom_method(ctx: MethodContext, django_context: Dja
 
     api = helpers.get_typechecker_api(ctx)
     annotated_type = get_annotated_type(api, django_model.typ, fields_dict=new_td)
-    return default_return_type.copy_modified(args=[annotated_type, annotated_type])
+    new_args: list[MypyType] = [annotated_type]
+    if len(default_return_type.args) > 1:
+        original_row = get_proper_type(default_return_type.args[1])
+        if isinstance(original_row, Instance) and helpers.is_model_type(original_row.type):
+            new_args.append(annotated_type)
+        else:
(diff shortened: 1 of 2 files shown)


delfick:
Thanks! I can confirm this removes 72 errors from the django-stubs upgrade for us and we have no more "is not indexable" errors left after this 🥳

sobolevn (approved):
👍

PR #3385, opened 2026-05-13
Update mypy config to 2.0 defaults

sobolevn:


sobolevn:
Anyone who wants to pick this up is welcome, I don't have time right now to continue this work :(

PR #3387, opened 2026-05-13
Fix AdminSite.register typing.

noamkush:
This fixes the `AdminSite.register` calls that #3331 breaks (`error: Argument 2 to "register" of "AdminSite" has incompatible type`)

## AI Policy


tests/assert_type/contrib/admin/test_options.py
```diff
@@ -101,6 +101,9 @@ def a_method_action_with_message_user(self, request: HttpRequest, queryset: Quer
         self.message_user(request, _("Error message"), messages.ERROR)
 
 
+admin.site.register(FullAdminModel, FullModelAdmin)
```
sobolevn, line 108: Please, add more tests: 
- invalid functional registration
- decorator registration
- invalid decorator registration

sobolevn (commented):
Thanks!

tests/assert_type/contrib/admin/test_options.py
noamkush: Sure, replaced with yaml tests

sobolevn (commented):
It was a right place to test this, we just needed more examples :)

There's nothing mypy specific about these tests.

noamkush:
Done. Had to also change the decorator typing for that. There are some existing positive tests for the decorators already in `test_decorators.py`.

django-stubs/contrib/admin/decorators.pyi
```diff
@@ -61,5 +61,5 @@ def display(
     empty_value: str | None = ...,
 ) -> Callable[[_F], _F]: ...
 def register(
-    *models: type[Model], site: AdminSite | None = ...
-) -> Callable[[type[_ModelAdmin]], type[_ModelAdmin]]: ...
+    *models: type[_ModelT], site: AdminSite | None = ...
+) -> Callable[[type[ModelAdmin[_ModelT]]], type[ModelAdmin[_ModelT]]]: ...
```
sobolevn: This will change all types from precise ones to `type[ModelAdmin[ModelType]]`

Please, add a `assert_type` test for `DisplayModelAdmin`, it must be still `DisplayModelAdmin` after the `@register` decorator is applied

noamkush: I'm not sure how to achieve both. I can revert this but then using the wrong model doesn't give an error. Is there a way to achieve both or should I just revert this?

sobolevn: This is complex :(
I would say that we should keep the backwards compatibility and not break things :(

noamkush: Okay, I reverted this

tests/assert_type/contrib/admin/test_decorators.py
```diff
@@ -129,11 +129,11 @@ def method_action_http_response(self, request: HttpRequest, queryset: QuerySet[A
     def method_action_file_response(self, request: HttpRequest, queryset: QuerySet[ActionModel]) -> FileResponse: ...
 
 
-@admin.register(ActionModel)  # type: ignore[arg-type]  # pyrefly: ignore[no-matching-overload]  # ty: ignore[invalid-argument-type]  # pyright:ignore[reportArgumentType]
+@admin.register(ActionModel)
 class WrongModelAdmin(admin.ModelAdmin[DisplayModel]):
```
sobolevn, line 150: Please add a comment explaining that this is actually wrong, but we don't have any ways to represent the error right now.

noamkush: Added

sobolevn (approved):
Thanks!

PR #3388, opened 2026-05-13
Fix `related_model` type for mypy 2.1

UnknownPlatypus:
# I have made things!

I removed the `| Literal["self"]` because they where inconsistent. The [two methods](https://github.com/search?q=repo%3Adjango%2Fdjango%20%22def%20related_model%22&type=code) are proxying to `self.remote_field.model` which is typed as `type[Model]`

In practice user will never see this surface

## Related issues
Fixes #3385

## AI Policy


sobolevn (approved):

mypy.ini
```diff
@@ -8,8 +8,6 @@ incremental = true
 # TODO: add type args to all generics
 disallow_any_generics = false
 strict = true
-allow_redefinition_new = true
```
ngnpope, line 11: This should have been changed to the following rather than removed:

```suggestion
allow_redefinition = true
```

In mypy v2.0, `allow_redefinition` became `allow_redefinition_old` and `allow_redefinition_new` became `allow_redefinition`.

PR #3390, opened 2026-05-14
Re-enable `empty-body` error code for mypy and ty

ngnpope:
# I have made things!

Empty function bodies are unsound - they implicitly return `None` which makes them incompatible with non-`None` return types. This follows along with the approach taken in other projects such as these:

- https://github.com/python/typing/pull/2134
- https://github.com/sbdchd/django-types/pull/339

Instead of using `...` or `pass`, use `raise NotImplementedError`.

## Related issues

N/A

## AI Policy

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -11,7 +11,6 @@ strict = true
 strict_equality_for_none = true
 warn_unreachable = true
 
-disable_error_code = empty-body
 enable_error_code =
     deprecated,
     exhaustive-match,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -129,9 +129,6 @@ error-on-warning = true
 # Respect only `# ty: ignore` (ignore `# type: ignore`)
 respect-type-ignore-comments = false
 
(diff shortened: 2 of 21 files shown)


UnknownPlatypus:
What is the rationale? 
It seem to make maitenance a bit more complicated and to add noise in tests.
I don’t find it likely to catch any meaningful mistake ?

ngnpope:
Empty function bodies are unsound - they implicitly return `None` which makes them incompatible with non-`None` return types. This follows along with the approach taken in other projects such as these:

- https://github.com/python/typing/pull/2134
- https://github.com/sbdchd/django-types/pull/339

_(Sorry... GitHub must have lost some of what I wrote in the original description...)_

sobolevn (approved):
It looks really clunky but it is more correct.

UnknownPlatypus:
Thanks @ngnpope for the context, makes sense. I've amended the pr description, this is relevant context

PR #3393, opened 2026-05-15
Enable `unused-ignore` code in pyrefly

UnknownPlatypus:
# I have made things!
This is surprisingly not the default compared to the other typechecker.
It's really important in type assertion tests otherwise we miss regression

## AI Policy

### DIFF
--- a/tests/assert_type/utils/test_functional.py
+++ b/tests/assert_type/utils/test_functional.py
@@ -17,7 +17,7 @@ class Foo:
     def attr(self) -> list[str]:
         raise NotImplementedError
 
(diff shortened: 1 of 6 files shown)


tests/assert_type/utils/test_functional.py
```diff
@@ -58,7 +58,7 @@ def attr(self) -> list[str]:  # type: ignore[override]  # pyright: ignore[report
 assert_type("foo" + s, str)
 assert_type(s % "asd", str)
 
-s.nonsense  # type: ignore[attr-defined]  # pyrefly: ignore[no-attribute]
```
UnknownPlatypus, line 61: This is unfortunate but correct because the stub define `__getattribute__`:

```python
# Mypy requires this for the attribute hook to take effect
def __getattribute__(self, name: str, /) -> Any: ...  # type: ignore[explicit-override]
```

PR #3430, opened 2026-06-16
Fix only pk validation for abstract models

Lotram:
## Related issues

Basically, when a model is abstract, it has no `pk`, so the call to `self.get_meta().pk.name` inside `add_immediate_loading` fails. Test in  `tests/typecheck/managers/querysets/test_defer_only.yml has minimal example to reproduce. I've manually checked that it fails for Django 4 / 5 / 6. 

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -1238,7 +1238,7 @@ def validate_order_by(ctx: MethodContext, django_context: DjangoContext) -> Mypy
 def _validate_defer_only_fields(
     ctx: MethodContext, model_cls: type[Model], field_names: list[str], *, is_defer: bool
 ) -> None:
-    if not is_defer and model_cls._meta.abstract and "pk" in field_names:
+    if not is_defer and model_cls._meta.abstract:
         # Abstract models do not have a primary key field, but Query.add_immediate_loading()
         # assumes that one exists when resolving the "pk" alias.
         field_names = [field_name for field_name in field_names if field_name != "pk"]
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -1238,6 +1238,11 @@ def validate_order_by(ctx: MethodContext, django_context: DjangoContext) -> Mypy
 def _validate_defer_only_fields(
     ctx: MethodContext, model_cls: type[Model], field_names: list[str], *, is_defer: bool
 ) -> None:
+    if not is_defer and model_cls._meta.pk is None and "pk" in field_names:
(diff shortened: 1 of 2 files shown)


tests/typecheck/managers/querysets/test_defer_only.yml
```diff
@@ -218,6 +218,31 @@
...
 
+-   case: only_pk_on_abstract_model_no_crash
+    installed_apps:
+        - myapp
+    main: |
+        from django.db.models import QuerySet
+        from myapp.models import AbstractArticle
+
+        def test(qs: QuerySet[AbstractArticle]) -> None:
+            qs.only("pk")
```
sobolevn: Can you please use `reveal_type` here? What should be the result?

Lotram: I added a `reveal_type` call. 

> What should be the result?

I'm not sure I understand what you mean here. Are you asking me to add an assertion on the result type, or are you really wondering what the result should be ?

sobolevn: Sorry, I meant that "we should add a `reveal_type` call and specify the result with `# N: ...`" :)

Lotram:
I investigated the failing CI jobs, and it apparently fails because `pk` can't be None:
```
class Options(Generic[_M]):
    pk: Field
```
That's wrong for an abstract model, but I guess I can't simply change this to `pk: Field | None`, that would probably impact other files. 

In my specific case, I can replace `model_cls._meta.pk is None` by `model_cls._meta.abstract`, it should work the same

mypy_django_plugin/transformers/querysets.py
```diff
@@ -1238,6 +1238,11 @@ def validate_order_by(ctx: MethodContext, django_context: DjangoContext) -> Mypy
 def _validate_defer_only_fields(
     ctx: MethodContext, model_cls: type[Model], field_names: list[str], *, is_defer: bool
 ) -> None:
+    if not is_defer and model_cls._meta.abstract and "pk" in field_names:
```
```suggestion
    if not is_defer and model_cls._meta.abstract:
```

we can just always ignore `pk`

tests/typecheck/managers/querysets/test_defer_only.yml
```diff
@@ -218,6 +218,32 @@
                     tags = models.ManyToManyField(Tag)
 
 
+-   case: only_pk_on_abstract_model_no_crash
+    installed_apps:
+        - myapp
+    main: |
+        from django.db.models import QuerySet
+        from myapp.models import AbstractArticle
+        from typing import reveal_type
```
sobolevn:
```suggestion
        from typing_extensions import reveal_type
```

sobolevn (approved):

PR #3433, opened 2026-06-18
Make `validators` consistent

brianhelba:
* Use `Iterable` for parameters, since the input is never sliced as a sequence
* Use `list` for attributes, since it's always actually stored as that type
* Use consistent import of `_ValidatorCallable`

## AI Policy


sobolevn (approved):
Thanks!

PR #3434, opened 2026-06-19
Allow aggregate filters to use expressions

Lotram:
Django accepts expressions such as `Exists(...)` as aggregate filters `(e.g Count(..., filter=Exists(..))`, but the stubs typed these parameters as Q | None, causing false-positive arg-type errors.

I used `BaseExpression` which looked like the right base class, because `Lookup`, `ExpressionWrapper` and `Subquery` can all be used as `filter` argument. 

I created a new yaml test file because I was not sure where to put this

## AI Policy
- [X] I have read and agree to the [AI Policy](https://github.com/typeddjango/django-stubs/blob/master/.github/AI_POLICY.md), removed any "Co-Authored-By" lines attributing coding agents, and manually reviewed the final result

### DIFF
--- a/django-stubs/contrib/postgres/aggregates/general.pyi
+++ b/django-stubs/contrib/postgres/aggregates/general.pyi
@@ -60,7 +60,7 @@ class StringAgg(OrderableAggMixin, Aggregate):
         delimiter: Any,
         *,
         distinct: bool = False,
-        filter: Q | None = None,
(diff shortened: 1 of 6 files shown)


UnknownPlatypus (approved):
Thx, yaml test is fine, we don’t have a much better way yet when models are involved

PR #3444, opened 2026-06-23
Typecheck save()/asave() update_fields against model fields

federicobond:
Validates the `update_fields` argument of Model.save() and Model.asave() against the model's actual fields.

I also moved the field helpers to a new module called `field_validation.py` so they could be reused across `save.py` and `querysets.py`.

## AI Policy

### DIFF
--- a/mypy_django_plugin/lib/field_validation.py
+++ b/mypy_django_plugin/lib/field_validation.py
@@ -1,0 +1,65 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from django.core.exceptions import FieldDoesNotExist
+
+if TYPE_CHECKING:
+    from django.db.models.base import Model
+    from django.db.models.options import _AnyField
+    from mypy.plugin import MethodContext
+
+
+def try_get_field(
+    ctx: MethodContext, model_cls: type[Model], field_name: str, *, resolve_pk: bool = False
+) -> _AnyField | None:
+    opts = model_cls._meta
+    resolved_name = opts.pk.name if resolve_pk and field_name == "pk" else field_name
+    try:
+        return opts.get_field(resolved_name)
+    except FieldDoesNotExist as e:
+        ctx.api.fail(str(e), ctx.context)
+        return None
+
+
+def check_field_concrete(ctx: MethodContext, field: _AnyField, field_name: str, method: str) -> bool:
+    if not field.concrete or field.many_to_many:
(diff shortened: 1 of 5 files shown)


mypy_django_plugin/lib/field_validation.py
sobolevn, line 1: This does not look like a transformer, more like a utility 🤔 
Maybe move it to https://github.com/typeddjango/django-stubs/tree/master/mypy_django_plugin/lib ?

tests/typecheck/models/test_save.yml
```diff
@@ -0,0 +1,99 @@
...
+                class Article(models.Model):
+                    title = models.CharField(max_length=200)
+                    content = models.TextField()
+                    published = models.BooleanField(default=False)
+                    author = models.ForeignKey(Author, on_delete=models.CASCADE)
+                    category = models.ForeignKey(Category, on_delete=models.CASCADE)
+                    tags = models.ManyToManyField(Tag)
+
+
+-   case: save_update_fields_invalid
```
sobolevn, line 53: please, add to the test:

- a model method 
- a property
- a class var
- a manager

all of them should not be considered model fields.

tests/typecheck/models/test_save.yml
```diff
@@ -0,0 +1,99 @@
...
+        article = Article()
+
+        # Unknown field name
+        article.save(update_fields=["nonexistent"])  # E: Article has no field named 'nonexistent'  [misc]
+
+        # Primary key is not an updatable field
+        article.save(update_fields=["id"])  # E: "save()" does not support primary key fields. Got "id"  [misc]
+
+        # ManyToMany is not a concrete field
+        article.save(update_fields={"tags"})  # E: "save()" can only be used with concrete fields. Got "tags"  [misc]
```
sobolevn, line 69: Please, also test `author__name` as a field.

sobolevn (commented):
Great feature! Looking forward to it!

mypy_django_plugin/lib/field_validation.py
federicobond: Sure! Makes a lot of sense.

tests/typecheck/models/test_save.yml
federicobond: All done!

sobolevn (approved):
Awesome! Thanks a lot!

PR #3459, opened 2026-06-30
Fix order_by false positives with JSONField key transforms

### LINKED ISSUE #3457: order_by refuses JSONField transforms
# Bug report

## What's wrong

I think this was introduced by #3178, although I'm guessing a bit.

`models.py` in a test project:
```python
from django.db import models

class Item(models.Model):
    data = models.JSONField()

Item.objects.order_by("data__field")
```

```console
$ uv run -p 3.14 -w 'django-stubs==6.0.0,mypy' mypy src
src/ds/models.py:8: error: Cannot resolve keyword 'field' into field or transform on 'data'.  [misc]
Found 1 error in 1 file (checked 8 source files)
```

This is fine at run-time, and turns into a JSON transform.  I've reduced this from a project that I'm currently trying to upgrade from django-stubs 5.2.9 (where this worked) to 6.0.6; it's reproducible with all versions between 6.0.0 and 6.0.6 inclusive.

## How is that should be

```console
$ uv run -p 3.14 -w 'django-stubs==5.2.9,mypy' mypy src
Success: no issues found in 8 source files
```

## System information

- OS:
- `python` version: 3.14.6
- `django` version: 5.2.15
- `mypy` version: 2.1.0
- `django-stubs` version: 6.0.6
- `django-stubs-ext` version: 6.0.6

federicobond:
JSONField key transforms (`data__key`) are resolved via `get_transform()`, not `get_lookups()`, so the check added in #3178 wrongly rejected them.

## Related issues

Fixes https://github.com/typeddjango/django-stubs/issues/3457

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -12,7 +12,6 @@ from django.db.models.fields.related_descriptors import (
     ReverseOneToOneDescriptor,
 )
 from django.db.models.fields.reverse_related import ForeignObjectRel
-from django.db.models.lookups import Transform
 from django.db.models.sql.query import Query
 from mypy.errorcodes import NO_REDEF
 from mypy.nodes import (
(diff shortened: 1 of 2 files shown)


sobolevn (approved):

sobolevn:
Wow! This was rather easy 😅

PR #3460, opened 2026-07-01
Fix manager resolution when QuerySet overrides as_manager

### LINKED ISSUE #2758: Custom as manager makes the test fail
# Bug report
I am creating stubs for django-cte (named django-cte-stubs), but I have some issues since they have their own as_manager:
https://github.com/dimagi/django-cte/blob/v2.0.0/django_cte/cte.py#L189 overridden 

## What's wrong

This test fails :
```yml
-   case: test
    main: |
        reveal_type(MyModel.objects)  # N: Revealed type is "myapp.models.ManagerFromMyQuerySet[myapp.models.MyModel]"

    installed_apps:
        - myapp
    files:
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                from django.db import models
                from typing import ClassVar, TypeVar, Any

                    def just_int(self) -> int: ...

                    @classmethod
                    def as_manager(self) -> models.Manager[Any]:  # <- this is the important part
                        return super().as_manager()

                class MyModel(models.Model):


```

the output :
```
E     main:2: note: Revealed type is "myapp.models.UnknownManager[myapp.models.MyModel]" (diff)
E     myapp/models:13: error: Could not resolve manager type for "myapp.models.MyModel.objects"  [django-manager-missing] (diff)
E   Expected:
E     main:2: note: Revealed type is "myapp.models.ManagerFromMyQuerySet[myapp.models.MyModel]" (diff)
E     E: ...ed type is "myapp.models.ManagerFromMyQuerySet[myapp.models.MyModel]"
E     A: ...ed type is "myapp.models.UnknownManager[myapp.models.MyModel]"
E                                    ^

```

Note if we remove the 
```python
                    @classmethod
                    def as_manager(self) -> models.Manager[Any]:  # <- this is the important part
                        return super().as_manager()
```
the test works

i.e. Overriding as_manager in MyQuerySet causes MyQuerySet.as_manager() to raise a django-manager-missing mypy error. This does not happen if the method is not overridden.

## System information

- OS:
- `python` version: 3.9
- `django` version: 4.2.* (but the same thing with 5.*)
- `mypy` version: 1.17.0 but also 1.16.0
- `django-stubs` version: master (any in 5.*, the bug was not present in 4.*)
- `django-stubs-ext` version: same as django-stubs

federicobond:
Overriding `as_manager()` made the hook mistake the override for its own generated method and bail out, so the manager was never generated and `objects = MyQuerySet.as_manager()` fell back to `UnknownManager`.

Also extracted a `_get_or_create_manager_for_queryset` helper function since `add_as_manager_to_queryset_class` was growing too big. 

## Related issues

Fixes #2758

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -536,8 +536,8 @@ def add_as_manager_to_queryset_class(ctx: ClassDefContext) -> None:
     if existing_as_manager is not None and existing_as_manager.plugin_generated:
         # We generated `as_manager` on a previous pass; nothing left to do.
         return
-    # A user-declared `as_manager` override reaches here and still needs a
-    # generated manager; remember it so we don't overwrite it below.
+    # If `as_manager` is still present, the user declared it (ours returned above).
+    # Track it so we generate the manager but don't overwrite their method below.
     user_defined_as_manager = existing_as_manager is not None
 
     base_as_manager = queryset_info.get("as_manager")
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thank you!

PR #3461, opened 2026-07-03
Speed up test suite ~1.8x by caching package version lookups

federicobond:
report_config_data is called by mypy once per cached module, and each call ran three importlib.metadata.version() lookups, each scanning dist-info metadata from disk. On a warm run over a django-importing module graph that is ~1,300 disk scans per mypy invocation, accounting for ~90% of mypy wall time in profiling.

Cache the lookups per process, including the not-installed case for django-stubs-ext (functools.cache does not memoize exceptions, so the sentinel is a None return instead).

This roughly halves the duration of a single warm test run, and makes the full django-stubs test suite under pytest -n auto ~1.8x faster while using ~40% less CPU. Behavior is unchanged: the emitted config data is identical, including omitting django_stubs_ext_version when the package is absent.

## AI Policy

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import importlib.metadata
 import itertools
 import sys
-from functools import cached_property, partial
+from functools import cache, cached_property, partial
 from typing import TYPE_CHECKING, Any
 
 from mypy.build import PRI_MED, PRI_MYPY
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thank you! Amazing one :)

sobolevn:
Difference: 

- New: https://github.com/typeddjango/django-stubs/actions/runs/28645987523/job/84952572740?pr=3461 takes 1m 46s
- Old: https://github.com/typeddjango/django-stubs/actions/runs/28494702557/job/84458480057 take 2m 31s

ngnpope:
😮 Wild. Good find! Will have to report back on how much that saves me on a big monorepo.

PR #3467, opened 2026-07-08
Update mypy to 2.2

sobolevn:
Since the mypy has changed its rules: now `__new__`'s return type is used as-is.
It means that our old pattern of 

```yaml
-   case: subclass_request_changed_user_type
    main: |
        class MyRequest(HttpRequest):
            user: User # Override the type of user

        request = MyRequest()
        reveal_type(request) # N: Revealed type is "main.MyRequest"
        reveal_type(request.user) # N: Revealed type is "django.contrib.auth.models.User"
    custom_settings: |
        INSTALLED_APPS = ('django.contrib.contenttypes', 'django.contrib.auth')

```

does not work anymore. `HttpRequest` can't use `__new__` anymore with:

```python
    # The magic. If we instantiate HttpRequest directly somewhere, it has
    # mutable GET and POST. However, both ASGIRequest and WSGIRequest have immutable,
    # so when we use HttpRequest to refer to any of them we want exactly this.
    # Case when some function creates *exactly* HttpRequest (not subclass)
    # remain uncovered, however it's probably the best solution we can afford.
    def __new__(cls) -> _MutableHttpRequest: ...
    # When both __init__ and __new__ are present, mypy will prefer __init__
    # (see comments in mypy.checkmember.type_object_type)
    # def __init__(self) -> None: ...
```

Because all the subclasses will be ignored. `_MutableHttpRequest` will always be used instead of `MyRequest()`

So, we would have to break the feature with mutable `GET` and `POST`. At least for now.
Ideas on how to represent mutable / immutable `GET` / `POST` are welcome.


brianhelba:
@sobolevn Could you open an issue for the mutable `HttpRequest` regression?

It's a common pattern in testing to mutate `HttpRequest.GET` / `HttpRequest.POST`.

sobolevn:
I created https://github.com/typeddjango/django-stubs/issues/3565

PR #3476, opened 2026-07-10
Fix #3472: allow None for nullable FK/OneToOne _id lookups

### LINKED ISSUE #3472: Cannot filter with None on `_id`  attributes of FK/OneToOne
# Bug report

The validation of allowed filters seem to be incorrect for the `_id` variant of Fk/OneToOne fields

## What's wrong

```python
class MyModel(models.Model):
    prospect = models.OneToOneField(
        "foo.OtherModel"
        blank=True,
        null=True,
        on_delete=models.CASCADE,
    )

MyModel.objects.filter(prospect=None) # OK
MyModel.objects.filter(prospect_id=None) # error: Incompatible type for lookup 'prospect_id': (got "None", expected "str | int")  [misc]
```

## How is that should be

```python
MyModel.objects.filter(prospect=None) # OK
MyModel.objects.filter(prospect_id=None) # OK
```

## System information

- OS: ubuntu 24.04
- python version: 3.11
- django version: 5.2.15
- mypy version: 2.1.0
- django-stubs version: 6.0.6
- django-stubs-ext version: 6.0.6

usamashehab:
## Related issues

Fixes #3472

## Problem

Filtering on the `_id` column of a nullable ForeignKey or OneToOneField with `None` was incorrectly rejected by mypy:

```python
class MyModel(models.Model):
    prospect = models.OneToOneField(OtherModel, blank=True, null=True, on_delete=models.CASCADE)

MyModel.objects.filter(prospect_id=None)  # error: Incompatible type for lookup 'prospect_id': (got "None", expected "str | int")
```

This has two root causes in `_resolve_field_from_parts`:

1. **Field name detection**: The old check compared `field_part` against `model_name + "_id"` (the related model's lowercase name + `"_id"`). This only worked when the FK field name happened to match the related model name (e.g. `publisher` → `Publisher`). When the field name differs (e.g. `prospect` → `OtherModel`), the check `"prospect_id" == "othermodel_id"` failed, and the field stayed as the `RelatedField` instead of being resolved to the PK column field.

2. **Nullability**: Even when the `_id` variant was correctly detected (e.g. for `publisher_id`), the resolved PK field always has `null=False`. So `get_field_lookup_exact_type` returned `str | int` without `None`, even for nullable FKs where `prospect_id=None` is valid at the database level.

## Fix

- Replaced the `model_name + "_id"` check with `field.attname`, which is the actual column name Django assigns to the FK's `_id` variant. This works regardless of whether the field name matches the model name.
- Added `is_nullable_fk_id` tracking: when a FK `_id` attname is detected, the original FK's `null` attribute is preserved and passed through to `get_field_lookup_exact_type`, so `None` is accepted for nullable `_id` lookups.
- Excluded `ManyToManyField` via `field_part != field.name` (M2M fields have `attname == name`, no `_id` column variant exists).

## AI Policy

### DIFF
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -5,7 +5,7 @@ import sys
 from collections import defaultdict
 from contextlib import contextmanager
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, NamedTuple, Literal
+from typing import TYPE_CHECKING, Any, Literal, NamedTuple
 
 from django.core.exceptions import FieldDoesNotExist, FieldError
 from django.db import models
@@ -411,9 +411,7 @@ class DjangoContext:
 
         return related_model_cls
 
-    def _resolve_field_from_parts(
-        self, field_parts: Iterable[str], model_cls: type[Model]
-    ) -> ResolvedLookupField:
+    def _resolve_field_from_parts(self, field_parts: Iterable[str], model_cls: type[Model]) -> ResolvedLookupField:
         """
         Resolve ``field_parts`` (as produced by ``Query.solve_lookup_type``) to a
         concrete field and the model it belongs to.  ``is_nullable`` is ``True``
@@ -606,9 +604,7 @@ class DjangoContext:
                 return AnyType(TypeOfAny.explicit)
 
         if lookup_cls is None or issubclass(lookup_cls, Exact):
-            return self.get_field_lookup_exact_type(
-                helpers.get_typechecker_api(ctx), field, is_nullable=is_nullable
-            )
+            return self.get_field_lookup_exact_type(helpers.get_typechecker_api(ctx), field, is_nullable=is_nullable)
 
         if issubclass(lookup_cls, In):
             exact_type = self.get_field_lookup_exact_type(
--- a/mypy_django_plugin/django/context.py
+++ b/mypy_django_plugin/django/context.py
@@ -5,7 +5,7 @@ import sys
 from collections import defaultdict
 from contextlib import contextmanager
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, NamedTuple, Literal
 
 from django.core.exceptions import FieldDoesNotExist, FieldError
 from django.db import models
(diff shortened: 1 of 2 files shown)


federicobond (commented):
The bare positional bool returned from `_resolve_field_from_parts` reads a bit awkwardly, and is discarded at most call sites. It also carries a fairly narrow special-case name.

What do you think about introducing a small namedtuple `ResolvedLookupField(field, model, is_nullable)`, where `is_nullable` is the effective nullability of the looked-up column? That way the resolver owns the nullability decision and hands it down as a general fact.

usamashehab:
Agreed, the bare bool was awkward. Switched to a ResolvedLookupField NamedTuple with (field, model, is_nullable). The resolver now owns the nullability decision and passes it down as a general fact rather than a narrow special-case. All 21 filter tests still pass.

federicobond:
Look, if you want to continue contributing to the project I suggest you re-read the AI policy you agreed to.

If I have to respond to a comment written entirely by an LLM I would rather work with the model directly instead of through a slow human intermediary. It's not bad to use AI but if you are not willing to put in the time to write a short reply in your own words it's just not worth it to take your contribution seriously.

usamashehab:
> Look, if you want to continue contributing to the project I suggest you re-read the AI policy you agreed to.
> 
> If I have to respond to a comment written entirely by an LLM I would rather work with the model directly instead of through a slow human intermediary. It's not bad to use AI but if you are not willing to put in the time to write a short reply in your own words it's just not worth it to take your contribution seriously.

my skills in writing is not good, that's why i am using AI to write a good response but seems that is not the best behavior, will try to write on my own

UnknownPlatypus:
> my skills in writing is not good, that's why i am using AI to write a good response but seems that is not the best behavior, will try to write on my own

yes pls, it's not an issue if your writing is not perfect, and practice is the only way to make it better. 
Don't let AI steal that from you.

mypy_django_plugin/django/context.py
```diff
@@ -408,16 +432,25 @@ def _resolve_field_from_parts(
...
+                if field_part == field.attname and field_part != field.name:
+                    # The lookup uses the FK's ``_id`` suffix (e.g. ``prospect_id``).
+                    # Django's ``get_field`` returns the ``RelatedField`` for this
+                    # attname, but semantically it refers to the underlying column,
+                    # so resolve to the related model's primary key field instead.
+                    # Preserve the original FK's nullability — the PK field itself
+                    # is always ``null=False`` but the column is nullable when the
+                    # FK is.
+                    # ``field_part != field.name`` excludes ManyToManyField, whose
+                    # ``attname`` equals its ``name`` (no ``_id`` column variant).
```
UnknownPlatypus, line 444: Can you simplify this comment ? lot's of noise
It should focus on explaining the if conditional, mostly the last 2 lines I think ?

Also let the comment take more width and avoid breaking in the middle of sentences, it should be readable for a human as much as possible

PR #3477, opened 2026-07-11
Fix ngettext/npgettext number parameter type mismatch

### LINKED ISSUE #3454: ngettext has different annotations to ngettext_lazy
Hello,

I've noticed that the `number` argument to ngettext is different between the lazy and non lazy versions

https://github.com/typeddjango/django-stubs/blob/789d40fc36f6e26851c22c5f1bad5520425647cf/django-stubs/utils/translation/__init__.pyi#L12-L20

In the non lazy version it's `float` but in the lazy version it's `int | str | None`

Should they both be `int | str | None`?

usamashehab:
## Related issues

Closes #3454

## Summary

The `number` parameter type was inconsistent across the translation functions:

- `ngettext` used `float`
- `npgettext` used `int`
- `ngettext_lazy` and `npgettext_lazy` used `int | str | None`

The `float` vs `int` mismatch between the two non-lazy functions was an oversight. Both functions accept any numeric value at runtime for plural selection, so I aligned them to `int | float`.

The lazy variants are intentionally left unchanged. Their `str` and `None` types enable deferred number resolution at template rendering time, as documented in Django's [lazy translations and plural](https://docs.djangoproject.com/en/6.1/topics/i18n/translation/#lazy-translations-and-plural) docs.

Added YAML test cases covering:
- `ngettext`/`npgettext` accepting both `int` and `float`
- `ngettext_lazy`/`npgettext_lazy` accepting `int`, `str`, and `None`
- `ngettext` rejecting `str` for the `number` parameter

## AI Policy


tests/typecheck/test_translation.yml
```diff
@@ -0,0 +1,53 @@
+-   case: ngettext_accepts_int_and_float
```
federicobond: Could these work as `assert_type`-style tests? They don't really need the `models.py` file and are just standard function signature tests.

usamashehab:
Good point, done. Replaced the YAML tests with assert_type checks under tests/assert_type/utils/test_translation.py — no models.py or installed_apps needed.

django-stubs/utils/translation/__init__.pyi
```diff
@@ -9,9 +9,9 @@ class TranslatorCommentWarning(SyntaxWarning): ...
 
 def gettext_noop(message: str) -> str: ...
 def gettext(message: str) -> str: ...
-def ngettext(singular: str, plural: str, number: float) -> str: ...
+def ngettext(singular: str, plural: str, number: int | float) -> str: ...
```
ngnpope: It's unnecessary to union `int` with `float`: https://typing.python.org/en/latest/spec/special-types.html#special-cases-for-float-and-complex

```suggestion
def ngettext(singular: str, plural: str, number: float) -> str: ...
```

django-stubs/utils/translation/__init__.pyi
```diff
@@ -9,9 +9,9 @@ class TranslatorCommentWarning(SyntaxWarning): ...
...
 def pgettext(context: str, message: str) -> str: ...
-def npgettext(context: str, singular: str, plural: str, number: int) -> str: ...
+def npgettext(context: str, singular: str, plural: str, number: int | float) -> str: ...
```
ngnpope:
```suggestion
def npgettext(context: str, singular: str, plural: str, number: float) -> str: ...
```

tests/assert_type/utils/test_translation.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+
+from django.utils.functional import _StrPromise
+from django.utils.translation import ngettext, ngettext_lazy, npgettext, npgettext_lazy
+from typing_extensions import assert_type
+
+# ngettext / npgettext accept int and float
```
ngnpope: You can simplify these tests to remove the `int` cases. It's enough to test that `float` works.

UnknownPlatypus:
> Good point, done. Replaced the YAML tests with assert_type checks under tests/assert_type/utils/test_translation.py — no models.py or installed_apps needed.

Thanks for the quick fix! Just make sure to re-read the [AI Policy](https://github.com/typeddjango/django-stubs/blob/master/.github/AI_POLICY.md) you agreed too, we don't allow ai generated pr comment and `—` is quite telling.

It matters because review discussion is where contributors and maintainers actually learn from each other. A generated comment skips the part where you explain your thinking and I we get to respond to it. Use AI as you wish in your development process but make sure to keep this shared space a healthy one

Thanks, hope that makes sense

PR #3479, opened 2026-07-13
Preserve values() aliases through annotate() for order_by

### LINKED ISSUE #3471: Annotation lost when chaining `.values()` and `.annotate()`
# Bug report

We seem to loose some annotations when chaining `values` and `annotate`

## What's wrong

```python
MyModel.objects.values(name=F("id")).order_by("name") # OK
MyModel.objects.values(name=F("id")).annotate(foo=F("id")).order_by( # E: error: Cannot resolve keyword 'name' into field
    "name"
)
```

## How is that should be

```python
MyModel.objects.values(name=F("id")).order_by("name") # OK
MyModel.objects.values(name=F("id")).annotate(foo=F("id")).order_by( # OK
    "name"
)
```
## System information

- OS:
- `python` version: 3.11
- `django` version: 5.2.15
- `mypy` version: 2.1.0
- `django-stubs` version: 6.0.6
- `django-stubs-ext` version:  6.0.6

federicobond:
When `.values(...)` is followed by `.annotate(...)`, the row TypedDict was collapsed to `dict[str, Any]`, discarding the selected column names. This caused a false-positive "Cannot resolve keyword" error on a subsequent `.order_by()` referencing a values() alias.

Extend the existing values-row TypedDict with the annotation aliases instead.

## Related issues

Fixes #3471

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/querysets.py
+++ b/mypy_django_plugin/transformers/querysets.py
@@ -417,10 +417,10 @@ def _resolve_annotate_row_type(
         return annotated_model
     original_row_type = get_proper_type(default_return_type.args[1])
     if isinstance(original_row_type, TypedDictType):
-        return api.named_generic_type(
-            "builtins.dict",
-            [api.named_generic_type("builtins.str", []), AnyType(TypeOfAny.from_omitted_generics)],
+        annotation_td = helpers.make_typeddict(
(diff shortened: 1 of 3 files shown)


sobolevn (approved):
Great stuff! 👍

PR #3488, opened 2026-07-14
Auto-enable `django_stubs_ext.monkeypatch()` for assert_type tests

UnknownPlatypus:
# I have made things!
This will allow migrating yml test that where using `monkeypatch: true`.

## Related issues
Followup to #3483

## AI Policy


UnknownPlatypus:
No completely convinced it's the right call. Also confused stubtest that now thinks all of these are defined.
I think the proper fix is to upstream everything, I think django will be fine with that

PR #3489, opened 2026-07-15
Defer `_meta.get_field()` when the model is defined later in the module

### LINKED ISSUE #2833: New "has no field named" error after upgrading to 5.2.3+ (missing a deferral?)
# Bug report

apologies I haven't been able to narrow this down yet -- but it appears only after upgrading:

```
src/sentry/models/group.py:87: error: Group has no field named 'short_id'  [misc]
```

however this model does have this field:
- https://github.com/getsentry/sentry/blob/0fb99c104713410075b9be24d4882f73564cd4aa/src/sentry/models/group.py#L87
- https://github.com/getsentry/sentry/blob/0fb99c104713410075b9be24d4882f73564cd4aa/src/sentry/models/group.py#L634

debugging it looks like the symbol *does* exist but may need a deferral pass?

around hereish: https://github.com/typeddjango/django-stubs/blob/82c99e83e3406fad91d43ff8fdc920df08f8d46f/mypy_django_plugin/django/context.py#L88
```
((Pdb)) info.get(field_name)
<mypy.nodes.SymbolTableNode object at 0x14a1e0310>
((Pdb)) info.get(field_name).node
<Var 'sentry.models.group.Group.short_id' at 0x14a1cfe20>
((Pdb)) info.get(field_name).node.type  # None
((Pdb)) info.get(field_name).node.is_ready
False
```

(`Var` being a not-yet-resolved alias I believe?)

## System information

- OS:
- `python` version: 3.13.1
- `django` version: modern
- `mypy` version: 1.17.1
- `django-stubs` version: 5.2.5 (forked)
- `django-stubs-ext` version: 5.2.5

federicobond:
## Related issues

Fixes #2833

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/meta.py
+++ b/mypy_django_plugin/transformers/meta.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from django.core.exceptions import FieldDoesNotExist
+from mypy.nodes import Var
 from mypy.types import AnyType, Instance, TypeOfAny, get_proper_type
 from mypy.types import Type as MypyType
 
@@ -27,6 +28,15 @@ def return_proper_field_type_from_get_field(ctx: MethodContext, django_context: 
     field_type = get_field_type_from_model_type_info(model_type.type, field_name)
     if field_type is not None:
         return field_type
+
+    # Fields of a model defined further down the module aren't inferred yet; defer to a later pass.
+    field_sym = model_type.type.get(field_name)
+    if field_sym is not None and isinstance(field_sym.node, Var) and not field_sym.node.is_ready:
+        typechecker_api = helpers.get_typechecker_api(ctx)
+        # Deferral re-checks the enclosing top-level function, so without one there is nothing to defer.
+        if typechecker_api.scope.top_level_function() is not None:
+            typechecker_api.handle_cannot_determine_type(field_name, ctx.context)
+            return ctx.default_return_type
 
     if (django_model := DjangoModel.from_model_type(model_type, django_context)) is None:
         return ctx.default_return_type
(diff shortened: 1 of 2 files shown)


mypy_django_plugin/transformers/meta.py
```diff
@@ -28,6 +29,15 @@ def return_proper_field_type_from_get_field(ctx: MethodContext, django_context:
     if field_type is not None:
         return field_type
 
+    # Fields of a model defined further down the module aren't inferred yet; defer to a later pass.
+    field_sym = model_type.type.get(field_name)
+    if field_sym is not None and isinstance(field_sym.node, Var) and not field_sym.node.is_ready:
```
UnknownPlatypus, line 34: Do we have (or is there) an helper to handle these deferrals? Also I don't see any defer() call actually, is that expected ?

tests/typecheck/models/test_meta_options.yml
```diff
@@ -93,6 +93,29 @@
...
+        -   path: myapp/models.py
+            content: |
+                from typing_extensions import reveal_type
+                from django.db import models
+
+                def get_field_before_model_is_defined() -> None:
+                    # Definition order is the test: this body is checked before `MyModel`'s fields are inferred.
+                    reveal_type(MyModel._meta.get_field('short_id'))  # N: Revealed type is "django.db.models.fields.IntegerField[float | int | str | django.db.models.expressions.Combinable, int]"
+
+                # A module-level lambda has nothing to defer, so this falls back to the runtime lookup.
```
UnknownPlatypus, line 113: I'm not sure I understand why this is

sobolevn (commented):
LGTM

PR #3490, opened 2026-07-15
Fix #2591: relax ModelBackend.authenticate signature to match BaseBackend contract

### LINKED ISSUE #2591: `ModelBackend.auhenticate` signature should be more relaxed
# Enhancement

## What's wrong

The current type stubs define the `authenticate` method of the `ModelBackend` (`django.contrib.auth.backends.ModelBackend`) as such:
```py
def authenticate(
    self, 
    request: HttpRequest | None, 
    username: str | None = ..., 
    password: str | None = ..., 
    **kwargs: Any,
) -> _UserModel | None: ...
```

The parent class, `BaseBackend` defines it as such:
```py
def authenticate(self, request: HttpRequest | None, **kwargs: Any) -> _UserModel | None: ...
```

When Django looks for which authentication backend to use, it'll always call said backend using the parameters as `kwargs`, until it finds the first backend that accepts the parameters and returns a valid user (or `None` if they all fail). The documentation itself instructs the user to define it's own `kwargs` when defining a custom authentication backend ([relevant docs](https://docs.djangoproject.com/en/5.2/topics/auth/customizing/#writing-an-authentication-backend)).

However, a very common use case here is to subclass `ModelBackend` instead of `BaseBackend`, in cases where we want to make simple adjustments to how authentication _is done_, but not how it's handled once you have the specific user; in this cases, you'll subclass the `ModelBackend`, which already has all the code to handle permissions that `BaseBackend` does not, and simply overwrite the `authenticate` method so that instead of, for example, a username and password, it uses a token or a code or something of the sorts. For example:
```py
class MyCustomBackend(ModelBackend):

    def authenticate(self, code: Optional[str] = None, **kwargs: Any) -> Optional[User]:
        return User.get_user_for_code(code)
```

In this simple example Mypy will give an "override" error: `Signature of "authenticate" incompatible with supertype "ModelBackend"`, even though this backend does not use the username / password at all.

### Alternatives
There are already alternatives to getting around this of course; simply `#type: ignore[override]` will do, or adding the typed arguments to the definition even though they are not used at all, as such:
```py 
class MyCustomBackend(ModelBackend):

    def authenticate(
        self, 
        username: Optional[str] = None, 
        password: Optional[str] = None, 
        code: Optional[str] = None, 
        **kwargs: Any
    ) -> Optional[User]:
        return User.get_user_for_code(code)
```

## How is that should be
The stubs for the `ModelBackend` should not override the parent class' method definition at all (it should not define it at all and simply inherit). This will be inline to how all authentication backends are used, and allow users to override with their own type definitions without extra work.

## System information

- OS: -
- `python` version: 3.13.2
- `django` version: 5.1.7
- `mypy` version: 1.15.0

AxionAgent:
### Problem
`ModelBackend.authenticate` was stubbed with explicit `username` / `password` params, which is narrower than `BaseBackend.authenticate(self, request, **kwargs)`. Django always invokes auth backends with credentials passed as `**kwargs`, so a subclass that overrides `authenticate` with custom credentials (e.g. a token or code) raised a mypy `[override]` error even though it never uses `username`/`password`:

```py
class MyCustomBackend(ModelBackend):
    def authenticate(self, request=None, code: str | None = None, **kwargs):
        ...
# Signature of "authenticate" incompatible with supertype "ModelBackend" [override]
```

### Fix
Relax `ModelBackend.authenticate` / `aauthenticate` to inherit the backend contract from `BaseBackend`:

```py
def authenticate(self, request: HttpRequest | None, **kwargs: Any) -> _User | None: ...
```

Subclasses can now override with custom keyword credentials without being forced to accept unused `username`/`password` parameters. `RemoteUserBackend` (which keeps `# type: ignore[override]`) is unaffected.

### Verification
- Reproduced with `mypy==1.15.0` + local stubs: before the fix the override raised `[override]`; after the fix `mypy` reports `Success: no issues found in 1 source file`.
- Added regression test `tests/assert_type/contrib/auth/test_backends.py` (subclass overriding with `code` / async `token` credentials type-checks cleanly; base contract still accepts keyword credentials).

The fix is additive to the public signature shape (only the narrow `username`/`password` literals are dropped in favor of `**kwargs`); no existing call sites break.

By Axion (automated contribution pipeline).


django-stubs/contrib/auth/backends.pyi
```diff
@@ -23,14 +23,13 @@ class BaseBackend:
...
 
 class ModelBackend(BaseBackend):
+    # Inherit `authenticate` / `aauthenticate` from `BaseBackend` so subclasses
+    # can override with custom (e.g. token-based) credentials without an
+    # [override] error. Django calls all backends with credentials as `**kwargs`.
     @override
-    def authenticate(
-        self, request: HttpRequest | None, username: str | None = ..., password: str | None = ..., **kwargs: Any
-    ) -> _User | None: ...
+    def authenticate(self, request: HttpRequest | None, **kwargs: Any) -> _User | None: ...
```
sobolevn, line 30: I don't think that this is a correct solution.
1. If we inherit it from `BaseBackend`, we don't need to duplicate it here
2. It loses the known kwargs part. How can this be solved?

UnknownPlatypus:
Closing per our [AI Policy](https://github.com/typeddjango/django-stubs/blob/master/.github/AI_POLICY.md), we do not allow unsupervised agents submitting pr

PR #3501, opened 2026-07-21
Resolve dynamic manager methods via mypy's `find_member`

UnknownPlatypus:
# I have made things!

Replace the manual type var substitution in `get_method_type_from_dynamic_manager` (`_process_dynamic_method`, `_has_compatible_type_vars`, `_replace_type_var`) with a single `find_member` lookup on the queryset instance, parametrized with the manager's model. 

mypy maps the class type vars, binds `Self` and strips `self` for us, including overloads and decorated methods.

This also drops the `is_fallback_queryset` argument which was never read.

## Related issues
None

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Final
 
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Great refactoring!

PR #3502, opened 2026-07-21
Add type arguments to generic references in admin filters, utils and per-app admins

UnknownPlatypus:
## Related issues
Part of https://github.com/typeddjango/django-stubs/issues/64

## AI Policy


django-stubs/contrib/admin/filters.pyi
```diff
@@ -73,7 +75,7 @@ class FieldListFilter(FacetsMixin, ListFilter):
         request: HttpRequest,
         params: dict[str, list[str]],
         model: type[Model],
```
sobolevn, line 77: what about this part?

UnknownPlatypus: not sure I understand, there is no generic param to add here ?

sobolevn: Oh, my mistake :)

sobolevn (approved):

PR #3503, opened 2026-07-21
Add type arguments to generic references in `core`, `apps`, `urls` and `admindocs`

UnknownPlatypus:
## Related issues
Part of https://github.com/typeddjango/django-stubs/issues/64

## AI Policy


django-stubs/core/files/uploadedfile.pyi
```diff
@@ -1,10 +1,10 @@
-from typing import IO
+from typing import IO, Any, AnyStr
 
 from django.core.files.base import File
 from django.utils.functional import cached_property
 from typing_extensions import Self, override
 
-class UploadedFile(File):
+class UploadedFile(File[AnyStr]):
```
sobolevn: `AnyStr` would be deprecated since 3.16+, let's not use it in new code :)

UnknownPlatypus: Should I just use Any then ? what is the recomended alternative ?

sobolevn: We can create our own `AnyOfStrTypes` alias that does the same for now :)

sobolevn (approved):

PR #3511, opened 2026-07-23
Add plugin support for `apps.get_model()`

### LINKED ISSUE #304: Too broad return value for apps.get_model()

UnknownPlatypus:
# I have made things!

A rebased version of #1725 fixing ci issues. Also migrate tests to `tests/assert_type`

## Related issues
Closes: #1318
Closes: #304
Closes: #305
Closes: #814
Closes: #1725

## AI Policy

### DIFF
--- a/mypy_django_plugin/lib/fullnames.py
+++ b/mypy_django_plugin/lib/fullnames.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 ABSTRACT_BASE_USER_MODEL_FULLNAME = "django.contrib.auth.base_user.AbstractBaseUser"
 ABSTRACT_USER_MODEL_FULLNAME = "django.contrib.auth.models.AbstractUser"
 PERMISSION_MIXIN_CLASS_FULLNAME = "django.contrib.auth.models.PermissionsMixin"
+APPS_FULLNAME = "django.apps.registry.Apps"
 MODEL_METACLASS_FULLNAME = "django.db.models.base.ModelBase"
 MODEL_CLASS_FULLNAME = "django.db.models.base.Model"
 FIELD_FULLNAME = "django.db.models.fields.Field"
--- a/mypy_django_plugin/lib/fullnames.py
+++ b/mypy_django_plugin/lib/fullnames.py
@@ -4,6 +4,7 @@ ABSTRACT_BASE_USER_MODEL_FULLNAME = "django.contrib.auth.base_user.AbstractBaseU
 ABSTRACT_USER_MODEL_FULLNAME = "django.contrib.auth.models.AbstractUser"
 PERMISSION_MIXIN_CLASS_FULLNAME = "django.contrib.auth.models.PermissionsMixin"
 APPS_FULLNAME = "django.apps.registry.Apps"
+STATE_APPS_FULLNAME = "django.db.migrations.state.StateApps"
 MODEL_METACLASS_FULLNAME = "django.db.models.base.ModelBase"
 MODEL_CLASS_FULLNAME = "django.db.models.base.Model"
 FIELD_FULLNAME = "django.db.models.fields.Field"
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -64,7 +64,7 @@ if TYPE_CHECKING:
     from mypy.types import Type as MypyType
 
 
-_APPS_MODULES = frozenset({"django.apps", "django.apps.registry"})
+_APPS_MODULES: Final = frozenset({"django.apps", "django.apps.registry"})
 
 
 class NewSemanalDjangoPlugin(Plugin):
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -4,7 +4,7 @@ import importlib.metadata
 import itertools
 import sys
 from functools import cache, cached_property, partial
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Final
 
 from mypy.build import PRI_MED, PRI_MYPY
 from mypy.modulefinder import mypy_path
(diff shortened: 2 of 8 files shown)


mypy_django_plugin/lib/fullnames.py
```diff
@@ -3,6 +3,7 @@
 ABSTRACT_BASE_USER_MODEL_FULLNAME = "django.contrib.auth.base_user.AbstractBaseUser"
 ABSTRACT_USER_MODEL_FULLNAME = "django.contrib.auth.models.AbstractUser"
 PERMISSION_MIXIN_CLASS_FULLNAME = "django.contrib.auth.models.PermissionsMixin"
+APPS_FULLNAME = "django.apps.registry.Apps"
```
sobolevn, line 6: non blocking: we can mark all these as `Final` as a refactoring.

mypy_django_plugin/transformers/apps.py
```diff
@@ -0,0 +1,56 @@
...
+        apps.get_model("myapp.MyModel")        # -> type[myapp.models.MyModel]
+        apps.get_model("myapp", "MyModel")      # -> type[myapp.models.MyModel]
+
+    The model name is case-insensitive, so `"myapp.mymodel"` resolves too.
+
+    Non-literal arguments keep the permissive `type[Any]` fallback:
+
+        apps.get_model(app_label, model_name)   # -> type[Any]
+    """
+    if (lazy_reference := _get_model_lazy_reference(ctx)) and (
```
sobolevn, line 47: Question: `apps.get_model` can be used to test migrations like we do in https://github.com/wemake-services/django-test-migrations

Do we account for the possible changes in models?

mypy_django_plugin/main.py
```diff
@@ -63,6 +64,9 @@
     from mypy.types import Type as MypyType
 
 
+_APPS_MODULES = frozenset({"django.apps", "django.apps.registry"})
```
sobolevn:
```suggestion
_APPS_MODULES: Final = frozenset({"django.apps", "django.apps.registry"})
```

tests/assert_type/apps/test_registry/models.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class Second(models.Model):
+    pass
+
+
+def get_model_resolves_literal_references() -> None:
+    # Mypy plugin is needed to narrow `type[Any]` to type[MyModel]
+    # TODO: other type checker should also understand that
+    assert_type(apps.get_model("test_registry.First"), type[First])  # pyright: ignore[reportAssertTypeFailure]  # pyrefly: ignore[assert-type]  # ty: ignore[type-assertion-failure]
```
sobolevn, line 24: Let's move this to `yml` tests. This is pure plugin code.

UnknownPlatypus: I think it's worth keeping here because pyrefly and ty are working toward django support and this would test their builtin coverage. I'd like for our test suite to become mostly typechecker agnostic so it might help move the ecosystem forward at some point

mypy_django_plugin/transformers/apps.py
UnknownPlatypus: It would not, but supporting that seems very difficult since we would need to rebuild the state of the model based on the migration file we are checking.

We can maybe skip narrowing if we are working with the `StateApps` subclass used in migrations ?

mypy_django_plugin/lib/fullnames.py
UnknownPlatypus: Will send another pr for that

mypy_django_plugin/transformers/apps.py
UnknownPlatypus: I did that in bf9fe93ee2b760d243387907386fc56238dc3b66

sobolevn (approved):
Amazing work, as always! 👍

mypy_django_plugin/lib/fullnames.py
UnknownPlatypus: Done in #3515

cuu508:
After upgrading to django-stubs 6.0.8, I'm seeing new mypy warnings in my data migrations about missing fields and attributes in the models returned by `apps.get_model()`. These fields did exist in the historic versions of the models, but indeed do not exist in the current version. 

I think it's related to this PR because if I comment out this part in `mypy_django_plugin/main.py` as an experiment, the warnings disappear:

```python
        if (
            method_name == "get_model"
            and info.has_base(fullnames.APPS_FULLNAME)
            # `StateApps` returns historical models rebuilt from migration state whose fields
            # differ from the current definitions, so narrowing to the current model is wrong.
            and not info.has_base(fullnames.STATE_APPS_FULLNAME)
        ):
            return partial(apps.resolve_model_for_get_model, django_context=self.django_context)
```

I'm not sure what the solution could be here.

UnknownPlatypus:
@cuu508 Could you provide the `revealed_type` of `apps` in your case ?

This part `and not info.has_base(fullnames.STATE_APPS_FULLNAME)` was supposed to make it skip migration for the exact issue you describe 

We can do file based exclusion otherwise, anything under `migrations/**` should probably be skipped ?

cuu508:
```python
from typing import Any

from django.apps.registry import Apps


def create_projects(apps: Apps, schema_editor: Any) -> None:
    reveal_type(apps)
    Profile = apps.get_model("accounts", "Profile")
    Project = apps.get_model("accounts", "Project")
    Member = apps.get_model("accounts", "Member")
[...]
```

```
hc/accounts/migrations/0018_auto_20190112_1426.py:12: note: Revealed type is "django.apps.registry.Apps"
```

cuu508:
Ah, I guess I should be doing annotating apps as `StateApps`?

```
from django.db.migrations.state import StateApps


def create_projects(apps: StateApps, schema_editor: Any) -> None:
[...]
```

sobolevn:
Yes :)

cuu508:
I see, thanks and sorry for the noise, PEBKAC 😆

UnknownPlatypus:
No worries, not that easy to know that but yeah it's more correct

You can take the noop signature as an exemple if you want the full correct signature 
https://github.com/typeddjango/django-stubs/blob/master/django-stubs%2Fdb%2Fmigrations%2Foperations%2Fspecial.pyi

PR #3513, opened 2026-07-24
Add type arguments to generic references in `utils` and `templates`

UnknownPlatypus:
## Related issues
Part of #64

Also fixes the type of `TemplateResponse` which was leaking test attribute that we already define in `_MonkeyPatchedWSGIResponse`. I did a smoke test on my work codebase and it as correct 

## AI Policy


UnknownPlatypus:
For some reason, this revealed a cache issue we have :exploding_head: , will send a dedicated pr to fix it

sobolevn (approved):

PR #3514, opened 2026-07-24
Invalidate mypy cache when `settings.INSTALLED_APPS` change and cache full config computation

UnknownPlatypus:
The `_User` alias expansion and the additional deps go into cached files but their source settings (`INSTALLED_APPS`, `DEFAULT_AUTO_FIELD`) were not part of the reported config data, so mypy served stale results when they changed.

This was causing an issue in #3513 with invalid cache being served causing error message to change. Rebasing on this branch solved the issue

This also compute the config data once instead of on every call, since it is called for every module.

# I have made things!

## Related issues
Followup to #3461 by caching the whole config like [pydantic does](https://github.com/pydantic/pydantic/blob/a2a6577d4c329dd574a45dbb01a8feaa16b1ad3d/pydantic/mypy.py#L164) because it does not change based on the module inspected
Was the solution to the issue in #3513

## AI Policy

### DIFF
--- a/mypy_django_plugin/main.py
+++ b/mypy_django_plugin/main.py
@@ -349,17 +349,26 @@ class NewSemanalDjangoPlugin(Plugin):
                 return create_new_manager_class_from_from_queryset_method
         return None
 
-    @override
-    def report_config_data(self, ctx: ReportConfigContext) -> dict[str, Any]:
+    @cached_property
+    def _report_config_data(self) -> dict[str, Any]:
         # Cache would be cleared if any settings do change.
         extra_data = {
+            # The additional deps depend on the installed apps
+            "INSTALLED_APPS": list(self.django_context.settings.INSTALLED_APPS),
+            # The user model determines the `_User` alias expansion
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks! I hope this will speed up tests as well. And they will work with the cache, since we change it in mypy tests.

UnknownPlatypus:
> Thanks! I hope this will speed up tests as well. And they will work with the cache, since we change it in mypy tests.

I did some local benchmarking and it was slightly faster (1%) but not so much, the main perf gain was done in #3461. We just do one attribute lookup per module now instead of 3-5.

As for speed in ci, I would expect it to maybe be a bit slower but because it's more correct. Some tests will invalidate cache because of changes in installed apps, where it was not happening before (and causing [some issues](https://github.com/typeddjango/django-stubs/actions/runs/30076124966/job/89427130827?pr=3513))

I had claude compare ci timings and it's actually slowing ci (+30%) because test can alternate from different installed apps, hence regularly busting cache. I'm working on a pr to recover this adding a hash of the settings to the cache dir

PR #3516, opened 2026-07-24
Key test mypy cache dirs by settings content

UnknownPlatypus:
# I have made things!
Since #3514, test cases with different `INSTALLED_APPS` correctly invalidate the mypy cache, so cases alternating settings on the same worker thrash the shared cache dir and re-analyze the stubs from scratch (the "Run pytest tests" CI step increased by ~30%, summed over the matrix). 

Mitigate this by appending a hash of the generated settings and mypy config to the cache dir so identical-settings cases keep sharing a warm cache while differing ones get their own.

This recovers most of the slowdown in ci (now 10% slower, which makes sense given 40 tests use custom settings)

## AI Policy

### DIFF
--- a/scripts/tests_extension_hook.py
+++ b/scripts/tests_extension_hook.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import hashlib
 from typing import TYPE_CHECKING
 
 from pytest_mypy_plugins.collect import File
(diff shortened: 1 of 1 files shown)


scripts/tests_extension_hook.py
```diff
@@ -13,6 +14,12 @@ def django_plugin_hook(test_item: YamlTestItem) -> None:
     installed_apps = test_item.parsed_test_data.get("installed_apps", None)
     monkeypatch = test_item.parsed_test_data.get("monkeypatch", False)
 
+    # When the result of `Plugin.report_config_data` changes, the mypy cache is invalidated.
+    # This slows down the test suite because test can be executed in any order, regularly invalidating shared cache.
+    # To mitigate this, use different cache dirs based on the settings.
+    config_hash = hashlib.md5(f"{installed_apps}{custom_settings}".encode(), usedforsecurity=False).hexdigest()[:10]
```
sobolevn, line 20: Should we include `monkeypatch`?

UnknownPlatypus: I would say it should not affect typecheking and is mostly required for runtime to not crash but I'm not 100% sure 🤔 
We can maybe keep it like that for now and see if we encounter the issue at some point ?

sobolevn (approved):

PR #3519, opened 2026-07-24
Use mypy's `lookup_fully_qualified` instead of custom symbol lookup

UnknownPlatypus:
# I have made things!

I think it's an artifact of the past, we don't seem to generate fake fullnames like a.b.C[x.y.Z] anymore ?

## AI Policy

### DIFF
--- a/mypy_django_plugin/lib/helpers.py
+++ b/mypy_django_plugin/lib/helpers.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any, Literal, NamedTuple, TypedDict, cast
 from mypy import checker
 from mypy.checker import TypeChecker
 from mypy.checkmember import analyze_member_access as _mypy_analyze_member_access
+from mypy.lookup import lookup_fully_qualified
 from mypy.maptype import map_instance_to_supertype
 from mypy.mro import calculate_mro
 from mypy.nodes import (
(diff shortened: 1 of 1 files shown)


mypy_django_plugin/lib/helpers.py
```diff
@@ -149,54 +149,9 @@ class IncompleteDefnException(Exception):
     pass
 
 
-def lookup_fully_qualified_sym(fullname: str, all_modules: dict[str, MypyFile]) -> SymbolTableNode | None:
-    if "." not in fullname:
-        return None
-    if "[" in fullname and "]" in fullname:
```
sobolevn, line 155: Is this code branch already dead?
https://github.com/typeddjango/django-stubs/commit/cfd69c0acc4bb3c282cab4a411b014afe280d359

UnknownPlatypus: I think it became dead after 3f340c9df and 45364bbeb. 

We switched from `WithAnnotations[myapp.models.MyModel]` to `WithAnnotations[myapp__models__MyModel]` to `MyModel@AnnotatedWith` to represent annotated models

The full test suite pass without it, mypy primer too and I also did a smoke test on my work codebase by adding a `raise AssertionError` and it never raised. I also don't see any place in the plugin code generating a name generating `[`. We only generate `Model@AnnotatedWith`, `*_RelatedManager`, `*_ManyRelatedManager`

sobolevn (approved):
Thank you!

PR #3522, opened 2026-07-25
Add type arguments to generic `Field`'s

UnknownPlatypus:
## Related issues
Part of #64

Also aligns our definitions of `_ST` and `_GT` to set the proper variance

## AI Policy


UnknownPlatypus:
> + zerver/models/messages.py:180: error: Need type annotation for "search_tsvector"  [var-annotated]

This is expected, mypy primer does not run the plugin, so it would need explicit annotations here.
I'm not exactly sure why, I think it would be better ? https://github.com/hauntsaninja/mypy_primer/blob/eb1c48d6db2984f5ab083b8355f3647cb4d167a5/mypy_primer/projects.py#L364-L366

sobolevn (approved):

sobolevn:
Yes, we can try to update the zulip in mypy-primer :)

UnknownPlatypus:
> Yes, we can try to update the zulip in mypy-primer :)

I tried to do a poc of that with claude and it"s a mess. zulip has a bunch of dependencies that are hard to setup, most notably ldap. Probably still possible to make it work but I don't know zulip and don't really want to work on that

PR #2483, opened 2025-01-15
Fix `ContentType` model fields to be infered as `Any`, refs #2446

sobolevn:
Closes #2446

PR #3524, opened 2026-07-25
Declare fields of `LogEntry`, `FlatPage` and `Redirect` as assignments

UnknownPlatypus:
# I have made things!
Assignment-style field declarations let the mypy plugin infer the type correctly (instead of the current `Any` fallback).
This also makes our setup consistant, we use to mix ways of defining these.

## Related issues
Mostly implement the suggestion https://github.com/typeddjango/django-stubs/pull/2483#discussion_r1922014216

## AI Policy

### DIFF
--- a/django-stubs/contrib/admin/models.pyi
+++ b/django-stubs/contrib/admin/models.pyi
@@ -2,6 +2,8 @@ from collections.abc import Iterable
 from typing import Any, ClassVar, Literal, overload
 from uuid import UUID
 
+from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from django.db.models.base import Model
 
@@ -33,13 +35,13 @@ class LogEntryManager(models.Manager[LogEntry]):
     ) -> list[LogEntry]: ...
 
 class LogEntry(models.Model):
-    action_time: models.DateTimeField
-    user: models.ForeignKey
-    content_type: models.ForeignKey
-    object_id: models.TextField
-    object_repr: models.CharField
-    action_flag: models.PositiveSmallIntegerField
-    change_message: models.TextField
+    action_time = models.DateTimeField()
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    content_type = models.ForeignKey(ContentType, on_delete=models.SET_NULL, blank=True, null=True)
+    object_id = models.TextField(blank=True, null=True)
+    object_repr = models.CharField(max_length=200)
+    action_flag = models.PositiveSmallIntegerField()
+    change_message = models.TextField(blank=True)
     objects: ClassVar[LogEntryManager]
     def is_addition(self) -> bool: ...
     def is_change(self) -> bool: ...
--- a/django-stubs/contrib/admin/models.pyi
+++ b/django-stubs/contrib/admin/models.pyi
@@ -1,3 +1,4 @@
+import datetime as dt
 from collections.abc import Iterable
 from typing import Any, ClassVar, Literal, overload
 from uuid import UUID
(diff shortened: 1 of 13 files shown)


tests/assert_type/contrib/test_models.py
```diff
@@ -0,0 +1,42 @@
...
+from django.contrib.sites.models import Site
+from typing_extensions import assert_type
+
+
+def log_entry_fields_are_inferred() -> None:
+    # only the mypy plugin infers field descriptor types; the other type checkers
+    # can't solve the fields' type variables and need inline ignores
+    entry = LogEntry()
+    assert_type(entry.action_time, datetime)  # pyright: ignore[reportUnknownMemberType, reportAssertTypeFailure]  # ty: ignore[type-assertion-failure]
+    assert_type(entry.user, User)  # pyright: ignore[reportUnknownMemberType, reportAssertTypeFailure]  # pyrefly: ignore[assert-type]  # ty: ignore[type-assertion-failure]
```
UnknownPlatypus: this is quite cool the plugin get this one right 🥳

tests/assert_type/contrib/test_models.py
```diff
@@ -0,0 +1,40 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.contrib.flatpages.models import FlatPage
+from django.contrib.redirects.models import Redirect
+from django.contrib.sites.models import Site
+from typing_extensions import assert_type
+
+
+def log_entry_fields_are_inferred() -> None:
+    entry = LogEntry()
+    assert_type(entry.action_time, datetime)  # pyright: ignore[reportUnknownMemberType, reportAssertTypeFailure]  # ty: ignore[type-assertion-failure]
```
sobolevn: I don't quite like that `pyright` and others do not know about this attribute. Maybe we can try `action_time: models.DateTimeField = models.DateTimeField()`?

UnknownPlatypus: does not change anything, but I tried just binding the typevars and all typechekers get it, without the assignment so I think that an even better thing to do for the stubs

UnknownPlatypus: This is much better, also works with M2m field with a small stubs + plugin adjustment @sobolevn

django-stubs/contrib/admin/models.pyi
```diff
@@ -33,13 +37,13 @@ class LogEntryManager(models.Manager[LogEntry]):
...
 
 class LogEntry(models.Model):
-    action_time: models.DateTimeField
-    user: models.ForeignKey
-    content_type: models.ForeignKey
-    object_id: models.TextField
-    object_repr: models.CharField
-    action_flag: models.PositiveSmallIntegerField
-    change_message: models.TextField
+    action_time: models.DateTimeField[str | dt.datetime | dt.date | Combinable, dt.datetime]
```
sobolevn, line 40: What do you think about providing default aliases for such types? We can expose them into the `ext` as well.

UnknownPlatypus: Why not, but I hope the changes to the way we infer fields will help make them not needed most of the time
So maybe it's not worth it ?

sobolevn: I know at least some places where it can be useful for me as well: https://github.com/wemake-services/django-modern-rest/blob/45331dc8cf3cbfced9b76a796432f73c67cf6fb6/dmr/security/jwt/blocklist/models.py#L15-L19

UnknownPlatypus: Yeah this would be useful here but is only really needed because the inference does not work without it.

I'd rather focus my efforts on #3317, I have a wip alternative branch trying to make it work with `__init__` overloads that seem promising

sobolevn: Ok, let's keep this private for now :)

sobolevn (approved):

PR #3526, opened 2026-07-27
Remove `tests/assert_type/test_import_all.py`

ngnpope:
# I have made things!

~These were missed when updating the stubs to support Django 6.0 or earlier versions.~

_Updated to remove this test module._

This was previously used to check that imports did not raise any errors, but the way we test things now is much better and would identify any issues anyway.

Also, because this test isn't generated, it's easy for it to become outdated, missing out new modules as they are added in each Django release.

## Related issues

N/A

## AI Policy


UnknownPlatypus:
I'm not sure I really get this test. Do you know what it tries to verify ?

I have a feeling it should either go or be generated but it feels odd to have it drift and require such fix ?

ngnpope:
It goes way back to 5c6be7ad121abb2bc8020b0a51c5c0a3981f7216, but it's not entirely clear why it was added. I don't know if it has helped expose any issues in the past or not. Thoughts @sobolevn?

Curiously it looks like there was originally [a script](https://github.com/typeddjango/django-stubs/pull/56/changes/36763314ac05476594701c16a34076be6f7cb7ac) to generate it.

UnknownPlatypus:
My guess is that stubtest is doing the job now (checking the stubs are matching the runtime) and that it just predate stubtest addition ?

ngnpope:
That's quite possible!

sobolevn:
Yes, it better to remove it now :)
It used to check that imports do not raise any errors. But, we do a lot of things to check this now.

sobolevn (approved):

PR #3527, opened 2026-07-27
Ensure ORM expressions return params as a tuple

ngnpope:
# I have made things!

Since Django 6.0, it is expected that `as_sql()` returns params as a tuple, never a list. Changing this helps to ensure that this incompatibility is picked up at static type checking time.

See https://docs.djangoproject.com/en/stable/releases/6.0/#custom-orm-expressions-should-return-params-as-a-tuple

## Related issues

N/A

## AI Policy

### DIFF
--- a/django-stubs/db/models/sql/compiler.pyi
+++ b/django-stubs/db/models/sql/compiler.pyi
@@ -12,13 +12,12 @@ from django.db.models.base import Model
 from django.db.models.expressions import BaseExpression, Expression, Ref
 from django.db.models.sql.query import Query
 from django.db.models.sql.subqueries import AggregateQuery, DeleteQuery, InsertQuery, UpdateQuery
-from django.utils.datastructures import _ListOrTuple
 from django.utils.functional import cached_property
 from typing_extensions import override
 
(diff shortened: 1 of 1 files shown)


ngnpope:
> Diff from [mypy_primer](https://github.com/hauntsaninja/mypy_primer), showing the effect of this PR on type check results on a corpus of open source code:
> 
> ```diff
> zulip (https://github.com/zulip/zulip)
> + zerver/models/lookups.py:12: error: Return type "tuple[str, list[str | int]]" of "as_sql" incompatible with return type "tuple[str, tuple[str | int, ...]]" in supertype "django.db.models.expressions.BaseExpression"  [override]
> + zerver/models/lookups.py:25: error: Return type "tuple[str, list[str | int]]" of "as_sql" incompatible with return type "tuple[str, tuple[str | int, ...]]" in supertype "django.db.models.expressions.BaseExpression"  [override]
> ```

This is good - exactly what we expect. Note that zulip currently targets Django 5.2, but they could update their `.as_sql()` methods to return tuples now ahead of updating to support Django 6.0+.

UnknownPlatypus (approved):

PR #3528, opened 2026-07-27
Improved typing of the checks framework

ngnpope:
# I have made things!

Makes the following changes:

- Prefer `**kwargs: object` in check functions

  Generally `object` should be preferred to `Any`, particularly where we're not actually using `*args` or `**kwargs` in the body of the function. See the following for more info:
    
    - https://mypy.readthedocs.io/en/stable/dynamic_typing.html#any-vs-object
    - https://adamj.eu/tech/2021/05/07/python-type-hints-use-object-instead-of-any/
    
    We should gradually extend this to other areas in the stubs.

- Always return `list[CheckMessage]` from check functions

  Although in many cases these will only return `Error` or `Warning`, there is no need to be so specific here and `CheckMessage` will do. Further, it avoids needing to change these if a different message type is returned in future versions of Django.

- Finish typing `django.contrib.admin.checks`

  The `must_be()`, `must_inherit_from()` and `refer_to_missing_field()` helpers can be typed more accurately.

## Related issues

N/A

## AI Policy


UnknownPlatypus:
> We should gradually extend this to other areas in the stubs.

Should we open an issue to track that ? 
It looks like a good idea in general but given this quote in mypy docs, there might some case where it can cause harm ?

> If you’re not sure whether you need to use [object](https://docs.python.org/3/library/functions.html#object) or Any, use [object](https://docs.python.org/3/library/functions.html#object) – only switch to using Any if you get a type checker complaint.

ngnpope:
> Should we open an issue to track that?

Yes, I'll open one. We should probably also use some of the typeshed aliases of `Any` to highlight whether it's intentionally `Any` or something that just hasn't been looked at yet...

> It looks like a good idea in general but given this quote in mypy docs, there might some case where it can cause harm ?

There are potentially some cases, yes. It often depends on how things are propagated through function calls and whether deeper calls expect a specific type. Usually the issue is use of `**kwargs` to be lazy and avoid redefining all function parameters. Sometimes there are other escape hatches to address this by use of `Unpack` and `TypedDict`. But generally, we should strive to use `object` where we can.

sobolevn (approved):
Looks really better!

ngnpope:
> > Should we open an issue to track that?
> 
> Yes, I'll open one.

#3532

PR #3533, opened 2026-07-27
Use typeshed aliases where possible

ngnpope:
# I have made things!

For consistency, we should use typeshed aliases where possible:

- Use `_typeshed.StrPath` for `str | os.PathLike[str]` replacing our custom `_PathCompatible`
- Use `_typeshed.ExcInfo` for `tuple[type[BaseException], BaseException, TracebackType]`
- Import `_typeshed.Self` aliased to `MetaclassSelf` for clarity

I did look into using `OpenBinaryMode | OpenTextMode` for `File.open(mode=...)`, `Storage.open(mode=...)`, etc. but it'll likely be a bit more complex as we need to look into overloads for binary/text which is quite involved.

## Related issues

N/A

## AI Policy


django-stubs/db/models/enums.pyi
```diff
@@ -17,13 +18,11 @@ else:
     class IntEnum(int, ReprEnum): ...  # type: ignore[misc]
     class StrEnum(str, ReprEnum): ...  # type: ignore[misc]
 
-_Self = TypeVar("_Self", bound=ChoicesType)
```
ngnpope, line 20: It didn't seem necessary to bind this one. No affect on the tests.

But for `ModelBase` we do need a custom type var bound to `Model`.

sobolevn (approved):

PR #3534, opened 2026-07-28
Infer type for `Form.fields` to a TypedDict

### LINKED ISSUE #1514: Infer type for Form.fields
https://stackoverflow.com/questions/73600700/best-practice-for-python-typing-out-of-dictionary-self-fields-type-in-django-mo

I'm looking to infer dictionary values type for this snippet

```py
class MerchantBrandsForm(forms.Form):
    brands = forms.ModelChoiceField(required=True, queryset=None)

    def __init__(self: merchant: Merchant, *args: Any, **kwargs: Any):
        self.fields['brands'].queryset = merchant.brands.all()
           # ^^^^^^^^^^
           # This would be a type error since forms.ModelChoiceField is generalized to forms.Field
```

self.fields type above should be inferred to

```py
class Fields(TypedDict):
    brands: forms.ModelChoiceField
```

randspace0:
This should solve the related issue. I have implemented what I have describe in the issue description.

## Related issues

Fixes #1514

## AI Policy


mypy_django_plugin/lib/fullnames.py
```diff
@@ -39,6 +39,7 @@
 
 
 BASEFORM_CLASS_FULLNAME: Final = "django.forms.forms.BaseForm"
+FORM_FIELD_FULLNAME: Final = "django.forms.fields.Field"
```
UnknownPlatypus, line 42:
```suggestion
FORM_FIELD_CLASS_FULLNAME: Final = "django.forms.fields.Field"
```

To match the other names pattern

tests/typecheck/test_forms.yml
```diff
@@ -181,3 +181,57 @@
...
+
+        class BaseForm(forms.Form):
+            name = forms.CharField()
+
+        class MyForm(BaseForm):
+            age = forms.IntegerField()
+            name = forms.EmailField()  # override parent field with a different type
+
+        form = MyForm()
+        reveal_type(form.fields["age"])  # N: Revealed type is "django.forms.fields.IntegerField"
```
sobolevn, line 199: please, assert the type of `form.fields` instead :)

tests/typecheck/test_forms.yml
```diff
@@ -181,3 +181,57 @@
...
+        reveal_type(form.fields["name"])  # N: Revealed type is "django.forms.fields.EmailField"
+        form.fields["missing"]  # E: "missing" is not a valid TypedDict key; expected one of ("age", "name")  [misc]
+
+        # dynamic reassignment of the whole dict is still allowed
+        form.fields = {}
+
+        # a plain, undeclared Form still falls back to the default 'dict[str, Field]' typing
```
sobolevn, line 206: Can you please add a comment to the implementation: why is that? Some reasoning would be nice to have for the history

tests/typecheck/test_forms.yml
```diff
@@ -181,3 +181,57 @@
...
+-   case: form_fields_queryset_assignment_via_typed_dict
+    main: |
+        from typing import Any
+        from django import forms
+        from django.db.models import QuerySet
+        from myapp.models import Merchant
+        from typing_extensions import reveal_type
+
+        class MyForm(forms.Form):
+            brands = forms.ModelChoiceField(required=True, queryset=None)  # type: ignore[var-annotated]
```
sobolevn, line 219: Why do we have a type ignore here?

sobolevn (commented):
Thanks! Looks interesting :)

mypy_django_plugin/transformers/forms.py
```diff
@@ -15,3 +21,39 @@ def make_meta_nested_class_inherit_from_any(ctx: ClassDefContext) -> None:
...
+        for stmt in base_info.defn.defs.body:
+            if isinstance(stmt, AssignmentStmt) and len(stmt.lvalues) == 1 and isinstance(stmt.lvalues[0], NameExpr):
+                name = stmt.lvalues[0].name
+                if name not in seen:
+                    seen.add(name)
+                    yield name
+
+
+def transform_form_fields_attr_type(ctx: AttributeContext) -> MypyType:
+    """Create type `form.fields` as a TypedDict of the form's declared fields, instead of `dict[str, Field]`."""
```
UnknownPlatypus, line 39: Is a typedict a good idea ? I see a few potential issues with that (might need some test case to validate)

- `for f in self.fields.values(): f.widget.attrs[...] = ...` will raise "object" has no attribute "widget" because TypedDict is Mapping[str, object]
- `del self.fields["age"]` or `self.fields.pop("name", None)` will raise `TypedDict key "age" cannot be deleted`
- `self.fields[k] = ...` -> `TypedDict key must be a string literal [literal-required]`

I don't know if it's possible to have a loose typedict, maybe with overloads ?

mypy_django_plugin/transformers/forms.py
```diff
@@ -15,3 +21,39 @@ def make_meta_nested_class_inherit_from_any(ctx: ClassDefContext) -> None:
...
+    if ctx.is_lvalue:
+        return ctx.default_attr_type
+
+    object_type = ctx.type
+    if not isinstance(object_type, Instance):
+        return ctx.default_attr_type
+
+    field_types: dict[str, MypyType] = {}
```
UnknownPlatypus, line 47: I think we will miss the fields that are implicitly declared for exemple when using a ModelForm ?


```python
  class ArticleForm(forms.ModelForm):
      slug = forms.SlugField(required=False)
      class Meta:
          model = Article
          fields = ["name", "slug"]

  f.fields["name"]  # E: "name" is not a valid TypedDict key; expected one of ("slug")
```

mypy_django_plugin/transformers/forms.py
```diff
@@ -15,3 +21,39 @@ def make_meta_nested_class_inherit_from_any(ctx: ClassDefContext) -> None:
...
     else:
         meta_node.fallback_to_any = True
+
+
+def _iter_declared_field_names(info: TypeInfo) -> Iterator[str]:
+    """Yield each class-body-assigned attribute name once, starts from the most-derived class."""
+    seen: set[str] = set()
+    for base_info in info.mro:
+        for stmt in base_info.defn.defs.body:
+            if isinstance(stmt, AssignmentStmt) and len(stmt.lvalues) == 1 and isinstance(stmt.lvalues[0], NameExpr):
```
UnknownPlatypus, line 31: I think this will register annotation-only attrs as fields, for ex  `other: forms.IntegerField` but they won't be actual fields.

mypy_django_plugin/transformers/forms.py
```diff
@@ -15,3 +21,39 @@ def make_meta_nested_class_inherit_from_any(ctx: ClassDefContext) -> None:
             ctx.api.defer()
     else:
         meta_node.fallback_to_any = True
+
+
+def _iter_declared_field_names(info: TypeInfo) -> Iterator[str]:
+    """Yield each class-body-assigned attribute name once, starts from the most-derived class."""
+    seen: set[str] = set()
+    for base_info in info.mro:
```
UnknownPlatypus, line 29: We need to skip class in the mro if they don't have a form base. See https://github.com/django/django/blob/8362cdc7efc9c881fdba48ce19e89b9194de7c46/django/forms/forms.py#L38

mypy_django_plugin/main.py
```diff
@@ -351,6 +351,10 @@ def get_attribute_hook(self, fullname: str) -> Callable[[AttributeContext], Mypy
         if info.has_base(fullnames.STR_PROMISE_FULLNAME):
             return resolve_str_promise_attribute
 
+        # Lookup of the '.fields' attribute on a Form instance
+        if info.has_base(fullnames.BASEFORM_CLASS_FULLNAME) and attr_name == "fields":
+            return forms.transform_form_fields_attr_type
```
UnknownPlatypus, line 356: This might become a bit expensive since we recompute on every `fields` access. Maybe it's fine, people don't do that too often ?

Otherwise an option would be to do like `model._meta.get_fields()` by registering the result at the module scope so we can do simpel lookup to get it later. see `get_or_create_annotated_type`

mypy_django_plugin/transformers/forms.py
UnknownPlatypus: For `ModelForm` we could use django's `fields_for_model(model, fields, exclude)` using `django_context` but that might be better to redo it manually here to avoid more coupling with the django runtime ? 
We would like to be less coupled at some point. See #3376

UnknownPlatypus (commented):
Thanks for starting work on this. This will be a very cool and much needed feature.

I left some comments for things that could be improved.
I would also like that we check in more detail how we do similar stuff for django `Model._meta.get_fields()` because it feels a bit similar

mypy_django_plugin/transformers/forms.py
randspace0: Right. I'm thinking to just extend the `dict` type for flexibility. So we can just type the statically analyzed field and keep the dynamic field untyped.

Something like

```py
class MyForm(BaseForm):
    age = forms.IntegerField()
    name = forms.EmailField()

    class FieldsDict(dict[str, Field]):  # Created on the fly
        @overload
         def __getitem__(self, key: Literal["age"]) -> IntegerField: ...
         
         @overload
         def __getitem__(self, key: Literal["name"]) -> EmailField: ...
         
         @overload
         def __getitem__(self, key: str) -> Field: ...

    fields: FieldsDict  # Created on the fly
```

UnknownPlatypus: yes maybe, probably need to start with a few test from my previous comment to see if it works

PR #3559, opened 2026-08-04
Expand override manager create test + move to `assert_type`

UnknownPlatypus:
# I have made things!

Expand the coverage for maanger create overloads.
The unparametrized case exercise a bit of plugin but not the others two so I moved it to assert_type.

This also revealed that our typing of `Model.objects` as `ClassVar[Manager[Self]]` is already a bit problematic with pyright and pyrefly (see the type-ignore's). Given this exemple:

```python
class Book(models.Model):
    objects = BookManager()
    foo = BookManager() 
```

I get these revealed types
```
┌─────────────┬───────────────┬─────────────┐
│             │ Book.objects  │  Book.foo   │
├─────────────┼───────────────┼─────────────┤
│ mypy+plugin │ BookManager   │ BookManager │
├─────────────┼───────────────┼─────────────┤
│ ty          │ BookManager   │ BookManager │
├─────────────┼───────────────┼─────────────┤
│ pyright     │ Manager[Book] │ BookManager │
├─────────────┼───────────────┼─────────────┤
│ pyrefly     │ Manager[Book] │ BookManager │
└─────────────┴───────────────┴─────────────┘
```

To me our declaration in the stubs should not impact regular class level assignment inference and I would expect `objects` to have the same type as `foo`. The plugin should only apply the "does not exist on abstract model" bit, the rest should be regular inference.

I'm exploring how to improve that, maybe with a descriptor in a followup pr

## Related issues
Still trying to cut the bulk for #2776

## AI Policy


sobolevn (approved):
:(

PR #3562, opened 2026-08-04
State `pyrefly` / `pyright` / `ty` basic support

sobolevn:


README.md
```diff
@@ -6,6 +6,12 @@
 
 This package contains [type stubs](https://www.python.org/dev/peps/pep-0561/) and a custom mypy plugin to provide more precise static types and type inference for Django framework. Django uses some Python "magic" that makes having precise types for some code patterns problematic. This is why we need this project. The final goal is to be able to get precise types for most common patterns.
 
+## Supported type checkers
+
+- mypy: full and complete support with multiple advanced features with our custom mypy plugin
+- pyright: basic support, checked in CI
+- pyrefly: basic support, checked in CI
```
UnknownPlatypus, line 13: Should we add ty ?
We test similarly and have basic support too, probably similar to pyright with a few bugs cause more recent ?

UnknownPlatypus (approved):

README.md
sobolevn: I would say, that it is beyond basic? Since it has lots of ignores? Like `ty: initial support`?

ngnpope: I would say that we should give them all the same status:

```console
$ git grep '# *type: *ignore' tests/assert_type/ | wc  -l
107

$ git grep '# *pyrefly: *ignore' tests/assert_type/ | wc  -l
110

$ git grep '# *pyright: *ignore' tests/assert_type/ | wc  -l
174

$ git grep '# *ty: *ignore' tests/assert_type/ | wc  -l
163
```

There are actually more ignores for pyright than ty!

UnknownPlatypus: (I believe pyright has more because it run in strict mode and will raise for missing generic param for ex)

UnknownPlatypus: Also I think we should declare the same level of support for pyright/ty/pyrefly since we only have stubs for them so a bug is either a typechecker bug or a stubs bug affecting all most likely ?

sobolevn: Ok :)

README.md
```diff
@@ -6,6 +6,13 @@
...
+- ty: initial support, checked in CI, but has lots of known problems
```
sobolevn:
```suggestion
- ty: basic support, checked in CI
```

ngnpope (approved):

PR #3563, opened 2026-08-04
Bootstrap Django 6.1

ngnpope:
# I have made things!

Given we have a number of changes in the works such as using typevar defaults for fields, etc. I thought I'd just churn out the changes for 6.1 in one go to reduce lots of inevitable conflicts for many small pull requests coming in over the next few weeks.

**Things to do:**

- [x] Fix the plugin to support `GenericForeignKeyDescriptor`; fall back to `GenericForeignKey` for Django < 6.1
- [x] Check whether any more `@deprecated` need adding - I likely missed a few
- [x] Remove the temporary stubtest ignores for some post-RC1 fixes already applied
- [x] Update for any changes post-RC1, e.g. security and bug fixes
- [x] Check that plugin and stubs work reasonably with Django 6.0 & 5.2

## Related issues

N/A

## AI Policy


django-stubs/contrib/admin/options.pyi
```diff
@@ -35,19 +36,38 @@ from django.utils.functional import _StrOrPromise
...
+EMPTY_VALUE_STRING: Literal["-"]
 
 class ShowFacets(enum.Enum):
     NEVER = cast(str, ...)
     ALLOW = cast(str, ...)
     ALWAYS = cast(str, ...)
 
+class ActionLocation(enum.Enum):
+    CHANGE_FORM = ...
+    CHANGE_LIST = ...
```
```suggestion
    CHANGE_FORM = cast(str, ...)
    CHANGE_LIST = cast(str, ...)
```

django-stubs/contrib/admin/options.pyi
```diff
@@ -35,19 +36,38 @@ from django.utils.functional import _StrOrPromise
...
+
+@dataclass
+class Action:
+    func: Callable[..., str]
+    name: str
+    description: str
+    plural_description: str
+    locations: list[ActionLocation]
+    # RemovedInDjango70Warning.
+    def __iter__(self) -> Iterator[Callable[..., str] | str]: ...
```
sobolevn: Can we use `@deprecated` on such items maybe?

django-stubs/contrib/admin/options.pyi
```diff
@@ -35,19 +36,38 @@ from django.utils.functional import _StrOrPromise
...
+    CHANGE_LIST = ...
+
+HORIZONTAL: Literal[1]
+VERTICAL: Literal[2]
+
+_Direction: TypeAlias = Literal[1, 2]
+
+@dataclass
+class Action:
+    func: Callable[..., str]
```
sobolevn: It is called as `response = func(self, request, queryset)`, we can add param types here :)

UnknownPlatypus:
Are the stubs still supporting django 5.2 after these changes ?

django-stubs/contrib/auth/views.pyi
```diff
@@ -18,8 +18,8 @@ class RedirectURLMixin:
     redirect_field_name: str
     success_url_allowed_hosts: set[str]
     def get_success_url(self) -> str: ...
-    def get_redirect_url(self) -> str: ...
-    def get_success_url_allowed_hosts(self) -> set[str]: ...
+    def get_redirect_url(self, request: HttpRequest | None = ...) -> str: ...
```
sobolevn:
```suggestion
    def get_redirect_url(self, request: HttpRequest | None = None) -> str: ...
```

Let's use real defaults

django-stubs/contrib/auth/views.pyi
```diff
@@ -18,8 +18,8 @@ class RedirectURLMixin:
...
+    def get_success_url_allowed_hosts(self, request: HttpRequest | None = ...) -> set[str]: ...
```
sobolevn:
```suggestion
    def get_success_url_allowed_hosts(self, request: HttpRequest | None = None) -> set[str]: ...
```

django-stubs/contrib/contenttypes/fields.pyi
```diff
@@ -35,6 +35,18 @@ class GenericForeignKey(FieldCacheMixin, Field[Any, Any]):
     def __init__(
         self, ct_field: str = "content_type", fk_field: str = "object_id", for_concrete_model: bool = True
     ) -> None: ...
+    # class access
+    @overload
```
sobolevn, line 39: This needs to be tested with `assert_type`

django-stubs/db/models/functions/uuid.pyi
```diff
@@ -0,0 +1,30 @@
...
+class UUID7(Func):
+    output_field: ClassVar[models.UUIDField[Any, Any]]
+    def __init__(self, shift: dt.timedelta | None = None, **extra: Any) -> None: ...
+    @override
+    def as_sql(self, compiler: SQLCompiler, connection: BaseDatabaseWrapper, **extra_context: Any) -> _AsSqlType: ...  # type: ignore[override]
+    def as_postgresql(
+        self, compiler: SQLCompiler, connection: BaseDatabaseWrapper, **extra_context: Any
+    ) -> _AsSqlType: ...
+    @override
+    def as_sqlite(self, compiler: SQLCompiler, connection: BaseDatabaseWrapper, **extra_context: Any) -> _AsSqlType: ...
```
sobolevn, line 29: You can use a python-version based overload here. Return `Never` on <3.14

sobolevn (commented):
Amazing work! Thanks a lot!

ngnpope:
> Are the stubs still supporting django 5.2 after these changes ?

We currently target 6.0 with partial support for 5.2. I'd expect that we would make a final 6.0.x release before merging this and then release 6.1.0 supporting 6.1 with partial support for 6.0, 5.2 and possibly 5.1, 5.0 (but we need to decide whether to drop those as they're end-of-life.

I'll add an item to the task list to check that nothing is obviously broken with the old versions.

---

There's only [one small change](https://github.com/typeddjango/django-stubs/pull/3563/changes#diff-a0a340b31e8cf50dff33201dd7ec9bc4e065e4816ea73ac1d33380dd31d56793) needed to the plugin to support the new `GenericForeignKeyDescriptor` and I've left a fallback to `GenericForeignKey` for Django < 6.1.

django-stubs/contrib/admin/options.pyi
ngnpope: Yeah - second item on the task list in the description 🙂

> Check whether any more `@deprecated` need adding - I likely missed a few

UnknownPlatypus:
> I'll add an item to the task list to check that nothing is obviously broken with the old versions.

That's my main concern, we run 5.2 at work and will not go to 6 in the near future because we need a major python evrsion bump. I just want to make sure it's not breaking too much.
I can test this branch on my codebase when you think the pr is mature enough

django-stubs/db/models/functions/uuid.pyi
ngnpope: I'm not sure that makes sense. The function is still present for all versions of Python. It'll just be removed from Django when Python 3.14 is the minimum supported version. Even if this override is removed, its parent will still have `as_sqlite()` returning `_AsSqlType`, so returning `Never` here would imply that that parent function - when this override is removed - wouldn't be able to return anything which isn't true.

django-stubs/contrib/admin/options.pyi
ngnpope: See 29e40c810b112f14da507ffb937df98bb6d14c00 for an improvement.

django-stubs/contrib/contenttypes/fields.pyi
ngnpope: Added something basic in 531a695482b80c8e1ed360baca5c2dee7428c8e0.

tests/assert_type/contrib/contenttypes/test_fields.py
```diff
@@ -0,0 +1,19 @@
...
+
+from typing import Any
+
+from django.contrib.contenttypes.fields import GenericForeignKey, GenericForeignKeyDescriptor
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+from typing_extensions import assert_type
+
+
+class TaggedItem(models.Model):
```
UnknownPlatypus: If you want to assert stuff the plugin will add, you might wan to try the patern I describe in the CONTRIBUTING.md that makes these models collected by mypy runs.

django-stubs/contrib/admin/options.pyi
ngnpope: Added most of the easy ones in 5fcf66cd08cf87ea855f53283dbee86a894515a5. It might be possible to handle some others, but would probably need lots of `@overload`s.

tests/assert_type/contrib/contenttypes/test_fields.py
ngnpope: Improved in 283c1d83a2ee3c147bedbc3e66a32984674eca8b. Thanks!

ngnpope:
@sobolevn @UnknownPlatypus I think this is largely done now other than some final testing against older Django versions to make sure that they're not completely broken. Not sure how much confidence the CI tests give when running against Django 5.2.

I'm going to be away for a bit over a week on holiday, so feel free to finish this off!

sobolevn (approved):
Happy holidays! 

I think that we should make the last `6.0.8` release, then merge this and improve all the things that people will ask for.

`6.0.8` is on its way!

UnknownPlatypus:
I'll just do a last check for 5.2 compat, I really think we should not drop it, it's an LTS

sobolevn:
@UnknownPlatypus feel free to merge after the test :)
I rebased the PR to fix a conflict.

ngnpope:
> I'll just do a last check for 5.2 compat, I really think we should not drop it, it's an LTS

Of course - that hasn't been suggested - we're just doing what we normally do at each release 🙂 

Hopefully this will become less painful before too long anyway with the acceptance of [DEP 0020: Annual Release Cycle](https://github.com/django/deps/blob/main/accepted/0020-annual-release-cycle.rst)

UnknownPlatypus:
Pushed a commit to avoid issue with a breaking signature change in django 6+ that was affecting me at work. 
I've used py version gate because 6+ require 3.12 so we know for sure 3.11 means user has django 5.X
This will not work fur users on django 5 and py 3.12+ but nothing we can do for that, still better than doing nothing

CI does not want to run tho ?

ngnpope:
> CI does not want to run tho ?

https://www.githubstatus.com/incidents/qcvjkzcs7j74 🙈

UnknownPlatypus:
Ofc, forgot we are in the 2 nines era, thx ai

It started 7h ago you kidding me x)

https://softwareshould.work/

UnknownPlatypus (approved):
Great, tested on my work codebase (Django 5.2) and 0 regressions!

Good to merge once we land the small bug fixes for 6.0.8

sobolevn:
6.0.9 is on its way

sobolevn:
`6.0.9` is done, merging! Thanks again, @ngnpope!

adamchainz:
Just upgraded my "fleet" of open source projects, they were all happy :)

PR #3572, opened 2026-08-07
Detect every generic stub class in the generic consistency test

### LINKED ISSUE #3570: [6.0.8] `UploadedFile` is not subscriptable
Looks like we would need to release one more version of 6.0.x :)
We forgot to add `UploadedFile` to the list of patched classes in `ext`.

UnknownPlatypus:
# I have made things!

The old test only collected classes whose base was literally Generic[...]:
```python
  isinstance(base.value, ast.Name) and base.value.id == "Generic"
```
But most stub classes are generic implicitly, through a parameterized base — OrderedSet is declared as:
```python
class OrderedSet(MutableSet[_K]): ...
```

We were missing some with the consistency test, I've changed it so that it collects bases that are subsripted, then apply the monkypatching and finally try to subscript each one, we expect no crash if everything is monkey patched correctly.

This revealed we missed `OrderedSet` 

## Related issues
Fixes #3570

## AI Policy


sobolevn (approved):
Good find, thanks!

PR #3586, opened 2026-08-11
Support custom arguments in queryset update methods

WhyNotHugo:
Example in test should be pretty self-explanatory.

## AI Policy


sobolevn (approved):
Thanks! Can you please check if other methods are also affected by similar bugs?

PR #3587, opened 2026-08-11
Fix revealed type for DistanceField and AreaField

WhyNotHugo:
## AI Policy


sobolevn (approved):
Awesome, thanks!

WhyNotHugo:
Not entirely sure why this is failing — are these types different when using MySql?

django-stubs/contrib/gis/db/models/sql/conversion.pyi
```diff
@@ -9,6 +10,7 @@ _ST = TypeVar("_ST", contravariant=True)
 _GT = TypeVar("_GT", covariant=True)
 
 class AreaField(models.FloatField[_ST, _GT]):
+    _pyi_private_get_type: Area
```
sobolevn: this is an incompatible override from `FloatField._pyi_private_get_type`, it needs to be explicitly silenced. 
```suggestion
    _pyi_private_get_type: Area  # type: ignore[assignment]
```

django-stubs/contrib/gis/db/models/sql/conversion.pyi
```diff
@@ -20,6 +22,7 @@ class AreaField(models.FloatField[_ST, _GT]):
     def get_internal_type(self) -> Any: ...
 
 class DistanceField(models.FloatField[_ST, _GT]):
+    _pyi_private_get_type: Distance
```
sobolevn:
```suggestion
    _pyi_private_get_type: Distance  # type: ignore[assignment]
```

Same here :)

sobolevn (approved):

PR #3591, opened 2026-08-12
Allow database variants in on_delete for related fields.

### LINKED ISSUE #3590: New database variants for `on_delete` are not supported in ForeignKey.on_delete
# Bug report

## What's wrong

Using `on_delete=models.DB_CASCADE` for `ForeignKey`s raises `invalid-argument-type`:

```
    error[invalid-argument-type]: Argument to `ForeignKey.__init__` is incorrect
       --> src/someapp/models.py:294:9
        |
    294 |         on_delete=models.DB_CASCADE,
        |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Expected `(...) -> None`, found `DatabaseOnDelete`
    info: type `DatabaseOnDelete` has inferred callable type `(Field[Any, Any], QuerySet[Model, Model], str, /) -> None`
    info: Method defined here
       --> /workspace/.venv/lib/python3.14/site-packages/django-stubs/db/models/fields/related.pyi:186:9
        |
    186 |     def __init__(
        |         ^^^^^^^^
    187 |         self,
    188 |         to: type[Model] | str,
    189 |         on_delete: Callable[..., None],
        |         ------------------------------ Parameter declared here


``` 

## How is that should be

New database variants should be allowed.

## System information

- OS:
- `python` version: 3.14
- `django` version: 6.1
- `ty` version: 0.0.70
- `django-stubs` version: 6.1.0
- `django-stubs-ext` version: 6.1.0

felixxm:
# I have made things!

## Related issues

#3590 

## AI Policy


felixxm:
OK it's more complicated, tests pass without my fix and only crash when using `ty` :thinking:

UnknownPlatypus:
Does models.CASCADE also cause issue ?

felixxm:
> Does models.CASCADE also cause issue ?

Nope, it's only about new database variants. It's related with `__call__: _Collector`.

django-stubs/db/models/deletion.pyi
```diff
@@ -55,9 +55,8 @@ class DatabaseOnDelete:
     forced_collector: _Collector | None
 
     def __init__(self, operation: str, name: str, forced_collector: _Collector | None = ...) -> None: ...
-
-    __call__: _Collector
-
+    @staticmethod
+    def __call__(collector: Collector, field: Field[Any, Any], sub_objs: QuerySet[Model], using: str) -> None: ...
```
sobolevn: Can we use the same alias as the source does? https://github.com/django/django/blob/082b3df4067c3899dd4d57e8c2eca5baea9d07bb/django/db/models/deletion.py#L90C5-L90C26

```suggestion
    __call__ = DO_NOTHING
```

felixxm: Unfortunately not, it doesn't work on `ty` and `pyright` :disappointed:

sobolevn (approved):
Thanks!

felixxm:
I found a solution but it's a bit fishy.

PR #3593, opened 2026-08-13
Resolve objects through the metaclass and ban manager access on instances

### LINKED ISSUE #174: Incompatible types in assignment with User subclass
Heya, I've this class deriving from the (concrete) User class:
```python
from django.contrib.auth.models import User, UserManager

class DepositClient(User):
    objects = UserManager()
```
mypy with django-stubs 1.1.0 is complaining on the `objects` line that:
```
error: Incompatible types in assignment
(expression has type "UserManager[DepositClient]",
base class "User" defined the type as "UserManager[User]")
```
even if `DepositClient` is a subclass of `User` and the type variable in `Manager` is declared as covariant by django-stubs.

This should work out-of-the-box with django-stubs, right?
Or am I missing something here?

### LINKED ISSUE #1414: [Help wanted / bug ?] Trying to create UserModelMixins and ModelMixins; trouble typing manager.
I've tried the Gitter help link, but it appears to not be an active place (I left a question there a couple months ago and it never got an answer - and there's only about 5 more messages since then), so I'm trying my luck here.

For boilerplate purposes, I'm trying to provide multiple model Mixins, as well as a base abstract user model to be used in projects. This abstract user model should implement one of these Mixins (SoftDeleteMixin); this mixin implements it's own custom manager. That appears to be the start of the mess, as I can't make the types coherent.

Here are some snippets of the models and what I'm trying to achieve:
```py
# The mixin manager
SoftDeleteModel = TypeVar("SoftDeleteModel", bound=SoftDeleteMixin)
class SoftDeleteManager(models.Manager[SoftDeleteModel]):
    def get_queryset(self) -> models.QuerySet[SoftDeleteModel]: ...
    def include_deleted(self) -> models.QuerySet[SoftDeleteModel]: ...

# The mixin
class SoftDeleteMixin(models.Model):
     objects = SoftDeleteManager[Self]()

# The manager for the users - inherits from the other manager because the GenericUser type will subclass SoftDeleteMixin
GenericUser = TypeVar("GenericUser", bound=BaseAbstractUser)
class BaseUserManager(SoftDeleteManager[GenericUser], DjangoBaseUserManager[GenericUser]):
    def create_user(self, password: str, **fields: Any) -> GenericUser: ...
    def create_superuser(self, password: str, **fields: Any) -> GenericUser: ...

# The abstract user model
class BaseAbstractUser(PermissionsMixin, SoftDeleteMixin, DjangoAbstractBaseUser):
    objects = BaseUserManager[Self]()  # By default uses the base manager
```
Now this is generating loads of errors with mypy:
- `Incompatible types in assignment (expression has type "BaseUserManager[BaseAbstractUser]", base class "SoftDeleteMixin" defined the type as "SoftDeleteManager[SoftDeleteMixin]"`
  - This appears to be related to https://github.com/typeddjango/django-stubs/issues/174
- Subclasses have incompatible return type in the `objects` property; for example:
```py
class UserModel(BaseAbstractUser): ...
instance = UserModel.objects.create(...)  # instance will evaluate to BaseAbstractUser instead of UserModel
# Notably, Pylance on VSCode appears to correctly identify that `create` returns a UserModel, which seems weird to me.
```

Is this some sort of bug, or am I doing some kind of "anti-pattern" here, typing things incorrectly, etc? I've tried multiple changes with Generics and all that and always end up with some kind of issue in the end. Help would be greatly appreciated.

UnknownPlatypus:
### Do I need to annotate reverse relations?

Not if you use our mypy plugin. It reads your app registry and declares every reverse accessor for you, as
`RelatedManager[Model]`, or `ManyRelatedManager[Model, Through]` for a `ManyToManyField`.

pyright, pyrefly and ty do not run the plugin, so under those you declare them yourself.
Reach for `RelatedManager` rather than `Manager`: a reverse accessor is read from a model instance, whereas a regular manager such as `objects` only exists on the model class.

```python
from django.db import models
from django_stubs_ext.db.models.manager import ManyRelatedManager, RelatedManager


class Article(models.Model):
    category_set: RelatedManager[Category]
    tags: ManyRelatedManager[Tag, ArticleTag]
```

Make sure to import symbols from `django_stubs_ext.db.models.manager`. Django dynamically create these so they cannot be imported directly

---

# I have made things!

This fixes two issues around [`Model.objects`](https://docs.djangoproject.com/en/6.0/ref/models/class/#django.db.models.Model.objects):
- `MyModel().objects` was allowed while it's a runtime error -> I've used explicit `__get__(self, instance: None)` to solve that
- explicit `objects = BookManager()` are not overriding our stub-defined type for pyright/pyrefly so `objects` was a `Manager[Book]` and incorrect. -> I think I found a way to have our `objects` annotation to act as a true fallback using `__getattr__` on the `ModelBase` metaclass. By nature getattr is always checked last so now it's correct for every type checker!

I had to workaround two mypy issues that made this a bit more difficult but in the end it seem to work:
- https://github.com/python/mypy/issues/8203
- https://github.com/python/mypy/issues/13608

# Related issues
From https://github.com/typeddjango/django-stubs/pull/3559

> This also revealed that our typing of `Model.objects` as `ClassVar[Manager[Self]]` is already a bit problematic with pyright and pyrefly (see the type-ignore's). Given this exemple:
>
> ```python
> class Book(models.Model):
>     objects = BookManager()
>     foo = BookManager()
> ```
>
> I get these revealed types
>
> ```
> ┌─────────────┬───────────────┬─────────────┐
> │             │ Book.objects  │  Book.foo   │
> ├─────────────┼───────────────┼─────────────┤
> │ mypy+plugin │ BookManager   │ BookManager │
> ├─────────────┼───────────────┼─────────────┤
> │ ty          │ BookManager   │ BookManager │
> ├─────────────┼───────────────┼─────────────┤
> │ pyright     │ Manager[Book] │ BookManager │
> ├─────────────┼───────────────┼─────────────┤
> │ pyrefly     │ Manager[Book] │ BookManager │
> └─────────────┴───────────────┴─────────────┘
> ```
>
> To me our declaration in the stubs should not impact regular class level assignment inference and I would expect `objects` to have the same type as `foo`. The plugin should only apply the "does not exist on abstract model" bit, the rest should be regular inference.

Fixes #174
Fixes #1414
Also improves #579

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -1021,10 +1021,17 @@ class MetaclassAdjustments(ModelClassInitializer):
         if ctx.cls.fullname != fullnames.MODEL_CLASS_FULLNAME:
             return
 
-        for attr_name in ["DoesNotExist", "NotUpdated", "MultipleObjectsReturned", "objects"]:
+        for attr_name in ["DoesNotExist", "NotUpdated", "MultipleObjectsReturned"]:
             attr = ctx.cls.info.names.get(attr_name)
             if attr is not None and isinstance(attr.node, Var) and not attr.plugin_generated:
                 del ctx.cls.info.names[attr_name]
+
+        # `ModelBase.__getattr__` is the `objects` fallback for the other type checkers.
+        # mypy ignores its `Literal["objects"]` restriction (https://github.com/python/mypy/issues/8203),
+        # so it would swallow unknown attributes. `objects` is already inserted by the plugin.
+        metaclass = ctx.cls.info.metaclass_type
+        if metaclass is not None and "__getattr__" in metaclass.type.names:
+            del metaclass.type.names["__getattr__"]
 
     def get_exception_bases(self, name: str) -> list[Instance]:
         bases = []
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -1016,6 +1016,14 @@ class MetaclassAdjustments(ModelClassInitializer):
 
         Turn this setting off at your own risk.
         """
+        # Mypy does not properly support narrowing in `__getattr__` based on `Literal["objects"]`
+        # so the current `ModelBase.__getattr__` swallows every unknown attributes errors.
+        # `objects` is already inserted by the plugin, so the fallback is not necessary for mypy. We can drop it
+        # TODO: remove once https://github.com/python/mypy/issues/8203 is resolved
+        metaclass = ctx.cls.info.metaclass_type
+        if metaclass is not None and "__getattr__" in metaclass.type.names:
+            del metaclass.type.names["__getattr__"]
+
         if not plugin_config.strict_model_abstract_attrs:
             return
         if ctx.cls.fullname != fullnames.MODEL_CLASS_FULLNAME:
@@ -1025,13 +1033,6 @@ class MetaclassAdjustments(ModelClassInitializer):
             attr = ctx.cls.info.names.get(attr_name)
             if attr is not None and isinstance(attr.node, Var) and not attr.plugin_generated:
                 del ctx.cls.info.names[attr_name]
-
-        # `ModelBase.__getattr__` is the `objects` fallback for the other type checkers.
-        # mypy ignores its `Literal["objects"]` restriction (https://github.com/python/mypy/issues/8203),
-        # so it would swallow unknown attributes. `objects` is already inserted by the plugin.
-        metaclass = ctx.cls.info.metaclass_type
-        if metaclass is not None and "__getattr__" in metaclass.type.names:
-            del metaclass.type.names["__getattr__"]
 
     def get_exception_bases(self, name: str) -> list[Instance]:
         bases = []
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -1024,9 +1024,20 @@ class MetaclassAdjustments(ModelClassInitializer):
         if metaclass is not None and "__getattr__" in metaclass.type.names:
             del metaclass.type.names["__getattr__"]
 
-        if not plugin_config.strict_model_abstract_attrs:
-            return
         if ctx.cls.fullname != fullnames.MODEL_CLASS_FULLNAME:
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -66,7 +66,7 @@ pyrefly = [
   "psycopg2-binary"
 ]
 ty = [
-  "ty==0.0.69",
+  "ty==0.0.71",
```
UnknownPlatypus: This only works for `ty>0.0.69` but we don't really declare a minimal supported version. 
I suppose it's ok for now but we maybe should used `compatible-ty` extras like `compatible-mypy` at some point ? But it will be extra maintenance with test to run on both version, not sure we want to go down that road

mypy_django_plugin/transformers/models.py
```diff
@@ -1024,11 +1024,22 @@ def get_model_count(model_type: type[models.Model]) -> int:
         if metaclass is not None and "__getattr__" in metaclass.type.names:
             del metaclass.type.names["__getattr__"]
 
-        if not plugin_config.strict_model_abstract_attrs:
-            return
         if ctx.cls.fullname != fullnames.MODEL_CLASS_FULLNAME:
             return
 
+        if not plugin_config.strict_model_abstract_attrs:
```
UnknownPlatypus, line 1037: Not happy with that but seem to be required to honor this setting documentation
See [failing test](https://github.com/typeddjango/django-stubs/actions/runs/31736524419/job/94569374207?pr=3593) prior to that

sobolevn (approved):
Looks solid! Thanks a lot, this is much better now!

UnknownPlatypus:
Should we apply a similar pattern to our defined manager ? For ex on the User model we still use classvar

sobolevn:
We should unify all type-checker results as much as possible :)

UnknownPlatypus:
> We should unify all type-checker results as much as possible :)

I think it's worth merging as is, will send a followup. 
It was not that easy actually. The plugin still tries to insert manager classvar for these so I was getting conficts, probably need more plugin work (or in fact, to remove some plugin work)

sobolevn (approved):

PR #3597, opened 2026-08-14
Fix bogus name-defined error for cross-module as_manager overrides

### LINKED ISSUE #3596: Subclassing another library's QuerySet triggers a false "Name ... is not defined" error since 6.0.7
# Bug report

## What's wrong

I was upgrading django-stubs 6.0.5 to 6.0.7 (also unblocking mypy 2.1 to 2.3, which 6.0.5's `compatible-mypy` pin holds back), and mypy started failing with:

```
app/models.py:182: error: Name "_All" is not defined  [name-defined]
Found 1 error in 1 file
```

`_All` appears nowhere in my code. It's a module-private `TypeVar` in [django-polymorphic](https://github.com/django-commons/django-polymorphic), used in the return annotation of `PolymorphicQuerySet.as_manager` in [`polymorphic/query.py`](https://github.com/django-commons/django-polymorphic/blob/main/src/polymorphic/query.py). I triggered the error just by subclassing that queryset in my models module.

Two more things about the message are misleading:

- **The line number is polymorphic's, not mine.** 182 is where `as_manager` is declared in polymorphic's `query.py`, printed against *my* filename.
- **The reported name is arbitrary.** It's whatever the consuming module happens not to have in scope. Remove `from django.db import models` from the consumer below and the same bug reports `Name "models" is not defined`. (This may be why it hasn't been reported under one recognisable signature.)

### Failing test case

A pytest-mypy-plugins case for `tests/typecheck/managers/querysets/test_as_manager.yml` (django-polymorphic isn't required; a tiny stand-in library is enough):

```yaml
-   case: overridden_as_manager_in_other_module_with_module_private_typevar
    main: |
    installed_apps:
        - django.contrib.contenttypes
        - myapp
    files:
        -   path: lib/__init__.py
        -   path: lib/query.py
            content: |
                from typing import TypeVar

                from django.db import models
                from typing_extensions import override

                _T = TypeVar("_T", bound=models.Model)

                class LibQuerySet(models.QuerySet[_T]):
                    @classmethod
                    @override
                    def as_manager(cls) -> "models.Manager[_T]":
                        return models.Manager.from_queryset(cls)()
        -   path: myapp/__init__.py
        -   path: myapp/models.py
            content: |
                from django.contrib.contenttypes.models import ContentType
                from django.db import models

                from lib.query import LibQuerySet

                class MyModel(models.Model):
                    ctype = models.ForeignKey(ContentType, on_delete=models.CASCADE)

                class MyQuerySet(LibQuerySet):
                    pass
```

On current `master` (`8980afff`) it fails with:

```
E   Actual:
E     myapp/models:11: error: Name "_T" is not defined  [name-defined] (diff)
E   Expected:
E     (empty)
```

(Note `myapp/models:11`: line 11 is where `as_manager` is declared in `lib/query.py`, printed against the consumer's filename.)

### Standalone reproducer

The regression comes from PR #3460 ("Fix manager resolution when QuerySet overrides as_manager"), first shipped in 6.0.7: stubs 6.0.5 is clean and 6.0.7 errors, on both mypy 2.1.0 and 2.3.0. The same setup as a plain project:

`lib/query.py`, standing in for any library that overrides `as_manager` with a module-private TypeVar in its annotation:

```python
from typing import TypeVar

from django.db import models

_T = TypeVar("_T", bound=models.Model)


class LibQuerySet(models.QuerySet[_T]):
    @classmethod
    def as_manager(cls) -> "models.Manager[_T]":
        return models.Manager.from_queryset(cls)()
```

`app/models.py`:

```python
from django.contrib.contenttypes.models import ContentType
from django.db import models

from lib.query import LibQuerySet


class Thing(models.Model):
    ctype = models.ForeignKey(ContentType, on_delete=models.CASCADE)


class ThingQuerySet(LibQuerySet):
    pass
```

`config/settings.py`: `INSTALLED_APPS = ["django.contrib.contenttypes", "app"]`; `pyproject.toml` enables `mypy_django_plugin.main` with that settings module.

There's no `from_queryset`, no `as_manager()` call, no manager assignment: declaring the subclass is enough. The `ContentType` FK is load-bearing: the plugin's `get_additional_deps` injects a reverse dependency from a `django.*` module into `app.models`, which puts consumer and library in the same SCC with the consumer scheduled first (`AUTH_USER_MODEL` pointing at a model in that module does the same).

### Root cause

`add_as_manager_to_queryset_class` (since PR #3460) re-resolves the inherited override's return annotation with `semanal_api.anal_type(...)`, which resolves names against the **current** module. When the defining module hasn't been analyzed yet (`base_as_manager.node.type is None`), the annotation is still a raw `UnboundType`, so the library's module-private names fail to resolve in the consumer's namespace. 6.0.5 bailed out early in exactly that state, which is why this is a 6.0.7 regression.

## How is that should be

No error (the 6.0.5 behaviour). I have the failing test case above and a candidate fix (suppress resolution errors from that `anal_type` call and bail, keeping #3460's same-module fix intact) ready, and will open a PR shortly.

## System information

- OS: Linux (Fedora 43)
- `python` version: 3.13
- `django` version: 6.0.6 (also reproduced on 6.1)
- `mypy` version: 2.1.0 and 2.3.0 (both affected with 6.0.7; both clean with 6.0.5)
- `django-stubs` version: 6.0.7 broken / 6.0.5 clean
- `django-stubs-ext` version: 6.0.2

BenA-SA:
Since #3460, [add_as_manager_to_queryset_class](https://github.com/typeddjango/django-stubs/blob/8980afff2f297758809b1300257bde35a9c24bec/mypy_django_plugin/transformers/managers.py#L371) resolves an inherited as_manager override's return annotation with [semanal_api.anal_type()](https://github.com/typeddjango/django-stubs/blob/8980afff2f297758809b1300257bde35a9c24bec/mypy_django_plugin/transformers/managers.py#L422). 

That resolves names against the current module. When the override lives in another module that mypy has not analyzed yet (consumer scheduled before the library inside the same SCC), the annotation is still an UnboundType, and any name private to the defining module - typically a TypeVar - fails to resolve, producing a spurious 'Name "..." is not defined' error reported against the consumer's file with the library's line number.

Suppress resolution errors from that call and bail out without generating a manager, restoring the pre-6.0.7 behaviour for the cross-module case while keeping #3460's fix for same-module overrides.

## Related issues
Fixes #3596

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/managers.py
+++ b/mypy_django_plugin/transformers/managers.py
@@ -419,7 +419,12 @@ def add_as_manager_to_queryset_class(ctx: ClassDefContext) -> None:
             return
         # A user override's return annotation may still be unanalyzed here, so
         # resolve it explicitly (a no-op if already analyzed); defer if incomplete.
-        base_as_manager_ret_type = semanal_api.anal_type(base_as_manager_type.ret_type)
+        # Suppress resolution errors: a not-yet-analyzed cross-module override may
+        # reference names private to its defining module — bail out, don't report.
+        with semanal_api.msg.filter_errors() as resolution_errors:
+            base_as_manager_ret_type = semanal_api.anal_type(base_as_manager_type.ret_type)
+        if resolution_errors.has_new_errors():
+            return
         if base_as_manager_ret_type is None:
             return _defer()
         base_as_manager_ret_type = get_proper_type(base_as_manager_ret_type)
--- a/tests/typecheck/managers/querysets/test_as_manager.yml
+++ b/tests/typecheck/managers/querysets/test_as_manager.yml
@@ -98,6 +98,42 @@
                 class MyModel(models.Model):
                     objects = MyQuerySet.as_manager()
 
+-   case: overridden_as_manager_in_other_module_with_module_private_typevar
+    main: |
+        from myapp.models import MyModel
+    installed_apps:
+        - django.contrib.contenttypes
+        - myapp
+    files:
+        -   path: lib/__init__.py
+        -   path: lib/query.py
+            content: |
+                from typing import TypeVar
+
+                from django.db import models
+                from typing_extensions import override
+
+                _T = TypeVar("_T", bound=models.Model)
+
+                class LibQuerySet(models.QuerySet[_T]):
+                    @classmethod
+                    @override
+                    def as_manager(cls) -> "models.Manager[_T]":
+                        return models.Manager.from_queryset(cls)()
+        -   path: myapp/__init__.py
+        -   path: myapp/models.py
+            content: |
+                from django.contrib.contenttypes.models import ContentType
+                from django.db import models
+
+                from lib.query import LibQuerySet
+
+                class MyModel(models.Model):
+                    ctype = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+
+                class MyQuerySet(LibQuerySet):
+                    pass
+
 -   case: declares_manager_type_like_django
     main: |
         from typing_extensions import reveal_type


mypy_django_plugin/transformers/managers.py
```diff
@@ -419,7 +419,12 @@ def _defer() -> None:
             return
         # A user override's return annotation may still be unanalyzed here, so
         # resolve it explicitly (a no-op if already analyzed); defer if incomplete.
-        base_as_manager_ret_type = semanal_api.anal_type(base_as_manager_type.ret_type)
```
UnknownPlatypus, line 422: Isn't there already a flag in anal_type to suppress such error ?

BenA-SA: I tested this with the two error-related flags [the plugin-facing `anal_type` exposes](https://github.com/python/mypy/blob/v2.3.0/mypy/plugin.py#L324-L333) (`allow_unbound_tvars` and `report_invalid_types`), each on its own and both together. My suggested test case fails identically in every combination.

I think that's because those flags never reach this error. It's a name lookup failure rather than an invalid type: [`TypeAnalyser.lookup_qualified` calls `api.lookup_qualified` with `suppress_errors=False` hardcoded](https://github.com/python/mypy/blob/v2.3.0/mypy/typeanal.py#L286-L289), so the `Name "..." is not defined` error is emitted by [`SemanticAnalyzer.name_not_defined`](https://github.com/python/mypy/blob/v2.3.0/mypy/semanal.py#L7536) before either flag is consulted. As far as I can tell, [`report_invalid_types` only gates the "Invalid type: try using Literal[...]" family of messages](https://github.com/python/mypy/blob/v2.3.0/mypy/typeanal.py#L1433-L1440), and `allow_unbound_tvars` applies to TypeVars that resolve but are not bound in scope.

I think [`filter_errors()`](https://github.com/python/mypy/blob/v2.3.0/mypy/messages.py#L193) is the closest tool for the job, but happy to switch if there's a better mechanism I've missed.

PR #3606, opened 2026-08-23
Annotate `FileField.save_form_data`

brianhelba:
`FileField.save_form_data` can have a more specific annotation than its parent `Field.save_form_data`.

Per the code comments and behavior:
https://github.com/django/django/blob/03988c5a5ba248c3b9b11ea96fd4fda5e98849aa/django/db/models/fields/files.py#L363-L371 the `data` parameter can be:
* `None`
* `False`
* a `File` instance (the normal case)
* a string (implicitly accepted, as `False` internally transforms to `""`, and the ultimate field value upon save is a string)

## AI Policy

### DIFF
--- a/django-stubs/db/models/fields/files.pyi
+++ b/django-stubs/db/models/fields/files.pyi
@@ -104,6 +104,8 @@ class FileField(Field[Any, Any]):
     @override
     def contribute_to_class(self, cls: type[Model], name: str, **kwargs: Any) -> None: ...  # type: ignore[override]
     def generate_filename(self, instance: Model | None, filename: StrPath) -> str: ...
+    @override
+    def save_form_data(self, instance: Model, data: File[Any] | str | bool | None) -> None: ...
 
 class ImageFileDescriptor(FileDescriptor):
     field: ImageField


django-stubs/db/models/fields/files.pyi
```diff
@@ -104,6 +104,9 @@ class FileField(Field[Any, Any]):
     @override
     def contribute_to_class(self, cls: type[Model], name: str, **kwargs: Any) -> None: ...  # type: ignore[override]
     def generate_filename(self, instance: Model | None, filename: StrPath) -> str: ...
+    @override
+    def save_form_data(self, instance: Model, data: File[Any] | str | bool | None) -> None:
```
brianhelba: For comparison, here's the parent that this now overrides: https://github.com/brianhelba/django-stubs/blob/0b54fe159afb510122895cf2ab10a2d21686cbb0/django-stubs/db/models/fields/__init__.pyi#L254

sobolevn: You are missing `...` here

brianhelba: Fixed.

brianhelba:
@sobolevn Fixed the issue (sorry for the silly mistake). PTAL again.

UnknownPlatypus (approved):
Looking good thanks

PR #3608, opened 2026-08-23
Don't build a ManyRelatedManager from a placeholder _Through

### LINKED ISSUE #3605: Plugin crashes on mypy >= 2.2: NotImplementedError: Cannot serialize PlaceholderType instance
### What's wrong

On a large Django codebase (~15,200 files checked), the `mypy_django_plugin` crashes during cache/interface serialization under mypy >= 2.2:

```
NotImplementedError: Cannot serialize PlaceholderType instance
```

The crash is fatal — mypy exits 2, so it takes out a gating CI check.

Full traceback (mypy 2.3.1, `--num-workers 1 --no-incremental --show-traceback`):

```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "mypy/build_worker/worker.py", line 126, in main
  File "mypy/build_worker/worker.py", line 228, in serve
  File "mypy/build.py", line 4900, in process_stale_scc_interface
  File "mypy/build.py", line 3604, in write_cache
  File "mypy/build.py", line 2325, in write_cache
  File "mypy/nodes.py", line 599, in write
  File "mypy/nodes.py", line 5169, in write
  File "mypy/nodes.py", line 5074, in write
  File "mypy/nodes.py", line 4321, in write
  File "mypy/nodes.py", line 1760, in write
  File "mypy/types.py", line 4553, in write_type_list
  File "mypy/types.py", line 742, in write
  File "mypy/types.py", line 318, in write
NotImplementedError: Cannot serialize PlaceholderType instance
```

### The crash is not tied to any particular module

This is the part that makes it awkward to reduce. mypy attributes the crash to one module, but **that module is arbitrary**. When I stripped the reported file down to two `StrEnum` classes and a one-field model, the crash did not go away — it **relocated** to a structurally unrelated repository module in a different app:

| Run | Crash attributed to |
| --- | --- |
| unmodified tree (6 runs) | `contracts_v3/workflow_analytics/data/models.py` |
| after gutting that file (2 runs) | `inline_editor_v2/editable_document/data/db_repo.py` |

Neither file is unusual — plain `SDBaseDBModel` subclasses / repository helpers. Nothing about their contents appears to matter; what matters is that *something* in the graph leaves a `PlaceholderType` behind.

I suspect this is the same failure mode as python/mypy#21640, where @ilevkivskyi explained the contract for plugins:

> Whenever you store some type (that was analyzed/constructed by you, as opposite to mypy internals) as an attribute of a symbol, you should check if it `has_placeholders()`. If it does, you check if `final_iteration` is `True`. If it is, you give an error about cyclic definition, otherwise you call `.defer()`

That issue was closed because the reporter fixed their own plugin. If `mypy_django_plugin` has the same gap, this would be it surfacing on a codebase large enough to hit the cyclic case.

### Version bisect

Bisected against a fixed codebase and fixed django-stubs 6.1.0:

| mypy | Crash | `explicit-any` reported |
| --- | --- | --- |
| 1.5.1 | no | 0 |
| 1.20.2 | no | 1325 |
| 2.0.0 | no | 1325 |
| 2.1.0 | no | 1325 |
| **2.2.0** | **yes** | 138 |
| **2.3.1** | **yes** | 138 |

The crash appears in exactly the release that changed placeholder handling. (The `explicit-any` column is incidental but shows 2.2 also changed `dataclass_transform` handling substantially.)

### Ruled out

Each of these still crashes:

- **django-stubs version** — 6.0.9 and 6.1.0 both crash identically, so it is not a 6.1 regression.
- **Incremental cache** — `--no-incremental` (the serialization happens for worker IPC regardless).
- **Parallel workers** — `--num-workers 1`.
- **`disallow_any_explicit`** — setting it to `False` globally.
- **A single offending file** — see the migration table above.

### How to reproduce

I can't share the codebase, and I haven't been able to reduce it — by its nature the trigger is a graph-level cycle rather than a construct, and editing files just moves the crash. Happy to run any diagnostic patch or instrumented build against the real tree and report back, which is probably the fastest route to a reduction.

### System information and files

- `mypy` 2.3.1 (also 2.2.0)
- `django-stubs` 6.1.0 (also 6.0.9)
- `djangorestframework-stubs` 3.18.0
- Django 5.2.17
- Python 3.12.12
- Plugins: `mypy_django_plugin.main`, `mypy_drf_plugin.main`
- ~15,200 files checked

codebreach:
# I have made things!

## Related issues

Fixes #3605

## What was wrong

`ProcessManyToManyFields.create_many_related_manager` (`mypy_django_plugin/transformers/models.py`) reuses the `_Through` `TypeVar` from `ManyRelatedManager` when it synthesises `<Model>_ManyRelatedManager`:

```python
through_type_var = self.many_related_manager.defn.type_vars[1]
assert isinstance(through_type_var, TypeVarType)
generic_to_many_related_manager = Instance(self.many_related_manager, [model, through_type_var.copy_modified()])
related_manager_info = helpers.add_new_class_for_module(...)
related_manager_info.defn.type_vars = [through_type_var.copy_modified()]
related_manager_info.add_type_vars()
```

`_Through` is `TypeVar("_Through", bound=Model, default=Model)` in `django-stubs/db/models/fields/related_descriptors.pyi`. If that stub module is semantically analysed while its own `from django.db.models.base import Model` is still a `PlaceholderNode`, `_Through.upper_bound` is a `PlaceholderType`. Nothing here checks for that, and `set_many_to_many_manager_info` means the class is never revisited, so the placeholder is still there at `write_cache` time:

```
  File "mypy/build.py", line 4900, in process_stale_scc_interface
  File "mypy/build.py", line 3604, in write_cache
  File "mypy/build.py", line 2325, in write_cache
  File "mypy/nodes.py", line 599, in write        # MypyFile.write -> self.names.write(...)
  File "mypy/nodes.py", line 5169, in write       # SymbolTable.write
  File "mypy/nodes.py", line 5074, in write       # SymbolTableNode.write -> self.node.write(data)
  File "mypy/nodes.py", line 4321, in write       # TypeInfo.write -> self.defn.write(data)
  File "mypy/nodes.py", line 1760, in write       # ClassDef.write -> write_type_list(data, self.type_vars)
  File "mypy/types.py", line 4553, in write_type_list
  File "mypy/types.py", line 742, in write        # TypeVarType.write -> self.upper_bound.write(data)
  File "mypy/types.py", line 318, in write
NotImplementedError: Cannot serialize PlaceholderType instance
```

Both other places that store type vars taken from another class already guard against this — `create_manager_class` and `reparametrize_generic_class`, in `mypy_django_plugin/transformers/managers.py` — this one didn't. It is the same plugin contract described in python/mypy#21640.

## Evidence

Walking every module symbol table at the end of a build over a ~15,200 file codebase (mypy 2.3.1, django-stubs 6.1.0, `--num-workers 1`) found 39 classes whose `defn.type_vars` still contained a placeholder. All 39 are `<Model>_ManyRelatedManager`, including django's own:

```
POISONED django.contrib.auth.models.Group_ManyRelatedManager       tv=_Through bound=<placeholder django.db.models.fields.related_descriptors.Model>
POISONED django.contrib.auth.models.Permission_ManyRelatedManager  tv=_Through bound=<placeholder django.db.models.fields.related_descriptors.Model>
POISONED sd_organizations.models.CompanyProfile_ManyRelatedManager tv=_Through bound=<placeholder django.db.models.fields.related_descriptors.Model>
```

Logging the analysis order shows a narrow window between `related_descriptors`' first pass and its re-analysis. Every M2M manager built inside that window is affected; the ones built after are fine, which is why the module mypy blames moves around when you edit files:

```
line 18658  ManyRelatedManager analysed, _Through.bound = <placeholder ...Model>
line 18669  Group_ManyRelatedManager created, bound = <placeholder ...>
   ...      39 managers created inside the window
line 18836  ManyRelatedManager re-analysed, _Through.bound = django.db.models.base.Model
```

## The fix

Check `has_placeholder` and raise `helpers.IncompleteDefnException` — `process_model_defn` already turns that into `ctx.api.defer()`, exactly as `create_manager_class` does for the same situation.

## Test

No test, per review — reproducing the SCC ordering from scratch needs thousands of modules. I tried ~400 synthetic project shapes and could open the window (`_Through.upper_bound` observably becomes a `PlaceholderType`), but never with an M2M manager being built inside it. Verified against the real codebase instead, below.

## Verification

* The ~15,200 file codebase from #3605: `INTERNAL ERROR ... Cannot serialize PlaceholderType instance` (exit 2) before, `Found 1399 errors in 854 files (checked 15231 source files)` (exit 1) after. No crash, and no new error mentioning `ManyRelatedManager`.
* `pytest tests`: same result before and after the change locally (442 passed, 12 failed — the 12 are postgres/gis extras missing in my environment, unrelated to this change).
* Repeated runs are now stable: before this change the crash relocated between unrelated modules from run to run, and the build never reached a summary line.

## One thing I noticed but did not change

`reparametrize_generic_class` in `mypy_django_plugin/transformers/managers.py` calls `ctx.api.defer()` when it finds placeholders and then falls through and assigns them anyway:

```python
if any(has_placeholder(type_var) for type_var in type_vars):
    if not ctx.api.final_iteration:
        ctx.api.defer()
    else:
        return

parent_class.args = tuple(type_vars)
class_info.node.defn.type_vars = type_vars
```

That fires 84 times on the codebase above and self-heals, because mypy's `setup_type_vars` resets `defn.type_vars` when the class is re-analysed. So it is not this bug, but the `defer()` branch storing placeholders looks unintended. Happy to send a separate PR if you want it tightened.

## AI Policy

### DIFF
--- a/mypy_django_plugin/transformers/models.py
+++ b/mypy_django_plugin/transformers/models.py
@@ -23,6 +23,7 @@ from mypy.nodes import (
 )
 from mypy.plugins import common
 from mypy.semanal import SemanticAnalyzer
+from mypy.semanal_shared import has_placeholder
 from mypy.typeanal import TypeAnalyser
 from mypy.types import AnyType, Instance, ProperType, TypedDictType, TypeOfAny, TypeType, TypeVarType, get_proper_type
 from mypy.types import Type as MypyType
@@ -971,6 +972,12 @@ class ProcessManyToManyFields(ModelClassInitializer):
         # class X_ManyRelatedManager(ManyRelatedManager[X, _Through], type(X._default_manager), Generic[_Through]): ...
         through_type_var = self.many_related_manager.defn.type_vars[1]
         assert isinstance(through_type_var, TypeVarType)
+        if has_placeholder(through_type_var):
+            # 'ManyRelatedManager' hasn't been fully analysed yet, so '_Through' still
+            # carries a 'PlaceholderType' bound. Copying it into the class we're about to
+            # create would store an unserializable type on a synthetic class that is never
+            # re-analysed, so defer instead and pick it up on a later iteration.
+            raise helpers.IncompleteDefnException()
         generic_to_many_related_manager = Instance(self.many_related_manager, [model, through_type_var.copy_modified()])
         related_manager_info = helpers.add_new_class_for_module(
             module=self.api.modules[model.type.module_name],


tests/test_m2m_placeholder_serialization.py
sobolevn, line 1: please, convert this test to our regular `.yml` based test

mypy_django_plugin/transformers/models.py
```diff
@@ -971,6 +972,12 @@ def create_many_related_manager(self, model: Instance) -> None:
         # class X_ManyRelatedManager(ManyRelatedManager[X, _Through], type(X._default_manager), Generic[_Through]): ...
         through_type_var = self.many_related_manager.defn.type_vars[1]
         assert isinstance(through_type_var, TypeVarType)
+        if has_placeholder(through_type_var):
```
sobolevn, line 975: Question: how can it be both `TypeVarType` and `PlaceholderType`?

codebreach:
Confirming this fixes the crash on the tree that produced #3605.

Built an env pinned to this branch (`django-stubs @ git+…@fix/3605-m2m-placeholder-serialization`, `django-stubs-ext` 6.1.0, `djangorestframework-stubs` 3.18.0, mypy 2.3.1) and ran the full tree — Django 5.2.17, Python 3.12, ~15,200 files, `mypy_django_plugin.main` + `mypy_drf_plugin.main`:

```
Found 1393 errors in 853 files (checked 15231 source files)
```

**No `INTERNAL ERROR`, and the build completes.** For comparison, every run before this branch died without ever reaching a summary line — the `Cannot serialize PlaceholderType instance` failure aborted the build, so the error counts those runs printed were partial. This is the first complete run of this codebase on mypy ≥2.2, so the jump from "158 errors" to 1393 is the crash no longer truncating the build, not a regression introduced here.

Also worth recording, since it may help anyone else who lands on #3605: the crash **migrates between modules** when unrelated files change. On our tree it was attributed to `contracts_v3/workflow_analytics/data/models.py` across six runs, then relocated to `inline_editor_v2/editable_document/data/db_repo.py` after I gutted the first file to two enums and a one-field model. Neither file is unusual, and neither has an M2M — which is consistent with your diagnosis, since the poisoned `_Through` lives on the synthetic `<Model>_ManyRelatedManager` rather than on whichever module happens to be serialised first.

Things that made no difference, in case they save someone a bisect: `--no-incremental`, `--num-workers 1`, `disallow_any_explicit = False`, and django-stubs 6.0.9 (so this is not a 6.1 regression). Bisected cleanly to mypy 2.2.0.

Happy to re-run against any further iteration of this branch.

UnknownPlatypus:
See @sobolevn review pls, happy to merge once it's addressed

mypy_django_plugin/transformers/models.py
ilevkivskyi: Upper bound of a `TypeVar` can have placeholders (and it is actually quite common, when people use a class defined later in file as a `TypeVar` upper bound).

ilevkivskyi: To be clear I am not saying this is what happens here, just explaining why `has_placeholder()` returning `True` for a type variable is not surprising at all :-)

sobolevn: Thank you! Indeed :)

codebreach: it seems this can happen with circular imports or deeply nested imports where files depend in a circular fashion. Haven't been able to confirm but thats my hypothesis based on all the debugging

tests/test_m2m_placeholder_serialization.py
codebreach: 👍 incoming

tests/typecheck/fields/test_related.yml
```diff
@@ -1737,3 +1737,74 @@
...
+# 'django.db.models.fields.related_descriptors' can be analysed while its own
+# 'from django.db.models.base import Model' is still a 'PlaceholderNode'. That
+# leaves 'ManyRelatedManager._Through' with a 'PlaceholderType' upper bound for
+# one iteration, and any '<Model>_ManyRelatedManager' built during that window
+# used to copy it onto a class that is never re-analysed, so the placeholder
+# reached 'write_cache' and crashed mypy with
+# 'NotImplementedError: Cannot serialize PlaceholderType instance'.
+# Reproducing that ordering needs thousands of modules, so the plugin below
+# recreates the same intermediate state instead.
+-   case: test_m2m_manager_is_not_built_from_a_placeholder_through
```
sobolevn: This test seems overly complex, if you can try this change locally and it fixes the crash for you - I would say that this is enough.

codebreach: i have tried it locally and was able to reliably get mypy to pass (as opposed to fail at random places)

not sure how to add a test

sobolevn: ok, let's just remove it :)

codebreach: removed :)

PR #3610, opened 2026-08-24
Fix monkeypatch() erasing parametrization of real Generic subclasses

### LINKED ISSUE #3609: django_stubs_ext.monkeypatch() breaks real Generic subclasses of View (ext >= 6.0.8)
# Bug report

## What's wrong

Since `django-stubs-ext` 6.0.8, `monkeypatch()` adds `View` (and `TemplateResponseMixin`) to the patched classes - a follow-up to #2874 ("Make `View`, `TemplateView`, and `TemplateResponseMixin` generic on response type"). The patch assigns:

```python
el.cls.__class_getitem__ = classmethod(lambda cls, *args, **kwargs: cls)
```

This breaks any class that subclasses `View` **and** is a real `typing.Generic` at runtime. In such a class the MRO is `Subclass -> View -> Generic -> object`, so the patched `View.__class_getitem__` now shadows `Generic.__class_getitem__`, and `Subclass[SomeArg]` silently evaluates to the bare `Subclass` instead of a `_GenericAlias`. The parametrization is lost at runtime.

### Minimal repro

```python
import django

settings.configure()
django.setup()

import django_stubs_ext

django_stubs_ext.monkeypatch()

from typing import Generic, TypeVar, get_args, get_origin

from django.views import View

T = TypeVar("T")


class Controller(View, Generic[T]):
    pass


alias = Controller[int]
print(alias)              # <class '__main__.Controller'>  (expected: __main__.Controller[int])
print(get_origin(alias))  # None                           (expected: Controller)
print(get_args(alias))    # ()                             (expected: (<class 'int'>,))
```

With `django-stubs-ext <= 6.0.7` the same script prints the expected values.

### Real-world impact

`django-modern-rest` (dmr) builds its whole controller API on `class Controller(View, Generic[_SerializerT_co])` and infers the serializer from the type args of `types.get_original_bases(cls)` in `__init_subclass__`. With ext >= 6.0.8 the parametrization is erased, so every project that calls `django_stubs_ext.monkeypatch()` in settings (as the docs recommend) crashes on import of any dmr controller:

```
dmr.exceptions.UnsolvableAnnotationsError: Type args () are not correct for
<class '...views.BlogPostCreate'>, at least 1 type arg must be provided
```

The failure is non-obvious to diagnose because the traceback points at the downstream library, not at the monkeypatch, and the breakage is silent for anyone who merely loses the parametrization without a hard check.

## How it should be

The runtime shim should preserve normal `__class_getitem__` resolution for subclasses that provide a real implementation later in the MRO. In particular, patching `View` should not shadow `Generic.__class_getitem__` in a subclass such as `class Controller(View, Generic[T])`.

One possible approach is to delegate to the remainder of the subclass MRO first, and fall back to the current no-op behavior only when no other `__class_getitem__` is available, e.g. conceptually:

```python
def _make_class_getitem(patched_cls):
    def __class_getitem__(cls, item):
        try:
            return super(patched_cls, cls).__class_getitem__(item)
        except AttributeError:
            return cls
    return classmethod(__class_getitem__)
```

so that `Controller[int]` (where `Generic` follows `View` in the MRO) still resolves via `Generic.__class_getitem__`, while plain `View[HttpResponse]` keeps returning `View` as today.

The exact implementation may differ; the important part is that adding runtime subscription support to `View` should not erase existing generic semantics of its subclasses.

## System information

- OS: Linux (Docker, python:3.14 image)
- `python` version: 3.14
- `django` version: 6.0.8
- `django-stubs` version: 6.1.0
- `django-stubs-ext` version: broken on 6.0.8, 6.0.9, 6.1.0; last good: 6.0.7
- `mypy` version: n/a (runtime-only issue)

mylittletraf:
This should fix the related issue. I implemented the approach described in #3609.

## Related issues

Fixes #3609 

## AI Policy

### DIFF
--- a/ext/django_stubs_ext/patch.py
+++ b/ext/django_stubs_ext/patch.py
@@ -134,14 +134,11 @@ if VERSION >= (6, 0):
 
 def _make_class_getitem(patched_cls: type) -> Any:
     def __class_getitem__(cls: type, item: Any) -> Any:
-        # `cls` may be a subclass that also inherits from `typing.Generic`.
-        # In that case the patched class shadows `Generic.__class_getitem__` in
-        # the MRO, which would silently erase the runtime parametrization of
-        # `cls[...]` (https://github.com/typeddjango/django-stubs/issues/3609).
-        # Delegate to the next `__class_getitem__` in the MRO if there is one,
-        # and only fall back to the no-op behavior of returning `cls` as is.
+        # A patched class may shadow `Generic.__class_getitem__` in a subclass MRO.
+        # Delegate to the next implementation when present, otherwise keep the
+        # existing no-op behavior.
         mro = cls.__mro__
-        for base in mro[mro.index(patched_cls) + 1 :]:
+        for base in mro[mro.index(patched_cls) + 1 : -1]:
             getitem = base.__dict__.get("__class_getitem__")
             if getitem is not None:
                 return getitem.__get__(None, cls)(item)
--- a/ext/django_stubs_ext/patch.py
+++ b/ext/django_stubs_ext/patch.py
@@ -132,6 +132,24 @@ if VERSION >= (6, 0):
     )
 
 
+def _make_class_getitem(patched_cls: type) -> Any:
+    def __class_getitem__(cls: type, item: Any) -> Any:
+        # `cls` may be a subclass that also inherits from `typing.Generic`.
+        # In that case the patched class shadows `Generic.__class_getitem__` in
+        # the MRO, which would silently erase the runtime parametrization of
+        # `cls[...]` (https://github.com/typeddjango/django-stubs/issues/3609).
+        # Delegate to the next `__class_getitem__` in the MRO if there is one,
+        # and only fall back to the no-op behavior of returning `cls` as is.
+        mro = cls.__mro__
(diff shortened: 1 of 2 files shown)


ext/django_stubs_ext/patch.py
```diff
@@ -132,6 +132,24 @@ def __repr__(self) -> str:
...
+def _make_class_getitem(patched_cls: type) -> Any:
+    def __class_getitem__(cls: type, item: Any) -> Any:
+        # `cls` may be a subclass that also inherits from `typing.Generic`.
+        # In that case the patched class shadows `Generic.__class_getitem__` in
+        # the MRO, which would silently erase the runtime parametrization of
+        # `cls[...]` (https://github.com/typeddjango/django-stubs/issues/3609).
+        # Delegate to the next `__class_getitem__` in the MRO if there is one,
+        # and only fall back to the no-op behavior of returning `cls` as is.
+        mro = cls.__mro__
+        for base in mro[mro.index(patched_cls) + 1 :]:
```
sobolevn: We can skip the last one, `object.__class_getitem__` is not defined, that's for sure

sobolevn (approved):
Thanks!

PR #3615, opened 2026-08-27
Add defaults to the form type parameters of generic editing views

brianhelba:
This is a usability improvement: simple usages like `CreateView[Article]` or `DeleteView[Article]` no longer need to spell out a form type parameter, and instead default to the form type that Django actually provides at runtime when `form_class` is unset.

For example, the minimal usage changes from:
```py
class ArticleCreateView(CreateView[Article, ModelForm[Article]]): ...
```
to:
```py
class ArticleCreateView(CreateView[Article]): ...
```

The two defaults are justified differently:

- `_ModelFormT = ModelForm[_M]` (`CreateView`, `UpdateView`, and their bases): when `form_class` is unset, `ModelFormMixin.get_form_class()` falls back to `modelform_factory(model, fields=...)`, which generates a `ModelForm` subclass for the view's model, so `ModelForm[_M]` is what Django provides. This does require `fields` to be set (otherwise Django raises `ImproperlyConfigured`), a condition the type system cannot express, but that misconfiguration fails loudly at runtime.

- `_DeleteFormT = Form` (`BaseDeleteView`, `DeleteView`): `BaseDeleteView` unconditionally sets `form_class = Form`, so `Form` is always the runtime default. This is a new TypeVar rather than a default on the shared `_FormT`, because `FormView` also uses `_FormT` but inherits `form_class = None` and therefore has no runtime default to mirror.

## AI Policy

### DIFF
--- a/tests/assert_type/views/test_generic.py
+++ b/tests/assert_type/views/test_generic.py
@@ -7,7 +7,7 @@ from django.db import models
 from django.db.models import F
 from django.db.models.functions import Upper
 from django.views.generic.detail import SingleObjectMixin
-from django.views.generic.edit import DeleteView
+from django.views.generic.edit import CreateView, DeleteView, UpdateView
 from django.views.generic.list import ListView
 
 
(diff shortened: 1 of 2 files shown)


tests/assert_type/views/test_generic.py
```diff
@@ -46,6 +46,36 @@ class MyDeleteView(DeleteView[MyModel, ConfirmForm]): ...
...
 
 
+# When no form type parameter is given, the form types default to what Django provides at runtime.
+class SimpleCreateView(CreateView[MyModel]):
+    model = MyModel
+    fields = "__all__"
+
+
+simple_create_view = SimpleCreateView()
+assert_type(simple_create_view.get_form_class(), type[forms.ModelForm[MyModel]])  # ty: ignore[type-assertion-failure]
```
brianhelba, line 56: All 4 of these `ty: ignore` are due to https://github.com/astral-sh/ty/issues/4408

They can be removed once Ty is fixed.

brianhelba:
I don't think this has any direct overlap with #3317, but it's worth mentioning.

sobolevn (approved):
Thanks!

PR #3616, opened 2026-08-28
Allow Py3.15, please

milos-korenciak:
# Metadata allow Py3.15
Helps with testing things on Py3.15 (pre-release). No new feature gonna affect typing.

## AI Policy
- [X] I have read and agree to the [AI Policy](https://github.com/typeddjango/django-stubs/blob/master/.github/AI_POLICY.md), removed any "Co-Authored-By" lines attributing coding agents, and manually reviewed the final result

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
   "Programming Language :: Python :: 3.13",
   "Programming Language :: Python :: 3.14",
+  "Programming Language :: Python :: 3.15",
   "Typing :: Typed",
   "Framework :: Django",
   "Framework :: Django :: 5.2",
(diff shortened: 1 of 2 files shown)


UnknownPlatypus (changes requested):
Thanks for the pr but we don't test against 3.15 currently, pls update ci to test against this before declaring suport for it

pyproject.toml
```diff
@@ -21,6 +21,7 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
   "Programming Language :: Python :: 3.13",
   "Programming Language :: Python :: 3.14",
+  "Programming Language :: Python :: 3.15",
```
intgr, line 24: (Lack of) Python classifiers doesn't actually prevent installation.

If you're getting an error, there must be a different issue - post the messages you're getting.

PR #3620, opened 2026-09-02
Fix `@admin.display` overload to specify `function`

ngnpope:
# I have made things!

Fixes an inconsistency in overload definitions.

Follow up to 1f73fcff3bf8a06ea79df756111828a15dbd9f92 and fd3a2b8ac0d609e55c8abfa4831ce520bf7a6dec.

## Related issues

N/A

## AI Policy

### DIFF
--- a/django-stubs/contrib/admin/decorators.pyi
+++ b/django-stubs/contrib/admin/decorators.pyi
@@ -58,6 +58,7 @@ def display(
 ) -> Callable[[_F], _F]: ...
 @overload
 def display(
+    function: None = None,
     *,
     boolean: None = ...,
     ordering: str | Combinable | BaseExpression | None = ...,


django-stubs/contrib/admin/decorators.pyi
```diff
@@ -58,6 +58,7 @@ def display(
 ) -> Callable[[_F], _F]: ...
 @overload
 def display(
+    function: None = None,
```
sobolevn, line 61: wait, is there any real difference with the previous overload here? except `empty_value: None` vs `empty_value: str | None`

ngnpope: Yes. See the first of the commits mentioned in the description. They make `boolean` and `empty_value` mutually exclusive.

PR #3622, opened 2026-09-03
Relax `AbstractBaseUser.natural_key` return type

Kakadus:
Hi again :wave: !

# I have made things!

## Related issues

[I can open a ticket if wanted]

Django's `natural_key` magic function is intended to customize the natural key used for serialization. I noticed that `AbstractBaseUser`'s return type of it is overly restrictive when subclassing. It is allowed that subclasses return a tuple of multiple field values, or even the empty tuple to opt out of natural keys (since Django 6.1).

See https://docs.djangoproject.com/en/6.1/topics/serialization/#serialization-of-natural-keys

I noticed declarations of the `natural_keys` function at other places, which may have similar problems (this may become the ticket, I guess). 

## AI Policy

(No models were harmed in the creation of this small PR)


tests/assert_type/contrib/auth/test_models.py
sobolevn, line 1: please, remove the test, the change is simple enough to test it :)

sobolevn (approved):

PR #3629, opened 2026-09-07
fix(utils): allow _StrOrPromise in Truncator.__init__

Meghwin-Dave:
Accept lazy translation strings (\_StrOrPromise\) in \Truncator.__init__\ to match Django runtime behavior where lazy translation objects can be passed into Truncator.


UnknownPlatypus:
See https://github.com/typeddjango/django-stubs/blob/master/.github/AI_POLICY.md, this is not acceptable behavior

PR #3636, opened 2026-09-07
fix(views): use generic _ViewResponse return type for View.dispatch

Meghwin-Dave:
Use the generic type variable \_ViewResponse\ for \View.dispatch\ return type instead of \HttpResponseBase\ to preserve precise response typing when subclassing or overriding \dispatch\.


UnknownPlatypus:
- [ ]

PR #3642, opened 2026-09-08
Add a mutable HttpRequest typing helper

### LINKED ISSUE #3565: HttpRequest is no longer mutable after mypy 2.2 update
Source: https://github.com/typeddjango/django-stubs/pull/3467

New versions of mypy added a new feature that changed how `__new__` works.
We used a hack to separate `HttpRequest()` and `request: HttpRequest` objects to fit the Django runtime.

Since the mypy has changed its rules: now `__new__`'s return type is used as-is.
It means that our old pattern of 

```yaml
-   case: subclass_request_changed_user_type
    main: |
        class MyRequest(HttpRequest):
            user: User # Override the type of user

        request = MyRequest()
        reveal_type(request) # N: Revealed type is "main.MyRequest"
        reveal_type(request.user) # N: Revealed type is "django.contrib.auth.models.User"
    custom_settings: |
        INSTALLED_APPS = ('django.contrib.contenttypes', 'django.contrib.auth')

```

does not work anymore. `HttpRequest` can't use `__new__` anymore with:

```python
    # The magic. If we instantiate HttpRequest directly somewhere, it has
    # mutable GET and POST. However, both ASGIRequest and WSGIRequest have immutable,
    # so when we use HttpRequest to refer to any of them we want exactly this.
    # Case when some function creates *exactly* HttpRequest (not subclass)
    # remain uncovered, however it's probably the best solution we can afford.
    def __new__(cls) -> _MutableHttpRequest: ...
    # When both __init__ and __new__ are present, mypy will prefer __init__
    # (see comments in mypy.checkmember.type_object_type)
    # def __init__(self) -> None: ...
```

Because all the subclasses will be ignored. `_MutableHttpRequest` will always be used instead of `MyRequest()`

So, we would have to break the feature with mutable `GET` and `POST`. At least for now.
Ideas on how to represent mutable / immutable `GET` / `POST` are welcome.

I propose adding new `MutableHttpRequest` type to `django_stubs_ext`, so we can support request mutation for testing purposes.

SIBTAIN-ASAD:
Fixes #3565

Django's concrete HttpRequest keeps GET and POST mutable, while WSGIRequest and ASGIRequest expose immutable query dictionaries. Recent mypy releases changed __new__ handling, so the existing stub workaround could no longer preserve custom HttpRequest subclasses.

This adds a public MutableHttpRequest helper in django-stubs-ext. At runtime it aliases Django's HttpRequest; during type checking it exposes mutable QueryDict attributes for tests and request factories without changing HttpRequest subclass inference.

Verified with:
- uv run pytest ext -q
- uv run pytest tests/typecheck/test_request.yml -q
- uv run mypy ext/django_stubs_ext
- targeted prek hooks

### DIFF
--- a/ext/django_stubs_ext/types.py
+++ b/ext/django_stubs_ext/types.py
@@ -13,9 +13,10 @@ if TYPE_CHECKING:
         GET: QueryDict  # type: ignore[assignment]
         POST: QueryDict  # type: ignore[assignment]
 else:
-    # The runtime class is Django's regular HttpRequest; mutability is a
-    # type-checking distinction only.
-    from django.http import HttpRequest as MutableHttpRequest
+    # The runtime class is Django's regular HttpRequest; mutability is a type-checking distinction only.
+    from django.http import HttpRequest
+
+    MutableHttpRequest = HttpRequest
 
 
 # Used internally by mypy_django_plugin.
@@ -24,6 +25,3 @@ class AnyAttrAllowed(Protocol):
 
     @override
     def __setattr__(self, item: str, value: Any) -> None: ...
-
-
-__all__ = ["AnyAttrAllowed", "MutableHttpRequest"]
--- a/ext/django_stubs_ext/types.py
+++ b/ext/django_stubs_ext/types.py
@@ -1,8 +1,21 @@
 from __future__ import annotations
 
-from typing import Any, Protocol
+from typing import TYPE_CHECKING, Any, Protocol
 
 from typing_extensions import override
+
+if TYPE_CHECKING:
+    from django.http.request import HttpRequest, QueryDict
+
+    class MutableHttpRequest(HttpRequest):
+        """A request with mutable query dictionaries, as used in tests."""
+
+        GET: QueryDict  # type: ignore[assignment]
+        POST: QueryDict  # type: ignore[assignment]
+else:
+    # The runtime class is Django's regular HttpRequest; mutability is a
+    # type-checking distinction only.
+    from django.http import HttpRequest as MutableHttpRequest
 
 
 # Used internally by mypy_django_plugin.
@@ -11,3 +24,6 @@ class AnyAttrAllowed(Protocol):
 
     @override
     def __setattr__(self, item: str, value: Any) -> None: ...
+
+
+__all__ = ["AnyAttrAllowed", "MutableHttpRequest"]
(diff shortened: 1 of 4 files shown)


ext/django_stubs_ext/types.py
```diff
@@ -1,13 +1,29 @@
...
 
 # Used internally by mypy_django_plugin.
 class AnyAttrAllowed(Protocol):
     def __getattr__(self, item: str) -> Any: ...
 
     @override
     def __setattr__(self, item: str, value: Any) -> None: ...
+
+
+__all__ = ["AnyAttrAllowed", "MutableHttpRequest"]
```
sobolevn: I prefer not to have `__all__`

ext/tests/typecheck/test_aliases.yml
```diff
@@ -14,3 +14,11 @@
 
                 class Blog(models.Model):
                     created_at = models.DateTimeField()
+
+-   case: mutable_http_request
```
UnknownPlatypus: Can you move this to an assert_type test ?

ext/django_stubs_ext/types.py
```diff
@@ -1,13 +1,29 @@
...
+    from django.http.request import HttpRequest, QueryDict
+
+    class MutableHttpRequest(HttpRequest):
+        """A request with mutable query dictionaries, as used in tests."""
+
+        GET: QueryDict  # type: ignore[assignment]
+        POST: QueryDict  # type: ignore[assignment]
+else:
+    # The runtime class is Django's regular HttpRequest; mutability is a
+    # type-checking distinction only.
```
UnknownPlatypus:
```suggestion
    # The runtime class is Django's regular HttpRequest; mutability is a type-checking distinction only.
```

UnknownPlatypus (changes requested):

SIBTAIN-ASAD:
Thanks for the review. I removed the module-level __all__, moved the regression coverage into tests/assert_type/http/test_request.py, and kept the runtime alias explicit so django_stubs_ext.MutableHttpRequest resolves to Django's HttpRequest at runtime.

I re-ran the ext tests, HttpRequest typecheck cases, mypy on the stubs, and pyright/pyrefly/ty for the assert_type file; all pass. The follow-up commit is 77cc1fc.

sobolevn (approved):
Thank you!

PR #3643, opened 2026-09-09
Loosen `Widget.format_value` and remove override ignores

brianhelba:
`Widget.format_value` returns whatever the template needs, and Django's own subclasses return a variety of types. Previously, the annotation of the base class as returning `str | None` (while this was the precise behavior of the implementation) forced `# type: ignore[override]` on many subclasses (including downstream third-party widgets).

Instead, this returns `Any` from the base class and annotates only those child classes having explicit overrides with implementation-accurate types. This is the most accurate representation of the type contract for `Widget.format_value`.

This is a trade-off, as subclasses without an explicit override (e.g. `TextInput`) lose the more precise `str | None` return type of their actual implementation. However, the alternative is to annotate every single subclass with a `format_value` method, which seems too burdensome.

## AI Policy


UnknownPlatypus:
Could you add an `assert_type` test showcasing what that this resolves for you ?

It's not super clear to me and I'm afraid it currently regresses the type of most builtin fields ?

brianhelba:
@UnknownPlatypus I'm happy to add a test if needed, but maybe I can just better explain the situation:

---
Before this change, we have the hierarchy:
```python
class Widget:
    def format_value(self, value: Any) -> str | None: ...

class ChoiceWidget(Widget):
    def format_value(self, value: Any) -> list[str]: ...  # type: ignore[override]
```

Note how `ChoiceWidget.format_value` needs an ignore, because it has a different return type and violates Liskov substitution.

Worse, any user-defined `Widget` subclass in third-party code is implicated. [Django says that `format_value`](https://docs.djangoproject.com/en/6.1/ref/forms/widgets/#django.forms.Widget.format_value):
> Cleans and returns a value for use in the widget template.

So, user-defined widgets will frequently need `# type: ignore`:
```python
class MyAgeWidget(Widget):
    def format_value(self, value: Any) -> int: ...  # type: ignore[override]
```

---
This change makes it so `Widget.format_value` describes the *contract* for all subclasses, instead of describing just what the particular default implementation does.
```python
class Widget:
    # format_value subclasses can return any value usable in a template
    def format_value(self, value: Any) -> Any: ...

class ChoiceWidget(Widget):
    def format_value(self, value: Any) -> list[str]: ...

class MyAgeWidget(Widget):
    def format_value(self, value: Any) -> int: ...
```
Now subclasses no longer need a `# type: ignore` to declare `format_value` in Django's intended way. From a polymorphism perspective, this is now correct too.

---
Why'd I say there was any tradeoff? Look at:
```python
class Widget:
    # format_value subclasses can return any value usable in a template
    def format_value(self, value: Any) -> Any: ...

class Input(Widget): ...
```
If I call `Input.format_value(...)`, what does it return? At runtime, since it inherits the default implementation from `Widget.format_value`, it will be `str | None`, but now the type system doesn't know this anymore.

---
With your feedback, I'd be fine to make this PR instead add annotations for all the direct descendants of `Widget` that don't actually declare their own `format_value`:
```python
class Input(Widget):
    def format_value(self, value: Any) -> str | None: ...

class Textarea(Widget):
    def format_value(self, value: Any) -> str | None: ...

class MultiWidget(Widget):
    def format_value(self, value: Any) -> str | None: ...

class RelatedFieldWidgetWrapper(forms.Widget):
    def format_value(self, value: Any) -> str | None: ...
```
Note that none of these methods actually exist, but we're legally narrowing the contract so that these widgets (and their descendants) now have more specific type information.

This would be more extra code. Also, I believe it's more common for users to declare a custom widget with a `format_value` (wherein we want to spare them from type errors) than it is to actually directly invoke an existing `format_value` (this is usually called by Django's internal form machinery). So, I decided to go with the simpler approach, but I'm happy to reconsider.

UnknownPlatypus:
Ok I see, maybe a loose type is fine indeed, I think it's drill a good idea to add a few test of user defined custom widgets to ensure it's working fine with no extra type ignore needed

brianhelba:
@UnknownPlatypus I just added a test case.

Let me know if you'd like other particular tests, only this one was obvious to me.

UnknownPlatypus:
@brianhelba Awesome thanks, this is great because the test it also document the reason in case someone try to change this behavior one day, he'll end up here.

Great work!

PR #3655, opened 2026-09-15
Fix lookup expected type for transform chains (e.g. start_date__year__in)

### LINKED ISSUE #3411: __year__in (transform + __in) regressed in 6.0.5: rejects list[int], expects Iterable[str | date]

manysuq:
Fixes #3411

## Summary

The expected type for a lookup was resolved against the original field, ignoring intermediate transforms in the chain. For `start_date__year__in` on a `DateField` this incorrectly expected `date` values instead of `int`, since the `year` transform outputs an `IntegerField`.

The fix walks the intermediate transforms of the lookup chain and re-resolves the field via each transform's `output_field`, falling back to the original field when a transform has no concrete `output_field`.

## Validation

- The reproducer from #3411 now typechecks correctly: `start_date__year__in` accepts `list[int]` and still rejects incompatible values like `(Event(),)`.
- Existing lookup/field behavior is unchanged (exact lookups, plain `__in`, related lookups all resolve as before).
- Full typecheck test suite passes locally (the only failure is a pre-existing GIS one caused by missing GDAL in this environment, reproduced on clean master).

Happy to add a regression test case to the filter yml if you'd like one.


sobolevn (commented):
This would need a test case :)

manysuq:
Fair point, added a test case in `test_filter.yml` (`transform_then_lookup_uses_transform_output_type`).

It covers both directions: ints are now accepted for `__year` lookups (the false positive from the issue) while dates are rejected against the transform's output type. I also threw in a truncation chain (`created_at__date__in`) since it goes through the same code path with a per-instance output field.

The test fails on `main` the way the issue describes, so it should be a decent regression guard going forward.

tests/typecheck/managers/querysets/test_filter.yml
```diff
@@ -428,3 +428,40 @@
...
+-   case: transform_then_lookup_uses_transform_output_type
+    main: |
+        from datetime import date
+
+        from myapp.models import Event
+
+        def by_int_years(years: list[int]) -> None:
+            # `year` transform outputs an int, so lookups on it expect ints
+            Event.objects.filter(start_date__year__in=years)
+            Event.objects.filter(start_date__year__exact=2000)
```
sobolevn, line 440:
```suggestion
            Event.objects.filter(start_date__year__exact=2000)
            Event.objects.filter(start_date__year=2000)
```

should be also tested to be the same thing.

manysuq: Done in 652f83d0 — added a bare transform lookup case:\n\n	python\n\tEvent.objects.filter(start_date__year=2000)\n\nIt resolves the same way as `__year__exact` (`solve_lookup_type` returns `['year']` in `lookup_parts` and the final lookup falls back to `Exact`), so the expected type is now `str | int` for both. Test passes locally, full suite is green (only the pre-existing GIS/GDAL failures, also failing on master).

sobolevn (approved):
Thanks!

PR #3661, opened 2026-09-17
Allow a sequence of ActionLocation in admin.action's location

knyghty:
# I have made things!

Without this the checker will complain if you try to use the `@admin.action` decorator with [a sequence of `ActionLocation`s](https://docs.djangoproject.com/en/6.1/ref/contrib/admin/actions/#controlling-where-actions-are-available).

## Related issues

I didn't make an issue but I will if it's needed.

## AI Policy


UnknownPlatypus (commented):
> I didn't make an issue but I will if it's needed.

No need, thanks for the PR !

